PR #176, opened 2025-11-29
feat(desktop): add cloud workspaces using yolocode.ai integration

caffeinum:
## Summary

- Add cloud workspace button (blue cloud icon) in header next to the regular "+" button
- Creates E2B sandbox via yolocode API and opens two embedded webview tabs:
  - **Cloud Agent** - Claude chat UI at port 7030
  - **Cloud Terminal** - WebSSH terminal at port 8888 (pre-filled with localhost/user credentials)
- Uses electron `<webview>` tag instead of iframe to properly embed external URLs
- Cloud sandbox status shown on workspace tabs (blue cloud icon = running, yellow warning = stopped)
- Dangling sandboxes (running but not linked to a workspace) shown in tab bar for management

## Architecture

- **Cloud tabs**: Dedicated `TabType.Cloud` with `createCloudTab()` factory that extracts sandbox ID from URL for tab naming
- **tRPC**: All cloud operations migrated to tRPC router (`cloud.createSandbox`, `cloud.deleteSandbox`, `cloud.listSandboxes`, etc.)
- **Status polling**: Each `WorkspaceItem` fetches its own cloud status via `getWorktreeCloudStatus` query (30s polling)
- **Sandbox linking**: Sandboxes stored on worktree in local db, auto-deleted when workspace is deleted
- **Split tabs**: Cloud tabs fully support split view - can be split via context menu, keyboard shortcuts, or drag-and-drop

## Auth

Currently uses user's GitHub token (`gh auth token`) which is sent to yolocode API. Users will need to authenticate with yolocode directly instead.

## Setup

Requires `CLAUDE_CODE_OAUTH_TOKEN` in `.env` file. Get it by running:

```bash
claude setup-token
```

## TODO

- [x] Draft implementation for cloud worktrees using yolocode.ai
- [x] Show content of AI chat and cloud terminal from yolocode sandbox
- [x] Cloud status indicators on workspace tabs
- [x] Dangling sandbox management
- [x] Migrate IPC handlers to tRPC
- [x] Split tabs support for Cloud tabs
- [ ] Proper yolocode auth (not GitHub token)

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/TopBar/WorkspaceTabs/WorkspaceItemContextMenu.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/TopBar/WorkspaceTabs/WorkspaceItemContextMenu.tsx
@@ -12,7 +12,7 @@ import {
 } from "@superset/ui/hover-card";
 import { toast } from "@superset/ui/sonner";
 import { type ReactNode, useState } from "react";
-import { HiMiniCloud } from "react-icons/hi2";
+import { HiMiniCloud, HiMiniTrash } from "react-icons/hi2";
 import { trpc } from "renderer/lib/trpc";
 import { trpcClient } from "renderer/lib/trpc-client";
 import { useAddCloudTab } from "renderer/stores/tabs";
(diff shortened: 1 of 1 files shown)


apps/desktop/src/lib/trpc/routers/workspaces/workspaces.ts
```diff
@@ -134,11 +135,17 @@ export const createWorkspacesRouter = () => {
...
 			const activeProjects = db.data.projects.filter(
 				(p) => p.tabOrder !== null,
 			);
 
+			// Fetch live sandbox statuses
+			const sandboxResult = await cloudApiClient.listSandboxes();
+			const liveSandboxes = new Map(
+				(sandboxResult.sandboxes ?? []).map((s) => [s.id, s.status]),
+			);
+
```
caffeinum: this doesnt look good, probably it should be a separate place where this happens

PR #349, opened 2025-12-13
feat(marketing): load env from monorepo root and add cookie domain default

Kitenite:
## Changes

- Add dotenv package to load .env from monorepo root in development mode
- Set NEXT_PUBLIC_COOKIE_DOMAIN default to 'localhost' for local development
- Load root .env file with override flag to ensure proper environment variable loading

## Why

This ensures the marketing app can properly load environment variables from the monorepo root .env file during development, matching the pattern used in other apps in the monorepo.

## Summary by CodeRabbit

* **Chores**
  * Added dotenv as a dev dependency and improved dev environment loading.
  * Added a default cookie-domain value to simplify local development.

* **Style**
  * Refined responsive spacing, padding and typography across header, CTA, download button, platform dropdown and social links for improved small-screen layout.
  * Shortened some button/link labels on small viewports and adjusted dropdown sizing/positioning for better alignment.
* **Bug Fixes**
  * Hidden sign-in link by default in the not-logged-in state.

### DIFF
--- a/apps/marketing/src/env.ts
+++ b/apps/marketing/src/env.ts
@@ -1,12 +1,15 @@
-import { resolve } from "node:path";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
 import { createEnv } from "@t3-oss/env-nextjs";
 import { vercel } from "@t3-oss/env-nextjs/presets-zod";
 import { config } from "dotenv";
 import { z } from "zod";
 
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
 if (process.env.NODE_ENV === "development") {
 	// Load .env from monorepo root
-	config({ path: resolve(process.cwd(), "../../.env"), override: true });
+	config({ path: resolve(__dirname, "../../.env"), override: true });
 }
 
 export const env = createEnv({


apps/marketing/src/env.ts
```diff
@@ -1,7 +1,17 @@
...
 import { createEnv } from "@t3-oss/env-nextjs";
 import { vercel } from "@t3-oss/env-nextjs/presets-zod";
+import { config } from "dotenv";
 import { z } from "zod";
 
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+if (process.env.NODE_ENV === "development") {
+	// Load .env from monorepo root
+	config({ path: resolve(__dirname, "../../.env"), override: true });
```
Kitenite, line 12: @saddlepaddle fyi I just have this load .env from root. I think it's annoying to have more than 1 especially as an open-source monorepo

apps/marketing/src/app/components/CTAButtons/CTAButtons.tsx
```diff
@@ -9,37 +9,40 @@ export async function CTAButtons() {
...
 		);
 	}
 
 	return (
-		<div className="flex items-center gap-3">
+		<div className="flex items-center gap-2 sm:gap-3">
 			<a
 				href={`${env.NEXT_PUBLIC_WEB_URL}/sign-in`}
-				className="px-4 py-2 text-sm font-normal text-muted-foreground hover:text-foreground transition-colors"
+				className="hidden px-2 sm:px-4 py-2 text-sm font-normal text-muted-foreground hover:text-foreground transition-colors"
```
Kitenite, line 36: @saddlepaddle fyi hiding this because it is not functional and can cause confusion

apps/marketing/src/env.ts
saddlepaddle: tru we can do that too! i have direnv set up fyi, but this is reasonable too to add to each project 👍🏼

PR #356, opened 2025-12-13
fix(desktop): bundle git binary to avoid system git dependency

saddlepaddle:
## Summary

- Bundles a portable git binary using [dugite](https://github.com/desktop/dugite) (same approach as GitHub Desktop)
- Eliminates dependency on system git, fixing the "Xcode license agreements" error on macOS
- Includes git-lfs in the bundle, so LFS repositories work out of the box

## Problem

Users on macOS were seeing this error when trying to use the app:

> Failed to initialize git repository: You have not agreed to the Xcode license agreements. Please run 'sudo xcodebuild -license' from within a Terminal window to review and agree to the Xcode and Apple SDKs license.

This happened because the app relied on the system git, which on macOS comes from Xcode Command Line Tools and requires accepting the license.

## Solution

Bundle git with the app using dugite, the same solution GitHub Desktop uses. The app now ships with its own git binary (~40-50MB added to app size) and works out of the box without any system dependencies.

## Test plan

- [x] Typecheck passes
- [x] All tests pass (313/313)
- [ ] Manual test: Open a project on a fresh macOS install without Xcode
- [ ] Manual test: Clone a repo with LFS files

## Summary by CodeRabbit

* **Improvements**
  * Desktop app now uses an embedded Git binary for repository operations, improving Git LFS detection, worktree handling, and error classification.
* **Chores**
  * Bundled Git added to packaging and runtime; environment‑based Git handling removed.
* **Tests**
  * Tests and mocks updated to validate bundled Git behavior, precise path matching, and related edge cases.
* **Build**
  * Native module copy logic updated to include bundled Git files and dereference symlinks.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/workspaces/utils/git.ts
+++ b/apps/desktop/src/lib/trpc/routers/workspaces/utils/git.ts
@@ -9,8 +9,35 @@ import {
 	animals,
 	uniqueNamesGenerator,
 } from "unique-names-generator";
+import { checkGitLfsAvailable, getShellEnvironment } from "./shell-env";
 
 const execFileAsync = promisify(execFile);
+
+/**
+ * Builds the merged environment for git operations.
+ * Takes process.env as base, then overrides only PATH from shell environment.
+ * This preserves runtime vars (git credentials, proxy, ELECTRON_*, etc.)
+ * while picking up PATH modifications from shell profiles (e.g., homebrew git-lfs).
+ */
+async function getGitEnv(): Promise<Record<string, string>> {
+	const shellEnv = await getShellEnvironment();
+	const result: Record<string, string> = {};
+
+	// Start with process.env as base
+	for (const [key, value] of Object.entries(process.env)) {
+		if (typeof value === "string") {
+			result[key] = value;
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


AviPeltz:
the change view shows "unable to load changes" 
-> debugging
-> it works locally in dev but not in the built release think i need to do the __dirname thing so it bundles so trying to futs with out dugite is being bundled

saddlepaddle:
> the change view shows "unable to load changes" -> debugging -> it works locally in dev but not in the built release think i need to do the __dirname thing so it bundles so trying to futs with out dugite is being bundled

LOL rip, production app behaves so differently from dev :')

AviPeltz:
> > the change view shows "unable to load changes" -> debugging -> it works locally in dev but not in the built release think i need to do the __dirname thing so it bundles so trying to futs with out dugite is being bundled
> 
> LOL rip, production app behaves so differently from dev :')

yeah I think the Dugite's JS code loads from inside the electron bundled asar archive, so when it uses __dirname to find the git binary, it returns a path inside the asar - but you can't execute binaries from inside an asar. So hopefully if i just point to the unpacked version of the app.asar file the binary will be there and it will work

AviPeltz:
i think this should bundle but ill check out the above coderabbit error

PR #421, opened 2025-12-18
fix(desktop): separate env files by context to prevent renderer crashes

saddlepaddle:
## Summary

- **Fixed blank screen issue in production builds** caused by `...process.env` spread being bundled into renderer code
- **Split env.ts into context-specific files**:
  - `src/main/env.main.ts` - Main process env vars (can use t3-env with `...process.env`)
  - `src/renderer/env.renderer.ts` - Renderer env vars (uses Zod with Vite-injected values)
  - `src/shared/env.shared.ts` - Shared env vars (only NODE_ENV, works in both contexts)
- **Updated CI workflow** to pass PostHog env vars at build time

### Why this works

Vite can only replace **individual** `process.env.X` property accesses at build time. The spread `...process.env` requires an actual object at runtime, which doesn't exist in the browser. By splitting the env files:

1. Main process uses t3-env with runtime `process.env` (Node.js has this)
2. Renderer uses Zod validation on build-time injected values (Vite `define` block)
3. Shared code imports from `env.shared.ts` which only accesses `process.env.NODE_ENV` (replaced by Vite)

## Test plan

- [x] Verified typecheck passes
- [x] Tested deep links work in production build
- [x] Verified PostHog initializes with key from CI secrets
- [x] Confirmed app doesn't crash on launch

## Summary by CodeRabbit

* **Chores**
  * Configured production environment settings and analytics tracking integration for the desktop application.
  * Updated environment variable configuration to support customizable API and web service URLs.
  * Reorganized internal environment configuration management for improved application stability.

### DIFF
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -57,7 +57,7 @@ jobs:
         working-directory: apps/desktop
         env:
           NEXT_PUBLIC_POSTHOG_KEY: ${{ secrets.NEXT_PUBLIC_POSTHOG_KEY }}
-          NEXT_PUBLIC_POSTHOG_HOST: https://us.i.posthog.com
+          NEXT_PUBLIC_POSTHOG_HOST: ${{ secrets.NEXT_PUBLIC_POSTHOG_HOST }}
         run: bun run compile:app
 
       # Build the Electron app for macOS
--- a/apps/desktop/src/main/lib/api-client.ts
+++ b/apps/desktop/src/main/lib/api-client.ts
@@ -1,7 +1,7 @@
 import type { AppRouter } from "@superset/trpc";
 import { createTRPCClient, httpBatchLink } from "@trpc/client";
+import { env } from "main/env.main";
 import superjson from "superjson";
-import { env } from "../env.main";
 import { authService } from "./auth";
 
 /**
--- a/apps/desktop/src/main/lib/auth/auth-service.ts
+++ b/apps/desktop/src/main/lib/auth/auth-service.ts
@@ -1,8 +1,8 @@
 import { EventEmitter } from "node:events";
 import { TOKEN_CONFIG } from "@superset/shared/constants";
 import { type BrowserWindow, shell } from "electron";
+import { env } from "main/env.main";
 import type { AuthProvider, AuthSession, SignInResult } from "shared/auth";
-import { env } from "../../env.main";
 import { pkceStore } from "./pkce";
 import { tokenStorage } from "./token-storage";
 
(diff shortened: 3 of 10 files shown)


.github/workflows/release-desktop.yml
```diff
@@ -54,6 +55,9 @@ jobs:
 
       - name: Compile app with electron-vite
         working-directory: apps/desktop
+        env:
+          NEXT_PUBLIC_POSTHOG_KEY: ${{ secrets.NEXT_PUBLIC_POSTHOG_KEY }}
+          NEXT_PUBLIC_POSTHOG_HOST: https://us.i.posthog.com
```
Kitenite: maybe don't hard code this?

apps/desktop/src/main/lib/auth/auth-service.ts
```diff
@@ -2,7 +2,7 @@ import { EventEmitter } from "node:events";
 import { TOKEN_CONFIG } from "@superset/shared/constants";
 import { type BrowserWindow, shell } from "electron";
 import type { AuthProvider, AuthSession, SignInResult } from "shared/auth";
-import { env } from "../../../env";
+import { env } from "../../env.main";
```
Kitenite: should we just do the alias? main/env.main instead of relative?

apps/desktop/src/main/lib/auth/deep-link-handler.ts
```diff
@@ -1,6 +1,6 @@
 import type { AuthSession } from "shared/auth";
 import { PROTOCOL_SCHEMES } from "shared/constants";
-import { env } from "../../../env";
+import { env } from "../../env.main";
```
Kitenite: same should we just do the alias? main/env.main instead of relative?

apps/desktop/src/main/lib/api-client.ts
```diff
@@ -1,7 +1,7 @@
 import type { AppRouter } from "@superset/trpc";
 import { createTRPCClient, httpBatchLink } from "@trpc/client";
 import superjson from "superjson";
-import { env } from "../../env";
+import { env } from "../env.main";
```
Kitenite: should we just do the alias? main/env.main instead of relative?

apps/desktop/src/main/lib/auto-updater.ts
```diff
@@ -1,7 +1,7 @@
 import { app, type BrowserWindow, dialog } from "electron";
 import { autoUpdater } from "electron-updater";
 import { PLATFORM } from "shared/constants";
-import { env } from "../../env";
+import { env } from "../env.main";
```
Kitenite: should we just do the alias? main/env.main instead of relative?

apps/desktop/src/main/lib/sound-paths.ts
```diff
@@ -1,7 +1,7 @@
 import { existsSync } from "node:fs";
 import { join } from "node:path";
 import { app } from "electron";
-import { env } from "../../env";
+import { env } from "../env.main";
```
Kitenite: should we just do the alias? main/env.main instead of relative?

.github/workflows/release-desktop.yml
saddlepaddle: LOOL ur so valid for that - mb i missed that

PR #536, opened 2025-12-29
fix(desktop): shift+enter adding stray \ characters in Claude Code

maddada:
## Description

Changed the Shift+Enter terminal behavior to send a plain newline (\n) instead of backslash-newline (\\n) to avoid adding unnecessary slash characters in cli tool input boxes.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [x] Other (please describe): Small UX tweak

## Testing

Verified that the change works well manually on codex, claude code, gemini-cli, opencode.

2025-12-29_Superset_15-08-05

2025-12-29_Superset_15-09-05

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal input behavior: Shift+Enter now sends the correct continuation sequence to the backend to ensure proper line continuation.
* **Documentation**
  * Clarified docs/comments describing Shift+Enter behavior and the sequence used for line continuation.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/Terminal.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/Terminal.tsx
@@ -355,7 +355,7 @@ export const Terminal = ({ tabId, workspaceId }: TerminalProps) => {
 		};
 
 		const cleanupKeyboard = setupKeyboardHandler(xterm, {
-			onShiftEnter: () => handleWrite("\n"),
+			onShiftEnter: () => handleWrite("\x1b\r"), // ESC + CR for line continuation without '\'
 			onClear: handleClear,
 		});
 
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts
@@ -173,7 +173,7 @@ export function createTerminalInstance(
 }
 
 export interface KeyboardHandlerOptions {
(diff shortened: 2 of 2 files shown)


andreasasprou (commented):
Fixing the stray `\` showing up in TUI inputs is a good UX improvement. That said, this is a behavior change:

- Previously, Shift+Enter injected `\\\n` (forced shell line continuation). Now it injects a plain `\n`, and we still intercept the key. This likely changes Shift+Enter from “continue” to effectively “execute” in shells. It also bypasses `xterm.onKey`, so `commandBufferRef` and tab auto-title clearing may get out of sync.
- Can you clarify the intended UX?
  - Should Shift+Enter behave like Enter everywhere?
  - Or is this specifically to avoid the backslash in full-screen TUIs?
    - If the goal is “behave like Enter”, we should remove the **Shift+Enter** interception entirely and let xterm send the normal Enter sequence (`\r`) instead of manually writing `\n`.
    - If the goal is “TUI-specific”, consider only altering behavior in the alternate buffer (full-screen apps) and keeping the continuation behavior in the normal buffer.
- Docs need updating. `helpers.ts` still says **Shift+Enter** “creates a line continuation (like iTerm)” (see comments around `helpers.ts:204` and `helpers.ts:257`).

maddada:
@andreasasprou Good points thank you.
For the alternate buffer suggestion, that won't work in Claude Code since they don't use alternate buffer.

I checked how Claude Code modifies the keybindings on Ghostty when I ran /terminal-setup and it adds this keybinding:
`keybind = shift+enter=text:\x1b\r`

So I used that instead of \n in order to still have continuation work in normal terminal but not have the \ appear before each newline when using shift + enter in Claude Code and other Agentic CLI tools.

https://github.com/user-attachments/assets/f6af3c9e-1ab3-4edb-96d3-3b6b6d832bf0

Adjusted the text in the comments too.

Hope this is better.

PR #752, opened 2026-01-14
fix(desktop): prevent blank file viewer on workspace switch

andreasasprou:
**Links**
- ExecPlan: `apps/desktop/plans/done/20260114-1739-fix-file-pane-cross-workspace.md`
- Follow-up ExecPlan: `apps/desktop/plans/20260114-2050-workspace-scoped-tabs-state.md`

Demo of bug and fix: https://www.loom.com/share/26e180759f1143219a47950ffb99b654

## Summary
- Prevent cross-workspace tab rendering so workspace switching can’t show a stale file viewer pane.
- Resolve file viewer `worktreePath` from `tab.workspaceId` and sanitize persisted tab pointers on hydration.
- Add unit tests for the active-tab resolver and guard against cross-workspace pane moves.

## Why / Context
Switching workspaces while a file viewer pane was open could render an empty/blank file pane in the target workspace. The root cause was an invariant violation: the route `workspaceId` and the rendered tab’s `workspaceId` could drift (stale `activeTabIds` / history), so the UI rendered a tab from another workspace.

## How It Works
- Introduces `resolveActiveTabIdForWorkspace` (active → MRU → first tab fallback).
- Uses the resolver in the store + UI render path, and sanitizes persisted pointers (`activeTabIds`, `tabHistoryStacks`, `focusedPaneIds`) at hydration.
- Makes `TabView` fetch the workspace by `tab.workspaceId` for file viewer panes.
- Disallows moving panes across workspaces (`movePaneToTab` guard).

## Manual QA Checklist
- [x] Open a file viewer in workspace A, switch to workspace B with a terminal open → workspace B shows its own panes (no blank file viewer)
- [x] Switch back to A → file viewer still present
- [x] Repeat switching a few times (no blank panes)
- [x] Restart app and repeat

## Testing
- `bun run lint`
- `bun test apps/desktop/src/renderer/stores/tabs/utils.test.ts`
- Note: `bun test` (repo-wide) currently fails due to `apps/cli/src/lib/storage/lowdb-adapter.test.ts` (unrelated).

## Risks / Rollback
- Risk: on startup, corrupted persisted tab pointers may be auto-corrected to a different tab than before; this is intended to avoid cross-workspace panes.
- Rollback: revert commits `570a819f` and `b800faf4`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed blank file viewer panes that appeared when switching between workspaces
  * Improved tab consistency and recovery when navigating across workspaces
  * Prevented invalid pane transfers between different workspaces

* **Tests**
  * Added regression tests for workspace tab resolution and workspace-switching scenarios


saddlepaddle:
Thanks for grabbing this! This definitely needs to be fixed so i'm fine with merging to make sure the bug is gone, but if you have a moment it may be time to organize the workspace state under a workspace instead of it being global and having things like all the tabs in all workspaces at the top-level.i.e.:
```
// tabs state is now isolated per workspace
  {
    tabsState: Record<workspaceId, {
      tabs: Tab[],
      panes: Record<paneId, Pane>,
      activeTabId: string | null,
      focusedPaneIds: Record<tabId, paneId>,
      tabHistoryStack: string[]
    }>
  }
```

If you have the time it'd be nice to solve the root issue, otherwise ye we can merge this in 👍🏼

saddlepaddle (approved):

PR #813, opened 2026-01-18
fix(desktop): make scroll to bottom instant by default

Kitenite:
## Summary
- Change terminal scroll to bottom behavior from smooth to instant for immediate navigation
- Make the scroll behavior configurable via a parameter that defaults to "instant"
- Rename `smoothScrollToBottom` to `scrollToBottom` to reflect the new default behavior

## Test plan
- [ ] Open a terminal with content scrolled up
- [ ] Click the scroll to bottom button and verify it jumps instantly (no animation)
- [ ] Use the keyboard shortcut for scroll to bottom and verify instant behavior


AviPeltz:
shit

AviPeltz:
we both just did this

PR #856, opened 2026-01-20
feat(desktop): Add drag-and-drop panes to tab bar

Kitenite:
## Summary
- Allow dragging panes to existing tabs in the tab bar to move them
- Allow dragging panes to empty tab bar space to create a new tab
- Visual feedback with accent highlighting during drag-over

## Test plan
- [x] Split a terminal into multiple panes
- [x] Drag a pane by its toolbar to another tab - verify pane moves to that tab
- [x] Drag a pane to empty space in the tab bar - verify new tab is created
- [x] Verify visual feedback (highlight) appears when dragging over valid drop targets
- [x] Verify dropping a pane on its own tab does nothing

## New Features
* Drag-and-drop panes across the workspace: drag a pane and drop it onto tabs to reorganize layouts.
* Drop onto a tab moves the pane into that tab; dropping on the strip can create a new tab for the pane.
* Visual feedback highlights valid drop targets and the strip while dragging.
* Drag state is cleaned up automatically when dragging ends to keep behavior consistent.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/GroupStrip/GroupItem.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/GroupStrip/GroupItem.tsx
@@ -52,16 +52,13 @@ export function GroupItem({
 				return { handled: true };
 			},
 			canDrop: () => {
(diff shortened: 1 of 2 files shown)


Kitenite:
Closing because this is buggy

PR #892, opened 2026-01-22
fix(desktop): kill all terminal child processes on close

### LINKED ISSUE #893: [bug] terminals not fully cleaned up
#### Describe the bug

I'm noticing an issue where if I Command-W a window while I have Claude running, that Claude instance remains running. This keeps crashing my computer as it accumulates up to a roughly 80GB of memory before my computer crashes lol

Other associated issues include my next apps not cleaning up their lock files

My laptop can't take these loose Claudes roaming my laptop so reproduced and submitted a PR to resolve this by getting all of the associated processes and killing them all, rather than just the shell process.

chasemcdo:
## Summary

Fixes terminal child processes (Claude Code, npm dev servers, Next.js, etc.) not being killed when closing a terminal tab with Cmd+W. Previously, only the shell process was terminated while child processes continued running in the background.

## Problem

When closing a terminal in the desktop app:
- Only the shell (zsh/bash) receives the termination signal
- Child processes like Claude Code, `npm run dev`, etc. survive and accumulate
- This causes resource leaks
- Next.js lock files aren't cleaned up, blocking new dev server starts

## Reproduction

1. Open a terminal in the desktop app
2. Run `claude` to start Claude Code
3. Inside Claude Code, run `echo $PPID` to get the shell's PID (e.g., `12345`)
4. Close the terminal with Cmd+W
5. Check if the process is still running: `ps aux | grep 12345`
6. **Before this fix:** The Claude Code process is still running
7. **After this fix:** The process is terminated

## Root Cause

The original implementation used node-pty's `ptyProcess.kill(signal)` which only sends the signal to the shell's PID directly - not to any child processes.

A naive fix would be to use `process.kill(-pid, signal)` to kill by process group. However, this doesn't work because interactive shells with job control (zsh, bash) give foreground jobs their **own process groups**:

```
zsh (PID 100, PGID 100)  ← shell's process group
└─ claude (PID 101, PGID 101)  ← zsh gave claude its OWN process group!
```

So `process.kill(-100, signal)` only kills processes in PGID 100 (the shell), not PGID 101 (claude).

## Solution

Use [tree-kill](https://www.npmjs.com/package/tree-kill) to terminate the entire process tree by traversing parent-child relationships (PPID). This handles job control correctly because the parent-child link is preserved even when zsh assigns a new process group.

An alternative approach is TTY-based killing (`pkill -t <tty>`), which catches all processes sharing the controlling terminal. tree-kill is simpler (no shelling out to `ps`/`pkill`) and sufficient here—the edge cases where TTY wins (daemonized processes that keep the TTY) don't apply to interactive terminal commands.

## Changes

- Added `tree-kill` dependency
- Updated `handleKill()` to use tree-kill with SIGTERM → SIGKILL escalation
- Updated `handleDispose()` to use tree-kill
- `handleSignal()` unchanged—Ctrl+C should only signal the foreground process, not kill the tree

## Testing

1. Open a terminal, run `claude` (or `npm run dev`)
2. Close with Cmd+W
3. Verify process is gone: `pgrep -f claude` should not show the terminated instance

Resolves #893

---

> [!NOTE]
> Ensures terminal child processes are reliably terminated when a tab is closed.
> 
> - Replace direct `ptyProcess.kill()` with `tree-kill` in `handleKill()` to terminate the full process tree; escalate to `SIGKILL` after 2s and force-exit if `onExit` never fires
> - Update `handleDispose()` to `tree-kill` the process tree with `SIGKILL` on exit
> - Keep `handleSignal()` behavior unchanged (signals like `SIGINT` only)
> - Add `tree-kill` dependency and bump desktop app version to `0.0.59`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d386714a2763a634779f26572e6573e918eb751e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal process termination to reliably stop all related processes and ensure proper cleanup when closing terminal windows, with enhanced error handling and fallback mechanisms.

### DIFF
--- a/apps/desktop/src/main/terminal-host/pty-subprocess.ts
+++ b/apps/desktop/src/main/terminal-host/pty-subprocess.ts
@@ -372,7 +372,7 @@ function handleKill(payload: Buffer): void {
 	// tree-kill traverses by PPID to find all descendants
 	treeKill(pid, signal, (err) => {
 		if (err) {
-			// Process may already be dead, that's fine
+			console.error("[pty-subprocess] Failed to kill process tree:", err);
 		}
 	});
 
@@ -383,7 +383,7 @@ function handleKill(payload: Buffer): void {
 
 		treeKill(pid, "SIGKILL", (err) => {
 			if (err) {
-				// Process may already be dead
+				console.error("[pty-subprocess] Failed to SIGKILL process tree:", err);
 			}
 		});
 
@@ -445,7 +445,11 @@ function handleDispose(): void {
 		const pid = ptyProcess.pid;
 
 		// Kill the process tree - fire and forget since we're exiting
-		treeKill(pid, "SIGKILL");
+		treeKill(pid, "SIGKILL", (err) => {
+			if (err) {
+				console.error("[pty-subprocess] Failed to kill process tree:", err);
+			}
+		});
 		ptyProcess = null;
 	}
 


Kitenite:
I think this actually happens inconsistently which points to a race condition or we're not sending the proper kill signal. perhaps tree-kill would be a cleaner solution

chasemcdo:
> I think this actually happens inconsistently which points to a race condition or we're not sending the proper kill signal. perhaps tree-kill would be a cleaner solution

from my testing it _seems_ to happen every time, but perhaps I am just very unlucky haha

chasemcdo:
@Kitenite though you are probably right that we should use tree-kill. Just updated and for the issues I have reproduced this seems to be equivalently functional but definitely cleaner

PR #984, opened 2026-01-27
feat(desktop): add GitHub PR shortcut (⌘P)

chasemcdo:
## Summary

- Adds a new keyboard shortcut `⌘⇧R` to quickly open or create a GitHub PR from the current workspace
- Inspired by GitHub Desktop's convenient PR shortcut - wanted the same quick access here
  - GitHub Desktop does this via `⌘R` but that is presently taken by the refresh command

## Behavior

- If a PR already exists for the current branch → opens it in the browser
- If no PR exists → creates one (pushing the branch if needed) and opens GitHub's compare page

## Test plan

- [x] Open a workspace with an existing PR, press `⌘⇧R` → should open the PR in browser
- [x] Open a workspace without a PR, press `⌘⇧R` → should create PR and open GitHub compare page
- [x] Verify hotkey appears in keyboard shortcuts panel under "Workspace" category

## Summary by CodeRabbit

* **New Features**
  * Added a keyboard shortcut (Cmd/Ctrl+Shift+P) to open or create the GitHub pull request for the current workspace; success and error notifications are shown.
  * Added a dedicated PR button in the commit UI to create/view PRs and copy the PR URL.
* **UX Improvements**
  * PR creation now integrates with the push flow so PRs are created after pushing when needed.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/Sidebar/ChangesView/components/CommitInput/CommitInput.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/Sidebar/ChangesView/components/CommitInput/CommitInput.tsx
@@ -18,8 +18,15 @@ import {
 	HiArrowUp,
 	HiCheck,
 	HiChevronDown,
+	HiPlus,
 } from "react-icons/hi2";
+import { LuGitPullRequest } from "react-icons/lu";
 import { electronTrpc } from "renderer/lib/electron-trpc";
+import {
+	formatHotkeyText,
+	getCurrentPlatform,
+	getHotkey,
+} from "shared/hotkeys";
 
 interface CommitInputProps {
 	worktreePath: string;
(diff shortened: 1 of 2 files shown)


chasemcdo:
> CMD+Shift+R is the hard refresh shortcut for chrome (good for some debugging). Is there any other command you'd prefer?

Fair enough - could move it to something like (⌘P for PR). Technically this would be print, but afaik there wont ever really be a _print_ in this app. And really we are printing to a PR :P

Kitenite:
My concern is it's not a very common action. 

Do you think it might be better if we ship a similar button to the PR button in Conductor instead?

chasemcdo:
> My concern is it's not a very common action.
> 
> Do you think it might be better if we ship a similar button to the PR button in Conductor instead?

As in like [this](https://docs.conductor.build/workflow#create-pr-+-merge) so ⌘⇧P with the below button updated?

Pre-superset days I would use GitHub Desktop as it had some nice built in shortcuts for merging, pushing, PRing - only one missing right now is this shortcut. ⌘⇧P is fine by me if you want to make it less prominent

Converting this button into a general create PR action would also be pretty sweet if image

In GH Desktop this shortcuts serves to view your PR (or create it if it doesn't presently exist)

chasemcdo:
Just downloaded Conductor to see how that button worked and it is doing way too much LOL

PR #1047, opened 2026-01-29
fix (desktop): first pass auth flash fix

AviPeltz:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Summary by CodeRabbit

* **New Features**
  * Offline detection and an "Offline" indicator in the top bar.
  * Offline UI with a Retry action when a local token exists but no network.
  * Non-blocking background session validation; UI renders promptly while validation runs.
  * Temporary full-screen loading when a local token is present and session validation is pending.

* **Bug Fixes**
  * Authentication works offline when a stored token is still valid.
  * Session validation failures no longer block rendering; improved sign-in navigation when no local token.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -11,7 +11,7 @@
 		"typecheck": "tsc --noEmit"
 	},
 	"dependencies": {
-		"@anthropic-ai/sdk": "^0.71.2",
+		"@anthropic-ai/sdk": "^0.72.1",
 		"@electric-sql/client": "https://pkg.pr.new/@electric-sql/client@3724",
 		"@linear/sdk": "^68.1.0",
 		"@modelcontextprotocol/sdk": "^1.25.3",
--- a/apps/desktop/src/renderer/index.tsx
+++ b/apps/desktop/src/renderer/index.tsx
@@ -41,8 +41,9 @@ const handleDeepLink = (path: string) => {
 	console.log("[deep-link] Navigating to:", path);
 	router.navigate({ to: path });
 };
-if (window.ipcRenderer?.on && window.ipcRenderer?.off) {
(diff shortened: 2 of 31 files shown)


apps/desktop/src/renderer/routes/_authenticated/layout.tsx
```diff
@@ -66,6 +69,21 @@ function AuthenticatedLayout() {
...
+		return (
+			<div className="flex h-screen w-screen items-center justify-center bg-background">
+				<Spinner className="size-8" />
+			</div>
+		);
+	}
+
+	// If session is loading but no local token, redirect to sign-in
+	if (isPending && !hasLocalToken) {
+		return <Navigate to="/sign-in" replace />;
```
saddlepaddle: Can merge this and below

apps/desktop/src/renderer/providers/AuthProvider/AuthProvider.tsx
```diff
@@ -29,10 +29,18 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
-		// If token exists, set it in memory and refetch session
+		// If token exists and isn't expired locally, set it and validate in background
 		if (storedToken?.token && storedToken?.expiresAt) {
-			setAuthToken(storedToken.token);
-			refetchSession();
+			const isExpired = new Date(storedToken.expiresAt) < new Date();
+
+			if (!isExpired) {
+				setAuthToken(storedToken.token);
+				// Validate session in background - don't block UI on network
```
saddlepaddle: Would be good to remove logs / comments if they're excessive, not sure which should stay tho

PR #1064, opened 2026-01-30
fix: prevent postinstall infinite loop on bun install

### LINKED ISSUE #1063: Build from source: postinstall spawns infinite electron-builder processes
## Problem

When building from source, running `bun install` triggers the postinstall script which spawns hundreds of `electron-builder install-app-deps` processes in an infinite loop, never completing.

After ~15-20 minutes, the system has 370+ zombie `electron-builder` processes and the install never finishes.

## Environment

- macOS 24.6.0 (Darwin)
- Bun 1.2.12
- Apple Silicon (arm64)

## Root Cause

The postinstall script in `package.json`:
```json
"postinstall": "sherif; bun run --filter=@superset/desktop install:deps"
```

The `install:deps` script runs `electron-builder install-app-deps`, which appears to recursively trigger more installs in some edge case with the bun workspace configuration.

## Workaround

```bash
# Skip the broken postinstall
bun install --ignore-scripts

# Manually rebuild native modules
cd apps/desktop
npm rebuild better-sqlite3 node-pty --build-from-source

# Symlink them so the copy script finds them
mkdir -p node_modules
ln -sf ../../../node_modules/better-sqlite3 node_modules/
ln -sf ../../../node_modules/node-pty node_modules/
ln -sf ../../../node_modules/bindings node_modules/
ln -sf ../../../node_modules/file-uri-to-path node_modules/

# Build
cd ../..
SKIP_ENV_VALIDATION=1 bun run build
```

## Suggested Fix

A few options:
1. Add a guard in the postinstall to prevent recursive invocation
2. Move native module compilation to the build step rather than postinstall
3. Check if `electron-builder install-app-deps` is being called from within itself before spawning

Happy to submit a PR if you point me in the right direction.

spanishflu-est1918:
## Summary

- Fixes the infinite process spawn loop during `bun install`
- Adds environment variable guard to prevent recursive postinstall invocations

## Problem

When running `bun install`, the postinstall script spawns 370+ `electron-builder install-app-deps` processes in an infinite loop, causing the install to never complete.

**Root cause:** `electron-builder install-app-deps` triggers nested bun installs which re-run postinstall, creating a recursive loop.

## Solution

Created a `scripts/postinstall.sh` wrapper that:
1. Checks for `SUPERSET_POSTINSTALL_RUNNING` environment variable
2. Exits early if already in a postinstall run
3. Sets the guard variable before running the actual postinstall commands

## Test plan

- [ ] Run `bun install` from a clean clone - should complete without hanging
- [ ] Run `bun run build` - should build the desktop app successfully
- [ ] Verify native modules (better-sqlite3, node-pty) are properly compiled

Fixes #1063

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed postinstall script to prevent recursive execution, ensuring installation runs exactly once and completes reliably.

* **Chores**
  * Reorganized installation script structure for improved maintainability and clarity during the package setup process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
 		"format:check": "biome format .",
 		"typecheck": "turbo typecheck",
 		"ui-add": "turbo run ui-add",
-		"postinstall": "sherif; bun run --filter=@superset/desktop install:deps",
+		"postinstall": "./scripts/postinstall.sh",
 		"clean": "git clean -xdf node_modules",
 		"clean:workspaces": "turbo clean",
 		"release:desktop": "./apps/desktop/create-release.sh",
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Prevent infinite recursion during postinstall
+# electron-builder install-app-deps can trigger nested bun installs
+# which would re-run postinstall, spawning hundreds of processes
+
+if [ -n "$SUPERSET_POSTINSTALL_RUNNING" ]; then
+  exit 0
+fi
+
+export SUPERSET_POSTINSTALL_RUNNING=1
+
+# Run sherif for workspace validation
+sherif
+
+# Install native dependencies for desktop app
+bun run --filter=@superset/desktop install:deps


Kitenite (approved):
I can't consistently reproduce it but since this is a harmless/reversible change that doesn't affect behavior I can merge. Thanks @spanishflu-est1918

PR #1198, opened 2026-02-04
fix(desktop): sync workspace branch names and hover card branch display

ascrazy:
## Description

Syncs workspace branch names from `git status` to avoid stale branch labels after external `git switch`, and fixes hover card to show the actual branch (separate from worktree name).

## Type of Change

- [x] Bug fix
- [-] New feature
- [-] Documentation
- [-] Refactor
- [-] Other (please describe):

## Testing

- `cd apps/desktop && SKIP_ENV_VALIDATION=1 bun run dev`
- Add a workspace and create a worktree
- Open a worktree and a terminal tab inside it 
- `git switch -c test-new-branch` in the terminal tab
- Hover on the worktree in the sidebar
- Branch name on the worktree item in sidebar should update
- Worktree hover card should show correct branch name
 
| Before | After |
| --- | --- |
| ezgif-4747c1f8969e6b6f | ScreenRecording2026-02-04at21 10 20-ezgif com-video-to-gif-converter | 

## Additional Notes

- Hover card now uses `branchName` from `getWorktreeInfo`; `worktreeName` remains the path segment for compatibility.
- Branch updates are persisted to local DB on `changes.getStatus`.

## Summary by CodeRabbit

* **New Features**
  * Automatic branch synchronization keeps workspace branch state aligned with the active Git branch.
  * Workspace info now surfaces branch name; hover cards and views display the branch.

* **Bug Fixes / Stability**
  * UI now refreshes related views and clears stale data when branches diverge to prevent inconsistent displays.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/changes/status.ts
+++ b/apps/desktop/src/lib/trpc/routers/changes/status.ts
@@ -1,5 +1,5 @@
 import { projects, workspaces, worktrees } from "@superset/local-db";
-import { and, eq, isNull } from "drizzle-orm";
+import { and, eq, isNull, not } from "drizzle-orm";
 import { localDb } from "main/lib/local-db";
 import type { ChangedFile, GitChangesStatus } from "shared/changes-types";
 import simpleGit from "simple-git";
@@ -118,53 +118,40 @@ function syncWorkspaceBranch({
 
 	try {
 		const worktree = localDb
-			.select()
+			.select({ id: worktrees.id })
 			.from(worktrees)
 			.where(eq(worktrees.path, worktreePath))
 			.get();
 
 		if (worktree) {
-			if (worktree.branch !== currentBranch) {
-				localDb
-					.update(worktrees)
-					.set({ branch: currentBranch })
-					.where(eq(worktrees.id, worktree.id))
-					.run();
-			}
+			localDb
+				.update(worktrees)
+				.set({ branch: currentBranch })
+				.where(
+					and(
+						eq(worktrees.id, worktree.id),
+						not(eq(worktrees.branch, currentBranch)),
+					),
+				)
+				.run();
 
-			const workspacesForWorktree = localDb
-				.select({ branch: workspaces.branch })
-				.from(workspaces)
+			localDb
+				.update(workspaces)
+				.set({ branch: currentBranch })
 				.where(
 					and(
 						eq(workspaces.worktreeId, worktree.id),
 						isNull(workspaces.deletingAt),
+						not(eq(workspaces.branch, currentBranch)),
 					),
 				)
-				.all();
-
-			const hasWorkspaceMismatch = workspacesForWorktree.some(
-				(ws) => ws.branch !== currentBranch,
-			);
-
-			if (hasWorkspaceMismatch) {
-				localDb
-					.update(workspaces)
-					.set({ branch: currentBranch })
-					.where(
-						and(
-							eq(workspaces.worktreeId, worktree.id),
-							isNull(workspaces.deletingAt),
-						),
-					)
-					.run();
-			}
+				.run();
 
 			return;
 		}
 
 		const project = localDb
-			.select()
+			.select({ id: projects.id })
 			.from(projects)
 			.where(eq(projects.mainRepoPath, worktreePath))
 			.get();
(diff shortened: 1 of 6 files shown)


apps/desktop/src/renderer/screens/main/components/WorkspaceView/RightSidebar/ChangesView/ChangesView.tsx
```diff
@@ -69,6 +70,17 @@ export function ChangesView({ onFileOpen, isExpandedView }: ChangesViewProps) {
...
 
+	useEffect(() => {
+		if (!workspace || !status?.branch || status.branch === "HEAD") return;
+		if (workspace.branch !== status.branch) {
+			utils.workspaces.getAllGrouped.invalidate();
+			utils.workspaces.get.invalidate({ id: workspace.id });
+			utils.workspaces.getWorktreeInfo.invalidate({
+				workspaceId: workspace.id,
+			});
+		}
```
saddlepaddle: **Medium: Duplicate invalidation logic + over-broad dependency**

Two issues here:

**1. Duplicate logic (Medium):** This `useEffect` is nearly identical to the one in `WorkspaceListItem.tsx:138-147`. Both detect branch mismatches and invalidate the same set of queries. Consider extracting a shared hook:

```typescript
// e.g., hooks/useBranchSyncInvalidation.ts
function useBranchSyncInvalidation({
  gitBranch,
  workspaceBranch,
  workspaceId,
}: {
  gitBranch: string | undefined;
  workspaceBranch: string | undefined;
  workspaceId: string;
}) {
  const utils = electronTrpc.useUtils();
  useEffect(() => {
    if (!gitBranch || gitBranch === "HEAD" || !workspaceBranch) return;
    if (gitBranch !== workspaceBranch) {
      utils.workspaces.getAllGrouped.invalidate();
      utils.workspaces.get.invalidate({ id: workspaceId });
      utils.workspaces.getWorktreeInfo.invalidate({ workspaceId });
    }
  }, [gitBranch, workspaceBranch, workspaceId, utils]);
}
```

**2. Over-broad dependency (Low):** Using the full `workspace` object as a dependency means this effect re-evaluates on any workspace property change (e.g., `updatedAt`, `lastOpenedAt`). The guards prevent unnecessary invalidations, but narrowing prevents unnecessary effect executions:

```diff
-}, [status?.branch, utils, workspace]);
+}, [status?.branch, utils, workspace?.id, workspace?.branch]);
```

apps/desktop/src/lib/trpc/routers/changes/status.ts
```diff
@@ -94,6 +101,111 @@ export const createStatusRouter = () => {
...
+					and(
+						eq(workspaces.worktreeId, worktree.id),
+						isNull(workspaces.deletingAt),
+					),
+				)
+				.all();
+
+			const hasWorkspaceMismatch = workspacesForWorktree.some(
+				(ws) => ws.branch !== currentBranch,
+			);
```
saddlepaddle: **Nit: Redundant read-before-update pattern**

The function reads workspaces to check for mismatches, then runs the same `WHERE` clause again in the `UPDATE`. Since the `UPDATE` is a no-op when there are no matching rows, you could skip the read step:

```typescript
// Instead of read-check-update, just update directly:
localDb
  .update(workspaces)
  .set({ branch: currentBranch })
  .where(
    and(
      eq(workspaces.worktreeId, worktree.id),
      isNull(workspaces.deletingAt),
      not(eq(workspaces.branch, currentBranch)),
    ),
  )
  .run();
```

This avoids the extra `SELECT` on every status poll. The same pattern applies to the project-level branch workspace update below (lines 178-202).

Not blocking — the current approach is clear and correct, this is just a performance nit since `getStatus` is polled every 2.5s.

apps/desktop/src/renderer/screens/main/components/WorkspaceSidebar/WorkspaceListItem/WorkspaceListItem.tsx
```diff
@@ -136,6 +136,16 @@ export function WorkspaceListItem({
...
 
+	useEffect(() => {
+		const latestBranch = localChanges?.branch;
+		if (!latestBranch || latestBranch === "HEAD") return;
+		if (latestBranch !== branch) {
+			utils.workspaces.getAllGrouped.invalidate();
+			utils.workspaces.get.invalidate({ id });
+			utils.workspaces.getWorktreeInfo.invalidate({ workspaceId: id });
+		}
+	}, [localChanges?.branch, branch, id, utils]);
```
saddlepaddle: **Note:** This effect only fires after hover (because `localChanges` depends on `hasHovered`), so it handles the case where the user externally switches branches in a *non-active* workspace. For the *active* workspace, `ChangesView.tsx` handles it via its 2.5s polling.

See inline comment on `ChangesView.tsx` regarding extracting a shared hook to deduplicate this logic.

saddlepaddle (commented):
## Code Review: fix(desktop): sync workspace branch names and hover card branch display

### Summary

The PR addresses a real UX bug — stale branch labels after external `git switch`. The approach is sound: `syncWorkspaceBranch` runs synchronously inside the `getStatus` tRPC query to update the local DB, and React effects on the client detect the mismatch to trigger cache invalidation.

Overall this is clean and well-structured. A few items to address below.

---

### Key Observations

**`syncWorkspaceBranch` is synchronous, not fire-and-forget.** Since better-sqlite3 operations (`.get()`, `.run()`, `.all()`) are synchronous, the function executes fully before the subsequent `Promise.all`. This is correct — no `await` is needed and there's no race condition with the response.

**Detached HEAD guard is good.** The early return for `currentBranch === "HEAD"` correctly avoids clobbering stored branch names during detached HEAD states.

**Error handling is appropriate.** The try/catch with `console.warn` ensures a failed sync never breaks the status query. Correct tradeoff for a best-effort sync.

---

### Issues

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Medium | `WorkspaceListItem.tsx`, `ChangesView.tsx` | Duplicate cache invalidation logic — extract to shared hook |
| 2 | Low | `ChangesView.tsx:81` | Over-broad `workspace` object in useEffect dependency array |
| 3 | Nit | `status.ts` | Read-before-update pattern queries workspaces twice (once to check, once to update) — can be simplified |

See inline comments for details.

PR #1201, opened 2026-02-04
fix(desktop): show branch subtitle for main workspace list item when branch name differs

ascrazy:
## Description

Show the branch subtitle for the **main (branch) workspace menu item** only when the workspace name differs from the branch, so we avoid duplicate labels while still exposing the current branch when a custom name is used.

## Type of Change

- [x] Bug fix
- [-] New feature
- [-] Documentation
- [-] Refactor
- [-] Other (please describe):

| Before | After |
| --- | --- |
| Screenshot 2026-02-04 at 21 28 40 | Screenshot 2026-02-04 at 21 25 19 | 

## Release Notes

* **Bug Fixes**
  * Improved branch name display logic in workspace sidebar to more accurately show branch information in tooltips when branch names differ from branch labels.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceSidebar/WorkspaceListItem/WorkspaceListItem.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceSidebar/WorkspaceListItem/WorkspaceListItem.tsx
@@ -289,7 +289,7 @@ export function WorkspaceListItem({
 	const showDiffStats = !!diffStats;
 
 	// Determine if we should show the branch subtitle
-	const showBranchSubtitle = !isBranchWorkspace;
+	const showBranchSubtitle = !isBranchWorkspace || (!!name && name !== branch);
 
 	// Collapsed sidebar: show just the icon with hover card (worktree) or tooltip (branch)
 	if (isCollapsed) {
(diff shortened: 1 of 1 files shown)


ascrazy:
This addresses a big onboarding pain for folks adopting Superset on an existing repo.

**The problem** In my case I added the repo when the main copy was on a feature branch, so the main workspace is now forever stuck named after that branch. Even though I have long ago moved it to main branch

Is it intentional that the *main workspace* item can’t be renamed?  

**Possible direction:** I think the best long‑term fix is to allow renaming the main workspace (like worktrees). 

Also, having visibility into the *actual* branch (what this PR suggests) is valuable — I’m not sure though if it's a good idea to show it conditionally, and only have this condition for main worktree item. It feels like it it adds a lot of small inconsistencies.

saddlepaddle:
> This addresses a big onboarding pain for folks adopting Superset on an existing repo.
> 
> **The problem** In my case I added the repo when the main copy was on a feature branch, so the main workspace is now forever stuck named after that branch. Even though I have long ago moved it to main branch
> 
> Is it intentional that the _main workspace_ item can’t be renamed?
> 
> **Possible direction:** I think the best long‑term fix is to allow renaming the main workspace (like worktrees).
> 
> Also, having visibility into the _actual_ branch (what this PR suggests) is valuable — I’m not sure though if it's a good idea to show it conditionally, and only have this condition for main worktree item. It feels like it it adds a lot of small inconsistencies.

Honestly I feel like it's in a weird spot in general, I kinda like the Conductor approach of hiding the concept of the "main branch" entirely and just treating everything as a worktree. I think your approach is reasonable as a short-term fix though!

ascrazy:
>hiding the concept of the "main branch" entirely

Please no! :)

I like Superset specifically for the fact it feels lower level in such things, gives me more control, etc...

PR #1213, opened 2026-02-05
fix: nested hidden folders not showing when 'Show Hidden Files' enabled

### LINKED ISSUE #1193: [bug] nested hidden folders do not show even when "Show Hidden Files" is enabled
There is the toggle in the File Browser to "Show Hidden Files".

This will toggle the display of .hiddenfolders and .hiddenfiles perfectly ... at the top level.

But if I have:

```
/foldername
    /foldername
        /.hiddenfoldername
```

And "Show Hidden Files" is enabled, the nested `.hiddenfoldername` will not show.

I expect even the nested hidden folders and files to show when "Show Hidden Files" is enabled.

michaelgold3n:
## Problem

Fixes #1193

When toggling 'Show Hidden Files' in the file browser, hidden folders at the root level would correctly appear, but nested hidden folders inside non-hidden directories would not show.

## Root Cause

The `useFileTree` hook caches children of expanded folders in `childrenCache`. When `showHiddenFiles` changed, the root directory query correctly re-fetched with `includeHidden: true`, but the cached children of expanded folders were not invalidated. This meant nested directories kept using stale cached data that was fetched without hidden files.

## Solution

Added a `useEffect` that clears `childrenCache` whenever `showHiddenFiles` changes. This ensures all nested directories will be re-fetched with the correct `includeHidden` setting when the toggle is changed.

## Testing

- Toggle 'Show Hidden Files' now correctly shows/hides hidden folders at all nesting levels

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where nested hidden files and folders would not display correctly when toggling the show hidden files setting. The file tree now properly refreshes when this preference changes.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/RightSidebar/FilesView/hooks/useFileTree.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/RightSidebar/FilesView/hooks/useFileTree.ts
@@ -56,6 +56,12 @@ export function useFileTree({
 		}));
 	}, [rootEntries]);
 
+	// Clear children cache when showHiddenFiles changes to ensure nested
+	// hidden files/folders are properly shown or hidden
+	useEffect(() => {
+		setChildrenCache({});
+	}, [showHiddenFiles]);
+
 	const buildTree = useCallback(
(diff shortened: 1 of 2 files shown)


michaelgold3n:
Closing - submitted in error. Apologies!

PR #1246, opened 2026-02-06
fix(desktop): false positive LFS detection from .lfsconfig

### LINKED ISSUE #1244: [bug] Creating workspace fails with Git LFS error for repos that don't use LFS
#### Describe the bug

When adding a new workspace for a repository that has a `.lfsconfig` file (but doesn't actually track any files with LFS), the app reports:

> Failed to create worktree: This repository uses Git LFS, but git-lfs was not found or failed. Please install git-lfs (e.g., 'brew install git-lfs') and run 'git lfs install'.

The superset.sh repo itself triggers this - it has a `.lfsconfig` with `locksverify = false` but no LFS-tracked files.

Installing git-lfs is a workaround, but the error shouldn't occur since no files are tracked with LFS.

Here is a screenshot of trying to create a new workspace using the superset repo:

image

de1mat:
## Description

The `repoUsesLfs()` function checks `.lfsconfig` for the string `[lfs]` and treats it as evidence that the repo tracks files with LFS. However, `.lfsconfig` is a configuration file for LFS behaviour (server URLs, lock settings) - the `[lfs]` string is just an INI section header present in every `.lfsconfig` file.

The superset.sh repo itself has a `.lfsconfig` with only `locksverify = false`, which triggers a false positive. This blocks workspace creation with a "please install git-lfs" error even though no files are tracked with LFS.

Here is a screenshot of trying to create a new workspace using the superset repo:

image

This PR removes `.lfsconfig` from the attribute file check. The remaining detection methods are sufficient:
- `.git/lfs/` directory (LFS objects present)
- `.gitattributes` with `filter=lfs` (files configured for LFS tracking)
- `.git/info/attributes` with `filter=lfs`
- `git check-attr filter` on sample files

## Related Issues

Fixes #1244

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

- Updated the "detects LFS via .lfsconfig" test to verify that `.lfsconfig` alone does **not** trigger LFS detection
- All 12 existing tests pass

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/workspaces/utils/git.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/workspaces/utils/git.test.ts
@@ -71,17 +71,17 @@ describe("LFS Detection", () => {
 		expect(content.includes("filter=lfs")).toBe(true);
 	});
 
-	test("detects LFS via .lfsconfig", async () => {
+	test("does not detect LFS from .lfsconfig alone", async () => {
 		const repoPath = createTestRepo("lfs-config-test");
 
-		// Create .lfsconfig
+		// .lfsconfig configures LFS behaviour but does not indicate file tracking
 		writeFileSync(
 			join(repoPath, ".lfsconfig"),
-			"[lfs]\n\turl = https://example.com/lfs\n",
+			"[lfs]\n\tlocksverify = false\n",
 		);
 
-		const content = await Bun.file(join(repoPath, ".lfsconfig")).text();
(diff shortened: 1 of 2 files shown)


de1mat:
@Kitenite found this bug today when I went to try and create a new workspace in my superset repo, and it would not create because of the above error.

Kitenite:
My bad claude code decided to make those comments. Don't worry, i actually review PRs haha

Kitenite:
Thanks @de1mat !

PR #1257, opened 2026-02-06
fix(desktop): auto-scroll sidebar to active workspace on keyboard navigation

kkjcheng:
## Summary
- When navigating workspaces via keyboard shortcuts (Cmd+1-9, Cmd+Alt+Up/Down), the active workspace could end up off-screen in the sidebar
- Adds `scrollIntoView({ block: "nearest", behavior: "smooth" })` to `WorkspaceListItem` when it becomes active
- `block: "nearest"` ensures no-op when already visible, and minimal scroll distance otherwise
- Works in both expanded and collapsed sidebar modes

## Test plan
- [x] Open app with enough workspaces to require scrolling in the sidebar
- [x] Use Cmd+1 through Cmd+9 to jump — sidebar should scroll to show the target
- [x] Use Cmd+Alt+Up/Down to cycle — sidebar should follow
- [x] Manually click a visible workspace — no jarring scroll should occur
- [x] Test in collapsed sidebar mode as well

## Release Notes

* **New Features**
  * Workspace sidebar items now automatically scroll into view when activated, improving navigation and visibility.


Kitenite:
Awesome thanks for both PRs @kkjcheng !

PR #1286, opened 2026-02-07
fix(desktop): simplify chat re-render fix

Kitenite:
## Summary
- Removes the `startSessionRef` / `stopSessionRef` ref workaround from #1283 and replaces it with a simple `eslint-disable` on the dependency array
- tRPC mutation objects are stable transports (we only call `.mutate()`), so they don't need to be reactive deps — an eslint suppression is clearer than 4 lines of ref boilerplate
- Keeps the `sessionReady` state improvement from #1283

## Test plan
- [ ] Open the chat pane — should load without crashing (no infinite re-render)
- [ ] Switch sessions/workspaces and confirm chat reconnects correctly

## Summary by CodeRabbit

No user-visible changes in this release. This update includes internal code optimizations to improve maintainability.


Kitenite:
Closing — the ref-based approach in #1283 is actually the better pattern. An eslint-disable suppresses all dep warnings for the effect, which can mask real bugs if someone later adds reactive values inside it.

PR #1525, opened 2026-02-16
feat(desktop): implement session refetch subscription and update billing success URLs

Ipriyankrajai:
- Added a new subscription for session refetch requests in the auth router to handle session updates after specific events, such as Stripe checkout.
- Emitted a session refetch event upon successful billing deep link navigation.
- Updated success URLs in billing components to use a protocol scheme constant for consistency and improved link handling.
- Refactored the usePaywall hook to simplify user plan determination based on session data.

## Summary by CodeRabbit

* **New Features**
  * Added a session-refetch subscription so the app can refresh session data in response to external events.
* **Improvements**
  * Billing success deep links now trigger an immediate session refresh, reflecting subscription and account updates without manual refresh.
  * Billing upgrade redirects use the app protocol scheme for more reliable in-app navigation.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/settings/billing/components/BillingOverview/BillingOverview.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/settings/billing/components/BillingOverview/BillingOverview.tsx
@@ -7,6 +7,7 @@ import { HiArrowRight } from "react-icons/hi2";
 import { env } from "renderer/env.renderer";
 import { authClient } from "renderer/lib/auth-client";
 import { useCollections } from "renderer/routes/_authenticated/providers/CollectionsProvider";
+import { PROTOCOL_SCHEME } from "shared/constants";
 import {
 	isItemVisible,
 	SETTING_ITEM_ID,
@@ -67,7 +68,7 @@ export function BillingOverview({ visibleItems }: BillingOverviewProps) {
 					referenceId: activeOrgId,
 					annual,
 					seats: memberCount,
-					successUrl: `${env.NEXT_PUBLIC_WEB_URL}/settings/billing?success=true`,
+					successUrl: `${PROTOCOL_SCHEME}://settings/billing?success=true`,
 					cancelUrl: env.NEXT_PUBLIC_WEB_URL,
 					disableRedirect: true,
 				},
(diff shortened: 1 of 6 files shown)


Kitenite:
I've pushed the fix for the unstable `collections` dependency to `origin/pri/ss-5-update-side-bar-task-state`. The change extracts `collections.subscriptions` into a stable local reference so `useLiveQuery` doesn't re-execute on every render:

```diff
 	const collections = useCollections();
+	const subscriptions = collections.subscriptions;

 	const { data: subscriptionsData } = useLiveQuery(
-		(q) => q.from({ subscriptions: collections.subscriptions }),
-		[collections],
+		(q) => q.from({ subscriptions }),
+		[subscriptions],
 	);
```

You can cherry-pick the commit: `ba1f113ec`

Kitenite:
we should probably refetch the state session instead of using sessions imo. probably when we come back from stripe with deeplink we should just refetch the session state. 

real-time collections can be flaky atm

Ipriyankrajai:
@Kitenite Made the changes and used deeplink approach

apps/desktop/src/renderer/routes/_authenticated/settings/billing/components/BillingOverview/BillingOverview.tsx
```diff
@@ -67,7 +68,7 @@ export function BillingOverview({ visibleItems }: BillingOverviewProps) {
 					referenceId: activeOrgId,
 					annual,
 					seats: memberCount,
-					successUrl: `${env.NEXT_PUBLIC_WEB_URL}/settings/billing?success=true`,
+					successUrl: `${PROTOCOL_SCHEME}://settings/billing?success=true`,
```
Kitenite, line 71: What's the reason we're changing this?

Ipriyankrajai: We want it to redirect it to our app instead of the web

Ipriyankrajai: Where we will referch the session

PR #1563, opened 2026-02-18
fix(desktop): suppress reparenting navigation events to preserve browser query params

Kitenite:
## Summary
- When a `<webview>` is reparented between DOM containers (hidden ↔ visible during tab/workspace switches), Electron can fire spurious navigation events that overwrite the store with a stale URL (missing query params) and corrupt the browser history
- Added an `isReparenting` guard that suppresses `did-start-loading`, `did-stop-loading`, `did-navigate`, and `did-navigate-in-page` events for 500ms after reclaiming a webview from the hidden container
- `syncStoreFromWebview` still runs synchronously before any async events fire, so legitimate agent-driven navigation while hidden is still picked up

## Test plan
- [ ] Navigate to a page with query params (e.g. `localhost:3000/page?foo=bar`)
- [ ] Switch to a different tab, then switch back — verify query params preserved in URL bar
- [ ] Switch workspaces and back — verify query params preserved
- [ ] Navigate through several pages, switch tabs and back — verify back/forward history still works
- [ ] Have an agent navigate the browser while on a different tab — verify the new URL is picked up when switching back

## Summary by CodeRabbit

* **Bug Fixes**
  * Webview no longer shows stale or incorrect addresses when moved between view areas.
  * Suppresses stray navigation/loading events during view moves to keep URL display and history consistent.

* **Refactor**
  * Internal lifecycle and event handling streamlined to improve reliability when views are relocated.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/TabView/BrowserPane/hooks/usePersistentWebview/usePersistentWebview.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/TabView/BrowserPane/hooks/usePersistentWebview/usePersistentWebview.ts
@@ -2,12 +2,7 @@ import { useCallback, useEffect, useRef } from "react";
 import { electronTrpc } from "renderer/lib/electron-trpc";
 import { useTabsStore } from "renderer/stores/tabs/store";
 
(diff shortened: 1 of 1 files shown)


Kitenite:
Closing as stale: created in Jan-Mar with no recent activity. If still relevant, re-open or re-create from a fresh branch. Bulk audit 2026-05-06.

PR #1682, opened 2026-02-22
fix(api): make durable streams env vars required

Kitenite:
## Summary
- Made `DURABLE_STREAMS_URL` and `DURABLE_STREAMS_SECRET` required (non-optional) in the API env schema
- These were optional, causing silent 500 errors (`Invalid URL: 'undefined/sessions/...'`) on all chat stream routes when not configured
- The API app will now fail at startup with a clear validation error if either var is missing

## Test plan
- [ ] Verify API app starts successfully with both env vars set
- [ ] Verify API app fails at startup with a clear error if either var is missing
- [ ] Ensure deployed environments (Vercel) have both env vars configured

## Summary by CodeRabbit

* **Breaking Changes**
  * Environment variables DURABLE_STREAMS_URL and DURABLE_STREAMS_SECRET are now required at startup (previously optional).

* **Style**
  * Minor import formatting adjusted for consistency.

* **Tests**
  * Removed an unused type import from a test file (no behavioral changes).

### DIFF
--- a/apps/desktop/src/renderer/stores/tabs/utils.test.ts
+++ b/apps/desktop/src/renderer/stores/tabs/utils.test.ts
@@ -1,6 +1,5 @@
 import { describe, expect, it } from "bun:test";
 import type { MosaicNode } from "react-mosaic-component";
-import type { FileStatus } from "shared/changes-types";
 import type { Tab } from "./types";
 import {
 	buildMultiPaneLayout,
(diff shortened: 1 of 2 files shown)


Kitenite:
@saddlepaddle missing some context on the right pattern for the URL and secret in electric, leaving to your review and to make the CI pass

PR #1782, opened 2026-02-25
Improve first-terminal responsiveness (write path + runtime prewarm)

Kitenite:
## Summary by CodeRabbit

* **Performance**
  * Terminal startup now performs a background warmup to reduce first-use latency.
  * Locale detection is cached to avoid repeated system probes.

* **Refactor**
  * Terminal connection and hot-path operations reworked to improve responsiveness and error handling.
  * Background warmups are deduplicated to prevent redundant work and transient failures.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -30,7 +30,10 @@ import { loadWebviewBrowserExtension } from "./lib/extensions";
 import { localDb } from "./lib/local-db";
 import { ensureProjectIconsDir, getProjectIconPath } from "./lib/project-icons";
 import { initSentry } from "./lib/sentry";
-import { prewarmTerminalRuntime, reconcileDaemonSessions } from "./lib/terminal";
+import {
+	prewarmTerminalRuntime,
+	reconcileDaemonSessions,
+} from "./lib/terminal";
 import { disposeTray, initTray } from "./lib/tray";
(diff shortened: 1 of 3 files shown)


Kitenite:
Addressed the startup-blocking concern in a follow-up commit:

- Removed synchronous locale probing from terminal env setup (execSync)
- Locale prewarm is now async/background-only (exec + cached result)
- getLocale() is now non-blocking and returns UTF-8 fallback immediately when needed

This keeps eager prewarm behavior while avoiding main-thread startup stalls.

Kitenite:
Pulled all comments/checks and applied the addressable items:

- `prewarmInFlight` now resets to `null` in a `.finally(...)` block, so transient failures can be retried on later calls.
- Fixed current lint failures (formatting) in:
  - `apps/desktop/src/main/index.ts`
  - `apps/desktop/src/main/lib/terminal/index.ts`
  - `apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalConnection.ts`

Validation rerun locally:
- `bun run lint`
- `bun --cwd apps/desktop typecheck`
- `bun --cwd apps/desktop test src/main/lib/terminal/env.test.ts src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.test.ts`

PR #1837, opened 2026-02-26
fix(desktop): replace all placeholder instances in notify-hook template

zoumo:
## Summary
- Fix `getNotifyScriptContent()` to use `replaceAll()` instead of `replace()` for template placeholders
- The template has two `{{DEFAULT_PORT}}` instances (debug and non-debug curl branches), but `replace()` only substituted the first one
- This caused the generated `notify.sh` to fail, breaking Claude Code hooks

## Test plan
- [x] Verify both `{{DEFAULT_PORT}}` placeholders in template are replaced in generated script
- [x] Confirm hook notifications work in both debug and production modes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where notification configuration was not fully applied in certain scenarios, ensuring all necessary replacements are properly processed.


zoumo:
@Kitenite

Kitenite:
This was my bad, verified fixed. thanks @zoumo !

PR #1900, opened 2026-02-27
fix: map Option+P to Meta+P escape sequence on macOS (#1359)

### LINKED ISSUE #1359: [bug] Alt+P shortcut in Claude Code outputs π instead of triggering action
#### Describe the bug

Claude Code uses `Alt+P` as a keyboard shortcut (to switch model/effort level while the prompt has text). Inside Superset's terminal, pressing `Alt+P` (Option+P) outputs the character `π` instead of sending the escape sequence `\x1bp`.

Claude Code shows a hint: *"To enable Alt+P run /terminal-setup"*, but `/terminal-setup` cannot run inside Superset ("Terminal setup cannot be run from Superset").

#### Root cause

Likely related to #899 — Superset's xterm.js configuration for `macOptionIsMeta`. When Option is not treated as Meta, macOS produces special characters (π, ∫, etc.) instead of escape sequences that CLI tools expect.

#### Steps to reproduce

1. Open Claude Code inside Superset terminal
2. Type some text in the prompt
3. Press `Alt+P` (Option+P)
4. Character `π` is inserted instead of triggering Claude Code's model picker

#### Expected behavior

`Alt+P` should send `\x1bp` so Claude Code (and other CLI tools relying on Alt-key shortcuts) can intercept it.

#### Environment

- **OS**: macOS
- **Terminal**: Superset desktop app
- **CLI tool**: Claude Code (Anthropic)

IkramBagban:
## Description

This PR fixes an issue where the Alt+P (Option+P) keyboard shortcut on macOS was outputting the special character π instead of triggering CLI actions. Specifically, this was breaking the model picker shortcut in Claude Code.

The fix involves intercepting the physical KeyP code combined with the altKey modifier in the xterm.js custom key event handler and manually sending the expected \x1bp (Meta+P) escape sequence.

## Related Issues
Closes #1359

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Summary by CodeRabbit

* **New Features**
  * New "Use Option as Meta key" toggle in Settings (Behavior) for macOS to control Option+Letter behavior in the terminal.
* **Bug Fixes**
  * Option+Letter in the terminal can now send the expected Escape+letter sequences when the setting is enabled.
* **Tests**
  * Added tests covering macOS Option+Letter handling.
* **Chores**
  * Added migration and default setting for the new option.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/settings/index.ts
+++ b/apps/desktop/src/lib/trpc/routers/settings/index.ts
@@ -21,6 +21,7 @@ import {
 	DEFAULT_AUTO_APPLY_DEFAULT_PRESET,
 	DEFAULT_CONFIRM_ON_QUIT,
 	DEFAULT_FILE_OPEN_MODE,
+	DEFAULT_OPTION_AS_META_KEY,
 	DEFAULT_OPEN_LINKS_IN_APP,
 	DEFAULT_SHOW_PRESETS_BAR,
 	DEFAULT_SHOW_RESOURCE_MONITOR,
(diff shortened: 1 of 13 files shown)


Kitenite:
Thanks for the fix! This correctly solves the Option+P → Meta+P mapping for Claude Code's model picker, and follows the existing pattern for Option+Left/Right.

However, we think this should be a more general solution rather than a per-key mapping. Every major terminal (iTerm2, VS Code, Alacritty, Warp) handles this with a global "Use Option as Meta key" setting that maps **all** Option+letter combos to `ESC + key` escape sequences. The per-key approach doesn't scale — other CLI tools use Meta+B, Meta+D, Meta+F, Meta+N, etc., and each would need another hardcoded case here.

Would you be open to implementing a global "Option as Meta" behavior setting instead? It would:

1. Add a toggle in Settings → Behavior (like the existing macOS menu bar icon setting)
2. When enabled, intercept all Option+letter keypresses on macOS and send `\x1b` + the letter instead of the Unicode dead character
3. Fix Option+P and every other Meta+key binding in one go
4. Work across keyboard layouts (the current `event.code` approach is correct for this)

The existing Option+Left/Right mappings could stay as unconditional defaults (since those are universally expected for word navigation), but everything else would be gated behind the setting — that way users who need Option for special characters (like `ø` or `ß`) can keep the default behavior.

emilbryggare:
@Kitenite I took a stab at this in this PR: https://github.com/superset-sh/superset/pull/2280. This is what has been blocking me from using Superset & Zellij.

IkramBagban:
> Thanks for the fix! This correctly solves the Option+P → Meta+P mapping for Claude Code's model picker, and follows the existing pattern for Option+Left/Right.
> 
> However, we think this should be a more general solution rather than a per-key mapping. Every major terminal (iTerm2, VS Code, Alacritty, Warp) handles this with a global "Use Option as Meta key" setting that maps **all** Option+letter combos to `ESC + key` escape sequences. The per-key approach doesn't scale — other CLI tools use Meta+B, Meta+D, Meta+F, Meta+N, etc., and each would need another hardcoded case here.
> 
> Would you be open to implementing a global "Option as Meta" behavior setting instead? It would:
> 
> 1. Add a toggle in Settings → Behavior (like the existing macOS menu bar icon setting)
> 2. When enabled, intercept all Option+letter keypresses on macOS and send `\x1b` + the letter instead of the Unicode dead character
> 3. Fix Option+P and every other Meta+key binding in one go
> 4. Work across keyboard layouts (the current `event.code` approach is correct for this)
> 
> The existing Option+Left/Right mappings could stay as unconditional defaults (since those are universally expected for word navigation), but everything else would be gated behind the setting — that way users who need Option for special characters (like `ø` or `ß`) can keep the default behavior.

That makes sense,  a global “Option as Meta” setting would definitely be the more scalable approach instead of handling individual keys.

I’ll take a look at implementing it so that Option+letter combinations map to ESC + key when the setting is enabled, while keeping the current default behavior for users who rely on Option for special characters. The existing Option+Left/Right mappings can remain as they are.

I’ll work on this when I get some time and update the PR once I have a draft ready. Thanks for the suggestion!

sandipb:
Please oh please get this out. This blocks the usability of multiple agents.

PR #1941, opened 2026-03-02
Fix race dropping agent prompt on new workspace init

Kitenite:
## Summary
- fix a race where `agentCommand` could be dropped for newly created workspaces when pending setup was rewritten after init effects had already started processing
- add `mutateAsyncWithPendingSetup` in `useCreateWorkspace` so pending setup overrides (including `agentCommand`) are attached atomically in the initial queue write
- update New Workspace modal and Task "Open in workspace" flows to pass `agentCommand` through that atomic path and remove late store rewrites
- harden concurrent create calls: pending setup overrides are now tracked per mutation call (unique call key) rather than by input object identity, preventing collisions when the same input object is reused

## Validation
- `bunx biome check apps/desktop/src/renderer/react-query/workspaces/useCreateWorkspace.ts apps/desktop/src/renderer/components/NewWorkspaceModal/NewWorkspaceModal.tsx 'apps/desktop/src/renderer/routes/_authenticated/_dashboard/tasks/$taskId/components/PropertiesSidebar/components/OpenInWorkspace/OpenInWorkspace.tsx'`

### DIFF
--- a/apps/desktop/src/renderer/react-query/workspaces/useCreateWorkspace.ts
+++ b/apps/desktop/src/renderer/react-query/workspaces/useCreateWorkspace.ts
@@ -27,22 +27,30 @@ export function useCreateWorkspace(options?: UseCreateWorkspaceOptions) {
 		(s) => s.addPendingTerminalSetup,
 	);
 	const updateProgress = useWorkspaceInitStore((s) => s.updateProgress);
-	const pendingSetupOverridesByInput = useRef(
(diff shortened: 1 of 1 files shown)


Kitenite:
Addressed the concurrency review feedback in 72f9f1da7: pending setup overrides are now keyed per mutation call (unique call key) instead of input object identity, so overlapping calls cannot overwrite each other. Re-ran desktop typecheck + biome checks.

PR #1984, opened 2026-03-03
fix: survive transient OS errors without killing all sessions

nickita-khylkouski:
## Summary

- The terminal-host daemon's `uncaughtException` handler unconditionally calls `stopServer()` → `terminalHost.dispose()`, which kills **every** active session
- For transient OS errors like `ENOSPC` (disk full), this is catastrophic — a momentary disk-full condition destroys all running terminal sessions
- This PR makes the daemon resilient: transient errors (`ENOSPC`, `ENOMEM`, `EMFILE`, `ENFILE`) are logged as warnings while sessions stay alive
- After 50 consecutive transient errors, the daemon still shuts down as a safety valve
- Non-transient uncaught exceptions behave exactly as before

## Problem

When the disk fills up, the daemon tries to write a log line, gets `ENOSPC`, and the `uncaughtException` handler fires:

```
[ERROR] Uncaught exception {
  error: 'ENOSPC: no space left on device, write'
}
[INFO] Terminal host disposed  (×19,745 times in 0.3 seconds)
```

Every terminal session — Claude Code, Codex, shells — is killed simultaneously. The user loses all active work across all tabs.

## Changes

- Added `isTransientError()` helper that checks error codes against a known list of transient OS errors
- `uncaughtException` and `unhandledRejection` handlers now skip `stopServer()` for transient errors
- Transient errors are logged at `warn` level with a running counter
- A consecutive error threshold (50) ensures the daemon still exits if the system is truly unrecoverable

## Test plan

- [ ] Verify daemon stays alive when a write fails with ENOSPC
- [ ] Verify daemon still exits on non-transient uncaught exceptions
- [ ] Verify daemon exits after 50+ consecutive transient errors
- [ ] Verify transient error counter resets on non-transient errors

---
## Summary by cubic
Make the terminal-host daemon resilient to transient OS errors so active sessions stay alive. Transient errors are warned and counted in a 60s sliding window; a centralized handler exits only after 50 within the window, while non-transient errors behave as before.

- **Bug Fixes**
  - Treat ENOSPC, ENOMEM, EMFILE, ENFILE as transient; keep sessions alive and skip stopServer() in uncaughtException/unhandledRejection.
  - Add a 60s sliding window counter (recordTransientErrorInWindow) with warn-level logs showing count; exit after 50; tests cover window behavior.
  - Centralize signal handling in signal-handlers.ts with an isShuttingDown guard and a 10s forced-exit timer to ensure a single shutdown; harden shutdown logging so failures can’t block exit; ignore further errors once shutdown begins.
  - Fix TypeScript type narrowing in signal handlers.

<sup>Written for commit ecb1d43481aabf36e66745579a4a388a3b92bc2f. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience to transient OS resource errors, reducing unnecessary shutdowns.
  * Sessions persist through transient errors; system performs forced shutdown only after repeated failures.
  * Enhanced logging and classification to surface transient vs. fatal conditions with counts and codes.
  * Safer, centralized shutdown flow prevents concurrent shutdowns and avoids interrupting active sessions.
* **Tests**
  * Added unit tests validating transient-error windowing, boundary handling, and pruning behavior.

### DIFF
--- a/apps/desktop/src/main/terminal-host/signal-handlers.ts
+++ b/apps/desktop/src/main/terminal-host/signal-handlers.ts
@@ -42,7 +42,7 @@ const TRANSIENT_ERROR_WINDOW_SECONDS = Math.floor(
 );
 const SHUTDOWN_TIMEOUT_MS = 10_000;
 
-function isTransientError(error: unknown): error is Error {
+function isTransientError(error: unknown): boolean {
 	if (error instanceof Error) {
 		return TRANSIENT_ERROR_CODES.some(
 			(code) =>
@@ -51,6 +51,14 @@ function isTransientError(error: unknown): error is Error {
 		);
 	}
 	return false;
+}
+
+function getTransientErrorIdentifier(error: unknown): string {
+	if (error instanceof Error) {
+		const code = (error as NodeJS.ErrnoException).code;
+		return code ?? error.message.split(",")[0];
+	}
+	return "unknown";
 }
 
 export function setupTerminalHostSignalHandlers({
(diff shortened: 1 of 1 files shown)


nickita-khylkouski:
Addressed remaining review feedback in commit 0445b835:

- `fatalShutdown` now guarantees process termination even if `stopServer()` rejects (catch + forced-exit timeout + finally exit).
- Added explicit error logging when transient unhandled rejections hit `MAX_TRANSIENT_ERRORS` before shutdown.

I also resolved the remaining review thread after this update.

Kitenite:
This is great thanks @nickita-khylkouski !

PR #2238, opened 2026-03-08
fix(desktop): align terminal link underlines for CJK text

### LINKED ISSUE #2231: [bug] Underline decoration misaligned with CJK/mixed-width text in terminal
## Problem

When the terminal displays text mixing CJK (Chinese/Japanese/Korean) characters with ASCII, ANSI underline decorations are visually offset from the text they should underline. The underline appears shifted rather than aligning with the underlined content.

## Screenshot

Image

The line reads: `本地 main 已恢复到 origin/main, 改动都在 fix/hardened-error-handling-and-consistency 分支上。`

The underline under `origin/main` and the branch name is shifted to the right relative to the actual text.

## Root Cause (likely)

CJK characters are full-width (2 cells), ASCII characters are half-width (1 cell). The Unicode11 addon is loaded and `unicode.activeVersion` is set to `"11"` (`helpers.ts:290`), so xterm.js knows the correct cell widths. However, the WebGL renderer (`@xterm/addon-webgl`) may be calculating underline pixel positions incorrectly when a line contains a mix of single-width and double-width characters — the underline start/end x-coordinates drift from the actual glyph positions.

This could also be font-related: the default font stack (`config.ts:16-32`) is Latin-focused (Meslo, Hack, Fira, JetBrains Mono, Menlo). CJK characters fall through to the system fallback font, and if its actual glyph width doesn't precisely equal 2× the primary font's character width, the rendered text and the cell-grid-based underline will diverge.

## Steps to Reproduce

1. Open Superset terminal
2. Run a command that outputs underlined text with CJK characters (e.g. a git status message in Chinese locale, or Claude Code output in Chinese)
3. Observe the underline position vs the text it decorates

## Environment

- Superset version: 1.1.2
- OS: macOS
- xterm.js: 6.1.0-beta.148
- Renderer: WebGL (default)

## Possible Mitigations

1. Test if switching to DOM renderer (`localStorage.setItem("terminal-renderer", "dom")`) resolves the issue — this would confirm it's WebGL-specific
2. If WebGL-specific, may need to be reported upstream to `@xterm/addon-webgl`
3. Consider adding a CJK-optimized monospace font (e.g. `Sarasa Mono`, `Noto Sans Mono CJK`) to the font fallback chain in `config.ts` to ensure consistent character widths

tt-a1i:
## Summary
- fix terminal link range positioning to use terminal cell width instead of UTF-16 length
- apply the same cell-width mapping to both URL and file-path link providers
- add regression tests for CJK-prefixed lines so link underlines start at the correct column

## Why
Terminal link underlines could appear shifted when a line contains wide characters (e.g. Chinese) before the link/path.

Fixes #2231

## Testing
- Added unit tests in:
  - apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/url-link-provider.test.ts
  - apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/file-path-link-provider.test.ts
- Local test run was not possible in this environment (bun is unavailable).

---
## Summary by cubic
Fix terminal link underline alignment when CJK/wide characters appear before links by using terminal cell width instead of UTF-16 length. Applies to URL and file-path links, including wrapped lines, and keeps common symbols single-width to avoid regressions. Fixes #2231.

- **Bug Fixes**
  - Added a cell-width utility mirroring `@xterm/xterm` (full-width=2; combining/control=0) and keeping common symbols single-width.
  - Used cell width for range math in `FilePathLinkProvider` and multi-line links, handling prev/current/next lines and wraps.
  - Added tests for CJK prefixes and symbol prefixes in URL and file-path links.

<sup>Written for commit 49b37e12bba7520bb744c2c13e7f42cab5fd3003. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal link detection and positioning to account for variable cell widths (CJK/full-width characters and combining marks), fixing inaccurate link boundaries and placement in wrapped lines and file-path links.

* **Tests**
  * Added tests verifying link start/end calculations with wide-character prefixes and symbol-prefix edge cases to ensure correct positioning.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/cell-width.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/cell-width.ts
@@ -1,5 +1,4 @@
 const COMBINING_MARK_REGEX = /\p{Mark}/u;
-const EXTENDED_PICTOGRAPHIC_REGEX = /\p{Extended_Pictographic}/u;
 
 /**
  * Mirrors xterm's cell width behavior for common cases:
@@ -16,7 +15,7 @@ function getCodePointCellWidth(char: string, codePoint: number): number {
 		return 0;
 	}
 
-	if (isFullWidthCodePoint(codePoint) || EXTENDED_PICTOGRAPHIC_REGEX.test(char)) {
+	if (isFullWidthCodePoint(codePoint)) {
 		return 2;
 	}
 
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/cell-width.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/cell-width.ts
@@ -1,0 +1,80 @@
+const COMBINING_MARK_REGEX = /\p{Mark}/u;
+const EXTENDED_PICTOGRAPHIC_REGEX = /\p{Extended_Pictographic}/u;
+
+/**
+ * Mirrors xterm's cell width behavior for common cases:
+ * - CJK/full-width chars occupy 2 cells
+ * - combining marks occupy 0 cells
+ * - control chars occupy 0 cells
+ */
+function getCodePointCellWidth(char: string, codePoint: number): number {
+	if (codePoint <= 0x1f || (codePoint >= 0x7f && codePoint <= 0x9f)) {
+		return 0;
+	}
+
+	if (COMBINING_MARK_REGEX.test(char)) {
+		return 0;
+	}
+
+	if (isFullWidthCodePoint(codePoint) || EXTENDED_PICTOGRAPHIC_REGEX.test(char)) {
+		return 2;
+	}
+
+	return 1;
+}
+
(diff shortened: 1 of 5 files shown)


tt-a1i:
Follow-up fix shipped in commit 49b37e12 to address the width-regression concern:

- Removed the blanket Extended_Pictographic => width 2 rule from apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/cell-width.ts (this was causing symbols like ©/®/™ to shift link ranges in some environments).
- Simplified apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/file-path-link-provider.test.ts line mocks to a neutral model (removed codePoint > 0xff => 2), so tests no longer assume over-broad double-width behavior.
- Added a regression test in apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/url-link-provider.test.ts to assert symbol-prefixed text (©®™ ) keeps single-width link start math.

PR remains Draft.

tt-a1i:
### Local test verification

All link provider tests pass on the latest commit:

```
bun test v1.3.1

 87 pass
 0 fail
 705 expect() calls
Ran 87 tests across 2 files. [32.00ms]
```

byrage:
I am experiencing the same issue. When will this be reflected?

PR #2469, opened 2026-03-14
fix(desktop): gate terminal stdin on shell readiness to prevent preset command loss

Roshvan:
## Summary

- Fix terminal preset commands (e.g. `claude`) being lost when the shell is still initializing (direnv, devenv, nvm, conda, oh-my-zsh, etc.)
- Session-owned shell readiness state machine gates `write()` — buffering stdin while the shell is pending, flushing in FIFO order once ready
- `createOrAttach()` returns immediately — terminal attaches and renders as soon as PTY spawns, no blocking
- Byte-level marker state machine prevents any marker bytes from leaking to xterm, even across chunk boundaries
- Terminal protocol responses (DA, DSR) dropped during shell init to prevent escape fragments appearing as typed text

## Problem

When a preset launches a terminal and sends a command, the command gets lost because the shell hasn't finished initializing. The PTY `Spawned` IPC frame fires immediately after `pty.spawn()` — **before** the shell has sourced RC files or run hook functions.

This affects any project where shell initialization takes non-trivial time:

| Tool | What it does at startup | Typical delay |
|---|---|---|
| **direnv / devenv** | Loads `.envrc`, evaluates Nix expressions, exports env vars | 1–5s |
| **nvm** | Detects `.nvmrc`, switches Node version, rehashes PATH | 100–500ms |
| **conda** | Activates environment, modifies PATH and prompt | 200–800ms |
| **oh-my-zsh** | Loads plugins (git, node, npm, etc.), compiles completions | 200–1000ms |
| **asdf / mise** | Detects `.tool-versions`, shims PATH entries | 100–300ms |
| **starship / powerlevel10k** | Renders prompt (git status, language versions) | 50–500ms |

### Approach: Session-owned stdin gating with OSC shell-ready marker

This follows the same pattern used by major terminal emulators for shell integration:
- **VS Code** — `OSC 133` sequences from precmd/PROMPT_COMMAND
- **iTerm2** — `OSC 1337` and `OSC 133` for shell integration
- **Warp** — `OSC 9001` proprietary sequences from shell hooks

We use `OSC 777` (private-use range, won't conflict with existing protocols).

### How it works

**1. Shell wrappers** (`shell-wrappers.ts`) inject a one-shot marker hook at the end of shell init:
- **zsh**: `precmd` function in `.zlogin` — placed **last** in `precmd_functions` so it fires after direnv and all other hooks
- **bash**: `PROMPT_COMMAND` function in rcfile — **appended** (not prepended) so it runs after existing hooks. Handles both scalar and array `PROMPT_COMMAND` (Bash 5.1+). Guards against empty `PROMPT_COMMAND`.
- **fish**: `--on-event fish_prompt` handler, self-removing via `functions -e`

**2. Session** (`session.ts`) owns shell readiness with a state machine:
- **States**: `pending | ready | timed_out | unsupported`
- `write()` buffers user commands in `preReadyStdinQueue` while `pending`, passes through otherwise
- Terminal protocol responses (escape sequences starting with `\x1b`) are **dropped** during init — they are stale DA/DSR responses from shell startup queries that would appear as typed text (e.g. `?62;4;9;22c`)
- Headless emulator responses (DA queries) also dropped during init — the renderer handles queries once a client attaches
- `resolveShellReady()` transitions state, clears timeout, flushes queue in FIFO order
- Marker detected via byte-level state machine that **withholds** potential marker bytes from output until confirmed — zero leakage to xterm even across PTY data frame boundaries
- Timeout (15s) started in `spawn()`, not in terminal-host
- Shells without markers (sh, ksh) start as `unsupported` — writes pass through immediately, no timeout

**3. TerminalHost** (`terminal-host.ts`) — unchanged:
- `createOrAttach()` returns immediately after PTY spawn — no blocking wait
- Terminal attaches and renders right away (user sees direnv loading in real time)
- All stdin gating is handled by Session internally

**4. Renderer** (`launch-command.ts`) — no changes needed

### Edge cases handled

- **Unsupported shell** (sh, ksh): Starts as `unsupported`, no buffering, no timeout
- **Very slow init** (>15s): Timeout fires, flushes queue, degrades to pre-fix behavior
- **Re-attach to existing session**: Only new sessions have pending state
- **Session exits/killed before ready**: `handleSubprocessExit` → `resolveShellReady('timed_out')` → queue flushed
- **Marker split across PTY frames**: Byte-level state machine holds partial matches
- **False marker starts**: Held bytes flushed as regular output on mismatch
- **DA responses from xterm**: Dropped during pending (stale startup query responses)
- **DA responses from headless emulator**: Dropped during pending (renderer takes over)
- **Empty PROMPT_COMMAND**: Guarded in bash scalar branch
- **Bash 5.1+ array PROMPT_COMMAND**: Detected via `declare -p`, both forms handled

## Files changed

| File | Change |
|---|---|
| `apps/desktop/src/main/terminal-host/pty-subprocess-ipc.ts` | `SHELL_READY_MARKER` constant with TSDoc |
| `apps/desktop/src/main/lib/agent-setup/shell-wrappers.ts` | One-shot marker hooks for zsh, bash, fish |
| `apps/desktop/src/main/terminal-host/session.ts` | Shell readiness state machine, write gating, marker stripping, DA response dropping |
| `apps/desktop/src/main/terminal-host/session-shell-ready.test.ts` | 19 focused unit tests |

## Test coverage (19 tests)

- **Write buffering**: buffer + flush after marker, unsupported shell pass-through, DA response dropping, timeout flush
- **Marker detection**: single frame, split across frames, start/end of frame, false marker starts
- **Kill/exit**: subprocess exit before marker, session killed during init
- **Shell matrix**: zsh/bash/fish buffered; sh/ksh/dash pass-through

## Test plan

- [x] 19 unit tests passing
- [x] `bun run typecheck` passes
- [ ] Open workspace with `.envrc` + direnv — terminal renders immediately, preset command runs after prompt
- [ ] Test without direnv — command executes promptly
- [ ] Test re-attaching to existing terminal — works instantly

---

### Initial implementation
Blocked `createOrAttach()` with `waitForShellReady()` — terminal wouldn't render until shell was ready (up to 15s). Used a carry-buffer approach for marker detection.

### Review feedback from @kitenite
Identified that blocking `createOrAttach()` was the wrong boundary — the terminal should attach immediately, and only stdin delivery should wait. Recommended:
- Move readiness ownership from TerminalHost into Session
- Buffer writes in Session, not block session creation
- Use a state machine instead of a promise for readiness
- Fix bash PROMPT_COMMAND ordering (was prepending, should append)
- Replace carry-buffer with a proper byte-level state machine

### Refactor to address feedback
Rewrote the implementation:
- Removed all blocking from `createOrAttach()` — returns immediately
- Session now owns readiness via `ShellReadyState` enum + `preReadyStdinQueue`
- `write()` gates on readiness; `resolveShellReady()` flushes queue
- Byte-level marker state machine prevents any marker bytes from reaching xterm
- Bash hook appended (not prepended) with empty guard and Bash 5.1+ array support
- Timeout started in `spawn()` where it belongs

### DA response fix
Discovered that headless emulator DA responses and renderer xterm protocol responses were reaching the shell during init, causing `?62;4;9;22c` to appear as typed text. Fixed by:
- Dropping emulator responses during pending state (renderer handles queries once attached)
- Dropping all escape sequence writes (`\x1b`-prefixed) during pending state — they are stale responses to shell startup queries

### Test suite
Added 19 focused unit tests covering all cases requested in review: write buffering, marker detection (including cross-frame splits), kill/exit before readiness, unsupported shell fallback, and DA response dropping.

### DIFF
--- a/apps/desktop/src/main/lib/agent-setup/shell-wrappers.test.ts
+++ b/apps/desktop/src/main/lib/agent-setup/shell-wrappers.test.ts
@@ -839,7 +839,7 @@ export SUPERSET_WORKSPACE_PATH="/wrong/path"
 			expect(args).toEqual([
 				"-l",
 				"--init-command",
-				`set -l _superset_bin "${TEST_BIN_DIR}"; contains -- "$_superset_bin" $PATH; or set -gx PATH "$_superset_bin" $PATH`,
+				`set -l _superset_bin "${TEST_BIN_DIR}"; contains -- "$_superset_bin" $PATH; or set -gx PATH "$_superset_bin" $PATH; function _superset_shell_ready --on-event fish_prompt; printf '\\033]777;superset-shell-ready\\007'; functions -e _superset_shell_ready; end`,
 			]);
 		});
 
@@ -853,7 +853,7 @@ export SUPERSET_WORKSPACE_PATH="/wrong/path"
 			expect(args).toEqual([
 				"-l",
 				"--init-command",
-				`set -l _superset_bin "/tmp/with space/quote\\"buck\\$slash\\\\bin"; contains -- "$_superset_bin" $PATH; or set -gx PATH "$_superset_bin" $PATH`,
+				`set -l _superset_bin "/tmp/with space/quote\\"buck\\$slash\\\\bin"; contains -- "$_superset_bin" $PATH; or set -gx PATH "$_superset_bin" $PATH; function _superset_shell_ready --on-event fish_prompt; printf '\\033]777;superset-shell-ready\\007'; functions -e _superset_shell_ready; end`,
 			]);
 		});
 	});


Roshvan:
@Kitenite — Addressed all your feedback. Here's a walkthrough of the implementation and how it maps to your plan.

## The Problem

When you open a terminal in Superset and a preset tries to run a command (like `claude`), the command gets lost. The shell (zsh/bash/fish) isn't ready yet — it's still loading things like direnv, nvm, oh-my-zsh, conda, etc. The command arrives while the shell is busy initializing and gets swallowed.

## What Was Wrong With the First Approach

The initial fix blocked `createOrAttach()` — the function that creates a terminal session and connects the UI to it. It waited up to 15 seconds for the shell to be ready before returning. This meant:

- The terminal wouldn't **render at all** until the shell finished loading
- The user couldn't see direnv output, loading messages, etc.
- It felt like the app was frozen

Your review correctly identified this: **the terminal should render immediately. Only the command delivery needs to wait.**

## How It Works Now

### 1. Shell-side: One-shot ready marker

Our shell wrappers now inject a tiny invisible signal — an OSC escape sequence — that fires right before the first prompt appears:

```
\033]777;superset-shell-ready\007
```

Same pattern VS Code (OSC 133), iTerm2 (OSC 1337), and Warp (OSC 9001) use for shell integration. We use 777 to avoid conflicts.

**Why a prompt hook?** Because `precmd` (zsh) / `PROMPT_COMMAND` (bash) / `fish_prompt` (fish) fire at exactly the right moment: after ALL initialization is done (direnv, nvm, conda, everything), right before the shell shows `$` and is ready for input.

**Why last in the hook chain?** direnv and other tools also register prompt hooks. We append ours at the end so it fires AFTER everything else finishes. If we prepended it, the marker would fire before direnv was done loading.

### 2. Session-side: State machine + write gating

Instead of blocking session creation, the `Session` class now owns a readiness state machine:

```
pending → ready       (marker detected in PTY output)
pending → timed_out   (15s elapsed, no marker — degrade gracefully)
unsupported           (sh/ksh — no marker injected, writes pass through immediately)
```

**How `write()` works now:**
- Shell ready? → send directly to PTY stdin
- Shell pending + user command? → buffer it in a queue
- Shell pending + escape sequence (`\x1b...`)? → drop it (stale terminal response)

When the marker arrives (or the timeout fires), all buffered commands flush to the PTY in order.

**Why drop escape sequences?** During shell init, zsh queries terminal capabilities. The renderer's xterm responds with something like `\x1b[?62;4;9;22c`. If we buffered and flushed that alongside the preset command, the shell would interpret those bytes as typed text — that's where the `?62;4;9;22c` garbage was coming from.

### 3. Marker detection: Byte-level state machine

We scan PTY output one character at a time, matching against the marker. Bytes that *might* be part of the marker are held back. Full match → discard held bytes and resolve readiness. Mismatch → flush held bytes as regular output. Zero marker bytes ever leak to xterm, even across chunk boundaries.

### 4. Headless emulator DA responses

The headless emulator responds to terminal queries when no renderer client is attached. During init we drop these — they'd pollute the stdin queue and appear as typed text once flushed. After a client attaches the renderer handles all queries.

## Mapping to Your Plan

| Your plan item | Implementation |
|---|---|
| 1. Session-scoped readiness state + stdin queue | `ShellReadyState` type (`pending \| ready \| timed_out \| unsupported`) + `preReadyStdinQueue` |
| 2. Start timeout in `spawn()`, not `createOrAttach()` | Timeout starts at end of `spawn()` method |
| 3. Remove blocking from `createOrAttach()` | `terminal-host.ts` has **zero changes** — completely unchanged from main |
| 4. Gate `Session.write()`, flush FIFO | `write()` checks state, buffers or drops; `resolveShellReady()` flushes queue |
| 5. Fix bash hook ordering | Appended (not prepended), handles scalar and Bash 5.1+ array `PROMPT_COMMAND` |
| 6. Replace carry buffer with state machine | Character-by-character scanner with held bytes — no leakage across chunks |
| 7. Focused test coverage | 19 unit tests: write buffering, marker split, DA dropping, kill/exit, shell matrix |

## Files Changed

- **`pty-subprocess-ipc.ts`** — `SHELL_READY_MARKER` constant
- **`shell-wrappers.ts`** — marker hooks for zsh/bash/fish
- **`session.ts`** — state machine, write gating, marker scanner
- **`session-shell-ready.test.ts`** — 19 tests
- **`terminal-host.ts`** — no changes

## User Experience

**Before:** Terminal opens, hangs for up to 15s (or command lost entirely).
**After:** Terminal opens instantly, direnv output visible in real time, preset command runs cleanly after prompt appears.

Happy to adjust further — let me know what you think or if there's another way you'd like to tackle this. Happy to pivot.

Roshvan:
@kiet this also blocks anyone trying to do Cmd+N or new workspace from firing off with Claude, since it will try to spin up and break.

PR #2526, opened 2026-03-17
fix(desktop): reduce terminal typing latency

kirankunigiri:
## Summary

Currently, it is really frustrating to type in superset terminals. There is a lot of input delay and it feels sluggish. This fixes it by making typing super smooth.

- replace the daemon PTY output batching timer with a next-turn flush
- remove the fixed ~30fps output cadence that was making terminal typing feel laggy
- keep burst coalescing within the same event-loop cycle without the extra timer delay

## Testing
- bun test apps/desktop/src/main/terminal-host/daemon.test.ts apps/desktop/src/main/terminal-host/session.test.ts --timeout 20000

---
## Summary by cubic
Reduce terminal typing latency in the Desktop app by flushing PTY output on the next event-loop turn instead of a 32 ms timer. This removes the ~30 fps cadence so interactive echo feels immediate while still coalescing bursts within the same cycle.

<sup>Written for commit 099fcd526ba2ca8d0120b46d32b56d705cfdba2e. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminal output now displays with reduced latency by switching to immediate flushing, resulting in faster and more responsive output display.

### DIFF
--- a/apps/desktop/src/main/terminal-host/pty-subprocess.ts
+++ b/apps/desktop/src/main/terminal-host/pty-subprocess.ts
@@ -57,7 +57,6 @@ const INPUT_QUEUE_HARD_LIMIT_BYTES = 64 * 1024 * 1024; // 64MB
 let outputChunks: string[] = [];
 let outputBytesQueued = 0;
 let outputFlushScheduled = false;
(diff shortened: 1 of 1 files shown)


Kitenite:
Thank you as always @kirankunigiri!

PR #2632, opened 2026-03-20
feat(desktop): focus prompt textarea when switching to a chat tab

michalkopanski:
When clicking on an existing chat tab, the prompt textarea is now automatically focused so users can start typing immediately without an extra click.

N/A

# QA
- [x] Desktop web
- [ ] Mobile web
- [ ] iOS
- [ ] Android
- [ ] API
- [ ] N/A

## QA steps
1. Open the desktop app with multiple chat tabs
2. Click on a chat tab that is not currently active
3. Verify the prompt textarea is focused (cursor appears in the input) immediately after switching tabs — no extra click required
4. Confirm this does not trigger when the app first loads or when a new tab is created and switched to programmatically

N/A

# Docs

---
## Summary by cubic
Automatically focuses the chat prompt textarea when you switch to an existing chat tab in the desktop app so you can type right away. It runs only on user-initiated tab switches and only for the focused pane in split view, not on initial load or when a tab is created or activated programmatically.

<sup>Written for commit dc7ddaa9e3b4954446de91dbbc7431bfce0853eb. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat input now reliably receives focus when switching to a chat tab and the pane is active, so users can begin typing immediately without extra clicks.
  * Improved behavior when returning to an active chat tab while the pane is focused, ensuring the input field is brought into focus consistently.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/TabView/ChatPane/ChatPane.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/TabView/ChatPane/ChatPane.tsx
@@ -73,14 +73,14 @@ export function ChatPane({
 	const chatContainerRef = useRef<HTMLDivElement>(null);
 
 	useEffect(() => {
-		if (isActiveTab && !prevIsActiveTabRef.current) {
+		if (isActiveTab && !prevIsActiveTabRef.current && isFocused) {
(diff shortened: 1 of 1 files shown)


Kitenite:
This is nice, could you consider the comments about multiple panes @michalkopanski ? thank you :)

michalkopanski:
Fixed in dc7ddaa — added `isFocused` as a guard so only the focused pane claims the textarea when switching tabs.

Kitenite:
Love it thanks for the adjustments!

PR #2653, opened 2026-03-20
fix(desktop): refocus chat textarea after @mention file selection

michalkopanski:
## Summary

After selecting a file via `@mention`, the prompt textarea loses focus because Radix's popover close steals it. The original fix used `document.querySelector('[data-slot=input-group-control]')` which is broken in split-pane layouts (React Mosaic) — multiple `PromptInputProvider` instances can be mounted simultaneously, and `document.querySelector` always picks the first match in document order.

The fix exposes an instance-scoped `focus()` method on `TextInputContext`, mirroring the existing `__registerFileInput` / `openFileDialog` pattern:

- `PromptInputProvider` holds a `textareaRef`, exposes `focus()`, and provides `__registerTextarea` for the textarea to register itself
- `PromptInputTextarea` registers its ref with the provider on mount via `useEffect`
- All consumers call `textInput.focus()` — always targeting the correct pane's textarea

Also moved the refocus from `handleOpenChange` (fires on Escape/outside-click too) to `handleSelectFile` (fires only on actual selection), so Escape no longer hijacks focus.

**Files changed:**
- `packages/ui` — `TextInputContext`, `PromptInputControllerProps`, `PromptInputProvider`, `PromptInputTextarea`
- Both `MentionPopover.tsx` variants — refocus moved to `handleSelectFile`
- Both `ChatInputDropZone.tsx` variants — `document.querySelector` → `textInput.focus()`
- Both `ChatShortcuts.tsx` variants — `inputRootRef.current?.querySelector` → `textInput.focus()`

## Test plan

- [ ] Type `@` in the chat input, search for a file, select with Enter or click → textarea is focused, can keep typing immediately
- [ ] Type `@`, open the mention popover, press Escape → textarea is **not** focused (intentional)
- [ ] Two split chat panes side by side → `@mention` in the right pane → only the right pane's textarea refocuses

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable input focus across the chat experience: focusing now consistently works after drag‑and‑drop, when using the focus hotkey, and after selecting file mentions from the suggestion popover.
  * Removed fragile focus behavior tied to surrounding elements, resulting in steadier keyboard interactions and fewer missed focus events.

### DIFF
--- a/apps/desktop/src/renderer/components/Chat/ChatInterface/components/ChatInputFooter/ChatInputFooter.tsx
+++ b/apps/desktop/src/renderer/components/Chat/ChatInterface/components/ChatInputFooter/ChatInputFooter.tsx
@@ -153,7 +153,6 @@ export function ChatInputFooter({
 										<ChatShortcuts
 											isFocused={isFocused}
 											setIssueLinkOpen={setIssueLinkOpen}
-											inputRootRef={inputRootRef}
 										/>
 										<IssueLinkCommand
 											open={issueLinkOpen}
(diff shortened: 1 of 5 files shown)


Kitenite:
Hey sorry i keep forgetting to remove myself as coauthor when merging, literally only fixed lint. Thanks for the great PR as always @michalkopanski :)

PR #2676, opened 2026-03-21
fix(desktop): restore native text editing shortcuts in chat and auto-focus on pane nav

michalkopanski:

Three related fixes to keyboard/focus behavior in the chat prompt textarea:

1. **Skip pane-nav shortcuts when input is focused** — `Cmd+Shift+Left/Right` was being intercepted by the `PREV_PANE`/`NEXT_PANE` hotkey handlers when the cursor was in a textarea or input. Added an early-return guard so native text selection is preserved.

2. **Stop modifier+arrow propagation in chat textarea** — Moved the fix to the source: the prompt textarea now calls `stopPropagation` on `Cmd/Ctrl+Arrow` keys so they perform native text navigation/selection rather than bubbling to pane-navigation handlers.

3. **Auto-focus chat textarea on pane nav** — When navigating to a chat pane via keyboard shortcut, the prompt textarea is now automatically focused (mirroring how the terminal auto-focuses on pane activation). The cursor is placed at the end of any existing draft text rather than selecting it all.

_To be added by developer_

# QA
- [ ] Desktop web
- [ ] Mobile web
- [ ] iOS
- [ ] Android
- [ ] API
- [x] N/A — Desktop app only

## QA steps

**Text editing shortcuts in chat textarea:**
1. Open the desktop app with multiple panes visible (e.g. chat pane + terminal pane).
2. Click into the chat prompt textarea and type several words.
3. Place the cursor in the middle of the text.
4. Press `Cmd+Left` — verify it moves the cursor to the start of the line (native behavior), **not** switching to the previous pane.
5. Press `Cmd+Right` — verify it moves the cursor to the end of the line, **not** switching to the next pane.
6. Press `Cmd+Shift+Left` — verify it selects text to the left of the cursor, **not** switching panes.
7. Press `Cmd+Shift+Right` — verify it selects text to the right, **not** switching panes.

**Pane navigation still works outside inputs:**
8. Click on an empty area (not a textarea/input) and press `Cmd+Shift+Left/Right` — verify pane navigation still works as expected.

**Auto-focus chat textarea on pane nav:**
9. With a chat pane and terminal pane open, click into the terminal so it has focus.
10. Press the pane navigation shortcut (`Cmd+Shift+Left` or `Cmd+Shift+Right`) to move focus to the chat pane.
11. Verify the chat prompt textarea is automatically focused (you can start typing immediately).
12. If there is existing draft text in the textarea, verify the cursor is placed at the **end** of the text (not selecting all).

N/A

# Docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced focus management for the prompt input when switching between panes
  * Improved text cursor positioning when focus is restored
  * Fixed handling of modifier key combinations (Ctrl/Cmd + arrow keys) in text input for better keyboard navigation

### DIFF
--- a/apps/desktop/src/renderer/components/Chat/ChatInterface/components/ChatInputFooter/ChatInputFooter.tsx
+++ b/apps/desktop/src/renderer/components/Chat/ChatInterface/components/ChatInputFooter/ChatInputFooter.tsx
@@ -4,13 +4,13 @@ import {
 	PromptInputAttachments,
 	type PromptInputMessage,
 	PromptInputTextarea,
-	usePromptInputController,
 } from "@superset/ui/ai-elements/prompt-input";
 import type { ThinkingLevel } from "@superset/ui/ai-elements/thinking-toggle";
 import type { ChatStatus, FileUIPart } from "ai";
 import type React from "react";
 import type { ReactNode } from "react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useRef, useState } from "react";
+import { useFocusPromptOnPane } from "renderer/components/Chat/ChatInterface/hooks/useFocusPromptOnPane";
 import { useHotkeyText } from "renderer/stores/hotkeys";
 import type { SlashCommand } from "../../hooks/useSlashCommands";
 import type { ModelOption, PermissionMode } from "../../types";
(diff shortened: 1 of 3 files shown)


Kitenite:
I don't think this is exactly right since the hotkeys can be rebound which would make this stale. wonder if we need to fix hotkeys at the api level or if we should fix the chat input to prevent bubbling to hotkeys instead ?

michalkopanski:
Hey @Kitenite this is ready for review now. I opted to prevent bubbling to hotkeys — good call!

PR #2913, opened 2026-03-26
fix(desktop): preserve file scroll position across workspace switches

Kitenite:
## Summary
- Preserve scroll position across workspace switches for all pane types: file viewer (raw, diff, rendered/tiptap), terminal (normal buffer), and changes tab.

## Why / Context
Switching workspaces and switching back resets scroll positions to the top (or bottom for terminal). This is disorienting for users who work across multiple workspaces. Reported by @matt.nathan.

## How It Works

Three mechanisms, each suited to its scroll system:

### 1. `useScrollPreservation` hook (plain DOM containers)
New shared hook at `WorkspaceView/hooks/useScrollPreservation/`. Attaches a scroll listener, caches `scrollTop` in a ref, saves to a module-level `Map` on cleanup, and restores via `requestAnimationFrame` on mount.

**Covers:** diff viewer, tiptap rendered mode, changes tab (InfiniteScrollView).

**TipTap fix:** The rendered mode had two nested scroll containers — TipTap's inner wrapper (`h-full overflow-y-auto`) filled the parent exactly and handled all scrolling internally, so `markdownContainerRef` never scrolled. Fixed with CSS override `[&>.markdown-renderer]:h-auto [&>.markdown-renderer]:!overflow-y-visible` to make the parent the sole scroll container.

### 2. CodeMirror raw mode (pixel-based `scrollTop`)
CodeMirror manages its own virtual scroll, so the hook can't be used directly. Instead:
- `CodeEditor` fires an `onScrollTopChange` callback from a DOM scroll listener on `view.scrollDOM`
- `FileViewerPane` caches the value in a ref, saves to a `rawScrollCache` Map on unmount
- On remount, restores via `editorRef.current.setScrollTop()` after content loads

New `getScrollTop()`/`setScrollTop()` methods added to `CodeEditorAdapter`.

### 3. xterm.js terminal (absolute viewportY)
xterm manages its own buffer-based scroll. On unmount:
- Saves `buffer.active.viewportY` (absolute line number) to `savedViewportOffset` Map — only when user is NOT at the bottom
- On restore, waits for xterm write queue to drain + 2 animation frames (for SIGWINCH prompt redraw), then calls `xterm.scrollToLine(savedViewportY)`
- If user was at bottom, default `scrollToBottom` behavior is preserved

**Note:** TUI apps (Claude Code, vim, etc.) use the alternate screen buffer where scroll position is managed by the TUI itself, not by xterm. Alt-screen scroll preservation is not possible at the terminal emulator level.

## Performance

No measurable impact:
- **Scroll listeners** are passive reads of `scrollTop`/`scrollDOM.scrollTop` into a ref — no state updates, no re-renders, no DOM writes on scroll.
- **Module-level Maps** hold a single number per pane/workspace — negligible memory, bounded by open pane count.
- **Save/restore** only runs on mount/unmount (workspace switch), not during normal use.
- **No new subscriptions, observers, or polling loops.**

### File Viewer — Raw Mode
- [ ] Open a large file, scroll down, switch workspace, switch back — scroll restored to exact position
- [ ] "Go to line" still works (revealPosition path unchanged)

### File Viewer — Diff Mode
- [ ] View a diff, scroll down, switch workspace, switch back — scroll restored

### File Viewer — Rendered/Tiptap Mode
- [ ] Open a markdown file in rendered mode, scroll down, switch workspace, switch back — scroll restored
- [ ] TipTap content renders correctly (no layout issues from CSS override)

### Changes Tab
- [ ] Expand changes view, scroll through diffs, switch workspace, switch back — scroll restored

### Terminal (normal buffer)
- [ ] Run a command that produces output, scroll up, switch workspace, switch back — scroll position restored
- [ ] Terminal already at bottom stays at bottom after switch
- [ ] New output appended while away doesn't shift the restored position

### Edge Cases
- [ ] File at top (no scroll) — no regression
- [ ] Opening a new file after returning works normally
- [ ] TUI app (Claude Code) reattach — alt-screen rehydrates normally (no scroll restore attempted)

## Testing
- `bun run typecheck` — pass
- `bun run lint:fix` — no fixes needed

## Design Decisions
- **Why three mechanisms instead of one**: CodeMirror and xterm manage their own virtual scroll — a generic DOM `scrollTop` save/restore doesn't survive their initialization. Plain scrollable divs (diff, tiptap, changes) share the `useScrollPreservation` hook.
- **Why pixel-based for raw mode**: The previous line-based approach used `revealPosition` which centers the line in the viewport. Pixel-based `scrollTop` gives exact restoration.
- **Why absolute viewportY for terminal**: New output appended at the bottom while on another workspace shouldn't shift the restored position. Absolute line numbers point at the same content regardless of new output.
- **Why CSS override for tiptap**: TipTap's wrapper div uses `h-full overflow-y-auto`, creating a nested scroll container inside `markdownContainerRef`. The override makes TipTap auto-height so the parent handles scrolling. Scoped to the file viewer via child selector — doesn't affect other TipTap usages.
- **Why not alt-screen scroll**: TUI apps manage their own viewport within the alternate screen buffer. xterm has no scrollback in alt-screen mode — the TUI controls what's displayed.

## New Files
- `hooks/useScrollPreservation/useScrollPreservation.ts` — shared hook for DOM scroll containers
- `hooks/useScrollPreservation/index.ts` — barrel export

## Summary by CodeRabbit

* **New Features**
  * Restores editor scroll position when returning to a file (raw, diff, and rendered views).
  * Preserves scroll position in the changes list (infinite scroll) across navigation.
  * Remembers and restores terminal viewport position after pane switches or resizes; falls back to bottom when appropriate.
  * Improves markdown rendering container behavior to reduce unexpected overflow during viewing.


Kitenite:
Closing as stale: created in March with no recent activity. Re-open or re-create from a fresh branch if still relevant. Bulk audit 2026-05-06.

PR #2996, opened 2026-03-29
fix(desktop): remove terminal backpressure warning

Kitenite:
## Summary

Remove the terminal-host backpressure warning while keeping the existing socket backpressure handling unchanged.

## What Changed

- delete the `console.warn` emitted when a client socket returns `false` from `write()`
- keep the existing pause/resume-on-drain flow intact

## Why

The warning was noisy and no longer worth carrying forward as a dedicated change. This leaves the existing behavior in place without logging repeated backpressure messages.

## Impact

- no terminal output delivery semantics change
- daemon logs no longer emit this backpressure warning

## Validation

```bash
bun test apps/desktop/src/main/terminal-host/session.test.ts
```

## Summary by CodeRabbit

* **Refactor**
  * Removed unnecessary console warnings from terminal session handling to reduce log noise while maintaining all backpressure functionality.

### DIFF
--- a/apps/desktop/src/main/terminal-host/session.ts
+++ b/apps/desktop/src/main/terminal-host/session.ts
@@ -55,12 +55,6 @@ const ATTACH_FLUSH_TIMEOUT_MS = 500;
 const MAX_SUBPROCESS_STDIN_QUEUE_BYTES = 2_000_000;
 
 /**
- * Minimum interval (ms) between backpressure warnings per session.
- * Prevents log flooding when a high-output pane keeps the socket buffer full.
- */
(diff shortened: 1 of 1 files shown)


Kitenite:
Updated summary for the final branch state:

- remove the terminal backpressure `console.warn`
- keep the existing pause/resume-on-drain handling unchanged
- no terminal output delivery semantics change
- no additional test changes are part of the final diff

Why this changed:

The earlier branch went through a few narrower versions while backing out the lossy streaming idea. The final PR is now intentionally minimal: remove the noisy backpressure warning and leave the rest of the behavior alone.

Validation:

```bash
bun test apps/desktop/src/main/terminal-host/session.test.ts
```

PR #3044, opened 2026-03-31
fix(desktop): correct terminal link underline positioning for CJK text

### LINKED ISSUE #2231: [bug] Underline decoration misaligned with CJK/mixed-width text in terminal

### LINKED ISSUE #2317: [bug] Clickable file paths in terminal don't work with non-ASCII (CJK)
#### Describe the bug

  According to the https://docs.superset.sh/terminal-integration, file paths
  in terminal output should be clickable and open in the editor.

  This works for English file paths, but does not work when the path contains
   CJK (Chinese/Japanese/Korean) characters.

  Steps to reproduce:
  1. Run Claude Code in Superset's terminal
  2. Claude Code outputs a file path containing Chinese characters, e.g.
  電馭工作流/筆記/直播腳本骨架-v2.md
  3. The path appears underlined/styled as a link
  4. Clicking it does nothing (neither plain click nor Cmd+click)

  Expected behavior:
  Clicking the file path should open the file in the editor, same as English
  paths.

  Environment:
  - macOS (Apple Silicon)
  - Superset latest version
  - Terminal running: Claude Code CLI

  Workaround:
  Currently using Cmd+P to search and open files manually.

greekr4:
## Summary
- CJK (Chinese/Japanese/Korean) characters occupy **2 terminal columns** but are only **1 character** in JavaScript strings
- Both `MultiLineLinkProvider` and `FilePathLinkProvider` used JS string char offsets directly as xterm column positions, causing link underlines to appear at wrong positions when CJK text precedes a link
- Added `charOffsetToColumn()` helper that walks xterm `IBufferLine` cells to compute accurate column positions
- Fixed off-by-one on link `end.x` (xterm range end is inclusive)

### Before
image

### After
image

## Test plan
- [x] All existing link provider tests pass (83 tests, 697 assertions)
- [x] Updated `end.x` test expectation to match inclusive semantics
- [ ] Manual: output text with Korean/CJK characters before a file path in terminal, verify underline aligns correctly

Fixes #2231
Fixes #2317

---
## Summary by cubic
Fixes misaligned terminal link underlines when CJK, emoji, or tab-indented text precede URLs or file paths. Links now align to the correct terminal columns across wrapped lines. Fixes #2231 and #2317.

- **Bug Fixes**
  - Added `charOffsetToColumn()` in `MultiLineLinkProvider` and `FilePathLinkProvider` to map JS offsets to terminal columns by walking buffer cells (handles wide CJK, zero-width cells, and UTF‑16 surrogate pairs).
  - Updated range calculations to use this mapping for start/end, treat `end.x` as inclusive, and include tab-width leading trim on continuation lines; added regression tests for CJK, emoji, and tab-indented continuations.

<sup>Written for commit 944a97332f1543099f778d52c0591671b1df949c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate link detection and coordinate calculation in the terminal for paths and URLs that include wide characters (CJK, emoji) and when links span wrapped lines; range endpoints now compute correctly across wrapped boundaries.

* **Tests**
  * Added and updated tests covering wide-cell scenarios, surrogate-pair emoji, and wrapped-line URL/path detection to validate coordinates and link merging.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/file-path-link-provider.test.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/file-path-link-provider.test.ts
@@ -2,21 +2,41 @@ import { describe, expect, it, mock } from "bun:test";
 import type { IBufferLine, ILink, Terminal } from "@xterm/xterm";
 import { FilePathLinkProvider } from "./file-path-link-provider";
 
-function createMockLine(text: string, isWrapped = false): IBufferLine {
+interface MockCellSpec {
+	chars: string;
+	width: number;
+}
+
+function createMockCell(chars: string, width: number) {
+	return {
+		getChars: () => chars,
+		getWidth: () => width,
+	};
+}
+
+function createMockLine(
+	text: string,
+	isWrapped = false,
+	cells?: MockCellSpec[],
+): IBufferLine {
+	const mockCells = cells?.map((cell) =>
+		createMockCell(cell.chars, cell.width),
+	);
+
 	return {
 		translateToString: () => text,
 		isWrapped,
-		length: text.length,
-		getCell: mock(() => null),
-		getCells: mock(() => []),
+		length: mockCells?.length ?? text.length,
+		getCell: mock((index: number) => mockCells?.[index] ?? null),
+		getCells: mock(() => mockCells ?? []),
 	} as unknown as IBufferLine;
 }
 
 function createMockTerminal(
-	lines: Array<{ text: string; isWrapped?: boolean }>,
+	lines: Array<{ text: string; isWrapped?: boolean; cells?: MockCellSpec[] }>,
 ): Terminal {
 	const mockLines = lines.map((l) =>
-		createMockLine(l.text, l.isWrapped ?? false),
+		createMockLine(l.text, l.isWrapped ?? false, l.cells),
 	);
 
 	return {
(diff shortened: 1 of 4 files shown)


Kitenite:
@greekr4 could you take a look at the PR comments and see if they should be addressed? Thanks for the PR :)

greekr4:
@Kitenite thank u, I’ll give it a try

greekr4:
Addressed the review comments in commit `c0c4d939e`.

What changed:
- fixed `charOffsetToColumn()` in both link providers to count UTF-16 code units from xterm cells, so ranges stay aligned when emoji / surrogate pairs appear before a link
- updated `MultiLineLinkProvider.offsetToPosition()` to convert `leadingTrim + localOffset` in a single `charOffsetToColumn()` call
- strengthened the regression harness with real cell-width mocks for CJK wide cells, emoji surrogate pairs, and tab-indented URL continuations

Validation:
- `bun test apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/file-path-link-provider.test.ts`
- `bun test apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers/url-link-provider.test.ts`
- `bun test apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/link-providers`

All passing locally.

PR #3054, opened 2026-03-31
fix: forward DA1 query responses regardless of attached clients

### LINKED ISSUE #3028: Add Primary Device Attribute (DA1) query response support
## Summary

Fish shell sends a Primary Device Attribute (DA1) escape sequence (`ESC [ c`) on startup to detect terminal capabilities. Superset doesn't appear to respond to this query, which causes fish to display a warning after a 10-second timeout:

```
fish could not read response to Primary Device Attribute query after waiting for 10 seconds.
This is often due to a missing feature in your terminal.
See 'help terminal-compatibility' or 'man fish-terminal-compatibility'.
This fish process will no longer wait for outstanding queries, which disables some optional features.
```

## Expected Behavior

The terminal should respond to the DA1 query (`ESC [ c`) with a standard VT100/xterm-compatible response (e.g., `ESC [ ? 6 4 ; ... c`), allowing fish to detect supported features without timing out.

## Impact

- 10-second delay on every new fish shell session while waiting for the query to time out
- Fish disables optional features (cursor shape changes, reflow detection) for the rest of the session

## Reference

- [Fish terminal compatibility docs](https://fishshell.com/docs/current/interactive.html#terminal-compatibility)
- DA1 is part of the VT100 spec and supported by most modern terminals (iTerm2, Alacritty, Kitty, WezTerm, Terminal.app, etc.)

## Workaround

Users can suppress the query by adding to `~/.config/fish/config.fish`:

```fish
set -g fish_handle_reflow 0
```

But native DA1 support would be the proper fix.

zombopanda:
## Summary

Fixes #3028 — fish shell 10-second startup delay caused by missing DA1 response.

The `emulator.onData` callback in `session.ts` only forwarded terminal query responses (DA1, DSR) to the PTY subprocess when `attachedClients.size === 0`. When a renderer client is attached (the normal case when viewing a terminal), the headless emulator's DA1 response was silently dropped.

**Previous fix #3030** removed the `shellReadyState === "pending"` guard but left the `attachedClients.size === 0` check intact. The renderer's xterm also generates DA1 responses, but those go through `write()` which drops all escape sequences during shell init — so neither code path delivered the DA1 response to fish.

### The fix

Remove the `attachedClients.size === 0` condition so the headless emulator always forwards query responses to the subprocess via `sendWriteToSubprocess`. This is safe because it writes directly to the PTY via IPC, bypassing the renderer's `write()` path entirely.

### Before
```
warning: fish could not read response to Primary Device Attribute query after waiting for 10 seconds.
```

### After
Fish starts instantly with no warning.

## Test plan
- [x] Verified locally on macOS (Superset v1.4.6, fish 4.x, Apple Silicon)
- [ ] Verify fish shell starts without DA1 warning in terminal pane
- [ ] Verify no duplicate input/echo when typing in terminal with client attached
- [ ] Verify zsh/bash terminals still work correctly

**Note:** The terminal-host daemon is a detached long-running process. After deploying this fix, the daemon must be restarted (kill process from `~/.superset/terminal-host.pid`) for the change to take effect.

---
## Summary by cubic
Always forward DA1/DSR query responses from the headless emulator to the PTY, even when a renderer client is attached, bypassing the renderer write path. This fixes fish’s 10-second startup delay and warning.

- **Migration**
  - Restart the terminal-host daemon after deploy (kill the process using the PID at ~/.superset/terminal-host.pid).

<sup>Written for commit 84d34573a554ee60ba23730c9b92224351cf30b0. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal data handling in headless mode to ensure subprocess output is properly processed when the terminal is ready, regardless of active client connections.


SaladFork:
In my testing, this PR fixes #3028. Thank you!

Kitenite:
Thanks for this @zombopanda !

PR #3098, opened 2026-04-01
Fix app crash when run commands contain non-string values

### LINKED ISSUE #3095: App crashes with 'command.trim is not a function' when config uses object format for run commands
## Description

The app crashes on startup with a black screen showing "Something went wrong!" when `.superset/config.json` contains objects (with `name` and `command` properties) in the `setup`, `teardown`, or `run` arrays instead of plain strings.

## Steps to Reproduce

1. Create `.superset/config.json` in a project with the following format:
```json
{
  "setup": [],
  "teardown": [],
  "run": [
    {
      "name": "My Script",
      "command": "npm run dev"
    }
  ]
}
```
2. Open the project in Superset.
3. The app crashes immediately with a black screen.

## Expected Behavior

The app should either:
- Support the object format (`{ name, command }`) in addition to plain strings, or
- Gracefully handle invalid config values with a helpful error message instead of crashing the entire app.

## Actual Behavior

The app crashes with the following error in the developer console:

```
TypeError: command.trim is not a function
    at WorkspaceRunButton.tsx:53:24
    at Array.some (<anonymous>)
    at WorkspaceRunButton2 (WorkspaceRunButton.tsx:52:52)
    at renderWithHooks (react-dom-client.production.js:4351:21)
    at updateFunctionComponent (react-dom-client.production.js:6497:15)
    ...
```

## Root Cause

In `WorkspaceRunButton`, the code assumes `commands` is always an array of strings:

```js
const hasRunCommand = (runConfig?.commands ?? []).some(
    (command) => command.trim().length > 0
);
```

When the config contains objects instead of strings, `command` is an object and `.trim()` is not a function on it.

Notably, `hasConfiguredScripts` already guards against this with a `typeof` check:

```js
const run = Array.isArray(config?.run) ? config.run.filter(
    (s) => typeof s === "string" && s.trim().length > 0
) : [];
```

But the `getResolvedRunCommands` resolver that feeds `WorkspaceRunButton` does not apply the same validation.

## Suggested Fix

Validate/coerce commands to strings in `getResolvedRunCommands` before returning them to the UI — consistent with how `hasConfiguredScripts` already handles it. Additionally, a `typeof` guard in the component would prevent a full app crash from invalid data.

## Workaround

Use plain strings in the config instead of objects:

```json
{
  "setup": [],
  "teardown": [],
  "run": ["npm run dev"]
}
```

denjalonso:
## Summary

- Fixes `TypeError: command.trim is not a function` crash in `WorkspaceRunButton` when `.superset/config.json` contains objects instead of plain strings in the `run` array
- Filters `config.run` entries in `getResolvedRunCommands` to only include non-empty strings, consistent with how `hasConfiguredScripts` already handles this

Fixes #3095

## Root Cause

`getResolvedRunCommands` returned `config.run` directly without validation:

```ts
return { commands: config?.run ?? [] };
```

When a user writes their config as:
```json
{ "run": [{ "name": "Dev", "command": "npm run dev" }] }
```

The renderer receives objects instead of strings, and `WorkspaceRunButton` crashes calling `.trim()` on them — making the entire app unusable (black screen).

## Fix

Filter to only valid string entries before returning, matching the existing pattern in `hasConfiguredScripts`:

```ts
const commands = Array.isArray(config?.run)
    ? config.run.filter(
        (s): s is string => typeof s === "string" && s.trim().length > 0,
    )
    : [];
return { commands };
```

## Test plan

- [ ] Create `.superset/config.json` with object-format run commands (`{ name, command }`) — app should no longer crash
- [ ] Create config with valid string run commands — run button should work as before
- [ ] Create config with empty `run: []` — no crash, run button shows "Set Run"
- [ ] Create config with mixed valid/invalid entries — only valid strings are used

---
## Summary by cubic
Fixes an app crash when `.superset/config.json` contains non-string entries in `run` by returning only non-empty string commands from getResolvedRunCommands, preventing `.trim()` errors in the Workspace Run button.

- **Bug Fixes**
  - Filter `config.run` to non-empty strings; ignore objects and blanks.
  - Aligns with existing `hasConfiguredScripts` validation and preserves normal run button behavior.

<sup>Written for commit 68cf238bee31a8ded1141cba7a2c2cf927464d8d. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation and handling of run command configuration to ensure proper processing and improve system stability.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/workspaces/procedures/query.ts
+++ b/apps/desktop/src/lib/trpc/routers/workspaces/procedures/query.ts
@@ -342,9 +342,14 @@ export const createQueryProcedures = () => {
 					projectId: project.id,
 				});
 
-				return {
-					commands: config?.run ?? [],
-				};
+				const commands = Array.isArray(config?.run)
+					? config.run.filter(
+							(s): s is string =>
+								typeof s === "string" && s.trim().length > 0,
+						)
+					: [];
+
+				return { commands };
 			}),
 	});
 };


denjalonso:
Hey @kirankunigiri, could you take a look at this? The crash originates from the `getResolvedRunCommands` procedure you added in #2511 — it passes `config.run` to the renderer without filtering for strings, so non-string entries (like objects) cause `WorkspaceRunButton` to crash on `.trim()`.

PR #3108, opened 2026-04-02
feat(desktop): terminal in v2 pane

Kitenite:
## Summary

- **Terminal runtime registry** (`terminal-runtime-registry.ts`): App-level singleton keyed by `paneId` that manages xterm instances, websocket connections, and DOM attach/detach. The xterm wrapper div is reparented between containers — surviving tab/workspace switches without losing scrollback. Serializes buffer to localStorage on detach via `@xterm/addon-serialize` so scrollback survives renderer restarts.

- **Server session registry** (`terminal.ts`): Route changed from `/terminal/:workspaceId` to `/terminal/:paneId`. PTY lifetime is now independent of socket lifetime — socket close = detach (buffer output up to 64KB), reconnection replays buffer, explicit `dispose` message kills PTY. `workspaceId` is init metadata passed as a query param.

- **TerminalPane as attach/detach wrapper**: Reduced from 168 to ~50 lines. Calls `registry.attach()` on mount, `registry.detach()` on unmount. No xterm/websocket creation logic.

- **Global terminal lifecycle hook** (`useGlobalTerminalLifecycle`): Mounted once in authenticated layout. Watches `v2WorkspaceLocalState` across all workspaces. Disposes terminals only when a `paneId` disappears from every persisted workspace layout, with a 500ms re-check delay for cross-workspace moves.

## Lifecycle model

| Event | Action |
|---|---|
| Tab switch | detach only |
| Workspace switch | detach only |
| Route unmount | detach only |
| Pane move (same paneId) | no dispose |
| workspaceId change alone | no dispose |
| paneId gone from all persisted layouts | dispose (after 500ms re-check) |

## Test plan

- [ ] Open a terminal, type commands, switch tabs — output preserved on return
- [ ] Switch workspaces and back — terminal output and process state preserved
- [ ] Close a terminal pane — PTY is killed, localStorage entry cleaned up
- [ ] Move a terminal pane between workspaces — no interruption
- [ ] Restart the Electron app — scrollback restored from localStorage, server replays buffered output

---
## Summary by cubic
Decouples the terminal runtime from the view lifecycle so terminals persist across tab/workspace switches and renderer restarts without losing scrollback. Adds a per-pane runtime+transport registry and a new `/terminal/:paneId` server route with buffered replay and explicit disposal.

- **New Features**
  - Renderer split: `terminal-runtime.ts` (per-pane `@xterm/xterm`, wrapper reparenting, buffer/dimension persistence, full repaint on reattach), `terminal-ws-transport.ts` (idempotent `connect` — no-op if already open or connecting to the same URL; resize/input/dispose; closes displaced socket with code 4000; guards handlers against displaced sockets; connection state observable), `terminal-runtime-registry.ts` (per-`paneId` orchestrator; attach/detach only; always calls `connect()` — idempotency handled in transport; keeps WS alive across detach; eager entry creation for `onStateChange`).
  - `TerminalPane` now takes `paneId`, attaches/detaches only, and uses `useSyncExternalStore` to show connection state; shows a small “Disconnected” bar when closed.
  - Server sessions keyed by `paneId`; PTY lifetime is independent of the socket; up to 64KB is buffered and replayed on reconnect; reconnect displaces the previous socket; `onClose/onError/onMessage` are guarded by socket reference; `dispose` kills the PTY.
  - Global lifecycle disposes only when a `paneId` disappears from all persisted workspace layouts (500ms grace for cross-workspace moves).

- **Migration**
  - Use `/terminal/:paneId` and pass `workspaceId` as a query param.
  - Ensure `TerminalPane` receives `paneId` and `workspaceId`.
  - Behavior: tab/workspace switches now detach only; terminate by removing the pane from all layouts.

<sup>Written for commit 3259e181f116cb9a5e336a6bfd1b89b5c32e4078. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Per-pane terminal instances with preserved scrollback and dimensions when switching panes
  * Global lifecycle management to automatically dispose idle terminals after a short delay
  * Reconnect-friendly behavior: buffered output and seamless socket detach/reattach

* **Bug Fixes / UI**
  * Simplified terminal view with clearer "Disconnected" state and more reliable connection handling

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/terminal-runtime-registry.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-runtime-registry.ts
@@ -43,17 +43,7 @@ class TerminalRuntimeRegistryImpl {
 			sendResize(transport, runtime.terminal.cols, runtime.terminal.rows);
 		});
 
-		// Only connect/reconnect when the socket is not already open for this URL.
-		// On a simple tab-switch the transport stays alive, so we skip the reconnect
-		// and just send a resize to sync dimensions.
-		const isAlreadyConnected =
-			transport.connectionState === "open" && transport.currentUrl === wsUrl;
-
-		if (isAlreadyConnected) {
-			sendResize(transport, runtime.terminal.cols, runtime.terminal.rows);
-		} else {
-			connect(transport, runtime.terminal, wsUrl);
-		}
+		connect(transport, runtime.terminal, wsUrl);
 	}
 
 	/**
--- a/apps/desktop/src/renderer/lib/terminal/terminal-ws-transport.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-ws-transport.ts
@@ -42,6 +42,12 @@ export function connect(
 	terminal: XTerm,
 	wsUrl: string,
 ) {
+	// Idempotent: skip if already connected/connecting to the same endpoint.
+	const isActive =
+		transport.connectionState === "open" ||
+		transport.connectionState === "connecting";
+	if (isActive && transport.currentUrl === wsUrl) return;
+
 	if (transport.socket) {
 		transport.socket.close();
 		transport.socket = null;


apps/desktop/src/renderer/routes/_authenticated/components/GlobalTerminalLifecycle/GlobalTerminalLifecycle.tsx
```diff
@@ -0,0 +1,6 @@
+import { useGlobalTerminalLifecycle } from "../../hooks/useGlobalTerminalLifecycle";
+
+export function GlobalTerminalLifecycle() {
```
saddlepaddle, line 3: nit: you should be able to mount the hook wherever you want without doing this

apps/desktop/src/renderer/routes/_authenticated/components/GlobalTerminalLifecycle/hooks/useGlobalTerminalLifecycle/useGlobalTerminalLifecycle.ts
```diff
@@ -0,0 +1,86 @@
+import type { WorkspaceState } from "@superset/panes";
+import { useLiveQuery } from "@tanstack/react-db";
+import { useEffect, useRef } from "react";
+import { terminalRuntimeRegistry } from "renderer/lib/terminal/terminal-runtime-registry";
+import { useCollections } from "../../providers/CollectionsProvider";
+
+/** Cross-workspace moves temporarily remove a paneId then re-add it. Wait before disposing. */
+const DISPOSE_DELAY_MS = 500;
+
```
saddlepaddle, line 9: I'd move the lib/terminal stuff under apps/desktop/src/renderer/routes/_authenticated/hooks/useGlobalTerminalLifecycle/useGlobalTerminalLifecycle/lib imo, to follow rules of colocation

apps/desktop/src/renderer/routes/_authenticated/layout.tsx
```diff
@@ -174,6 +175,7 @@ function AuthenticatedLayout() {
 	return (
 		<DndProvider manager={dragDropManager}>
 			<CollectionsProvider>
+				<GlobalTerminalLifecycle />
```
saddlepaddle, line 178: ah this is why it needs to be a thin component - maybe move `hooks/useGlobal...` under `.../routes/_authenticated/components/GlobalTerminalLifecycle/GlobalTerminalLifecycle/hooks/...`?

apps/desktop/src/renderer/routes/_authenticated/components/GlobalTerminalLifecycle/GlobalTerminalLifecycle.tsx
Kitenite: The hook calls `useCollections()` which requires being inside `CollectionsProvider`. In the layout, the provider is rendered as a child — so calling the hook directly in `AuthenticatedLayout` would be outside the provider context. The thin component is needed to mount the hook inside the provider tree.

apps/desktop/src/renderer/routes/_authenticated/components/GlobalTerminalLifecycle/hooks/useGlobalTerminalLifecycle/useGlobalTerminalLifecycle.ts
Kitenite: `lib/terminal/` is imported by multiple consumers at different route levels — `useGlobalTerminalLifecycle`, `TerminalPane` (deep in workspace route), plus 6 other files import `launch-command`. Co-locating under one consumer would violate the project's co-location rule ("if used 2+ times, promote to highest shared parent"). `renderer/lib/terminal/` is the correct shared location.

apps/desktop/src/renderer/routes/_authenticated/layout.tsx
Kitenite: Yeah the thin component is needed for provider context (see reply above). The current structure `GlobalTerminalLifecycle/hooks/useGlobalTerminalLifecycle/` is already co-located — adding another nesting level would just be unnecessary depth.

PR #3176, opened 2026-04-05
feat(desktop): define v2 terminal env contract, stop leaking process.env

Kitenite:
## Summary
- Replace raw `process.env` passthrough in v2 terminals with a defined env contract
- Strip Electron / Superset / host-service internals via denylist before spawning PTYs
- Inject small public terminal surface (`TERM_PROGRAM=Superset`, `COLORTERM`, `LANG`, `PWD`)

## Why / Context

The v2 workspace terminal in host-service was spreading raw `process.env` into PTYs. This leaked Electron runtime vars (`ELECTRON_RUN_AS_NODE`, `NODE_OPTIONS`), host-service secrets (`HOST_SERVICE_SECRET`, `AUTH_TOKEN`), Chromium internals (`GOOGLE_API_KEY`), and build-system config (`VITE_*`, `TURBO_*`) into user terminal sessions.

V1 terminals solve this with an allowlist (`buildSafeEnv`), but v2 needs a different approach: the host-service process env already contains the user's shell-derived env (merged by the desktop app via `getProcessEnvWithShellPath` before spawning host-service), so we use a denylist to strip only the known internal vars while preserving everything the user's shell provides.

## How It Works

**Denylist approach** (not allowlist) on the already shell-derived base env:

1. **Prefix strips**: `ELECTRON_`, `SUPERSET_`, `VITE_`, `NEXT_PUBLIC_`, `TURBO_`, `npm_`, `CHROME_`
2. **Exact strips**: `HOST_SERVICE_SECRET`, `AUTH_TOKEN`, `GOOGLE_API_KEY`, `NODE_OPTIONS`, `NODE_ENV`, etc.
3. **Public contract injection**: `TERM=xterm-256color`, `TERM_PROGRAM=Superset`, `TERM_PROGRAM_VERSION`, `COLORTERM=truecolor`, `LANG=<utf8>`, `PWD=<cwd>`

User shell env (PATH, SSH_AUTH_SOCK, NVM_DIR, GOPATH, proxy vars, etc.) passes through unchanged.

No v1 hook metadata (`SUPERSET_PANE_ID`, `SUPERSET_TAB_ID`, `SUPERSET_PORT`, `SUPERSET_HOOK_VERSION`) is injected unless a v2 feature explicitly needs it.

**Two implementations, same contract:**
- `packages/host-service/src/terminal/env.ts` — runtime builder (reads from `process.env`)
- `apps/desktop/src/main/lib/terminal/env.ts` — canonical contract (takes `shellEnv` param, for desktop-side v2 integration)

The v2 env builder is applied inside `createTerminalSessionInternal` — the single place where `node-pty` is spawned in the new session-based terminal architecture.

## Testing
- `bun test apps/desktop/src/main/lib/terminal/env.test.ts` — 100 tests pass (28 new v2 tests)
- `bun run typecheck` — full monorepo typecheck passes
- Tests cover: prefix stripping, exact var stripping, user env preservation, public contract vars present, legacy `SUPERSET_*` absent, input immutability, host-service secret stripping

## Manual QA Checklist

- [ ] v2 workspace terminal spawns and works normally
- [ ] `env` command in v2 terminal shows `TERM_PROGRAM=Superset`
- [ ] `env` command does NOT show `SUPERSET_PANE_ID`, `HOST_SERVICE_SECRET`, `ELECTRON_RUN_AS_NODE`
- [ ] `node --version`, `git --version`, `ssh -T git@github.com` work (PATH/SSH preserved)
- [ ] Version managers (nvm, pyenv) work in the terminal

## Design Decisions
- **Why denylist instead of allowlist for v2**: The host-service `process.env` already contains the shell-derived env (desktop app merges it via `getProcessEnvWithShellPath` before spawning). An allowlist would duplicate the v1 approach and risk missing legitimate user vars. A targeted denylist strips only what we know shouldn't be there.
- **Why two implementations**: Host-service can't import from `apps/desktop`. Both follow the same contract; the desktop version is the canonical definition with tests.

## Follow-ups
- Wire `buildV2TerminalEnv` into desktop terminal session when v2 panes use the desktop terminal system
- Add `TERM_PROGRAM_VERSION` from actual app version (currently defaults to `1.0.0` in host-service)


dgilperez:
Fresh repro on Desktop 1.8.5 canary, macOS Sequoia 15.6 — bug this PR fixes is still live.

Inside any v2 agent terminal:

```
$ env | grep SSH_AUTH_SOCK
(empty)

$ launchctl getenv SSH_AUTH_SOCK
/private/tmp/com.apple.launchd.<random>/Listeners
```

So launchd has it published, the v2 PTY env doesn't. Every `git push` over SSH from agent terminals fails with *"Could not open a connection to your authentication agent"*, while the same command works in macOS Terminal. Same shape as #4030 — different symptom (ssh-agent socket vs gh keyring), same root cause class.

Worth sanity-checking before merge: if `getProcessEnvWithShellPath` runs `zsh -c env` (non-interactive), `.zshrc` doesn't source and `SSH_AUTH_SOCK` won't be there for users who rely on launchd injection. A darwin-only `launchctl getenv SSH_AUTH_SOCK` fallback closes this deterministically.

Local workaround until merge:

```bash
# in each ~/.superset/bin/<agent>, before exec
[ -z "$SSH_AUTH_SOCK" ] && export SSH_AUTH_SOCK="$(launchctl getenv SSH_AUTH_SOCK 2>/dev/null)"
```

Happy to test a build.

mdrxy:
Strong support for this direction. Setting an explicit `TERM_PROGRAM=Superset` as part of a defined env contract (rather than leaking raw `process.env`) is the right call — it fixes the "claiming to be a terminal we're not" class of bugs and avoids leaking Electron/host-service internals into user shells. 👍

PR #3181, opened 2026-04-05
feat(desktop): add agent notification status indicators to v2 workspaces

Kitenite:
## Summary
- Wire agent lifecycle notifications into v2 workspace UI — tab strip status dots and sidebar workspace icons now show real status instead of mock data
- New standalone `useV2PaneStatusStore` keeps status tracking decoupled from `@superset/panes`
- Separate `useV2AgentHookListener` hook avoids modifying the v1 listener (zero regression risk)

## Why / Context

The notification pipeline (shell hook → HTTP server → EventEmitter → tRPC subscription → OS notification + sound) already works end-to-end for v2 workspaces. The missing piece was the renderer: `useAgentHookListener` only writes to the v1 `useTabsStore`, so v2 panes never got status indicators. The dashboard sidebar was displaying fake status from `getWorkspaceRowMocks()`.

## How It Works

**Store** (`renderer/stores/v2-pane-status/store.ts`): Minimal Zustand store mapping `paneId → PaneStatus` and `paneId → workspaceId`. Panes are registered/unregistered as v2 workspaces mount (registration persists after unmount so the global listener can still process events).

**Listener** (`useV2AgentHookListener.ts`): Runs globally in `_authenticated/layout.tsx` alongside the v1 listener. Both subscribe to the same tRPC subscription — each silently ignores the other's panes via `isV2Pane()` check. Status mapping mirrors v1: Start→working, Stop→review/idle, PermissionRequest→permission.

**Tab strip**: `renderTabAccessory` on `<Workspace>` renders `<StatusIndicator>` per tab, aggregating pane statuses via `getHighestPriorityStatus()`.

**Sidebar**: `DashboardSidebarWorkspaceItem` reads from the store via `selectWorkspaceStatus()`. `workspaceStatus` is passed as a direct prop to child components (no longer through mockData).

### New Files
- `renderer/stores/v2-pane-status/store.ts` — Zustand store
- `renderer/stores/v2-pane-status/index.ts` — barrel export
- `renderer/stores/v2-pane-status/useV2AgentHookListener.ts` — notification listener

### Modified Files
- `_authenticated/layout.tsx` — call `useV2AgentHookListener()` globally
- `useV2WorkspacePaneLayout.ts` — pane registration effect (persists on unmount)
- `v2-workspace/$workspaceId/page.tsx` — `renderTabAccessory` + `TabStatusAccessory`
- `DashboardSidebarWorkspaceItem.tsx` — read real status from store
- `DashboardSidebarExpandedWorkspaceRow.tsx` — accept `workspaceStatus` as direct prop
- `getWorkspaceRowMocks.ts` — remove `workspaceStatus` from mock data

## Design Decisions
- **Separate hook vs v2 fallback in v1 listener**: Separate hook keeps v1 completely untouched (zero regression risk). Two subscriptions to the same tRPC endpoint is negligible cost.
- **Store outside `@superset/panes`**: Keeps the panes package generic. Status is a domain concept that doesn't belong in a layout engine.
- **No unregister on unmount**: Pane entries persist so the global listener and sidebar can process events even when the workspace page isn't mounted.

## Known Limitations / Blockers

**v2 terminals don't set `SUPERSET_PANE_ID`**: v2 terminals connect via WebSocket to the host service — they don't use the v1 `terminal.createOrAttach` tRPC call that sets `SUPERSET_PANE_ID` via `buildTerminalEnv()`. As a result, `notify.sh` fires without a pane ID that matches any registered v2 pane, and events are silently dropped by `isV2Pane()`. The renderer-side plumbing is complete, but the host service needs to inject `SUPERSET_PANE_ID` into v2 terminal environments for this to work end-to-end.

## Testing
- Could not run `bun run typecheck` — worktree missing `node_modules`
- Manual type review done against all imports and interfaces
- v1 notifications verified unchanged

## Follow-ups
- Host service: inject `SUPERSET_PANE_ID` (and `SUPERSET_WORKSPACE_ID`) into v2 terminal environments so `notify.sh` can relay the correct pane ID


apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/page.tsx
```diff
@@ -192,141 +191,107 @@ function WorkspaceContent({
...
+				query={commandPalette.query}
+				scope={commandPalette.scope}
+				searchResults={commandPalette.searchResults}
+				workspaceName={workspaceName}
 			/>
 		</>
 	);
 }
+
+function TabStatusAccessory({ tab }: { tab: Tab<PaneViewerData> }) {
```
saddlepaddle, line 338: nit: should be a subcomponent

Issue #3235, opened 2026-04-07
feat(desktop): reduce EDR/security agent CPU usage from idle terminals and active agent sessions


felixbarny:
## Background

Users on machines with kernel-level EDR agents (e.g. Elastic Endpoint) experience sustained high CPU usage (>100–200%) from the security agent even when Superset is completely idle. EDR agents intercept and inspect every file open, stat, and process spawn system-wide, so Superset's background activity directly drives the EDR's CPU load.

## Root causes identified

Two independent sources were found via `fs_usage` and `sample` profiling on macOS:

**1. Scrollback file — continuous writes on every PTY event**

`HistoryWriter.write()` calls `stream.write()` on every PTY data chunk. With active Claude Code agent sessions producing dense terminal output, this causes `scrollback.bin` to be modified hundreds of times per second per session. Each mtime update causes the EDR to re-stat and re-open the file for inspection.

**2. Port scanner — `ps` + `lsof` spawned every 2.5s per session**

`PortManager` runs a periodic scan every 2,500ms that calls `pidtree` (spawning `ps`) for each registered terminal session, then runs `lsof` across all discovered PIDs. With 13 open sessions this produces ~14 process spawns every 2.5 seconds. EDR agents intercept every process spawn and monitor the files those processes open — `lsof` in particular opens thousands of file descriptors per invocation as it enumerates all sockets system-wide.

## Proposed fixes

1. **Scrollback write debounce** (`terminal-history.ts`): Buffer PTY output in memory and flush to disk at most once every 250ms. Reduces file modification events by batching many small PTY chunks into a single write. No impact on terminal responsiveness — the in-memory buffer is the source of truth; disk writes are just for cold restore.

2. **Port scan interval** (`port-manager.ts`): Increase from 2,500ms → 10,000ms. The hint-based scan (fires within 500ms of seeing "listening on port X" in output) handles prompt detection; the periodic scan is just a safety-net sweep.

3. **Skip idle sessions in periodic scan** (`port-manager.ts`): Skip any session that has no currently-detected ports and has not seen port-hint output in the last 60 seconds. A machine with 13 idle terminal sessions running no dev servers produces zero `ps`/`lsof` spawns per cycle. Hint scans still fire immediately when a dev server starts.

## Result

[...]

PR #3237, opened 2026-04-07
perf(desktop): debounce scrollback writes to reduce EDR file-inspection load

felixbarny:
## Summary

`HistoryWriter.write()` calls `stream.write()` on every PTY data chunk, causing `scrollback.bin` to be modified hundreds of times per second during active agent sessions. Each mtime update triggers kernel-level security agents (e.g. Elastic Endpoint) to re-stat and re-open the file for inspection.

Buffer PTY output in memory and flush to disk at most once every 250ms. This batches many small PTY events into a single file write, reducing the rate of mtime changes without affecting terminal responsiveness — the in-memory buffer is the source of truth; disk writes are only for cold restore.

Part of #3235

---
## Summary by cubic
Debounced terminal scrollback writes by buffering PTY output and flushing at most every 250ms to cut `scrollback.bin` mtime churn and reduce EDR re-inspection load. Addresses Linear #3235 without impacting terminal responsiveness or cold-restore behavior.

- **Performance**
  - Buffer PTY data in memory and batch disk writes every 250ms.
  - Flush buffer on drain, close, and reinitialize; cancel timers appropriately.
  - Keep backpressure behavior and `MAX_PENDING_WRITE_BYTES` cap; drop when capped.

<sup>Written for commit a3f85609815c6eb69269d80f9c049412e49b3ba0. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Optimized terminal output writing to batch and schedule writes at intervals, improving performance and responsiveness when handling rapid terminal output.

### DIFF
--- a/apps/desktop/src/main/lib/terminal-history.ts
+++ b/apps/desktop/src/main/lib/terminal-history.ts
@@ -22,6 +22,10 @@ import { SUPERSET_DIR_NAME } from "shared/constants";
 const MAX_HISTORY_BYTES = 5 * 1024 * 1024; // 5MB per session
 const MAX_PENDING_WRITE_BYTES = 256 * 1024; // cap in-memory backlog when disk is slow
 const DRAIN_TIMEOUT_MS = 1000;
+// Debounce interval for flushing writes to disk. Batching writes reduces the
+// rate of file modification events, which prevents security agents (e.g.
+// Elastic Endpoint) from continuously re-inspecting the scrollback file.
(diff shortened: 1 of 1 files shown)


felixbarny:
Superseded by #3236, which covers the same fix with a better implementation (array-based debounce buffer, exported constant for testability, and a pre-existing flush() hang bug fix we missed). Closing in favour of that PR.

Issue #3309, opened 2026-04-09
[bug] Codex terminal state corrupts on pane reattach after tab/workspace switch


nikhilshinday:
#### Describe the bug

When returning to an existing Superset terminal pane that is running Codex CLI, the terminal can corrupt its visible state instead of simply repainting cleanly.

The most obvious symptom is that a stray newline gets inserted into the Codex input area and previously rendered text gets overwritten or shifted. This is different from the already-reported "blank output after Cmd+Tab" issue: in this case the pane is still visible, but the rendered terminal state is wrong.

This appears to be specific to Superset's terminal host / pane reattach path. The same Codex session behavior is fine outside Superset.

This also reproduces when launching the real Codex binary directly inside Superset (`/opt/homebrew/bin/codex`), so it does not appear to require Superset's `~/.superset/bin/codex` wrapper.

#### Expected behavior

Returning to a Codex pane after switching away should preserve terminal state exactly. No extra newline should be inserted, and previously rendered lines should not be overwritten or shifted.

#### Actual behavior

After switching away from the Codex pane and coming back:

- a new blank line may appear in the input area
- previously rendered text can be overwritten or visually shifted
- the pane can look partially corrupted even though Codex is still running

#### Reproduction

I do not yet have the smallest possible public repro, but I can reproduce this consistently with Codex inside Superset:

1. Open a Superset terminal pane.
2. Start Codex CLI and let it render a normal TUI session.
3. Switch away from that pane by changing tabs/workspaces and/or app focus.
4. Return to the same Codex pane.
5. Observe that the terminal state is no longer identical to before the switch; an extra newline may appear and previous text may be overwritten.

Control case:

- The same Codex workflow is fine outside Superset, so this does not look like a plain zsh / prompt issue.
- Launching `/opt/homebrew/bin/codex` directly inside Superset still reproduces, so this does not look specific to the Superset Codex wrapper script.

#### Environment

- Superset Desktop: 1.4.7
- Electron runtime: 40.2.1
- macOS: 26.1
- Hardware: Apple Silicon (M4 Max)
- Codex CLI: 0.118.0

#### Evidence

The local daemon log shows repeated pane reattach activity for the same session IDs during the failing window:

```text
[2026-04-09T21:53:36.559Z] [terminal-host] [INFO] Creating/attaching session: pane-1775770830804-560flwa60
[2026-04-09T21:53:36.562Z] [terminal-host] [INFO] Session pane-1775770830804-560flwa60 attached
[...]

Issue #3208, opened 2026-04-06
[bug] when focusing back on terminal, content gets corrupted


aliak00:
#### Describe the bug

When switching between tabs or worktrees and coming back to the terminal the content in the terminal is corrupted. The example below shows what happens with a CLI I am building. In this example the content is at least not overwritten, but when the output is bigger and you scroll around etc, the context actually gets clobbered as well.

This is an image of the CLI process in question when this particular app is run and the terminal it's being run in within superset is always in focus.

<img width="1492" height="1516" alt="Image" src="https://github.com/user-attachments/assets/a2275601-99a3-486d-bd4d-7dca1af5eb34" />

And then after I go to another workspace and come back 5 times:

<img width="1498" height="1532" alt="Image" src="https://github.com/user-attachments/assets/4c4eb12d-9d37-434c-b5f3-8c83f1d4b44c" />

The rendering just gets more messed up with each in focus.

I guess this might be the same? https://github.com/superset-sh/superset/issues/1830

I have access to the CLI in the images above if you want me to help in anyway by trying out things or telling you how the rendering works.

Kitenite:
**Related reports** (cross-referenced during issue triage — same area, may or may not share a root cause):

- #3794 — Terminal renders garbled characters (WebGL texture atlas corruption)
- #3572 — [bug] Intermittent garbled text rendering with multiple Claude Code tabs and Chinese content
- #3527 — [bug] Terminal Output Corruption (Garbled CJK Characters) and Layout Glitches during Heavy Usage
- #3504 — [bug] macOS terminal rendering can corrupt and ghost into chat until fullscreen toggle
- #3668 — [bug] Claude code rendering issue in superset terminal
- #3208 — [bug] when focusing back on terminal, content gets corrupted
- #3321 — [bug] Terminal can still repaint into a stale visual state after focus/workspace restore
- #2968 — [bug] Desktop terminal temporarily freezes after tab/workspace switch with Droid TUI
- #1065 — [bug] TUI Flickering for Other Providers

Kitenite:
Closing as a duplicate of #5605, the consolidated tracker for stale `terminal-rendering` issues from the July 2026 stale-issue sweep (no activity since 2026-06-01). If this still applies on a current build, comment on #5605 and we'll reopen.

Issue #3080, opened 2026-04-01
[bug] TUI rendering broken: content duplication, text cutoff, and wasted horizontal space


omermosko4:
#### Describe the bug

Since approximately the last few releases (~1 week), the TUI rendering has become increasingly unreliable. Multiple rendering issues occur simultaneously:

1. **Content duplication** — Messages and UI elements render twice in the conversation view
2. **Text cutoff** — Long lines get clipped at the edge instead of wrapping properly
3. **Wasted horizontal space** — The content area renders in a narrow column, leaving a large unused blank area on the right side of the terminal
4. **General visual messiness** — The overall layout becomes garbled/messy during use

## Environment
- **Superset Desktop:** ~v1.4.3 through v1.4.6 (regression appeared around this range)
- **macOS:** Darwin 24.6.0
- **Agent:** Claude Code (Opus 4.6, 1M context)

## Steps to Reproduce
1. Open Superset Desktop
2. Start a Claude Code session in a terminal tab
3. Have a conversation with multiple messages / tool calls
4. Observe rendering issues — content gets duplicated, text cuts off, layout breaks

The issue appears consistently and does not require specific actions to trigger — normal usage is sufficient.

## Expected Behavior
- Messages should render once, not be duplicated
- Text should wrap correctly within the available terminal width
- The full terminal width should be utilized for content rendering
- The UI should remain clean and readable throughout the conversation

## Actual Behavior
- Messages frequently appear twice in the conversation view
- Text is cut off at terminal edges
- A large portion of the horizontal space is unused (content renders in a narrow left column)
- The overall TUI layout is garbled and hard to read

## Screenshots

<img width="1122" height="651" alt="Image" src="https://github.com/user-attachments/assets/f01850e3-2786-45b8-bb86-796a948fe5ca" />

<img width="1201" height="895" alt="Image" src="https://github.com/user-attachments/assets/8012f493-a181-4480-abc6-c7dc9287ea6d" />

<img width="1200" height="438" alt="Image" src="https://github.com/user-attachments/assets/f68915c8-dc23-4d09-b340-1db70342de87" />

## Additional Context
- **Reproducibility:** Frequent / nearly always
- **Regression window:** Started approximately 1 week ago across recent releases
- Potentially related: #1830 (duplicate input bars after tab switching), #3057 (viewport scrolling issues)

Kitenite:
[...]

Issue #2968, opened 2026-03-28
[bug] Desktop terminal temporarily freezes after tab/workspace switch with Droid TUI


ethanaggor:
#### Describe the bug

In Superset Desktop 1.4.4 on macOS 15.7.3, a terminal running the Droid CLI TUI becomes temporarily unresponsive after switching away from the tab/workspace and coming back.

The terminal eventually recovers, but there is a visible freeze / catch-up period before interaction resumes. I can reproduce this repeatedly on visibility changes.

This appears to be specific to Superset's terminal integration rather than Droid alone, because the same Droid TUI works smoothly in iTerm2 on the same machine.

Related context: this may be connected to #2961, but this report has a narrower and more reproducible trigger focused on visibility/reattach behavior.

#### Expected behavior

Returning to the tab/workspace should restore the terminal immediately and keep the TUI responsive.

#### Actual behavior

After returning to the tab/workspace, the terminal stalls for a period, then catches up and becomes interactive again.

The effect gets worse when Droid is repainting heavily, but I can still reproduce it even without a very large buffer.

#### Reproduction

1. Open a Superset terminal running Droid CLI.
2. Let the TUI render normally.
3. Switch to another tab/workspace or otherwise hide the terminal view.
4. Return to the Droid terminal.
5. Observe a freeze / delayed re-render before input becomes responsive again.

#### Control cases

- Droid in iTerm2: no issue
- Codex / Claude in Superset: not obviously reproducing in the same way
- This makes Droid look more like the trigger and Superset's terminal lifecycle / reattach path look like the failing host

#### Environment

- Superset Desktop: 1.4.4
- macOS: 15.7.3
- Hardware: Apple Silicon MacBook Pro
- Droid CLI: 0.89.0
- Superset terminal persistence setting: default persistent behavior enabled locally (`persist_terminal = 1`)

#### Evidence

Videos:
- Superset failing case: https://drive.google.com/file/d/1AYSfWmVd1kyf4drvrGv3Rx5PlZ8wwNk0/view?usp=drive_link
- iTerm2 positive control: https://drive.google.com/file/d/19YqWQFd9NQHncrn2_CHAJT0rziqXoVYG/view?usp=drive_link

Daemon log during the failing window:

```text
[Session pane-1774509268993-98gtz4ktl] Client socket buffer full, output may be delayed
[2026-03-28T21:04:46.424Z] [terminal-host] [INFO] Creating/attaching session: pane-1774509268993-98gtz4ktl
[...]

PR #3322, opened 2026-04-10
fix(desktop): retry terminal repaint after focus restore

### LINKED ISSUE #3321: [bug] Terminal can still repaint into a stale visual state after focus/workspace restore
## Summary

Returning to a terminal pane after switching app focus / window / workspace can still leave the terminal in a partially stale visual state even after the underlying text content has recovered.

In my local repro with Codex CLI running inside Superset Desktop on macOS, the most visible symptom was that the TUI content could repaint, but some terminal chrome remained stale for a beat longer than the text layer. In practice this showed up as blank strips / stale repaint artifacts, and in some runs the Codex input box styling lagged behind the recovered content.

This looks like a separate renderer lifecycle problem from the attach-time snapshot/socket overlap bug. The pane content itself may no longer be duplicated or shifted, but a single focus-time `fit() + refresh()` pass can still miss the final settled layout after the window becomes visible again.

## Environment

- Superset Desktop: built from current `main` as of 2026-04-10
- macOS: Apple Silicon
- Repro app inside terminal: Codex CLI

## Reproduction

1. Open a terminal pane in Superset Desktop.
2. Start Codex CLI or another full-screen / dense TUI.
3. Switch away from the app or switch workspaces/tabs long enough for the pane to fully unmount/occlude.
4. Return to the same terminal pane.
5. Observe that the text content may recover, but the terminal can still briefly retain stale visual state from before focus returned.

## Expected behavior

When the terminal becomes visible again, the entire pane should repaint into a fully consistent final state in one restore cycle, including terminal background/chrome and TUI styling.

## Actual behavior

- content can recover first
- stale blank space or stale visual styling can remain briefly afterward
- a single recovery pass on `window.focus` / `visibilitychange` does not always seem to be enough once layout finishes settling

## Suspected area

`apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.ts`

Specifically the focus / visibility recovery path around:

- `clearTextureAtlas()`
- `fitAddon.fit()`
- `xterm.refresh()`
- resize reporting after refocus

My local patched build stopped reproducing once I changed the restore path to run a short recovery burst instead of a single pass:

- immediately on visibility/focus restore
- again after ~120ms
- again after ~260ms

That suggests the terminal container is still settling for a short period after focus is restored, so a single repaint can happen too early.

## Related issues

- #1830
- #1873
- #2507
- #2968
- #3080
- #3208
- #3309

I am opening a PR with the small recovery-burst change that fixed the repro locally.

surlymochan:
## Summary

- add a short recovery burst when a terminal becomes visible again after focus / visibility changes
- keep the existing throttle logic, but retry repaint/fit twice after the initial restore attempt
- add a regression test that models the new burst behavior

## Why

The existing terminal restore path already does the right kinds of work on refocus:

- clear the WebGL texture atlas
- re-fit the terminal to its container
- force an xterm refresh

The remaining problem is timing. In practice, one refocus recovery can still happen slightly before the terminal container has fully settled after the app/window/workspace becomes visible again. That leaves stale blank space or stale terminal/TUI styling until some later repaint happens.

Running a short recovery burst fixed the repro locally:

- immediate restore
- second restore after ~120ms
- third restore after ~260ms

This keeps the change small and local to the focus/visibility recovery path, while giving xterm another chance to repaint once layout has actually stabilized.

## Test plan

- `bun test ./src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.test.ts`

## Related

- Closes #3321
- Related: #1830
- Related: #1873
- Related: #2507
- Related: #2968
- Related: #3080
- Related: #3208
- Related: #3309

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal reconnection reliability with an immediate recovery attempt plus additional quick follow-up retries.
  * Enhanced cleanup to cancel pending recovery retries when the component unmounts.

* **Tests**
  * Added test coverage verifying burst-style recovery triggers the immediate attempt plus follow-up retries.
  * Updated a throttle test description to label it as a regression test.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.test.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.test.ts
@@ -39,6 +39,7 @@ type SchedulerState = {
 
 function makeScheduler(runRecovery: (forceResize: boolean) => void): {
 	schedule: (forceResize: boolean) => void;
+	scheduleBurst: (forceResize: boolean) => void;
 	flush: () => void;
 	state: SchedulerState;
 } {
@@ -150,16 +151,11 @@ describe("scheduleReattachRecovery throttle — issue #1873", () => {
 	});
 
 	/**
-	 * REPRODUCTION TEST — this test currently FAILS, demonstrating the bug.
+	 * Regression test for the recovery throttle bug.
 	 *
-	 * Expected behaviour: when a recovery call is throttled, a retry should be
-	 * scheduled to run after the remaining throttle window expires. Without a
-	 * retry the terminal is permanently blank until the user resizes the window.
-	 *
-	 * Fix: in scheduleReattachRecovery (useTerminalLifecycle.ts), when the
-	 * throttle fires, add:
-	 *   const remaining = reattachRecovery.throttleMs - (now - reattachRecovery.lastRunAt);
-	 *   setTimeout(() => { if (!isUnmounted) scheduleReattachRecovery(reattachRecovery.pendingForceResize); }, remaining + 1);
+	 * Verifies that when a recovery request lands inside the 120ms throttle
+	 * window, the terminal schedules a retry after the remaining throttle delay
+	 * instead of silently dropping the repaint request.
 	 */
 	it("throttled recovery is retried after throttle window expires", async () => {
 		let calls = 0;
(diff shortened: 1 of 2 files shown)


surlymochan:
Follow-up update pushed to this PR.

I addressed the concrete review items in the regression test:
- added `scheduleBurst` to the `makeScheduler` return type
- rewrote the stale "currently FAILS" block comment into a normal regression-test comment

Validation run locally:
- `bun test apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.test.ts`
- `bunx biome check apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.test.ts`

I kept the PR scope limited to the repaint/focus-restore bug in #3321. The other downstream fixes I found while validating the patched build are being split into separate upstream candidates so this PR stays reviewable.

dorukardahan:
i have another repro in the same family here: #3504

on my mac, i can also get terminal content ghosting into the chat view, not only blank space inside the terminal. when i toggle fullscreen, it fixes the rendering right away.

that makes me think the repaint recovery path in this PR is very relevant for my case too.

i am closing my draft PR #3506 to avoid overlap. if this PR does not fully cover the chat ghosting and fullscreen case, i can send a smaller follow up later.

PR #3405, opened 2026-04-13
fix(hotkeys): match terminal Ctrl chords by event.code under non-Latin IME

### LINKED ISSUE #3365: [bug] Terminal-reserved Ctrl chords (Ctrl+C/D/Z…) fail under non-Latin IME (e.g. Korean)
## Summary

When a non-Latin IME (e.g. Korean 2-Set) is active, terminal-reserved chords like `Ctrl+C`, `Ctrl+D`, `Ctrl+Z` no longer reach the PTY. This is because `isTerminalReservedEvent()` checks `event.key`, which returns the IME-transformed character (e.g. `ㅇ` for `d`), not the physical key.

## Environment

- Superset Desktop v1.5.0
- macOS, Apple Silicon
- Korean (2-Set) input method active

## Reproduction

1. Switch keyboard input to a non-Latin IME (e.g. Korean).
2. Open a terminal pane.
3. Press `Ctrl+D` — observe it does not send EOF.
4. Press `Ctrl+C` — observe it does not send SIGINT.
5. Switch back to English input — both work again.

## Root cause

In `renderer/hotkeys/utils/utils.ts`:

```typescript
export function isTerminalReservedEvent(event: KeyboardEvent): boolean {
    if (!event.ctrlKey || event.metaKey || event.altKey || event.shiftKey)
        return false;
    const key = event.key.toLowerCase();
    return TERMINAL_RESERVED.has(`ctrl+${key}`);
}
```

`event.key` reflects the IME composition, so under Korean input `Ctrl+D` yields `event.key === "ㅇ"` → `"ctrl+ㅇ"` which is not in the reserved set.

Combined with the blanket `Ctrl`/`Cmd` interception from #3337, the chord gets swallowed entirely.

## Proposed fix

Use `event.code` (physical key, IME-independent) instead of `event.key`:

```typescript
const key = event.code.replace(/^Key/, "").toLowerCase();
return TERMINAL_RESERVED.has(`ctrl+${key}`);
```

AytuncYildizli:
## Summary / What changed
- `matchesHotkeyEvent` (`apps/desktop/src/shared/hotkeys.ts`): when the parsed hotkey is a single letter `a–z`, also match against `event.code` (`KeyA`–`KeyZ`). Mirrors the existing digit-key fallback at line 272.
- Two new regression tests in `hotkeys.test.ts` cover Turkish (`ç` + `KeyC`) and Korean (`ㅇ` + `KeyD`).

## Why
Originally reported by @eungkyu in #3365 for Korean 2-Set IME. Same bug reproduces on Turkish layout: `Ctrl+C` emits `event.key = "ç"` while `event.code` stays `"KeyC"`. Without the fallback, terminal-reserved chords (Ctrl+C/D/Z/V…) stop reaching the PTY whenever a non-Latin IME is active.

The patch is intentionally narrow: only single-letter `a–z` hotkeys gain the `event.code` fallback. This keeps existing matching behavior for everything else and follows the same pattern the code already uses for digits and `slash`.

## Tests
- `bun test apps/desktop/src/shared/hotkeys.test.ts`
- 15/15 pass (13 existing + 2 new IME regressions).

## Real-world validation
- Reproduced on Superset Desktop **v1.5.1** macOS, Turkish keyboard layout: before the patch `Ctrl+C` is swallowed; after the patch SIGINT reaches the PTY normally.
- Issue #3365 reporter (@eungkyu) confirms the same root cause for Korean 2-Set; the new test for `ㅇ` + `KeyD` matches their reproduction.

## Issue
Fixes #3365.

---
## Summary by cubic
Fixes terminal Ctrl letter chords failing under non-Latin IMEs by matching single-letter hotkeys against `event.code` (`KeyA`–`KeyZ`) in `matchesHotkeyEvent` (fixes #3365). Adds regression tests for Turkish (`ç` + `KeyC`) and Korean (`ㅇ` + `KeyD`).

<sup>Written for commit c4f5db08caa4b67a96d25a86d51d01ebec3bb614. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hotkey detection reliability when using non-Latin keyboard layouts or IME input methods. Hotkeys like Ctrl+C and Ctrl+D now work correctly even when keyboard input transforms key values.

### DIFF
--- a/apps/desktop/src/shared/hotkeys.test.ts
+++ b/apps/desktop/src/shared/hotkeys.test.ts
@@ -91,4 +91,32 @@ describe("isTerminalReservedEvent", () => {
 			}),
 		).toBe(true);
 	});
+
+	// Regression: superset-sh/superset#3365 — ctrl+c/d/z fail under non-Latin IME
+	// (Turkish, Korean, Russian) because event.key reflects IME-transformed char.
+	it("detects ctrl+c via event.code when IME mangles event.key (Turkish)", () => {
+		expect(
+			isTerminalReservedEvent({
+				key: "ç", // Turkish IME emits 'ç' instead of 'c'
+				code: "KeyC",
+				ctrlKey: true,
+				shiftKey: false,
+				altKey: false,
+				metaKey: false,
+			}),
+		).toBe(true);
+	});
+
+	it("detects ctrl+d via event.code when IME mangles event.key (Korean)", () => {
+		expect(
+			isTerminalReservedEvent({
+				key: "ㅇ", // Korean 2-Set IME emits Hangul jamo for 'd'
+				code: "KeyD",
+				ctrlKey: true,
+				shiftKey: false,
+				altKey: false,
+				metaKey: false,
+			}),
+		).toBe(true);
+	});
 });
--- a/apps/desktop/src/shared/hotkeys.ts
+++ b/apps/desktop/src/shared/hotkeys.ts
@@ -271,6 +271,15 @@ export function matchesHotkeyEvent(
 	// Use event.code to match digit keys when alt is pressed
 	if (/^[1-9]$/.test(key) && eventCode === `digit${key}`) return true;
 
+	// IME / non-Latin keyboard layout fallback (e.g. Turkish, Korean, Russian):
+	// `event.key` reflects the IME-transformed character (e.g. Turkish dead keys
+	// or Korean 2-Set), while `event.code` always reports the physical key.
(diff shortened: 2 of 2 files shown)


AytuncYildizli:
Closing — investigated further and realized this PR targets `apps/desktop/src/shared/hotkeys.ts` which was removed in the v1.5.x migration. The active code path is now `apps/desktop/src/renderer/hotkeys/utils/resolveHotkeyFromEvent.ts` which already uses `event.code` (fixed in #3391). Sorry for the noise. Will re-investigate the user-facing report and open a fresh issue if the bug is reproducible on the current code.

PR #3407, opened 2026-04-13
fix(desktop): keep OS clipboard chords (Cmd+C/V/X/A) bubbling out of v1 terminal

AytuncYildizli:
## Summary / What changed
- `apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts`: short-circuit OS clipboard chords (Cmd+C/V/X/A on macOS, Ctrl+Shift+C/V/X/A and Shift+Insert / Ctrl+Insert on other platforms) so they bubble out of the terminal handler instead of falling through to the PTY.

## Why
#3390 made the v1 terminal forward any chord not registered as an app hotkey to the PTY. That fixed Claude Code TUI shortcuts but regressed standard OS clipboard combos: `Cmd+C`, `Cmd+V`, `Cmd+X`, `Cmd+A` are not registered as app hotkeys, so they now go straight to xterm and never reach Electron's edit menu / xterm.js Selection. Result: text selected in the terminal pane can't be copied; clipboard paste also doesn't reach the input line.

The patch adds an explicit allow-bubble step before the "fall through to PTY" branch. Terminal-reserved chords (Ctrl+C SIGINT, Ctrl+D EOF, etc.) are still handled earlier in the same function, so PTY semantics are unchanged.

## Tests
No unit tests touched — the change is platform-conditional event routing inside the live keyboard handler. Manual repro was decisive enough.

## Real-world validation
- Reproduced on **Superset Desktop v1.5.1** (macOS, Apple Silicon, Turkish keyboard layout). Before: select text in a terminal pane, press `Cmd+C`, nothing happens; `Cmd+V` does nothing. After: selection copies to the system clipboard; paste reaches the PTY input line as expected.
- Verified `Ctrl+C` in PTY still sends SIGINT (terminal-reserved path runs first).

## Issue
Regression of #3390. No standalone issue yet — happy to file one if maintainers prefer that flow.

---
## Summary by cubic
Fixes the v1 terminal swallowing OS clipboard shortcuts so they bubble to the host, restoring copy, paste, cut, and select-all in terminal panes; PTY-reserved chords are unchanged. Also seeds observability blueprint docs and a triage runbook.

- **Bug Fixes**
  - Bubble OS clipboard chords: macOS Cmd+C/V/X/A; Linux/Windows Ctrl+Shift+C/V/X/A, Ctrl+Insert (copy), Shift+Insert (paste). Adds explicit Insert handling and excludes Ctrl+Delete/Shift+Delete so readline edits reach the PTY.
  - Addresses regression from #3390 where unregistered chords fell through to the PTY and bypassed Electron/xterm selection.

<sup>Written for commit 6d65367294be7d929a09c846c26fa867483d2507. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/3407?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated terminal keyboard handling to let standard OS clipboard shortcuts (copy, paste, cut, select-all) pass through correctly on macOS and Windows/Linux.
  * Restored platform-appropriate clipboard keys (Ctrl+Insert / Shift+Insert on Windows/Linux) while keeping delete-related shortcuts routed to the terminal.

* **Documentation**
  * Added observability system documentation, including event taxonomy, journey definitions, field registry, and query reference.

* **Runbooks**
  * Introduced an observability triage runbook for diagnosing missing telemetry data.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts
@@ -589,14 +589,16 @@ export function setupKeyboardHandler(
 		// OS-level clipboard chords must always bubble so the host (Electron /
 		// xterm.js Selection / OS) can fulfil copy / paste / cut / select-all.
 		// Without this, #3390's "unregistered → PTY" rule swallows Cmd+C/V/X/A
-		// (Mac) and Ctrl+Shift+C/V/X/A or Shift+Insert (Linux/Windows),
-		// breaking text copy from the terminal pane. Regression of #3390.
+		// (Mac) and the Linux/Windows clipboard chords, breaking text copy
+		// from the terminal pane. Regression of #3390.
 		const onlyMeta =
 			event.metaKey && !event.ctrlKey && !event.altKey && !event.shiftKey;
-		const onlyCtrl =
-			event.ctrlKey && !event.metaKey && !event.altKey && !event.shiftKey;
 		const ctrlShiftOnly =
 			event.ctrlKey && event.shiftKey && !event.metaKey && !event.altKey;
+		const onlyCtrl =
+			event.ctrlKey && !event.metaKey && !event.altKey && !event.shiftKey;
+		const onlyShift =
+			event.shiftKey && !event.ctrlKey && !event.metaKey && !event.altKey;
 		if (isMac && onlyMeta) {
 			// Cmd+C copy, Cmd+V paste, Cmd+X cut, Cmd+A select-all
 			if (
@@ -608,8 +610,8 @@ export function setupKeyboardHandler(
 				return false;
 			}
 		}
-		if (!isMac && (onlyCtrl || ctrlShiftOnly)) {
-			// Linux/Windows: Ctrl+Shift+C/V/X/A or Shift+Insert / Ctrl+Insert
+		if (!isMac) {
+			// Ctrl+Shift+C/V/X/A — Linux/Windows terminal clipboard chords.
 			if (
 				ctrlShiftOnly &&
 				(event.code === "KeyC" ||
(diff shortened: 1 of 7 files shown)


AytuncYildizli:
thanks @greptile-apps and @coderabbitai — both findings were valid. fixed in 9a2b3fa47:

- removed `Ctrl+Delete` from the bubble path; it's a shell readline chord (delete-word-forward), not clipboard.
- added explicit `Insert + (onlyCtrl || onlyShift)` so `Shift+Insert` (paste) actually matches now. `Ctrl+Insert` (copy) still works.
- left `Shift+Delete` out — terminals usually consume it as a readline edit, ambiguous as cut.

Kitenite:
Follow-up note: PR #3415 now exists to handle the broader terminal shortcut cases around this same area.

It tightens the platform guard so Linux fallback chords do not leak onto macOS, and aligns the v1 terminal behavior more closely with VS Code's terminal clipboard handling, including macOS `Cmd+A` select-all and boundary coverage for the non-clipboard PTY cases.

Link: https://github.com/superset-sh/superset/pull/3415

Kitenite:
Thank you for reporting a fixing!

Kitenite:
Closing: this only touches the legacy v1 UI (screens/main), which is being retired. only v1 Terminal helpers.ts; v2 has clipboard-shortcuts.ts (#3968). If the bug still shows in the v2 workspace, please open an issue.

Kitenite:
Thank you for digging into the clipboard chords, and for the careful event.code work in the sibling PR. The v1 terminal is on its way out, which is the only reason this isn't landing.

PR #3506, opened 2026-04-16
fix(desktop): refresh terminal renderer after reattach

### LINKED ISSUE #3504: [bug] macOS terminal rendering can corrupt and ghost into chat until fullscreen toggle
## Summary

On macOS, terminal rendering can become corrupted after normal pane switching or window state changes. I see garbled glyphs, stale terminal lines drawn in the wrong place, and sometimes background terminal content ghosting into the chat view.

This is not specific to one CLI. I reproduced it with Claude CLI, Codex CLI, OpenCode, and also inside Superset own conversation UI when a background terminal was running.

## Actual behavior

- Text in terminal panes becomes unreadable with wrong glyph shapes.
- Sometimes parts of a hidden or background terminal are painted over the chat area.
- Toggling fullscreen fixes it right away, but the problem comes back later.
- The problem looks like a renderer or compositor issue, not a shell encoding issue.

## Expected behavior

- Hidden or background terminal panes should never paint outside their own pane.
- Switching tabs, switching panes, or resizing the window should not corrupt terminal text.
- Fullscreen should not be needed as a recovery step.

## Repro notes

Environment I checked locally:

- Superset.app `1.5.5`
- Electron `40.8.5`
- macOS
- xterm `6.1.0-beta.195`
- `@xterm/addon-webgl` `0.20.0-beta.194`
- `@xterm/addon-ligatures` `0.11.0-beta.195`

What i usually do:

1. Open a workspace with one or more terminal panes.
2. Run a long lived CLI session like Claude CLI, Codex CLI, or OpenCode.
3. Switch between terminal and chat panes, or use the app in a non fullscreen window for a while.
4. At some point the terminal text becomes garbled, or a background terminal line is drawn over the chat view.
5. If i toggle fullscreen on and off, the rendering becomes normal again.

## Related findings

- I could not find an open issue for this exact symptom.
- `#1873` looks related because it also mentions terminal reattach problems and blank space after switching.
- xterm has an open issue here: `xtermjs/xterm.js#3303` about WebGL and ligature rendering artifacts. One comment there says manual `clearTextureAtlas()` makes the issue go away.

I also checked the current source:

- `apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts` always tries to load `LigaturesAddon` and then `WebglAddon`.
- `apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/v1-terminal-cache.ts` reattaches the cached wrapper and calls `xterm.refresh(...)`, but it does not clear or recreate the WebGL texture atlas on reattach or resize.
- `apps/desktop/src/lib/electron-app/factories/app/setup.ts` already has a macOS specific workaround with this comment: `occluded window throttling can corrupt GPU compositor layers`.

## Proposed fix

I think this is likely in the overlap between Superset terminal lifecycle and xterm WebGL rendering on macOS.

Possible directions:

1. Clear the WebGL texture atlas on reattach and resize, then force refresh.
2. Recreate or dispose the WebGL addon when a cached terminal is detached and attached again.
3. Add a safe DOM renderer fallback or a setting to disable WebGL and maybe ligatures on macOS.

If useful, i can try a follow up patch after maintainers confirm which direction they want.

dorukardahan:
## Summary

- add a shared terminal renderer refresh helper that clears the WebGL texture atlas before repaint
- call that helper after terminal reattach, resize, font changes, and WebGL context loss
- use the same recovery path in both terminal lifecycle implementations
- add a focused unit test for the renderer refresh helper

## Why

This keeps the terminal redraw path consistent after a pane was detached, reattached, resized, or occluded on macOS.

The goal is narrow. This PR does not disable WebGL, does not disable ligatures, and does not change renderer selection. It only makes the existing WebGL path refresh from a clean texture atlas before repainting.

Nearby risk: the app has two terminal lifecycle paths. This change updates both so the recovery logic does not drift between them.

## Tests

- `bun test apps/desktop/src/renderer/lib/terminal/terminal-renderer.test.ts`
- `bun test apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/**/*.test.ts apps/desktop/src/renderer/lib/terminal/terminal-renderer.test.ts apps/desktop/src/renderer/lib/terminal/*.test.ts`
- `cd apps/desktop && bun run typecheck`

## Real-world validation

- linked from the macOS rendering issue
- local source inspection matched the shipped app versions involved in the bug report
- fullscreen still acts as an immediate recovery in the reported broken state, which is consistent with this repaint path

## Issue

Fixes #3504

---
## Summary by cubic
Fixes stale or blank terminal frames on macOS by clearing the WebGL texture atlas and forcing a repaint after reattach, resize, font changes, or context loss. Unifies the recovery path across both terminal lifecycles and adds a focused test. Fixes #3504.

- **Bug Fixes**
  - Added `refreshTerminalRenderer` helper that clears the WebGL texture atlas (via `@xterm/addon-webgl`) and calls a full `refresh`.
  - Invoked the helper after attach, resize, font updates, and WebGL context loss in both terminal lifecycle paths.
  - Added a unit test for the helper; no change to renderer selection or ligatures.

<sup>Written for commit 9c2328722899d5af375e521c0d85dc43adaf9b00. Summary will update on new commits.</sup>

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/terminal-addons.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-addons.ts
@@ -6,10 +6,12 @@ import { SearchAddon } from "@xterm/addon-search";
 import { Unicode11Addon } from "@xterm/addon-unicode11";
 import { WebglAddon } from "@xterm/addon-webgl";
 import type { Terminal as XTerm } from "@xterm/xterm";
+import { refreshTerminalRenderer } from "./terminal-renderer";
 
 export interface LoadAddonsResult {
 	searchAddon: SearchAddon;
 	progressAddon: ProgressAddon;
+	refreshRenderer: () => void;
 	dispose: () => void;
 }
 
@@ -24,6 +26,7 @@ let suggestedRendererType: "webgl" | "dom" | undefined;
 export function loadAddons(terminal: XTerm): LoadAddonsResult {
 	let disposed = false;
 	let webglAddon: WebglAddon | null = null;
+	const refreshRenderer = () => refreshTerminalRenderer(terminal, webglAddon);
 
 	terminal.loadAddon(new ClipboardAddon());
 
@@ -51,7 +54,7 @@ export function loadAddons(terminal: XTerm): LoadAddonsResult {
 			webglAddon.onContextLoss(() => {
 				webglAddon?.dispose();
 				webglAddon = null;
-				terminal.refresh(0, terminal.rows - 1);
+				refreshRenderer();
 			});
 			terminal.loadAddon(webglAddon);
 		} catch {
@@ -63,6 +66,7 @@ export function loadAddons(terminal: XTerm): LoadAddonsResult {
 	return {
 		searchAddon,
 		progressAddon,
+		refreshRenderer,
 		dispose: () => {
 			disposed = true;
 			cancelAnimationFrame(rafId);
--- a/apps/desktop/src/renderer/lib/terminal/terminal-renderer.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-renderer.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it, mock } from "bun:test";
+import { refreshTerminalRenderer } from "./terminal-renderer";
+
+describe("refreshTerminalRenderer", () => {
+	it("clears the WebGL texture atlas before forcing a repaint", () => {
+		const terminal = {
+			rows: 4,
+			refresh: mock(() => {}),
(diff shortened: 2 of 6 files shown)


dorukardahan:
i am closing this draft to avoid overlap with #3322.

my separate repro is tracked in #3504. if the current repaint recovery work does not fully fix the chat ghosting and fullscreen case on mac, i can send a narrower follow up PR later.

PR #3521, opened 2026-04-17
docs(desktop): v2 project create/import flow design

Kitenite:
## Summary

Three companion design docs for the v2 cloud-driven project create/import flows:

- **`v2-project-create-import.md`** — state matrix (cloud / host-service / disk / remote), host-service as the single orchestrator, backing derivation via two sources (local host-service `project.list` + Electric `v2_workspaces ⋈ v2_hosts` for remote online hosts), sidebar reconciliation with existing localStorage pin collections, phased rollout.
- **`v2-project-create-import-types.md`** — type contracts. Reuses existing Drizzle/tRPC shapes; only introduces `HostBacking` as a client-side composition plus new host-service endpoints (`project.list`, `project.create`, additive `acknowledgeWorkspaceInvalidation` on `project.setup`).
- **`v2-project-user-journeys.md`** — five scenarios walked against the state machine: new user in new org, joining an existing org, adding a second host, multi-host workflow, host going offline.

No code changes — design only.

## Test plan

- [ ] Review the state matrix and confirm cell 1–5 coverage matches expected flows
- [ ] Confirm host-service as sole orchestrator is the right architectural direction
- [ ] Resolve the strict-vs-permissive pin-sync question flagged in journey 3 / open question #6
- [ ] Resolve auto-pin-on-create question flagged in open question #5

---
## Summary by cubic
Implements Phases 1–4 of the v2 cloud-driven project create/import: backing-aware sidebar and workspaces, centralized create/import/setup modals, inline Set up/Repair flows, and a remote‑host workspace stub. Projects derive backing from the local host-service and the cloud `v2_host_projects` table; workspace actions prompt setup when needed and retry automatically.

- **New Features**
  - Cloud + APIs: added `v2_host_projects` (schema, migration, relations, Electric sync) with tRPC `v2HostProject.upsert/delete`; `v2Project.findByRemote`; host-service `project.create/setup` support clone/import, persist local project, upsert cloud host backing, and surface `PROJECT_NOT_SETUP`; `project.list` returns `pathStatus` for stale‑path detection.
  - Desktop: centralized `AddRepositoryModals` host for New project, Folder‑first Import, and Pin & set up (supports conflict re‑point and force re‑point); sidebar shows backing via `ProjectBackingStateIndicator` with inline “Set up” for unbacked and “Repair” for stale paths (collapsed rows route clicks accordingly); Workspaces tab adds “Available projects” and host‑type chips; remote‑device workspaces render a “Not on this device” stub with a Set up CTA; pending workspace creation catches `PROJECT_NOT_SETUP`, opens Pin & set up, and retries on success.

- **Bug Fixes**
  - Case‑insensitive GitHub owner/name match in `v2Project.create` and `v2Project.findByRemote`.
  - Fixed sidebar re‑render loops by pinning stable query fallbacks; tightened tooltip copy.

<sup>Written for commit 610706b2371882861b532d5ce84b2c9e90b23a72. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design and implementation-plan docs for the v2 project create/import experience: new cloud-backed project flows, per-host project states (Normal, Host offline, Not set up here), repair/setup and create/import actions, folder-first import, sidebar visibility and CTAs (pin-driven), an “Available” discovery section with create/pin/setup actions, phased rollout plan, user journeys and acceptance criteria.


docs/design/v2-project-create-import.md
```diff
@@ -0,0 +1,345 @@
...
+project.list() → Array<{
+  id: string              // matches v2_projects.id
+  repoPath: string
+  pathStatus: "healthy" | "missing"   // statSync(repoPath) at read time
+}>
+```
+
+One row per `host-service.projects` entry on the calling machine. Cell-3 detection stays server-side where `fs` lives.
+
+Renderer refetches after `project.create` / `project.setup` / `project.remove` mutations via React Query invalidation on the shared `["project", "list"]` key. No subscription — backing rarely changes and invalidation-on-mutation is sufficient.
```
saddlepaddle: nit: invalidation-on-mutation might not be sufficient

docs/design/v2-project-create-import.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Visibility rule
+
+**Pin alone.** A pinned project (`v2SidebarProjects` row) always renders. Backing health shows as row state, never as a filter. Users don't lose their place when a host goes offline. Pin-management (auto-pin, cross-device pin sync, unpin UX) is tuned separately — pin is a binary input to this design.
+
+### Backing derivation (client-side)
+
+Two sources, combined in `useDashboardSidebarData`. Nothing new gets synced.
+
+**Local backing** — authoritative, lag-free. Calls the local daemon:
```
saddlepaddle, line 191: why do we need two sources for the dashboard sidebar data? can't we rely just on the cloud rows?

docs/design/v2-project-create-import.md
```diff
@@ -0,0 +1,345 @@
...
+Both online and offline remote backings are surfaced — offline is what drives the "Host offline" row state. Direct signal, no workspace-count dependency.
+
+> ⚠️ **Implementation-time validation.** The two-source split (host-service tRPC for local, Electric live query for remote) is structurally motivated — `pathStatus` needs filesystem access, Electric gives push-based reactivity for remote changes. In practice, the combined hook may surface friction (ordering, invalidation timing, empty-state flashes). Revisit if the split causes pain during wiring.
+
+### Row state (per pinned project)
+
+| Row state | Condition | CTA |
+| --- | --- | --- |
+| Normal | local backing `healthy`, or any `remoteBacked.online` host | open / new workspace |
+| Stale path | local backing exists with `pathStatus: "missing"` | Repair (→ `project.setup` with `acknowledgeWorkspaceInvalidation`) |
```
saddlepaddle: do we care about stale path being a state? can't we just resolve on error in the main cases?

docs/design/v2-project-create-import.md
Kitenite: cloud row doesn't give whether a project has a host level backing or not, we have to check the host machine itself for the path itself. 2 layers: 
1. is project opened in host 
2. is project cloned on host (like does it exist or does the user have to set it up)

docs/design/v2-project-create-import.md
Kitenite: Good catch. Downgraded to pure DB read in Phase 1 (no pathStatus), so local host-service.projects only changes via our own mutations → invalidation-on-mutation is sufficient there. Proactive statSync + refetch interval moved to Phase 4 alongside the Stale-path row state (see 673eab0cc).

docs/design/v2-project-create-import.md
Kitenite: Cloud rows (v2_host_projects) tell us which hosts back a project, but not whether the clone actually exists on the local disk — filesystem access is the other half. Current host's disk health can only come from local host-service; Electric can't see disks. So the split is structural, not a choice. Added an asterisk in the doc to re-examine during wiring in case the split causes friction.

docs/design/v2-project-create-import.md
Kitenite: Agreed — pushed Stale-path entirely to Phase 4. Phase 1 project.list is a pure DB read, no statSync. Operations that hit a missing directory fail lazily; their error handlers invalidate the query so the sidebar can react. Proactive detection comes back in Phase 4 if users ask for it (673eab0cc).

packages/db/drizzle/0034_v2_host_projects.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE "v2_host_projects" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"organization_id" uuid NOT NULL,
```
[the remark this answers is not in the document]
Kitenite: Noted — skipping DB-level composite FK for now. The app layer (v2HostProjectRouter.resolveProjectHostOrg) already enforces same-org across project_id/host_id/organizationId on every write. Fixing only v2_host_projects would leave the pattern inconsistent with v2_workspaces, which has the same shape and no composite FK. Revisit as a cross-v2 cleanup.

apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/components/DashboardSidebarProjectSection/DashboardSidebarProjectSection.tsx
```diff
@@ -97,6 +98,7 @@ export function DashboardSidebarProjectSection({
 				<DashboardSidebarProjectRow
 					projectName={project.name}
 					githubOwner={project.githubOwner}
+					backingState={project.backingState}
```
saddlepaddle, line 126: WE should be able to remove this backing state as a concept from most places - instead, we can query if a host is online to see if the project is already set up in the new workspace creation modal and the eventual cron logic (which'll probably also share the project selector, so maybe we can consolidate some of this in a project selector component)?

apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/hooks/useDashboardSidebarData/useDashboardSidebarData.ts
```diff
@@ -44,6 +62,70 @@ export function useDashboardSidebarData() {
...
+
+	const locallyBackedProjectIds = useMemo(
+		() => new Set(localProjectList.map((p) => p.id)),
+		[localProjectList],
+	);
+
+	// Remote backing — v2_host_projects ⋈ v2_hosts, excluding rows for the
+	// current machine (current host's backing is covered by localProjectList
+	// above, which is authoritative and lag-free).
+	const { data: remoteBackingRows = EMPTY_REMOTE_BACKING_ROWS } = useLiveQuery(
```
saddlepaddle, line 96: the v2HostProjects can probably be removed

saddlepaddle: along with most of this logic, in favor of just doing an innerJoin on v2Hosts on workspaces

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspaces/components/FolderFirstImportModal/FolderFirstImportModal.tsx
```diff
@@ -0,0 +1,280 @@
+import { Button } from "@superset/ui/button";
```
saddlepaddle, line 1: seems like this component is added to the wrong directory? should be unrelated to v2-workspaces

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspaces/components/NewProjectModal/NewProjectModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+	onError?: (message: string) => void;
+}
+
+export function NewProjectModal({
+	open,
+	onOpenChange,
+	onSuccess,
+	onError,
+}: NewProjectModalProps) {
+	const { activeHostUrl } = useLocalHostService();
```
saddlepaddle, line 32: nit: may or may not need to provide an escape hatch to users for setting up projects on remote machines

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspaces/components/ParentDirectoryPicker/ParentDirectoryPicker.tsx
```diff
@@ -0,0 +1,49 @@
+import { Button } from "@superset/ui/button";
```
saddlepaddle, line 1: another one where we may need to move the file location to match where it's mounted

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspaces/page.tsx
```diff
@@ -23,15 +32,38 @@ function V2WorkspacesPage() {
 	}, [resetFilters]);
 
 	const { pinned, others, counts } = useAccessibleV2Workspaces({ searchQuery });
+	const { projects: availableProjects } = useAvailableV2Projects({
+		searchQuery,
+	});
 	const hasAnyAccessible = pinned.length > 0 || others.length > 0;
 
+	const openNewProject = useOpenNewProjectModal();
```
saddlepaddle, line 40: all probably needs to be mounted at the dashboard level

packages/trpc/src/router/v2-project/v2-project.ts
```diff
@@ -110,29 +115,91 @@ export const v2ProjectRouter = {
 			return { ...row, repoCloneUrl };
 		}),
 
-	create: protectedProcedure
+	findByRemote: jwtProcedure
```
saddlepaddle, line 118: nit: name might need to be clearer? maybe find by githubRemote?

packages/trpc/src/router/utils/host-access.ts
```diff
@@ -0,0 +1,33 @@
...
+		columns: { id: true },
+		where: and(
+			eq(v2UsersHosts.userId, userId),
+			eq(v2UsersHosts.hostId, hostId),
+		),
+	});
+	return !!row;
+}
+
+export async function requireHostAccess(
```
saddlepaddle, line 23: nit, this pattern kinda results in extra queries so we should probably start moving away from it (unless it encapsulates complex logic like it probably will later)

saddlepaddle: or actually jk maybe it's fine, we can just pass in a txid when we need it to be in one network round trip or w/e

PR #3543, opened 2026-04-18
fix(host-service): v2 workspace git correctness (upstream, 3-dot, numstat)

saddlepaddle:
## Summary

Fixes a cluster of v2-workspace git correctness bugs uncovered while chasing a report that "new workspaces start multiple commits behind origin/main":

1. **Stale local default branch on workspace creation.** New worktrees branched off `refs/heads/main` (local), which is rarely fast-forwarded. Now branches from the default branch's real configured upstream — fork-workflow safe (honors `upstream/main` when that's the canonical remote).
2. **Diff count grew as main advanced.** `getStatus` used a 2-dot diff, so every upstream merge inflated the sidebar count on untouched workspaces. Now 3-dot (merge-base), stable as main moves — matches GitHub's Files-changed semantics.
3. **Every base comparison was origin-hardcoded.** `getStatus`, `listBranches`, `listCommits`, `getDiff` all literally computed `origin/<default>` regardless of the local default branch's actual upstream. Centralized into one `resolveBaseComparison` helper that honors the real upstream first, falls back to `origin/<default>`.
4. **Renamed files showed +0 -0.** `parseNumstat` indexed by the raw numstat path string which for renames is `<old> => <new>`. Callers looked up by destination path from `--name-status` → never matched. Switched to `--numstat -z`, rewrote parser for the three-cell rename record, indexed both source and destination paths.
5. **Per-file diff view compared tip-to-tip, not merge-base.** For a file that both main and your branch touched unrelatedly, it showed main's unrelated edits as if you'd undone them. Now uses `git merge-base <base> HEAD` to get the fork-point commit and compares content there to HEAD.

## Documentation

`apps/desktop/docs/V2_WORKSPACE_DIFF_VIEWS.md` — plain-English explainer of every UI surface that shows "what changed," which comparison runs under the hood, and why. Written for people who don't know git; also useful for anyone who does and wants the mapping between UI and code paths.

## Scope notes

- **V1 untouched** per the v1→v2 port rule. V1 has parallel instances of these bugs in `apps/desktop/src/lib/trpc/routers/workspaces/utils/git.ts` — stays v1's problem.
- **`resolveRef`'s `"origin"` default** at `runtime/git/refs.ts:87` left as-is (low severity, not tied to any user-visible count).

## Test plan
- [ ] Reset local `main` 5 commits behind origin. Create new v2 workspace. `git -C <new-worktree> rev-parse HEAD` must equal `origin/main`.
- [ ] Add `upstream` remote; `git branch --set-upstream-to=upstream/main main`. Create a new v2 workspace. New worktree must branch off `upstream/main`, not `origin/main`. Sidebar counts must reflect `upstream/main...HEAD`.
- [ ] Open existing workspace on `main`. Sidebar count = 0. Merge a PR to upstream, run `git fetch` in the repo, reload the app. Count stays 0.
- [ ] `git mv src/foo.ts src/bar.ts` + change ~5 lines in the rename. Sidebar shows non-zero `+N -N` (not `+0 -0`).
- [ ] Pure rename (no edits): sidebar shows `+0 -0` correctly.
- [ ] Have your branch and main both edit the same file in unrelated places. Click that file in the Changes tab. Per-file diff must show only your changes (not main's unrelated edits).
- [ ] Non-default branch workspace creation still uses the local ref.
- [ ] Offline: create workspace off default. Targeted fetch catches error and proceeds with cached ref.
- [ ] `bun test` in `packages/host-service` — 120/120 pass (112 existing + 8 new numstat cases).

## Summary by CodeRabbit

* **New Features**
  * Improved upstream branch resolution for more accurate base selection.
  * Workspace creation now prefers remote-tracking start points when available.

* **Bug Fixes**
  * More accurate "against base" diffs using three-dot comparison semantics.
  * Corrected rename handling and diff counts (including non-ASCII and binary files).

* **Documentation**
  * Added detailed docs for v2 workspace diff view behaviors.

* **Tests**
  * Added integration and unit tests exercising git diff and parsing behaviors.


packages/host-service/src/runtime/git/refs.ts
```diff
@@ -143,3 +143,25 @@ export async function resolveDefaultBranchName(
 		return "main";
 	}
 }
+
+/**
+ * Resolve a local branch's upstream tracking info (`branch.<name>.remote`
+ * / `branch.<name>.merge`). Returns `null` if no upstream is configured.
+ */
+export async function resolveUpstream(
```
Kitenite, line 151: i think we actually have this already? can we check or its actually in my open PR that you're reviewing

packages/host-service/src/trpc/router/git/utils/git-helpers.ts
```diff
@@ -23,18 +24,38 @@ export function mapGitStatus(code: string): FileStatus {
 	}
 }
 
+/**
+ * Parse the NUL-delimited output of `git diff --numstat -z`. Renames
+ * appear as `<add>\t<del>\t\0<old>\0<new>\0` — three NUL-separated
+ * cells — and are indexed under both source and destination paths so
+ * callers keyed by either get a hit.
+ */
 export function parseNumstat(
```
Kitenite, line 33: i'd feel a lot safer if we have test coverage for this with a lot of edge cases, especially having it run it against your local gh

packages/host-service/test/numstat-parse.test.ts
```diff
@@ -0,0 +1,68 @@
+import { describe, expect, test } from "bun:test";
+import { parseNumstat } from "../src/trpc/router/git/utils/git-helpers";
+
+describe("parseNumstat", () => {
```
Kitenite: oh nvm there it is

Kitenite (changes requested):
Nah bro

saddlepaddle:
Addressed review feedback on top:

**Real correctness bugs caught by reviewers (fixed):**
- Non-ASCII filename regression: `--name-status` without `-z` quotes paths; `--numstat -z` doesn't. For `日本語.ts` or `ñ.ts`, the lookup silently returned `+0 -0`. Added `-z` to both and rewrote `parseNameStatus` for NUL records.
- `branch.<name>.remote = "."` (local branch tracking another local branch) produced invalid `./branch` refs. Now uses the bare branch name.

**Test coverage (Kiet's ask):**
- Moved `parseNumstat` tests next to `git-helpers.ts` per co-location rule.
- Added `parseNameStatus` unit tests (regular, rename, copy, non-ASCII, embedded tabs, mixed).
- Added parseNumstat rename-with-tabs + non-ASCII cases.
- Added a new integration test file (`git-helpers.integration.test.ts`) that spins up real on-disk repos with `simple-git` and covers:
  - `resolveUpstream` against real config: origin, fork-workflow (`upstream`), local-tracking (`remote = "."`), divergent remote branch name, nonexistent branch
  - `resolveBaseComparison` end-to-end: origin fallback, fork workflow, local-tracking, no default branch
  - `getDefaultBranchName` with `origin/HEAD` set / unset / pointing at `master`
  - `getChangedFilesForDiff` with real commits: adds/modifies/deletes, rename with edits, pure rename, **non-ASCII filename** (the regression above), binary file, and the 3-dot diff guarantee that base-side commits don't inflate the count

Full run: 161/161 host-service tests pass (112 pre-existing + 49 new).

**Housekeeping:**
- Doc moved to root `docs/` (general, not desktop-specific).
- Markdown fence language added.

docs/V2_WORKSPACE_DIFF_VIEWS.md
```diff
@@ -0,0 +1,190 @@
+# How v2 workspace diff views work
+
+An explainer for every place the UI shows "what changed" for a workspace, what comparison runs under the hood, and why.
+
+Written for people who don't know git. If you do know git: this doc maps each UI surface to the specific git command and explains why we picked that comparison over the alternatives.
```
Kitenite, line 5: How condescending, i know git and i don't understand anything either >:(

packages/host-service/src/trpc/router/git/utils/git-helpers.ts
```diff
@@ -23,38 +24,64 @@ export function mapGitStatus(code: string): FileStatus {
...
```
Kitenite, line 33: we should also use simple git instead of raw git yeah? if it has an equivalent

packages/host-service/src/trpc/router/workspace-creation/workspace-creation.ts
```diff
@@ -777,13 +755,48 @@ export const workspaceCreationRouter = router({
...
+						const remoteRef = asRemoteRef(
+							upstream.remote,
+							upstream.remoteBranch,
+						);
+						const remoteExists = await git
+							.raw([
+								"rev-parse",
+								"--verify",
+								"--quiet",
+								`${remoteRef}^{commit}`,
```
Kitenite, line 783: I think we moved away from `^{commit}` in general because its buggy, --no-track would be better, check https://github.com/superset-sh/superset/pull/3548

packages/host-service/src/trpc/router/workspace-creation/workspace-creation.ts
```diff
@@ -777,13 +755,48 @@ export const workspaceCreationRouter = router({
...
 					? buildStartPointFromHint(
 							input.composer.baseBranch,
 							input.composer.baseBranchSource,
 						)
 					: await resolveStartPoint(git, input.composer.baseBranch);
+
+			// Local default branches are rarely fast-forwarded; swap to the
+			// branch's configured upstream so we fork from the real tip, not a
+			// stale local ref. Non-default branches stay local-first by design.
+			if (startPoint.kind === "local") {
```
Kitenite, line 769: nit: consider refactor into a named function for readability onto what this block is doing instead of describing in comments

packages/host-service/src/trpc/router/git/git.ts
```diff
@@ -112,11 +105,13 @@ export const gitRouter = router({
...
 
-			const againstBase = await getChangedFilesForDiff(git, [baseRef, "HEAD"]);
+			const againstBase = await getChangedFilesForDiff(git, [
+				`${baseRef}...HEAD`,
+			]);
 
 			// Staged — use status.files index character for correct status
 			const stagedNumstat = parseNumstat(
-				await git.raw(["diff", "--numstat", "--cached"]).catch(() => ""),
+				await git.raw(["diff", "--numstat", "-z", "--cached"]).catch(() => ""),
```
Kitenite, line 114: What's -z for?

packages/host-service/src/runtime/git/refs.ts
saddlepaddle: resolveUpstream is new in this PR — grep shows the only definition is in runtime/git/refs.ts. Happy to move/consolidate if #3548 or another open PR introduced its own version.

docs/V2_WORKSPACE_DIFF_VIEWS.md
saddlepaddle: Fair — reworded to drop the framing, now just "maps each UI surface to the git command it runs".

packages/host-service/src/trpc/router/git/utils/git-helpers.ts
saddlepaddle: simple-git's diffSummary() tokenizes on whitespace and doesn't expose a -z / NUL-delimited equivalent, so it breaks on filenames with spaces/quotes/newlines (and pairs name-status with numstat ambiguously for renames). Falling back to git.raw([...-z]) is the reason this parser exists — switching to the simple-git helpers would regress the non-ASCII path coverage the tests added in this PR cover.

packages/host-service/src/trpc/router/workspace-creation/workspace-creation.ts
saddlepaddle: Different use site from #3548 — there we pass `${ref}^{commit}` as a worktree start-point (where `--no-track` is the right swap). Here we use it inside `rev-parse --verify --quiet` purely to test whether a remote-tracking ref resolves before preferring it over the local default. For a branch ref the `^{commit}` suffix is a no-op, so dropped it.

packages/host-service/src/trpc/router/workspace-creation/workspace-creation.ts
saddlepaddle: Good call — extracted as `preferUpstreamForLocalDefault(git, startPoint)` so the caller reads as a single step and the early-return conditions are named.

packages/host-service/src/trpc/router/git/git.ts
saddlepaddle: `-z` makes git output NUL-terminated records with raw (unquoted) paths. Without it, paths containing spaces/quotes/non-ASCII bytes come back as `"..."` with C-style escapes, which `parseNumstat` would have to re-decode — and rename pairs become ambiguous to line-split. -z also gives us the 3-cell shape for renames that `parseNumstat` keys on. Same reason it's used in the nameStatus/numstat reads in git-helpers.

PR #3604, opened 2026-04-21
fix(desktop): auto-derive project name on v2 folder import

AviPeltz:
## Summary
- Skip the "name this project" modal when a user imports a folder that doesn't match an existing cloud project — derive the name from the folder's basename instead
- Removes the `FolderFirstImportModal` component and its state machine; the hook now returns just `start()` and runs the create inline (matches the auto-setup path used when a cloud match is found)

## Test plan
- [ ] Open the v2 dashboard and trigger "Import existing folder"
- [ ] Pick a folder with no matching cloud project → project is created directly with the folder's basename as its name, no naming modal
- [ ] Pick a folder that does match a cloud project → setup runs immediately (unchanged)
- [ ] Cancel the folder picker → nothing happens, no modal

---
## Summary by cubic
Skip the “Name this project” modal on v2 folder import. If there’s no cloud match, we auto-create the project using the folder’s basename; matching cloud projects still set up immediately.

- **Refactors**
  - Removed `FolderFirstImportModal` and its index; deleted the naming flow.
  - Simplified `useFolderFirstImport` to derive names from the path and run create/setup inline; the hook now exposes only `start()`.

<sup>Written for commit fb44744542c7ac1f4710c6a0007fa77ffd76c34e. Summary will update on new commits.</sup>

## Release Notes

* **User Experience**
  * Folder import now automatically derives project names from folder paths instead of prompting for manual entry via dialog, streamlining the import workflow.


AviPeltz:
kiet did it too

PR #3648, opened 2026-04-22
feat(terminal): ack-based flow control + resize debouncing for v2 terminal

Kitenite:
## Summary

Two VSCode-derived performance improvements for the v2 terminal (PTY relayed over WebSocket). Both address pain points under network relay: bursty output floods the WS queue, and container drags cause repeated expensive reflows.

### 1. Ack-based flow control (vendored)

Without this, a fast producer (`npm install`, `find /`, big `cat`) queues unbounded frames in the WS buffer — Ctrl-C / typing lag spikes that don't recover until the producer stops.

- **Client** (`apps/desktop/src/renderer/lib/terminal/flow-control.ts`): `AckDataBufferer` + `FlowControlConstants`, byte-identical to VSCode's [`terminalProcessManager.ts`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/terminalProcessManager.ts) / [`terminal.ts`](https://github.com/microsoft/vscode/blob/main/src/vs/platform/terminal/common/terminal.ts). `const enum` → `const` object because our tsconfig has `isolatedModules: true`.
- **Server** (`packages/host-service/src/terminal/flow-control.ts`): pause/resume state from [`terminalProcess.ts`](https://github.com/microsoft/vscode/blob/main/src/vs/platform/terminal/node/terminalProcess.ts), factored from inline fields into `createFlowControlState` / `recordOutput` / `recordAck` / `clearUnacknowledged` so it's unit-testable without spawning a real pty. Upstream snippets preserved in JSDoc for diffability.
- **Wiring**:
  - Client creates a per-connection `AckDataBufferer`; `terminal.write(data, cb)` acks by `data.length` after xterm parses (matches VSCode's ack-after-parse pattern). The callback sends `{type:"ack",charCount}` upstream.
  - Server `pty.onData` calls `recordOutput` and pauses on crossing `HighWatermarkChars` (100k). New `"ack"` message calls `recordAck` and resumes below `LowWatermarkChars` (5k).
  - Socket disconnect clears unacked and force-resumes so a paused pty can't get stuck when the client vanishes (in-memory buffer already bounded by `MAX_BUFFER_BYTES`).

### 2. Resize debouncer (ported)

Currently the `ResizeObserver` calls `fitAddon.fit()` synchronously on every tick — during splitter/window drags this reflows the xterm grid on every frame.

- **Port** (`apps/desktop/src/renderer/lib/terminal/resize-debouncer.ts`): [`TerminalResizeDebouncer`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/terminalResizeDebouncer.ts)'s three-tier algorithm — immediate if buffer < 200 lines, `requestIdleCallback` if hidden, otherwise Y-immediate + X-debounced-100ms.
- Not an *exact* vendor because upstream depends on five VSCode base modules (`Disposable`/`MutableDisposable`, `getWindow`/`runWhenWindowIdle`, `@debounce(100)` decorator). Exact vendoring would pull ~500 lines of infrastructure and require TS experimental decorators. The port uses native `setTimeout` + `requestIdleCallback` and keeps the original source as JSDoc on each method for diffability.
- **Wiring**: `measureAndResize` splits `fitAddon.fit()` into `proposeDimensions()` + a debounced apply, with a no-change early-out. Initial attach / font change use `immediate: true`; ResizeObserver ticks use `immediate: false`. `detachFromContainer` flushes pending work so persisted dims are current.

## Test plan

- [x] `bun test` on new files — 35 pass across the three test files (flow-control client 9, flow-control server 14, resize debouncer 12)
- [x] `bun run typecheck` — 25/25 workspaces pass
- [x] `biome check` clean
- [ ] Manual smoke: `yes | head -n 1000000` stays responsive during, Ctrl-C bounded after
- [ ] Manual smoke: splitter drag across a terminal no longer reflows on every frame

## Notes

- Tests are fresh — VSCode ships no unit tests for either set of code (only empty `acknowledgeDataEvent() {}` stubs in mocks).
- Originally split across #3648 + #3657 (now closed); folded into one PR per request.

## Summary by CodeRabbit

* **New Features**
  * Ack-based terminal flow control with batched client acknowledgements and a resize debouncer to reduce redundant/expensive terminal reflows.

* **Bug Fixes**
  * Prevents terminals from remaining paused after disconnects and fixes edge cases around pause/resume thresholds and ack rounding.

* **Tests**
  * Added comprehensive tests for ack batching, pause/resume behavior, watermark constants, and resize debouncing.

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/flow-control.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/flow-control.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, mock, test } from "bun:test";
+import { AckDataBufferer, FlowControlConstants } from "./flow-control";
+
+// Upstream VSCode (terminalProcessManager.ts) ships `AckDataBufferer` without
+// dedicated unit tests. The tests below pin the documented behavior so we
+// notice if a future vendor refresh changes semantics.
+
+describe("AckDataBufferer", () => {
+	test("does not fire below CharCountAckSize", () => {
+		const cb = mock();
+		const bufferer = new AckDataBufferer(cb);
+		bufferer.ack(100);
+		bufferer.ack(1000);
+		bufferer.ack(3899);
+		expect(cb).not.toHaveBeenCalled();
+	});
+
+	test("does not fire at exactly CharCountAckSize (strictly-greater-than)", () => {
+		// Upstream uses `while (count > CharCountAckSize)`, not `>=`.
+		const cb = mock();
+		const bufferer = new AckDataBufferer(cb);
+		bufferer.ack(FlowControlConstants.CharCountAckSize);
+		expect(cb).not.toHaveBeenCalled();
+	});
+
+	test("fires once when crossing CharCountAckSize by a single char", () => {
+		const cb = mock();
+		const bufferer = new AckDataBufferer(cb);
+		bufferer.ack(FlowControlConstants.CharCountAckSize + 1);
+		expect(cb).toHaveBeenCalledTimes(1);
+		expect(cb).toHaveBeenCalledWith(FlowControlConstants.CharCountAckSize);
+	});
+
+	test("accumulates across multiple acks", () => {
+		const cb = mock();
+		const bufferer = new AckDataBufferer(cb);
+		bufferer.ack(3000);
+		expect(cb).not.toHaveBeenCalled();
+		bufferer.ack(3000);
+		expect(cb).toHaveBeenCalledTimes(1);
+		expect(cb).toHaveBeenCalledWith(FlowControlConstants.CharCountAckSize);
+	});
+
+	test("fires multiple times for a single large ack", () => {
+		const cb = mock();
+		const bufferer = new AckDataBufferer(cb);
+		// 15001 chars: 3 full batches of 5000 should drain.
+		bufferer.ack(FlowControlConstants.CharCountAckSize * 3 + 1);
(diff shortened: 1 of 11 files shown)


Kitenite:
Addressed the 3 substantive findings (4 bot comments, deduplicated — 3 unique issues):

**1. `FlowControlConstants` duplication** (greptile) — moved to `@superset/shared/terminal-flow-control`. Client and server now import the same source of truth; a future vendor refresh updates one file.

**2. `charCount` validation** (coderabbitai, cubic) — server ack handler now rejects non-finite, non-positive, or non-numeric values before calling `recordAck`. `NaN` would have poisoned the arithmetic and permanently disabled pause/resume for the session.

**3. Stale debounce timer on path transitions** (coderabbitai, cubic) — a real port-introduced bug. Upstream VSCode stores the debounce job and idle job in a single `MutableDisposable` slot; my two-slot port allowed a pending debounce to fire after the idle path applied the latest cols, reverting to the old value. Fixed by cancelling cross-path state on every transition (both directions, symmetric). Two regression tests added.

Also silently fixed the `"max" → "may"` JSDoc typo (both greptile + cubic flagged it). Upstream VSCode has the same typo; this is our first divergence from verbatim source, noted in the commit message.

All 37 flow-control + resize-debouncer tests pass.

Commit: 8659ec970

PR #3657, opened 2026-04-22
feat(terminal): port vscode resize debouncer for v2 terminal

Kitenite:
## Summary

- Ports VSCode's [`TerminalResizeDebouncer`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/browser/terminalResizeDebouncer.ts) to avoid reflow churn when the terminal container resizes rapidly (splitter drags, window resize, sidebar toggle). The three-tier algorithm:
  - **Small buffer / explicit immediate** → apply both cols and rows now
  - **Hidden** → defer to `requestIdleCallback` (with `setTimeout(0)` fallback)
  - **Visible + large buffer** → apply rows immediately (cheap), debounce cols 100ms (expensive reflow)
- Why a *port* and not an exact vendor like #3648's flow-control units: upstream depends on five VSCode base modules (`Disposable` / `MutableDisposable` from `base/common/lifecycle`, `getWindow` + `runWhenWindowIdle` from `base/browser/dom`, `@debounce(100)` from `base/common/decorators`). Exact vendoring would pull in ~500 lines of VSCode infrastructure and require TS experimental decorators. The port uses native `setTimeout` + `requestIdleCallback` and keeps the original source as JSDoc on each method for diffability against future VSCode revisions.
- **Not wired up yet.** Current `terminal-runtime.ts:189` calls `fitAddon.fit()` synchronously on every `ResizeObserver` tick. Follow-up will split that into propose-dimensions + `TerminalResizeDebouncer.resize(cols, rows, immediate)`.

## Test plan

- [x] `bun test apps/desktop/src/renderer/lib/terminal/resize-debouncer.test.ts` — 12 pass (36 expects)
- [x] `bun run typecheck` — 25/25 workspaces pass
- [x] `biome check` clean
- [ ] Manual smoke during splitter drag, once wired up in follow-up

## Notes

- Tests pin VSCode semantics explicitly: strict-`<` on the 200-line threshold, X-coalescing to last value under debounce, hidden-path coalescing via single-slot idle handle, `flush()` applies latest X/Y via `resizeBoth`, `dispose()` cancels pending without firing.
- Upstream ships no unit tests for this file.

---
## Summary by cubic
Ports VSCode’s terminal resize debouncer and wires it into the v2 terminal to cut reflow churn during fast resizes. Rows apply immediately, columns debounce for 100ms, and hidden resizes schedule via idle.

- **New Features**
  - Integrated into `terminal-runtime`: split `fitAddon.fit()` into `proposeDimensions()` + debounced apply with an early-out when dims are unchanged.
  - Initial attach and font changes use the immediate path; `ResizeObserver` ticks use the debouncer.
  - Flush on detach to persist the latest cols/rows; dispose cancels pending work.
  - Self-contained utility using native timers and `requestIdleCallback`, with unit tests pinning upstream semantics.

<sup>Written for commit 17823ac269322945204920cc09def10ee88b7ca1. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved terminal resize behavior: smarter debouncing that coalesces rapid resizes, applies immediate sizing when appropriate, avoids redundant work when hidden, and ensures initial and detach-time sizing are handled reliably.

* **Tests**
  * Added thorough tests covering immediate vs deferred resize paths, visibility-based scheduling, coalescing, flush behavior, and cleanup.

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/resize-debouncer.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/resize-debouncer.test.ts
@@ -1,0 +1,206 @@
+import { describe, expect, mock, test } from "bun:test";
+import {
+	DebounceMs,
+	type ResizeDebouncerCallbacks,
+	StartDebouncingThreshold,
+	TerminalResizeDebouncer,
+} from "./resize-debouncer";
+
+// Upstream VSCode (terminalResizeDebouncer.ts) ships without dedicated unit
+// tests. These tests pin the documented behavior so a future vendor refresh
+// surfaces any semantic drift.
+
+interface Harness {
+	debouncer: TerminalResizeDebouncer;
+	cb: {
+		isVisible: ReturnType<typeof mock>;
+		getBufferLength: ReturnType<typeof mock>;
+		resizeBoth: ReturnType<typeof mock>;
+		resizeX: ReturnType<typeof mock>;
+		resizeY: ReturnType<typeof mock>;
+	};
+}
+
+function createHarness(opts: {
+	visible?: boolean;
+	bufferLength?: number;
+	debounceMs?: number;
+}): Harness {
+	const cb = {
+		isVisible: mock(() => opts.visible ?? true),
+		getBufferLength: mock(() => opts.bufferLength ?? 1000),
+		resizeBoth: mock(),
+		resizeX: mock(),
+		resizeY: mock(),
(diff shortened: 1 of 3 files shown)


Kitenite:
Folded into #3648 per request to ship as one PR.

PR #3713, opened 2026-04-24
feat(desktop): port v1 workspace-creation loader to v2 pending page

AviPeltz:
## Summary
- Port v1's animated `KeypadLoader` + sliding `StepProgress` onto the v2 `/pending/$pendingId` route, replacing the plain dot-list.
- Split host-service progress from 3 buckets → 5 steps (`ensuring_repo` / `fetching_remote` / `creating_worktree` / `registering` / `finalizing`) so each key maps 1:1. Added `setProgress` calls are in-memory `Map` writes — **no added latency**.
- Add `useAnimatedInitStep` walker so fast local flows don't flash past the keypad animation; keypad stays mounted through the Electric-sync wait so there's no intermediate "Workspace ready — opening..." pane before `doNavigate` fires.

## Why
v2's pending page shipped with a minimal status line. v1's keypad/step UI is nicer and, with the finer host-service progress, now animates honestly — including the `git fetch` and the two cloud mutations that were previously buried inside `creating_worktree` / `registering`.

## Notes
- `StepProgress` gets an opt-out `animate` prop; v1 callers default to `animate={true}` and are unaffected.
- `fetching_remote` only fires when the start point is actually remote-tracking — local-only branches pass through that key pressed without an active beat, which matches reality.
- Walker target pins to `"ready"` on `status === "succeeded"` so the keypad reaches all-pressed even though host-service clears its progress map at the end of the mutation.

## Test plan
- [ ] Create a new workspace from a remote-tracking branch on a real repo — verify all 5 keys press sequentially with visible "active" beats.
- [ ] Create a new workspace from a local-only branch — verify key 2 (`fetching_remote`) passes through pressed without an active beat, others animate normally.
- [ ] Checkout an existing branch (non-PR) — verify same 5-key animation.
- [ ] Checkout a PR — verify key 2 is skipped (PR path uses `gh pr checkout` post-worktree-add), others animate.
- [ ] Adopt an existing worktree — verify no crash, UI still shows a sensible fallback (adopt has no host-side progress).
- [ ] Watch a slow-network create (throttle devtools) — walker should keep pace with real progress, not lag.
- [ ] On a fast local create, confirm there's no flash of a "Workspace ready — opening..." screen before route change.
- [ ] Retry after a failure — walker resets to `pending` (the `resetKey` threading via `pendingId`).

---
## Summary by cubic
Ports v1’s animated workspace-creation loader to the v2 pending page with a clear 5-step flow and a smoother handoff to the workspace. Navigation now opens as soon as the workspace is synced, without the “ready — opening…” flash.

- **New Features**
  - Reused `KeypadLoader` and `StepProgress` on `/pending/$pendingId`; added per-step messages and accessible labels.
  - Mapped host progress to 5 steps via `mapProgressToInitStep`; `useAnimatedInitStep` paces fast flows; keypad stays visible through sync; navigate as soon as synced.
  - Tightened progress polling to 200ms and included `pr-checkout`; keypad clicks honor mute/volume; success warnings are shown as toasts.
  - Host-service emits `fetching_remote` and `finalizing` and sets progress at each step; final keypad press is pinned when the mutation resolves.

- **Bug Fixes**
  - Removed the animation gate before navigation to avoid the intermediate “ready — opening…” screen.
  - Improved failed state with a `StepProgress` error icon, clearer copy, and a subtle Retry; walker resets when `pendingId` changes.

<sup>Written for commit 92e5e72d6e6a1f11532646691933da0b4cb63bb4. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added finer-grained progress stages for workspace creation and a mapping to the visible init steps.
  * New animated step controller to drive incremental progress displays.

* **Improvements**
  * Reworked initialization UI with animated progression, failure states, per-step messages, and improved accessibility labels.
  * Faster, broader progress polling and single-show warnings; keypad sound now honors saved mute/volume settings.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/pending/$pendingId/mapProgressToInitStep.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/pending/$pendingId/mapProgressToInitStep.ts
@@ -11,15 +11,13 @@ export interface HostServiceProgressStep {
  * `WorkspaceInitStep` enum so we can reuse the v1 KeypadLoader + StepProgress
  * components on the v2 pending page.
  *
- * Host-service emits five steps that line up 1:1 with the v1 keypad's five
- * keys. `fetching_remote` only fires when the start point is a remote-tracking
- * ref — for fully-local base branches, that key passes through pressed
- * without an active beat.
+ * Host-service emits five steps, but v2's meaningful user-facing work is in
+ * the latter three: worktree creation, registration, and finalizing/opening.
+ * Keep the first two keypad beats as quick preflight/fetch lead-ins, then map
+ * those latter three states onto the final three keypad keys.
  *
- * `finalizing` active maps to `ready` (all keys pressed) so the final key
- * visibly presses down before the pending row flips to `succeeded` and
- * navigates away; finalizing is fast (local sqlite insert + optional setup-
- * terminal spawn) and the brief pressed frame is what reads as "done".
+ * Navigation does not wait for this display mapping; it is purely
+ * presentational.
  */
 export function mapProgressToInitStep(
 	steps: HostServiceProgressStep[] | undefined,
@@ -40,7 +38,7 @@ export function mapProgressToInitStep(
 			case "registering":
 				return "copying_config";
 			case "finalizing":
-				return "ready";
+				return "finalizing";
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


AviPeltz:
Will change the animation to be fewer keycap steps once everyone has migrated to v2.

PR #3840, opened 2026-04-28
fix(desktop): listen for terminal pane removal at the store layer

Kitenite:
## Summary

The previous architecture tried to detect pane removals by diffing — either rendered tabs (Workspace.tsx `onRemoved` hook) or persisted `v2WorkspaceLocalState` rows (`useGlobalTerminalLifecycle`, 322 lines of grace timers + workspace-presence checks + a background-intent flag). Both layers conflated user-initiated close with incidental row movement (workspace switch, sleep/wake, sync hydration), so terminal cleanup either fired prematurely or accreted heuristics to avoid doing so.

This PR moves the listener to the only layer that knows the difference: the zustand store actions themselves.

- Add `store.subscribePaneRemovals(listener)` to `@superset/panes`. `closePane` / `removeTab` / `replacePane` notify; `replaceState` (workspace switch / sync hydration) does NOT.
- `Workspace.tsx` subscribes and dispatches to `PaneDefinition.onRemoved` — replacing the previous diff `useEffect`.
- Terminal `PaneDefinition.onRemoved` disposes the runtime and kills the session, consulting `terminal-background-intents` for the "move to background" path.
- Delete `useGlobalTerminalLifecycle` (-322 lines) and its layout mount.
- Browser lifecycle is unchanged — its global-sweeper hook predates this API and can be migrated in a follow-up.

Net: -214 lines, including 5 new store tests that lock in the no-fire-on-`replaceState` contract.

## Test plan

- [ ] Close a terminal pane (X button, hotkey, middle-click) — session is killed, runtime disposed
- [ ] Close a tab containing terminal panes — every terminal session in the tab is killed
- [ ] "Move to background" from the terminal pane header — pane closes, session stays alive, terminal still listed in the dropdown
- [ ] Switch sessions in the terminal dropdown — old runtime released without dispose, new session attaches
- [ ] Switch v2 workspaces back and forth — terminals in the destination workspace are NOT killed (this was the original failure mode of the in-renderer diff and persisted-state diff)
- [ ] Sleep / wake the machine with terminal panes open — no spurious kills
- [ ] Drag a terminal pane to another tab / split — runtime survives the move
- [ ] Browser pane lifecycle still works (open / close / cross-tab move)

## Summary by CodeRabbit

* **New Features**
  * Pane lifecycle wiring added so pane definitions receive post-close notifications.

* **Bug Fixes**
  * Terminal cleanup improved: renderer performs best-effort silent session termination to reduce noisy failures.

* **Refactor**
  * Centralized pane-removal handling via store-driven subscriptions; removed legacy global terminal lifecycle mounting.
  * Pane lifecycle callback semantics changed to fire after close (replacing the old removal hook).

* **Tests**
  * Added tests for pane-removal subscription behavior and resilience.

* **Documentation**
  * Clarified lifecycle/sweeper comments.


apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/usePaneRegistry.tsx
```diff
@@ -277,6 +295,15 @@ export function usePaneRegistry(
 			terminal: {
 				getIcon: () => <TerminalSquare className="size-3.5" />,
 				getTitle: () => "Terminal",
+				onRemoved: (pane) => {
```
saddlepaddle: we have an onBeforeClose, this should be onAfterClose

Kitenite: Done — renamed in 62f049d (onRemoved → onAfterClose) and the comment text was updated to match in 1f4c6b6e6.

packages/panes/src/react/components/Workspace/Workspace.tsx
```diff
@@ -39,20 +40,21 @@ export function Workspace<TData>({
 		}
 		for (const [prevId, prevPane] of previousPanesRef.current) {
 			if (!current.has(prevId)) {
-				registry[prevPane.kind]?.onRemoved?.(prevPane);
+				registry[prevPane.kind]?.onAfterClose?.(prevPane);
```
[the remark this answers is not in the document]
Kitenite: Mitigated by the per-workspace store change in 1f4c6b6e6. Switching workspaces no longer reuses one store with a `replaceState` swap — `WorkspaceContent` is keyed on `workspaceId` so each workspace has its own store, and switching unmounts/remounts cleanly. Workspace-switch transitions can no longer trigger this diff at all.

`replaceState` still runs for sync-driven hydration within a stable workspace. In that case, the diff firing for a removed pane is the desired behavior — another device authoritatively removed the pane, and the local cleanup (e.g., disposing a terminal runtime entry) is correct. Initial hydration goes from empty → populated, which the diff sees as additions only.

PR #3894, opened 2026-04-30
fix(desktop): suppress duplicate terminal query response leak in v2 terminal

### LINKED ISSUE #3499: [bug] terminal inside superset buggy when I/O required
Hey guys
I love the product, one week in and I am a power user now. That being said, I come across a lot of bugs and issues in daily usage, and I genuinely want this product to improve, so I thought instead of restarting every time I face an issue, I'll try sending it out to you guys.

Sending out a low-hanging but irritating issue that I just faced, I hope you guys see and fix this.

When using the terminal inside Superset, every time I run a command (it can be a third-party CLI) where, post-running the command, any input is required from me, I am unable to do that because, for some reason, the output or response from the third-party CLI is not handled correctly by the native Superset terminal. There is an ['\x1b''] overflow in the response every time.

Image

I faced it with gh, vercel and a few other cli's

tvrmsmith:
**Links**
- Closes: #3499

## Summary
- Wire the existing `suppressQueryResponses` helper into v2 terminal runtime so it matches v1's behavior. Without this, every v2 terminal pane lets renderer xterm auto-reply to DA/OSC color queries and the duplicate reply round-trips back through the PTY.
- Strengthen the helper to suppress at the **query** side (not just response side) for sequences where query and response share a final byte (DA1, DA2, OSC 10/11/12). Response-side suppression alone cannot stop the leak for these.
- Add 13 unit tests using `@xterm/headless` covering each query class and the cleanup path.

## Why / Context
In Superset's split terminal architecture, the daemon hosts a `HeadlessEmulator` that owns the canonical reply to terminal queries (DA, DSR, OSC 10/11/12, etc.) and writes the response back into the PTY. The renderer xterm sees the same PTY output and would also auto-reply via `terminal.onData → WebSocket → daemon Session.write() → PTY stdin`.

`session.ts` filters renderer-originated escape replies, but only while `shellReadyState === "pending"`. Once the shell is ready, those duplicate responses pass through. The kernel PTY driver then echoes them as visible text at the next interactive prompt of any line-edited CLI — e.g. `spacectl profile login` shows:

```
Enter your tenant subdomain: ^[[?62;4;9;22c^[]11;rgb:0c0c/0e0e/1414^[\
```

Symptom: every interactive CLI in a Superset v2 terminal silently corrupts its first prompt with escape garbage. OMP's interactive bash tool consistently fails to parse prompts as a result. v1 terminals were unaffected because they already called this helper in `helpers.ts:150`; v2 terminals never got the same treatment.

## How It Works
Two suppression strategies, picked per-sequence based on whether the query and response share a final byte:

1. **Suppress at the query side** — register a parser handler matching the **query's** final byte and return `true`. xterm's default handler is bypassed, no `onData` reply is generated, nothing reaches the PTY. Used for **DA1 (`CSI c`), DA2 (`CSI > c`), OSC 10/11/12 color queries**. Safe because the daemon's headless emulator still answers.
2. **Suppress at the response side** — register a handler matching the response's distinguishing bytes. Used for sequences where the response has a unique final byte vs. the query: **CPR (`CSI R`), focus reports (`CSI I/O`), DECRPM (`CSI $y`)**.

OSC color handlers gate on payload prefix: only suppress when payload starts with `?` (query). Set commands (`OSC 11;rgb:...`) still take effect so themes propagate.

The wiring in `terminal-runtime.ts` calls `suppressQueryResponses(terminal)` right after attaching the key event handler, and folds the disposer into `_disposeAddons` so cleanup runs alongside the rest of the runtime teardown.

### Terminal Features
- [x] `spacectl profile login` no longer corrupts its prompt; full interactive flow completes
- [x] Per-query repro shows zero reply bytes for DA1, DA2, OSC 10, OSC 11, OSC 12 individually
  ```
  bash -c 'stty -echo; printf "\e[c"; sleep 0.4; reply=""; while IFS= read -r -t 1 -n 1 ch; do reply+="$ch"; done; stty echo; echo "${#reply} bytes"'
  ```
- [x] Combined probe (`printf '\e[c\e[>c\e]10;?\a\e]11;?\a\e]12;?\a'; read -p`) — no `?62;4;9;22c` or `]11;rgb:...` text in prompt
- [x] DSR cursor position query (`CSI 6n`) still gets a CPR reply (regression guard — confirms suppression isn't over-broad)
- [x] OMP's interactive bash tool no longer trips on prompt parsing
- [x] Terminal pane closes/disposes cleanly (helper returns disposer, wired into `_disposeAddons`)

## Testing
- `bun run typecheck` (`@superset/desktop` — passes)
- `bunx biome check` on the 3 touched files (passes)
- `bun test apps/desktop/src/.../suppressQueryResponses.test.ts` — 13 pass
- Manual repro in dev electron with new dev `terminal-host` daemon

## Design Decisions
- **Why suppress at query side (not response side) for DA/OSC**: DA's query and response both end in `c`; OSC 10/11/12 query and set commands share the same OSC number. A response-side filter cannot distinguish them. Query-side suppression catches the input before xterm decides to reply, which is both simpler and removes the leak at its source.
- **Why keep response-side suppression for CPR/focus/DECRPM**: their response final bytes (`R`, `I`/`O`, `$y`) are distinct from any query, so response-side is unambiguous and avoids accidentally swallowing legitimate queries we don't own.
- **Why not suppress in the daemon's `Session.write()` filter**: that filter exists for the pre-shell window. Permanently filtering all renderer→PTY escape replies would break legitimate paste/typed escape input. Suppressing at the renderer parser level removes only the auto-replies we don't want, while keeping user input intact.

## Known Limitations
- Existing terminal pane runtimes created **before** the fix loaded keep their old, leak-prone runtime instance (the runtime registry caches per-pane state across HMR). Newly-opened panes after this PR ships pick up the fix automatically; users with a long-lived pane may need to close and reopen it once.

## Risks / Rollout
- **Risk:** Low. The helper is a renderer-only behavior change; it never touches the daemon, the PTY, or persistence. Worst-case rollback affects only what xterm does with 6 specific escape sequence classes.
- **Rollout:** Ship in next desktop release. No flag, no migration, no cross-package dependency.
- **Rollback:** Revert this commit. v2 terminals revert to leaking behavior; v1 terminals unaffected.

---
## Summary by cubic
Suppresses duplicate terminal query replies in v2 by installing parser hooks and cleaning them up with the runtime, preventing prompt corruption and matching v1 behavior. Also adds a design spec for a Vertex AI provider in `getSmallModel` (docs only). Closes #3499.

- **Bug Fixes**
  - Wired `suppressQueryResponses` into the v2 runtime and dispose via `_disposeAddons`.
  - Suppress queries for DA1 (`CSI c`/`CSI 0c`), DA2 (`CSI > c`), and OSC 10/11/12 when payload starts with `?`; set commands still work.
  - Keep response-side suppression for CPR (`CSI R`), focus reports (`CSI I/O`), and DECRPM (`CSI ?…$y`); DSR (`CSI 6n`) still replies.
  - Added `@xterm/headless` tests covering suppression behavior and cleanup.

- **Migration**
  - Close and reopen existing terminal panes to pick up the fix.

<sup>Written for commit 0a5eb9a9d58ee4229c0aab3de1f732bb84247321. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/3894?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded terminal suppression to cover additional terminal query responses (reducing duplicate/noisy escape output) while allowing legitimate settings to apply.
  * Ensured suppression is consistently cleaned up when the terminal runtime is disposed.

* **Tests**
  * Added tests exercising many escape-sequence queries, verifying suppression, expected non-suppressed replies, and disposal/resume behavior.

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/terminal-runtime.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-runtime.ts
@@ -3,6 +3,7 @@ import type { ProgressAddon } from "@xterm/addon-progress";
 import type { SearchAddon } from "@xterm/addon-search";
 import { SerializeAddon } from "@xterm/addon-serialize";
 import { Terminal as XTerm } from "@xterm/xterm";
+import { suppressQueryResponses } from "renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/suppressQueryResponses";
 import { DEFAULT_TERMINAL_SCROLLBACK } from "shared/constants";
 import {
 	applyTerminalFontFamilyCssVariable,
@@ -214,6 +215,12 @@ export function createRuntime(
 
 	installTerminalKeyEventHandler(terminal);
 
+	// Suppress duplicate query responses (DA1, DA2, OSC 10/11/12 query reply, etc.)
+	// — daemon HeadlessEmulator owns canonical replies; without this, renderer xterm's
+	// auto-reply leaks back through the PTY and gets echoed at the next interactive
+	// prompt of any line-edited CLI. See suppressQueryResponses.ts.
+	const disposeQuerySuppression = suppressQueryResponses(terminal);
+
 	// Activate Unicode 11 widths (inside loadAddons) before restoring the buffer,
 	// else CJK/emoji/ZWJ widths get baked wrong into the replay. (#3572)
 	const addonsResult = loadAddons(terminal);
@@ -241,7 +248,10 @@ export function createRuntime(
 		_disposeResizeObserver: null,
 		lastCols: cols,
 		lastRows: rows,
-		_disposeAddons: addonsResult.dispose,
+		_disposeAddons: () => {
+			disposeQuerySuppression();
+			addonsResult.dispose();
+		},
 		_disposeImagePasteFallback: disposeImagePasteFallback,
 	};
 }
--- a/apps/desktop/src/renderer/lib/terminal/terminal-runtime.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-runtime.ts
@@ -3,6 +3,7 @@ import type { ProgressAddon } from "@xterm/addon-progress";
 import type { SearchAddon } from "@xterm/addon-search";
 import { SerializeAddon } from "@xterm/addon-serialize";
 import { Terminal as XTerm } from "@xterm/xterm";
+import { suppressQueryResponses } from "renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/suppressQueryResponses";
 import { DEFAULT_TERMINAL_SCROLLBACK } from "shared/constants";
 import {
 	applyTerminalFontFamilyCssVariable,
@@ -213,6 +214,12 @@ export function createRuntime(
 
 	installTerminalKeyEventHandler(terminal);
 
+	// Suppress duplicate query responses (DA1, DA2, OSC 10/11/12 query reply, etc.)
+	// — daemon HeadlessEmulator owns canonical replies; without this, renderer xterm's
+	// auto-reply leaks back through the PTY and gets echoed at the next interactive
+	// prompt of any line-edited CLI. See suppressQueryResponses.ts.
+	const disposeQuerySuppression = suppressQueryResponses(terminal);
+
 	// Activate Unicode 11 widths (inside loadAddons) before restoring the buffer,
 	// else CJK/emoji/ZWJ widths get baked wrong into the replay. (#3572)
 	const addonsResult = loadAddons(terminal);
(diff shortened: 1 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Document Vertex AI small model design](https://stagereview.app/superset-sh/superset/pull/3894/chapters/1) |
| 2 | [Enhance terminal query suppression logic](https://stagereview.app/superset-sh/superset/pull/3894/chapters/2) |
| 3 | [Wire query suppression into terminal runtime](https://stagereview.app/superset-sh/superset/pull/3894/chapters/3) |
| 4 | [Verify suppression with headless terminal tests](https://stagereview.app/superset-sh/superset/pull/3894/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/3894">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 0a5eb9a9d58ee4229c0aab3de1f732bb84247321 on Jun 30, 2026 4:09pm UTC.</sub>

PR #3970, opened 2026-05-02
feat(desktop): release-desktop accepts a commit SHA (canary-style)

saddlepaddle:
## Summary
- Extends `apps/desktop/create-release.sh` with an optional positional commit SHA argument so a desktop release can be cut from any commit without touching the caller's branch — mirroring `scripts/release-canary.sh`.
- When a commit is passed, the script provisions a detached worktree at that commit, applies the version bump there, pushes a temp branch `release-desktop-v<version>-<short-sha>`, and tags the temp branch tip with `desktop-v<version>` to trigger `release-desktop.yml`.
- Existing zero-arg / version-only invocations are unchanged.

Usage:
```
./apps/desktop/create-release.sh 1.7.4 58a3f7e8        # release that commit as 1.7.4
./apps/desktop/create-release.sh 1.7.4 58a3f7e8 --publish
```

## Test plan
- [x] Triggered a draft release of commit `58a3f7e8` as `desktop-v1.7.4` via this script — workflow run https://github.com/superset-sh/superset/actions/runs/25239502686, temp branch `release-desktop-v1.7.4-58a3f7e8e` and tag `desktop-v1.7.4` both pushed at `2db93441` (bump on top of `58a3f7e8e`).
- [ ] Re-run with a no-op (commit already at target version) — verifies the "skipping bump" branch.
- [ ] Re-run without a commit arg — verifies the existing branch-based flow is untouched.

---
## Summary by cubic
Add commit-SHA support to `apps/desktop/create-release.sh` so we can cut a desktop release from any commit without switching branches, mirroring the canary flow. Also tightens input guards to prevent invalid usage.

- **New Features**
  - Accepts: [version] [commit] [--publish] [--merge] (commit is optional).
  - If commit is provided: creates a detached worktree, bumps version if needed, pushes `release-desktop-v<version>-<short-sha>`, and tags `desktop-v<version>` to trigger `release-desktop.yml`.
  - Skips the bump if the commit already matches the target version and cleans up the temp worktree.

- **Bug Fixes**
  - Rejects non-semver `version` to prevent accidental `./create-release.sh <sha>` misuse.
  - Warns that `--merge` has no effect with a commit SHA; the temp release branch remains on origin.
  - Restores the `jq && mv` guard in the commit-based path for consistency.

<sup>Written for commit 41af689caa3d1cf5d61d9f2aad52048ecb93363d. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Chores**
  * Release script now accepts an optional commit SHA to create commit-specific releases while preserving the caller’s working tree.
  * When releasing from a commit, the workflow bumps the version in a temporary worktree, pushes a temp release branch and tag, and leaves local changes untouched.
  * Passing --merge together with a commit SHA now warns that merge is not applied for commit-based releases.

* **Bug Fixes**
  * Added validation to prevent non-semver values from being treated as versions.

### DIFF
--- a/apps/desktop/create-release.sh
+++ b/apps/desktop/create-release.sh
@@ -4,12 +4,13 @@
 # Based on apps/desktop/RELEASE.md
 #
 # Usage:
-#   ./create-release.sh [version] [--publish] [--merge]
-#   Example: ./create-release.sh              # Interactive version selection
-#   Example: ./create-release.sh 0.0.1        # Explicit version
+#   ./create-release.sh [version] [commit] [--publish] [--merge]
+#   Example: ./create-release.sh                       # Interactive version selection
+#   Example: ./create-release.sh 0.0.1                 # Explicit version, current HEAD
+#   Example: ./create-release.sh 0.0.1 58a3f7e8        # Release commit 58a3f7e8 as 0.0.1
 #   Example: ./create-release.sh 0.0.1 --publish
-#   Example: ./create-release.sh --publish    # Interactive + auto-publish
-#   Example: ./create-release.sh --publish --merge  # Auto-publish and merge PR
+#   Example: ./create-release.sh --publish             # Interactive + auto-publish
+#   Example: ./create-release.sh --publish --merge     # Auto-publish and merge PR
 #
 # This script will:
 # 1. Prompt for version if not provided (patch/minor/major/custom)
@@ -28,6 +29,9 @@
 # - Draft by default for review before publishing
 # - Use --publish flag to auto-publish when build completes
 # - Use --merge flag to merge the PR and delete the branch after publishing
+# - Pass an optional commit SHA (canary-style) to release a specific commit
+#   without touching the current branch — the script provisions a temp release
+#   branch from that commit, applies the version bump there, and tags it.
 #
 # Requirements:
 # - GitHub CLI (gh) installed and authenticated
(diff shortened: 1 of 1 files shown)


saddlepaddle:
Addressed review feedback in 41af689ca:

- **CodeRabbit (semver validation)** — added a regex guard right after argparse so a non-semver positional value (e.g. `./create-release.sh 58a3f7e8`) errors out with a usage hint instead of tagging `desktop-v<sha>` and writing the SHA into `package.json`.
- **Greptile / cubic (`--merge` no-op)** — emit a `warn` when `--merge` is combined with a commit SHA, since the commit-based path never creates a PR and the temp release branch stays on origin until manually pruned. Kept it as a warning rather than a hard error so people can still use `--merge` for ergonomic reasons in mixed shell histories.
- **Greptile (jq/mv guard)** — restored the `&&` chain on the `jq → mv` in the commit-based path to match the existing style in the branch-based path.

Skipping CodeRabbit's other suggestion to "always create a unique commit even when the version is unchanged" — that loses the property that re-running the release at the same SHA produces the same output, which I think is more valuable than the workflow-lookup edge case it's working around.

Lint failure on this PR is pre-existing in `apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/hooks/useTerminalLifecycle.ts` (introduced in 58a3f7e8e on main) and unrelated to this change.

PR #4296, opened 2026-05-09
fix(trpc): prevent last-owner removal/demotion race condition via advisory lock

vaishcodescape:
## Solves Issue #4297 

Three organization mutation endpoints — `removeMember`, `leaveOrganization`, and `updateMemberRole` — each checked the owner count **before** performing the destructive operation, but without holding any lock. Two concurrent requests could both pass the count check and then both succeed, leaving the organization with zero owners.

## Solution

Introduce a `withOwnerMutationLock` helper that wraps the critical section in a PostgreSQL advisory transaction lock (`pg_advisory_xact_lock`). The lock key is derived from the organization ID via `hashtextextended`, making it org-scoped so concurrent mutations to **different** orgs are never serialized.

Under the lock, the member's role is re-read from the database before the count check, so a concurrent request that already demoted/removed the member is correctly handled without an error.

### Changes

- **`packages/trpc/src/router/organization/organization.ts`**
  - Import `dbWs` (the write-side Drizzle client) alongside the existing `db`.
  - Add `withOwnerMutationLock<T>(organizationId, fn)` — runs `fn` inside a transaction protected by a per-org advisory lock.
  - `removeMember`: wrap owner-removal path in `withOwnerMutationLock`; re-read role under lock before checking count; non-owners are removed on the fast path (no lock needed).
  - `leaveOrganization`: same pattern — owner path acquires lock and re-reads membership before the count check.
  - `updateMemberRole`: same pattern for demoting an owner — lock acquired only when `isDemotingOwner` is true; re-reads role under lock before count check.
  - Remove the now-redundant pre-lock `ownerCount` variable from `updateMemberRole`.

---
## Summary by cubic
Prevents orgs from losing their last owner during concurrent member changes by guarding owner-sensitive mutations with a per-org PostgreSQL advisory lock. Applies to `removeMember`, `leaveOrganization`, and `updateMemberRole`.

- **Bug Fixes**
  - Add `withOwnerMutationLock(orgId, fn)` using `pg_advisory_xact_lock` in a transaction via `dbWs` from `@superset/db/client`.
  - Re-read membership under the lock before counting owners; block actions that would remove or demote the last owner.
  - Use lock only on owner paths; non-owner paths stay fast.
  - Lock key is derived from org ID, so different orgs don’t block each other.

<sup>Written for commit 75b07eb89b0629ba95a8e3498fc94299d7f15c6c. Summary will update on new commits.</sup>

## Release Notes

* **Bug Fixes**
  * Enhanced organization member management to reliably prevent removal or demotion of the last owner during concurrent operations.
  * Improved validation when members leave an organization to ensure the operation succeeds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4296)

### DIFF
--- a/packages/trpc/src/router/organization/organization.ts
+++ b/packages/trpc/src/router/organization/organization.ts
@@ -1,5 +1,5 @@
 import { stripeClient } from "@superset/auth/stripe";
-import { db } from "@superset/db/client";
+import { db, dbWs } from "@superset/db/client";
 import { members, organizations } from "@superset/db/schema";
 import {
 	sessions as authSessions,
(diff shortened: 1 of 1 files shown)


vaishcodescape:
@Kitenite Please Review

PR #4492, opened 2026-05-13
chore(api): drop redundant matcher entry from CORS proxy

saddlepaddle:
## Summary

`apps/api/src/proxy.ts` had two matcher entries:

\`\`\`ts
matcher: [
  "/((?!_next|ingest|monitoring|<asset-exts>).*)",
  "/(api|trpc)(.*)",  // ← fully covered by the line above
],
\`\`\`

The second is redundant — the first wildcard already matches everything not in its exclusion list, including `/api/*` and `/trpc/*`. Next dedupes overlapping matchers so behaviour is identical, it's just dead config.

Dropping it leaves a single readable pattern.

## Scope

- No CORS or routing behaviour change.
- Originally this PR also added negative-lookahead exclusions for server-to-server endpoints (jwks, token, webhooks) to skip middleware on them and trim Vercel log volume. Pulled that back: chat-attachments turned out to be local-only per Avi's follow-up, so the matcher doesn't need finer-grained scoping right now. The /api log cleanup can come back later if the volume matters once #4490 + #4491 deploy.

## Test plan

- [ ] CI green
- [ ] Browser sign-in / tRPC still works from `app.superset.sh`, `admin.superset.sh`, desktop renderer (CORS unchanged)
- [ ] GitHub + Linear webhooks still deliver (still match the matcher exactly as before)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed middleware routing so proxy/CORS middleware applies to the correct set of requests while explicitly skipping internal Next.js paths, ingest/monitoring routes, and common static assets.

* **Performance**
  * Reduced unnecessary middleware execution on excluded paths to improve overall request handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4492)

### DIFF
--- a/apps/api/src/proxy.ts
+++ b/apps/api/src/proxy.ts
@@ -67,6 +67,5 @@ export default function proxy(req: NextRequest) {
 export const config = {
 	matcher: [
 		"/((?!_next|ingest|monitoring|[^?]*\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)",
-		"/(api|trpc)(.*)",
 	],
 };


saddlepaddle:
Misread the ask — going to land the endpoint removal in a fresh PR instead. Closing this.

PR #4565, opened 2026-05-14
fix(host-service): pin NODE_ENV=production at bundle time (fixes #4563)

### LINKED ISSUE #4563: Release CLI binaries ship with NODE_ENV baked in as "development"; dev-only branches in host-service run in production
## Summary

The released `superset-linux-x64` CLI binaries (`cli-v0.2.14` through `cli-v0.2.16`) have `process.env.NODE_ENV` substituted at bundle time with the literal string `"development"`. As a result, code paths in `packages/host-service` that are explicitly guarded for `NODE_ENV === "development"` in source — including the SIGTERM "clean iteration" handler in `serve.ts` and the non-detached, no-rotating-log pty-daemon spawn config in `DaemonSupervisor.ts` — run unconditionally in production builds.

Under any process supervisor (systemd, supervisord, pm2) this produces intermittent instability and prevents pty-daemon log rotation from ever happening on disk.

## Root cause

`packages/host-service/build.ts` invokes `Bun.build({...})` with no `define` option:

```ts
const result = await Bun.build({
  entrypoints: ["src/serve.ts"],
  target: "node",
  outdir,
  naming: "host-service.js",
  format: "esm",
  external: [/* native addons */],
});
```

Bun substitutes `process.env.NODE_ENV` references at bundle time based on the build process's own environment variable. `.github/workflows/build-cli.yml`'s "Build distribution" step sets `RELAY_URL`, `SUPERSET_API_URL`, `SUPERSET_WEB_URL` — but not `NODE_ENV` — so the bundle is produced with the runner's default (`development`).

## Source vs. bundled output

Source code is correct in all three places:

```ts
// packages/host-service/src/serve.ts:55
const isDev = process.env.NODE_ENV === "development";

// packages/host-service/src/daemon/DaemonSupervisor.ts:77
return env.NODE_ENV === "development";

// packages/host-service/src/daemon/DaemonSupervisor.ts:910
const isDev = process.env.NODE_ENV === "development";
```

After bundling, `dist/host-service.js` in the shipped tarball contains:

```js
// shipped host-service.js line 56704
const isProductionEnv = (process == null ? undefined : "development") === "production";

// shipped host-service.js line 378253 (DaemonSupervisor.ts:910 after substitution + constant folding)
const isDev2 = true;

// shipped host-service.js line 395231 (serve.ts:55 after substitution + constant folding)
const isDev2 = true;
```

Setting `NODE_ENV=production` at runtime — e.g. `Environment=NODE_ENV=production` in a systemd unit — has **no effect** because the value is constant-folded into the bundle. Confirmed at runtime by the startup log, which prints `NODE_ENV=development` regardless of the actual env:

```
[host-service] starting (org=…, port=…, NODE_ENV=development)
```

## Consequences in production deployments

1. **`devShutdown` handler runs on SIGTERM/SIGINT** (`serve.ts:55–~70`). Any SIGTERM — including a routine `systemctl restart` or transient signal — triggers:

   ```
   [host-service] dev-mode SIGTERM — stopping pty-daemon for clean iteration
   ```

   followed by `process.exit(0)`. This was designed for `bun dev` file-watcher reloads. In production it masks real shutdown semantics, can race with supervisor restart logic, and produces log churn that's tedious to grep through.

2. **pty-daemon spawned with `detached: false`** (`DaemonSupervisor.ts:940`) and `stdio: ["ignore", "pipe", "pipe"]` (line 912). The accompanying comment is unambiguous about the intent:

   > Prod: detached so PTYs survive host-service restarts via socket adoption. Dev: attached as defense-in-depth in case serve.ts's dev shutdown doesn't fire (e.g. host-service crash).

   In shipped binaries the prod branch never runs — pty-daemon dies with the parent every time, even on intentional production restarts. The rotating log file is never written; `pty-daemon.log` does not exist on disk in any of my installs. This likely contributes to #4003 (update killed all terminals), #3438 (keybindings lost after update), and #3611 (terminal/agent sessions wiped on rebuild).

## Reproduction

```bash
curl -fsSL https://superset.sh/cli/install.sh | sh
superset auth login
```

Run under systemd:

```ini
[Service]
Type=simple
Environment=NODE_ENV=production
ExecStart=/home/<user>/superset/bin/superset start
Restart=always
```

Then:

```bash
sudo journalctl -u superset | grep "host-service.*starting"
# → "[host-service] starting (org=…, port=…, NODE_ENV=development)"

sudo systemctl restart superset
sudo journalctl -u superset | grep "dev-mode SIGTERM"
# → "[host-service] dev-mode SIGTERM — stopping pty-daemon for clean iteration"

ls /home/<user>/.superset/host/*/pty-daemon.log
# → No such file or directory
```

## Fix

Two-line change in `packages/host-service/build.ts`:

```diff
 const result = await Bun.build({
   entrypoints: ["src/serve.ts"],
   target: "node",
   outdir,
   naming: "host-service.js",
   format: "esm",
+  define: {
+    "process.env.NODE_ENV": JSON.stringify("production"),
+  },
   external: [
     "better-sqlite3",
     ...
   ],
 });
```

This makes the release artifact deterministically production-mode regardless of the build host's environment, which is the safer default for a bundle that ships pre-compiled to end users. The `bun dev` workflow is unaffected because it runs `src/serve.ts` directly via `bun run`, not the bundled output.

The alternative — setting `NODE_ENV: production` in the "Build distribution" step of `.github/workflows/build-cli.yml` — also works but depends on CI configuration discipline and doesn't help anyone running `bun run build:host` manually.

## Possibly related open issues

- #4003 — Update killed all terminals (consistent with non-detached pty-daemon)
- #3438 — Custom keybindings lost after app update (same)
- #3611 — [bug] dev-mode daemon rebuild wipes terminal & agent sessions on every app restart (dev-mode behavior leaking into release artifacts)
- #4414 — Host goes offline silently after `TunnelClient.cleanupChannels` throws `InvalidAccessError` (separate root cause, but the dev-mode SIGTERM cascade can mask reconnect logic)

## Environment

- CLI: `cli-v0.2.14`, `cli-v0.2.15`, `cli-v0.2.16` (linux-x64) — all affected
- OS: Ubuntu 24.04.4 LTS and Debian 12 (bookworm), x86_64
- Hosts: 2, both under `systemd` with `Type=simple, Restart=always`
- Reproduced consistently across 5+ days of uptime

heidihowilson:
## Summary

Pins `process.env.NODE_ENV` to `"production"` in the host-service release bundle, fixing #4563.

`Bun.build` substitutes `process.env.NODE_ENV` references at bundle time based on the build host's env, which leaves it as `"development"` on CI runners. The result: dev-only branches in `serve.ts` (SIGTERM "clean iteration" handler) and `DaemonSupervisor.ts` (attached, non-rotating pty-daemon spawn) constant-fold to always-on in shipped binaries. Setting `NODE_ENV=production` at runtime can't override a constant-folded value.

## Change

Two-line addition to `packages/host-service/build.ts`:

```ts
define: {
  "process.env.NODE_ENV": JSON.stringify("production"),
},
```

The `bun dev` workflow is unaffected because it executes `src/serve.ts` directly via `bun run`, not the bundled output. Only the release-time `Bun.build()` path is changed.

## Verification

Built locally on `bun 1.3.8` against this branch. Comparison of relevant strings in `dist/host-service.js`:

**Before:**
```js
const isProductionEnv = (process == null ? undefined : "development") === "production";
const isDev2 = true;  // DaemonSupervisor.ts:910 path
const isDev2 = true;  // serve.ts:55 path
```

**After:**
```js
const isProductionEnv = (process == null ? undefined : "production") === "production";
const isDev2 = false;  // DaemonSupervisor.ts:910 path
const isDev2 = false;  // serve.ts:55 path
```

Unrelated occurrences of `process.env.MASTRA_DEV` and `env.NODE_ENV` (the t3-env runtime check) are unaffected — both correctly remain runtime-evaluated.

## Test plan

- [x] Local `bun run build:host` succeeds
- [x] Bundle output confirms expected substitution
- [ ] CI-built binary on a Linux host under systemd no longer prints `NODE_ENV=development` at startup
- [ ] CI-built binary no longer emits `[host-service] dev-mode SIGTERM` on `systemctl restart`
- [ ] CI-built binary writes `pty-daemon.log` to disk under `~/.superset/host/<org>/`

Last three checked items require a release build for CI to produce; happy to verify on a release tag if a maintainer can cut one.

---
## Summary by cubic
Pins `process.env.NODE_ENV` to "production" in the host-service release bundle so dev-only code paths never ship (fixes #4563), and adds a daily upstream sync GitHub Actions workflow.

<sup>Written for commit 88426170d016154791337f881240c971442c5b01. Summary will update on new commits.</sup>


heidihowilson:
Superseded by #4564, which was opened a few minutes earlier with the same fix plus a regression test and a refactor of `build.ts` to make `buildOptions` importable for testing. Closing this in favor of that one — no point making the maintainers pick.

For anyone reading: the underlying bug is still tracked in #4563, and the better PR to follow is #4564.

PR #4582, opened 2026-05-14
fix(desktop): stop preset-command leak in useV2WorkspaceRun

saddlepaddle:
## Summary

- Bisected renderer freezes on rapid workspace switches to commit `ac2dd469f` (#4335, "Add preset-backed workspace run")
- Root cause: `useV2WorkspaceRun` resolves preset commands through an async `useEffect` + `useState` pattern that produces a render cycle and an unbounded Promise/closure leak
- Fix: replace with a `useMemo` that calls the already-synchronous `resolvePresetCommands` directly — same output, no state, no effect, no cycle

## Root cause

In `useV2WorkspaceRun.ts`:

```ts
useEffect(() => {
  let cancelled = false;
  async function resolveCommands() {
    const entries = await Promise.all(matchedPresets.map(async (preset) => ({
      id: preset.id,
      commands: await resolvePresetCommands(preset),  // already sync, `await` is a no-op
    })));
    if (cancelled) return;
    const next: Record<string, string[]> = {};
    for (const entry of entries) next[entry.id] = entry.commands;
    setResolvedPresetCommandsById(next);  // fresh object every run → never equal → always re-renders
  }
  void resolveCommands();
  return () => { cancelled = true; };
}, [matchedPresets, resolvePresetCommands]);
```

In the parent (`useV2PresetExecution`):
- `matchedPresets = useMemo(..., [allPresets, projectId])` where `allPresets` comes from a `useLiveQuery` — every Electric sync emits a new array reference → new `matchedPresets` identity
- `resolvePresetCommands = useCallback(..., [agentCommandsById])` where `agentCommandsById` rebuilds on every `useV2AgentConfigs` refetch → new callback identity

Each tanstack-db update or agent-config refetch fires the effect; `setResolvedPresetCommandsById(newObject)` always re-renders; the re-render produces fresh upstream references; effect fires again. Each cycle allocates an async closure, a `Promise.all` chain, N preset-resolver promises, N PromiseReactions, and a state object. The `cancelled` flag suppresses stale `setState` but does **not** cancel in-flight promises — their captured closures live until GC.

The leak compounds on workspace switch because the `WorkspaceTrpcProvider` key remount forces every upstream query to refetch immediately.

## Diagnostic data

Heap snapshot diff between idle baseline and mid-leak showed:

- `Promise` instances: +513
- `PromiseReaction`: +454
- `EventListener`: +218 (retained, ~0.6/sec idle leak)
- `StackFrameInfo`: +1249 (V8 trace info from accumulated closures)

CDP `Performance.getMetrics` showed JS heap and listener counts climbing steadily during idle (no user input), then spiking 5–10× during a workspace switch.

`resolvePresetCommands` in `useV2PresetExecution.ts:76-84` is synchronous: it reads from an in-memory `Map<string, string>` keyed by `agentId`. There's no reason for the consumer to await it, store the resolved result, or rebuild that state via an effect — a single derived `useMemo` produces the same value with no side effects.

## Why this matches "introduced since 1.8.8"

`ac2dd469f` landed on May 9 2026, between v1.8.8 and v1.8.9. v1.8.8 has no `useV2WorkspaceRun`, so workspace switching in v1.8.8 does not trigger this code path. Bisect (`desktop-v1.8.8` good → `figure-out-crashes` bad) converged on this single commit.

## Test plan

- [ ] Open multiple v2 workspaces, switch rapidly between them — renderer stays responsive, no OOM
- [ ] Open a project with agent-backed presets, edit the agent command in `/settings/agents`, confirm preset commands still pick up the new value (lazy `useMemo` recomputes when `agentCommandsById` identity changes)
- [ ] Click the workspace Run button with no agent-backed preset → still runs preset.commands
- [ ] Click the workspace Run button with an agent-backed preset whose live agent has a custom command → still runs the agent command

---
## Summary by cubic
Fixes renderer freezes and a Promise/closure leak when switching v2 workspaces by making preset command resolution fully synchronous and removing the leaking async state in `useV2WorkspaceRun`. Commands now resolve via a memo using the `useV2AgentConfigs` cache, eliminating re-render loops.

- **Bug Fixes**
  - Replaced async `useEffect` + `useState` in `useV2WorkspaceRun` with `useMemo(resolvePresetCommands)`.
  - Made `resolvePresetCommands` synchronous and removed the live `getHostServiceClientByUrl(...).settings.agentConfigs.list.query()` fetch and no-op `await`s.
  - Updated types/callsites to return `string[]`; behavior unchanged—presets still reflect agent command updates.

<sup>Written for commit 30296d7d3f613df7fef95e3f11c1c5e8a31fe7aa. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Preset command resolution streamlined to run synchronously and rely on in-memory agent data, reducing background work and eliminating redundant lookups.
* **Bug Fixes / Reliability**
  * Fewer remote requests during preset execution, improving stability and making command resolution more consistent with fallback behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4582)

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/useV2PresetExecution/useV2PresetExecution.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/useV2PresetExecution/useV2PresetExecution.ts
@@ -5,7 +5,6 @@ import { useLiveQuery } from "@tanstack/react-db";
 import { useCallback, useMemo } from "react";
 import { useV2AgentConfigs } from "renderer/hooks/useV2AgentConfigs";
 import { buildAgentLaunchCommand } from "renderer/lib/agent-launch-command";
-import { getHostServiceClientByUrl } from "renderer/lib/host-service-client";
 import { useWorkspace } from "renderer/routes/_authenticated/_dashboard/v2-workspace/providers/WorkspaceProvider";
 import { useCollections } from "renderer/routes/_authenticated/providers/CollectionsProvider";
 import type { V2TerminalPresetRow } from "renderer/routes/_authenticated/providers/CollectionsProvider/dashboardSidebarLocal";
@@ -74,23 +73,16 @@ export function useV2PresetExecution({
 		[allPresets, projectId],
 	);
 
+	// `useV2AgentConfigs` is the cached source of truth for agent configs
+	// (`staleTime: Infinity`, invalidated on every Settings → Agents mutation),
+	// so resolving against the in-memory `agents` array is correct and
+	// synchronous. Re-fetching via the host-service client on every call would
+	// duplicate that query and pin this function async, which forced the
+	// previous consumer (`useV2WorkspaceRun`) into a re-render cycle.
 	const resolvePresetCommands = useCallback(
-		async (preset: V2TerminalPresetRow): Promise<string[]> => {
+		(preset: V2TerminalPresetRow): string[] => {
 			if (!preset.agentId) return preset.commands;
-
-			let resolveAgents = agents;
-			if (activeHostUrl) {
-				try {
-					resolveAgents =
-						await getHostServiceClientByUrl(
-							activeHostUrl,
-						).settings.agentConfigs.list.query();
-				} catch {
-					resolveAgents = agents;
-				}
-			}
-
-			const linkedAgent = findLinkedAgent(resolveAgents, preset.agentId);
+			const linkedAgent = findLinkedAgent(agents, preset.agentId);
 			const live =
 				linkedAgent && linkedAgent.command.trim().length > 0
 					? buildAgentLaunchCommand(linkedAgent)
@@ -98,7 +90,7 @@ export function useV2PresetExecution({
 			if (live) return [live];
 			return preset.commands;
 		},
-		[activeHostUrl, agents],
+		[agents],
 	);
 
 	const executePreset = useCallback(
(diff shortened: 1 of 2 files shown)


saddlepaddle:
Greptile caught a real bug — the first commit on this branch assumed `resolvePresetCommands` was sync, but it had been rewritten as async (with an inline live `agentConfigs.list.query()` fetch) in #4232. The `useMemo` was assigning `Promise<string[]>` to `commands`, which would have silently broken agent-backed workspace runs.

I pushed a follow-up that drops the inline live fetch and makes `resolvePresetCommands` sync again. The reasoning:

`useV2AgentConfigs` (the source for `agents`) is a `useQuery` with `staleTime: Infinity` and is explicitly invalidated on every Settings → Agents mutation — its own docstring says so:

> *"Configs only change via Settings → Agents mutations that invalidate this key — `staleTime: Infinity` keeps the startup prefetch warm across navigation instead of every consumer refetching on mount."*

The inline `getHostServiceClientByUrl(...).settings.agentConfigs.list.query()` inside `resolvePresetCommands` was calling the **same exact tRPC procedure** that hook already caches. It was duplicating the query on every consumer call, bypassing a cache that's already correct, and dragging the function into async-network-bound territory.

That async-ness is what forced the previous consumer (`useV2WorkspaceRun`) into the leaking effect+setState pattern in the first place — it was working around a problem that shouldn't have existed.

After this update:
- `resolvePresetCommands` is sync (lookup in the cached `agents` map)
- `useV2WorkspaceRun.resolvedMatchedPresets` is a plain `useMemo` — no state, no effect, no Promise
- `executePreset` drops the no-op `await`
- `UseV2WorkspaceRunArgs.resolvePresetCommands` type signature updated to return `string[]`

Re-trigger reviewers @greptile-apps when convenient.

PR #4611, opened 2026-05-15
[codex] Fix stale external worktree imports

### LINKED ISSUE #4608: [bug] External worktree silently hidden from import list when DB row exists at same path
#### Describe the bug

When a worktree is recreated externally (outside Superset) at a path that was previously imported, Superset silently excludes it from the "import existing worktree" list. The worktree shows up in `git worktree list --porcelain`, but Superset filters it out because a stale row in the `worktrees` table still references that path. There is no UI signal that anything is wrong — the worktree just never appears, even after refresh.

#### Steps to reproduce

1. Import an external worktree into Superset (creates a row in `worktrees` with that `path`).
2. Outside Superset, delete the worktree (e.g., `git worktree remove`, or a custom CLI script). The DB row is **not** cleaned up because Superset wasn't involved.
3. Outside Superset, recreate a worktree at the same path (different branch is fine).
4. In Superset, refresh the project's external worktree list.

**Expected:** the new worktree appears as available to import.
**Actual:** it never appears. No error, no warning.

#### Root cause

`getExternalWorktrees` in `apps/desktop/src/lib/trpc/routers/workspaces/procedures/git-status.ts` (around line 357–365) filters the live `git worktree list --porcelain` output against paths in the `worktrees` table:

```ts
if (trackedPaths.has(wt.path)) return false;
```

The DB row from the previous import is still present (its `id` no longer maps to anything on disk, but the path matches), so the new worktree is treated as "already imported" and dropped from the list.

#### Workaround

Delete the stale row directly:

```bash
sqlite3 ~/.superset/local.db "DELETE FROM worktrees WHERE path = '<worktree-path>';"
```

Then reload Superset.

#### Suggested fixes

A few options, roughly in increasing scope:

1. **Detect and surface the orphan**: when `getExternalWorktrees` finds a path that's in `trackedPaths` but the tracked worktree's `id` has no live correspondence (or the branch differs from what git reports), expose it in the response with a flag like `isOrphanedDbRecord: true`. The banner could then offer a "reconcile" action to drop the stale row and re-import.
2. **Startup/refresh sweep**: on project load or when the user hits refresh, prune `worktrees` rows whose `path` no longer exists on disk (`!existsSync(wt.path)`). Cheap and addresses the same problem.
3. **Full auto-sync** (#2662): broader feature request that would also cover this.

Option 1 or 2 would close the silent-failure gap without waiting on the bigger #2662 work.

#### Related issues

- #920 (closed) — same root cause (ghost rows in `worktrees`), but the trigger there is an in-app close+recreate sequence, not external worktree management.
- #2662 (open) — broader feature request for auto-detecting worktree changes.
- #2863 (open) — inverse direction: Superset-driven deletion leaves directories on disk.

#### Environment

- Superset desktop version: 1.8.1
- macOS (Darwin 25.3.0)
- Worktrees managed entirely via external CLI tooling

Kitenite:
## Summary

Fixes #4608.

This removes the stale path-only assumption from external worktree discovery. Superset now compares live Git worktrees against tracked rows by path and branch, so a worktree recreated externally at a previously imported path on a different branch can appear in the import list again.

Import/open flows reuse the existing row at that path instead of creating a duplicate. Active workspace paths remain protected: if a tracked worktree already has an open workspace, the external import path will not mutate that row out from under it. Direct open resolves the branch from `git worktree list`; the client no longer sends branch as input for that mutation.

## Validation

- `bun test apps/desktop/src/renderer/components/NewWorkspaceModal/components/PromptGroup/utils/resolveOpenableWorktrees/resolveOpenableWorktrees.test.ts apps/desktop/src/lib/trpc/routers/workspaces/utils/select-external-worktrees-for-import.test.ts apps/desktop/src/lib/trpc/routers/workspaces/utils/select-external-worktrees-for-import.integration.test.ts`
- `bun run lint`
- `bun run --cwd apps/desktop typecheck`
- `git diff --check`

## Summary by CodeRabbit

* **Refactor**
  * External worktree import now lists on-disk worktrees once and upserts tracked worktrees to preserve/refresh existing metadata instead of always inserting.
* **Bug Fixes**
  * Improved handling of active workspaces and branch mismatches to prevent duplicate or conflicting imports.
* **Tests**
  * Updated selection tests to reflect per-worktree tracked state (path, branch, active flag).
* **Breaking Changes**
  * Open-external-worktree input no longer accepts a branch; callers must supply only projectId and worktreePath.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4611?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ghost:

| | Title |
|---|-------|
| 1 | [Add worktree creation time utility](https://stagereview.app/superset-sh/superset/pull/4611/chapters/1) |
| 2 | [Update external worktree selection logic](https://stagereview.app/superset-sh/superset/pull/4611/chapters/2) |
| 3 | [Refactor worktree import and upsert logic](https://stagereview.app/superset-sh/superset/pull/4611/chapters/3) |
| 4 | [Resolve branch from disk in open flow](https://stagereview.app/superset-sh/superset/pull/4611/chapters/4) |
| 5 | [Expose active workspace status to UI](https://stagereview.app/superset-sh/superset/pull/4611/chapters/5) |
| 6 | [Update UI components and tests](https://stagereview.app/superset-sh/superset/pull/4611/chapters/6) |

<a href="https://stagereview.app/superset-sh/superset/pull/4611">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 1ced1ffa1dac47eda70bdece1ffb210cb25e72ae on May 16, 2026 10:12pm UTC.</sub>

PR #4612, opened 2026-05-16
fix(desktop): fall back from terminal WebGL on context loss

Kitenite:
## Summary
- marks terminal WebGL as unavailable after a WebGL context loss in both V1 and V2 terminal setup paths
- keeps the existing dispose + refresh behavior so the current terminal falls back to DOM rendering
- avoids clearTextureAtlas-based recovery

## Why
Clearing xterm's texture atlas can trigger the garbled glyph rendering issue in our app. A safer prevention path is to stop using WebGL after a context loss rather than rebuilding the atlas in place.

## Validation
- bun run lint -- apps/desktop/src/renderer/lib/terminal/terminal-addons.ts apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/helpers.ts
- bun run --cwd apps/desktop typecheck

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal stability by implementing automatic fallback to DOM rendering when WebGL context is lost. This ensures the terminal remains functional when hardware acceleration becomes unavailable, with subsequent instances using DOM rendering by default.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4612?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/terminal-addons.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-addons.ts
@@ -49,7 +49,6 @@ export function loadAddons(terminal: XTerm): LoadAddonsResult {
 		try {
 			webglAddon = new WebglAddon();
 			webglAddon.onContextLoss(() => {
-				suggestedRendererType = "dom";
 				webglAddon?.dispose();
 				webglAddon = null;
 				terminal.refresh(0, terminal.rows - 1);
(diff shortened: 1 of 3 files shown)


Kitenite:
Closing this after validation showed the automatic texture-atlas clear on terminal attach can itself trigger the garbled rendering issue. The branch now contains a revert and has no net diff; we should not merge this approach.

Kitenite:
Closing this after testing the actual WebGL context-loss path. Forcing `WEBGL_lose_context` shows the terminal already recovers on main via the existing xterm handler: the WebGL addon is disposed and the terminal refreshes.

The added fallback only changed future terminal creation after context loss, but it does not address the garbled glyph case we were trying to prevent. I removed that code so we do not ship an unproven behavior change.

PR #4706, opened 2026-05-19
fix: avoid heredoc prompt launches

### LINKED ISSUE #4705: [bug] Agent launch command uses bash-only heredoc syntax, fails in fish shell
#### Describe the bug

When launching an agent (e.g. Claude) in a workspace where the user's `$SHELL` is fish, Superset emits a bash-style command-substitution + heredoc to pass the prompt:

​```
claude --permission-mode acceptEdits "$(cat <<'SUPERSET_PROMPT_f96e72829be64afaa216cd1117284013'
test
SUPERSET_PROMPT_f96e72829be64afaa216cd1117284013
)"
​```

fish does not support heredocs — `<<` is an input-redirection operator — so the command fails immediately with:

​```
fish: Expected a string, but found a redirection
claude --permission-mode acceptEdits "$(cat <<'SUPERSET_PROMPT_f96e72829be64afaa216cd1117284013'
                                              ^
​```

The agent never starts. The same preset works fine when `$SHELL` is bash or zsh.

#### Expected behavior

Agent launch should either:
1. Always execute the prompt-transport command in a known-good POSIX shell (`bash -c '…'` / `sh -c '…'`) regardless of the user's interactive `$SHELL`, or
2. Detect the user's shell and emit shell-compatible syntax (fish has its own way to pass multi-line strings, e.g. `printf` piped via `psub`, or just env-var transport), or
3. Use a transport mode that doesn't depend on shell syntax at all — e.g. write the prompt to a temp file and pass the path, or pass via env var and have the agent read it.

#### To Reproduce

1. Set fish as the login shell: `chsh -s $(which fish)`
2. Open Superset, create a workspace
3. Launch the bundled Claude preset (or any agent preset using the heredoc prompt transport) with a non-empty prompt
4. Observe the fish parse error in the terminal pane; agent does not start

#### Environment

- Superset version: 1.9.7
- OS: macOS 15.x (Apple Silicon, MacBook Air M4)
- Shell: fish (version: `fish --version`)
- Agent: Claude Code (bundled preset, default config)

#### Additional context

Related: PR #4150 (`fix(host-service): don't crash startup when shell env probe fails`) suggests the host service is already shell-aware on startup, but the prompt-transport layer in the agent launch flow appears not to be.

Workaround for affected users: edit the agent preset's command to wrap it in `bash -c '…'`, or set the preset to use a non-heredoc prompt transport mode if one is exposed.

kgarg2468:
## Description

Fixes fish-shell agent launches by removing bash-only heredoc prompt transport from the generated agent commands.

- pass argv prompts as a single shell-quoted argument instead of `"$(cat <<...)"`
- pipe stdin prompts with `printf '%s\n' '<prompt>' | <agent>` instead of attaching a heredoc
- add shared and host-service regression coverage for quoted prompts and heredoc-free stdin launches
- update the V2 launch-context note that still described heredoc transport

## Related Issues

Closes #4705

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

- [x] `bun install --frozen --ignore-scripts`
- [x] `bun test packages/shared/src/agent-command.test.ts packages/shared/src/agent-launch-request.test.ts`
- [x] `bun test packages/host-service/src/trpc/router/agents/agents.test.ts`
- [x] `bash -lc <generated argv/stdin prompt commands>` and `zsh -fc <generated argv/stdin prompt commands>` smoke tests with multiline + single-quote prompt text
- [x] `bun run lint:fix`
- [x] `bun run lint`
- [x] `bun run --cwd packages/shared typecheck`
- [x] `bun run --cwd packages/host-service typecheck`
- [x] `bun run typecheck`
- [x] `bun run --cwd packages/shared test`
- [x] `bun run --cwd packages/host-service test`
- [x] `git diff --check`

N/A

## Additional Notes

`fish` is not installed in my local environment, so I could not execute the generated commands under fish directly. The generated commands no longer contain heredoc syntax or `$(cat ...)` for direct prompt launches, which is the syntax fish rejected in the reported failure.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4706">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes agent prompt launches in fish by removing heredocs and command substitution. Prompts are now shell-agnostic: argv prompts are a single quoted arg, stdin prompts use `printf`, and file prompts use `xargs`.

- **Bug Fixes**
  - stdin prompts: `printf '%s\n' '<prompt>' | <agent>` instead of heredocs.
  - argv prompts: pass prompt as one shell-quoted arg; escape single quotes.
  - file prompts (argv): `xargs -0 -I __SUP_PROMPT__ <command> __SUP_PROMPT__ < '<path>'` instead of `$(cat ...)`; stdin remains `< '<path>'`.
  - Updated command builders and tests in `packages/shared` and `packages/host-service`.
  - Updated `apps/desktop/docs/V2_LAUNCH_CONTEXT_GAPS.md` to reflect non-heredoc, non-substitution transport.

<sup>Written for commit 8c2ccda5b023a24367c28c3288ccd732d64137c6. Summary will update on new commits. <a href="https://cubic.dev/pr/superset-sh/superset/pull/4706?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prompt handling and shell-escaping so agent commands receive prompts safely and deterministically (no heredoc usage).

* **Tests**
  * Expanded and tightened tests to assert exact command strings and escaping across agent types and transports.

* **Documentation**
  * Updated user-facing docs to reflect the new prompt transmission and parsing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4706?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/desktop/docs/V2_LAUNCH_CONTEXT_GAPS.md
+++ b/apps/desktop/docs/V2_LAUNCH_CONTEXT_GAPS.md
@@ -35,9 +35,10 @@ empty strings. The pipeline otherwise works end-to-end.
    head branch. Prompt says so.
 3. **No body truncation** (or very high cap, e.g. 200 KB/source). Modern
    context windows are large. Don't cap aggressively.
-4. **No sanitization.** Prompt goes into a heredoc with a random
-   delimiter (no shell injection). Agent reads raw text, no HTML parser
-   downstream. V1's entity escaping was unnecessary.
+4. **No sanitization.** Prompt text is shell-quoted as a single argv value or
+   piped through `printf` for stdin transports (no prompt shell injection).
+   Agent reads raw text, no HTML parser downstream. V1's entity escaping was
+   unnecessary.
 5. **Attachments framing.** The `{{attachments}}` block includes a short
    header cueing the agent to read the files. Just paths; agent handles
    the rest.
--- a/packages/host-service/src/trpc/router/agents/agents.test.ts
+++ b/packages/host-service/src/trpc/router/agents/agents.test.ts
@@ -1,0 +1,48 @@
+import { describe, expect, it } from "bun:test";
+import { buildAgentCommandString } from "./agents";
+
+type AgentConfig = Parameters<typeof buildAgentCommandString>[0];
+
+function config(overrides: Partial<AgentConfig> = {}): AgentConfig {
+	return {
+		id: "agent-1",
+		presetId: "amp",
+		label: "Amp",
+		command: "amp",
+		args: [],
+		promptTransport: "stdin",
+		promptArgs: [],
+		env: {},
+		...overrides,
+	};
+}
+
+describe("buildAgentCommandString", () => {
+	it("pipes stdin prompts without heredoc syntax", () => {
+		const command = buildAgentCommandString(
+			config(),
+			"hello from fish\nit's safe",
+		);
+
+		expect(command).toBe(
+			"printf '%s\\n' 'hello from fish\nit'\\''s safe' | 'amp'",
+		);
+		expect(command).not.toContain("<<");
+	});
+
+	it("quotes argv prompts as a single shell argument", () => {
+		const command = buildAgentCommandString(
+			config({
+				command: "codex",
+				args: ["--dangerously-bypass-approvals-and-sandbox"],
+				promptArgs: ["--"],
+				promptTransport: "argv",
+			}),
+			"- prompt with spaces",
+		);
+
+		expect(command).toBe(
+			"'codex' '--dangerously-bypass-approvals-and-sandbox' '--' '- prompt with spaces'",
+		);
+	});
+});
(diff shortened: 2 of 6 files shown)


kgarg2468:
Thanks for the review. I pushed 8c2ccda5b to remove the `$(cat ...)` argv transport and use a shell-agnostic `xargs -0` handoff for file prompts, with regression coverage for quoted file paths and Codex.\n\nValidation run:\n- `bunx @biomejs/biome@2.4.2 check packages/shared/src/agent-prompt-launch.ts packages/shared/src/agent-command.test.ts`\n- `bun test packages/shared/src/agent-command.test.ts`

darivs:
Tested this branch under **fish** (the gap the PR description noted). Fix confirmed end-to-end. ✅

Ran `bun run dev:desktop` on `8c2ccda`, opened the desktop app, launched Claude Code with prompt `test` (argv transport) into a fish workspace. Generated command:

```
claude --permission-mode acceptEdits 'test'
```

fish parses the single-quoted argv form without complaint, `claude` boots, receives the prompt, and replies normally:

```
> test
● Received. What would you like to test or work on?
```

Pre-fix, the heredoc / `$(cat …)` transport would have errored on fish before claude was ever invoked. The shell-agnostic argv form works as intended.

PR #4713, opened 2026-05-19
perf(electric-proxy): index-back auth.apikeys shape via derived organization_id column

alco:
## Summary

The Electric shape for `auth.apikeys` currently filters with a leading-wildcard `LIKE` over the JSON-stringified `metadata` column:

```sql
"metadata" LIKE '%"organizationId":"' || $1 || '"%'
```

PR #4482 papered over the Postgres side with a `pg_trgm` GIN index, but two problems remain:

1. **Postgres**: trgm is a fuzzy text index — it produces a bitmap of *candidates* that still need rechecking against the full `LIKE` pattern. It is strictly worse than a B-tree equality lookup.
2. **Electric**: Electric maintains its own internal indexes over shape predicates, but only when the WHERE clause is one of a small set of recognised forms (see [Optimized Where clauses](https://electric.ax/docs/sync/guides/shapes#optimized-where-clauses)). The recognised forms require a **direct column reference**; expressions and JSON operators (`->`, `->>`, `@>`, `?`, …) are explicitly *not* optimized. That means even if we migrated `metadata` to JSONB and indexed `(metadata->>'organizationId')`, Electric would still treat the shape as un-indexed and scan every row on its side.

This PR fixes both at once by adding a stored generated column `auth.apikeys.organization_id uuid` derived from `metadata`, B-tree indexing it, and switching the proxy WHERE clause to `"organization_id" = $1` — the canonical optimized form on both sides.

## What changes

- **`packages/db/src/schema/auth.ts`** — adds a `STORED` generated column on `auth.apikeys`:

  ```sql
  organization_id uuid GENERATED ALWAYS AS (
    CASE
      WHEN metadata IS NULL OR metadata = '' THEN NULL
      WHEN (metadata::jsonb->>'organizationId') ~* '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
        THEN (metadata::jsonb->>'organizationId')::uuid
      ELSE NULL
    END
  ) STORED
  ```

  and a B-tree index `apikeys_organization_id_idx` on it.

- **`apps/electric-proxy/src/where.ts`** — the `auth.apikeys` case becomes:

  ```ts
  case "auth.apikeys": {
      const fragment = `"organization_id" = $1`;
      return { fragment, params: [organizationId] };
  }
  ```

## Why this approach

We considered, and rejected, migrating `metadata` to JSONB. The original Electric-side suggestion proposed it, but a JSONB migration alone does **not** unlock an indexed Electric shape: Electric still wouldn't optimize a predicate like `metadata->>'organizationId' = $1`. Worse, Better Auth's `apikey` plugin does `JSON.stringify` on write and `JSON.parse` on read, and node-postgres auto-parses JSONB into JS objects — so a JSONB column would break the plugin's read path with `SyntaxError: Unexpected token o in JSON at position 1`.

Keeping `metadata` as `text` preserves the Better Auth contract exactly. A `STORED` generated column extracted from `metadata` gives us:

- A real column Postgres can B-tree-index.
- A direct column reference that Electric's optimizer recognises.
- Automatic maintenance by Postgres on every INSERT/UPDATE of `metadata` — no trigger, no app-side hook, no risk of drift.

The `CASE` expression is defensive: it returns NULL for NULL/empty `metadata` and for any non-UUID-looking `organizationId`, so the STORED expression cannot raise on existing data. The regex guard is required because `text::uuid` raises on malformed input and that would block backfill.

## Validation we need from you before merging

Please run this query against production (or a recent restore) and confirm it returns zero rows:

```sql
SELECT count(*) AS bad_rows
FROM auth.apikeys
WHERE metadata IS NOT NULL
  AND metadata <> ''
  AND (
    NOT (metadata::jsonb IS NOT NULL)
    OR (
      (metadata::jsonb ? 'organizationId')
      AND (metadata::jsonb->>'organizationId') !~* '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
    )
  );
```

`bad_rows = 0` means the generated-column backfill will succeed for every existing row. If it returns non-zero, please share the offending shapes — we have two options on the table (backfill the rows, or relax the expression to coerce malformed values to NULL via a small `plpgsql` wrapper). The regex guard in the column expression already handles malformed *UUID strings*, so this query specifically surfaces rows where `metadata` itself is not parseable JSON.

## Migration

I have **not** generated the Drizzle migration in this PR because the project convention is to generate against a fresh Neon branch. To produce it on your side:

```bash
# point .env at a fresh Neon branch, then:
bunx drizzle-kit generate --name="apikeys_organization_id_generated_column"
```

The expected output is a single new file under `packages/db/drizzle/` containing the `ALTER TABLE ... ADD COLUMN ... GENERATED ALWAYS AS ... STORED;` plus `CREATE INDEX apikeys_organization_id_idx ON "auth"."apikeys" USING btree ("organization_id");`. If the generated SQL omits `STORED` (Postgres does not support `VIRTUAL` generated columns) or diverges materially, let me know and I'll adjust the Drizzle call — current Drizzle versions accept `.generatedAlwaysAs(sql, { mode: "stored" })` explicitly if needed.

## What is *not* in this PR (follow-up)

- The existing `apikeys_metadata_trgm_idx` is kept in place by this PR. Once you are comfortable that you will not need to roll back to the old `LIKE` predicate, drop it via a separate migration. I will prepare that as a small follow-up PR you can merge on your schedule.

## Open questions for you

1. (Validation) Does the query above return zero rows in production?
2. (API surface) Should `organizationId` be exposed on the Drizzle `apikeys` model? It will appear on `typeof apikeys.$inferSelect` automatically, which may or may not match how you prefer derived columns to surface in your type ecosystem. Happy to mark it internal/omit from selects if you'd prefer.

## Test plan

- [ ] `bun run typecheck`
- [ ] `bun run lint`
- [ ] `bun test`
- [ ] On a Neon branch: apply the generated migration, then `EXPLAIN (ANALYZE, BUFFERS) SELECT * FROM auth.apikeys WHERE organization_id = '<real-org-uuid>';` shows `Index Scan using apikeys_organization_id_idx` with sub-millisecond timing.
- [ ] On a Neon branch: subscribe to the `auth.apikeys` shape through the Electric proxy from a connected client and confirm no HTTP 400 from Electric (this was the failure mode that caused the revert in PR #4489 for the `auth.users` case; `"column" = $1` is the single best-supported Electric form and should be safe, but worth verifying end-to-end).
- [ ] Spot-check a handful of rows: `SELECT id, organization_id, (metadata::jsonb->>'organizationId') FROM auth.apikeys LIMIT 20;` — the two right-hand columns should agree on every row.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4713">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Replaced the `auth.apikeys` shape filter from a JSON `LIKE` on `metadata` to an indexed `"organization_id" = $1` equality for faster, reliable lookups across Electric and Postgres.

- **Refactors**
  - In `packages/db`, add STORED generated column `organization_id uuid` on `auth.apikeys` (guards NULL/empty/non-JSON and non-UUID) and index `apikeys_organization_id_idx`.
  - In `apps/electric-proxy/src/where.ts`, switch predicate to `"organization_id" = $1`.
  - Keep `apikeys_metadata_trgm_idx` for rollback.

- **Migration**
  - Include Drizzle migration `0054_apikeys_organization_id_generated_column.sql` to add the column and index. Apply normally.

<sup>Written for commit 56d655fe6e592cf89f62d45b2363ee265b2d00a4. Summary will update on new commits. <a href="https://cubic.dev/pr/superset-sh/superset/pull/4713?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Refactor**
  * Improved API key handling: added a stored organization identifier and index to make lookups by organization faster and more accurate, and updated query behavior to use that identifier for improved performance and reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4713?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/electric-proxy/src/where.ts
+++ b/apps/electric-proxy/src/where.ts
@@ -116,7 +116,7 @@ export function buildWhereClause(
 			return build(agentCommands, agentCommands.organizationId, organizationId);
 
 		case "auth.apikeys": {
-			const fragment = `"metadata" LIKE '%"organizationId":"' || $1 || '"%'`;
+			const fragment = `"organization_id" = $1`;
 			return { fragment, params: [organizationId] };
 		}
 
--- a/packages/db/drizzle/0054_apikeys_organization_id_generated_column.sql
+++ b/packages/db/drizzle/0054_apikeys_organization_id_generated_column.sql
@@ -1,0 +1,8 @@
+ALTER TABLE "auth"."apikeys" ADD COLUMN "organization_id" uuid GENERATED ALWAYS AS (CASE
+				WHEN metadata IS NULL OR metadata = '' THEN NULL
+				WHEN NOT (metadata IS JSON OBJECT) THEN NULL
+				WHEN (metadata::jsonb->>'organizationId') ~* '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
+					THEN (metadata::jsonb->>'organizationId')::uuid
+				ELSE NULL
+			END) STORED;--> statement-breakpoint
+CREATE INDEX "apikeys_organization_id_idx" ON "auth"."apikeys" USING btree ("organization_id");
--- a/packages/db/drizzle/meta/0054_snapshot.json
+++ b/packages/db/drizzle/meta/0054_snapshot.json
@@ -1,0 +1,6795 @@
+{
+  "id": "17bfa7f8-3ae9-4990-8264-bec535a04810",
+  "prevId": "c58833cb-bdb0-4663-81f9-ca66a01479a7",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "auth.accounts": {
+      "name": "accounts",
+      "schema": "auth",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "account_id": {
+          "name": "account_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider_id": {
+          "name": "provider_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "access_token": {
+          "name": "access_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token": {
+          "name": "refresh_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "id_token": {
+          "name": "id_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "access_token_expires_at": {
+          "name": "access_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token_expires_at": {
+          "name": "refresh_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "scope": {
+          "name": "scope",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "password": {
+          "name": "password",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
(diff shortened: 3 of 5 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Add generated organization_id column to apikeys](https://stagereview.app/superset-sh/superset/pull/4713/chapters/1) |
| 2 | [Generate database migration for apikeys index](https://stagereview.app/superset-sh/superset/pull/4713/chapters/2) |
| 3 | [Optimize Electric shape filter for apikeys](https://stagereview.app/superset-sh/superset/pull/4713/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/4713">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 56d655fe6e592cf89f62d45b2363ee265b2d00a4 on May 19, 2026 9:21pm UTC.</sub>

PR #4738, opened 2026-05-19
SUPER-752 host auth refresh

justincrich:
## Overview

Linear: https://linear.app/superset-sh/issue/SUPER-752/cli-host-service-silently-breaks-when-its-auth-session-expires-never

CLI-spawned host-service starts with an OAuth access token that can expire while the daemon keeps running. The host cloud client already retries once after a 401; this PR makes that retry path refresh from the CLI auth config instead of requiring a host restart or user re-login.

## What Changed

- Config-backed host auth now returns the stored config access token normally. After `JwtApiAuthProvider.invalidateCache()` is called by the existing 401 retry link, the next request refreshes with `auth.refreshToken`, single-flights concurrent callers, atomically updates only `auth`, preserves rotated refresh tokens, and skips writing if the on-disk auth changed during refresh.
  - Why: this is the bug fix. A daemon can keep running after its initial access token expires, and the retry path should recover by reading the current CLI auth config instead of surfacing a 401.
  - Review links: [existing 401 retry trigger](https://github.com/superset-sh/superset/blob/8da27d8842e02ef33eda44e26354840e07ba2de0/packages/host-service/src/api/createApiClient/createApiClient.ts#L21-L37), [config-backed token source](https://github.com/superset-sh/superset/pull/4738/files#diff-eec13e910b87d3b920903b4f51b2503d058b8d52e10762c74f0fae0f53bf486f), [provider invalidation hook](https://github.com/superset-sh/superset/pull/4738/files#diff-b9f59b11720898bcaf766949aa8c5bfa1726be36421da31eab60bc66fb2eef3a), [host wiring](https://github.com/superset-sh/superset/pull/4738/files#diff-fde3ca3f3114ccb0626fb3c6f7e17ccf0b19899734a22ec9d2367d198f1eb5a1).

- CLI-spawned host-service receives the CLI auth config path only for OAuth-backed sessions, while static token behavior remains unchanged when no config path is available.
  - Why: the host needs a durable auth source for OAuth refresh after startup, but API-key/static-token flows should keep their existing `AUTH_TOKEN` behavior.
  - Review links: [start command handoff](https://github.com/superset-sh/superset/pull/4738/files#diff-ac322143e28a21d441e5e7d1175ae9e2d9d3a94a6da468a2104a4799778402a2), [spawn env](https://github.com/superset-sh/superset/pull/4738/files#diff-0a68039ca9ae8c628e5c3d03158124b99de0334f39015c30e8070e2b1391fbee), [optional host env](https://github.com/superset-sh/superset/pull/4738/files#diff-3339b68360e683d9524775ce32a3daa4ae57159f21a6429dc3ddaf4e7a47d8c4).

- Config writes use unique temp files and sanitized OAuth failures avoid exposing raw response bodies.
  - Why: refresh recovery writes credentials from a long-running process, so writes must not corrupt an existing config on rename failure or concurrent writes. Failed OAuth responses can include token/cookie-bearing payloads, so user-visible errors should stay generic.
  - Review links: [atomic CLI config writes](https://github.com/superset-sh/superset/pull/4738/files#diff-b21d11a0e8068845af3a14740bb7fb12f9da77e22ea2a6f2849dde3ce52b7d7c), [sanitized CLI OAuth failures](https://github.com/superset-sh/superset/pull/4738/files#diff-3dab4620f11e3bc1439c54ee96b4dc2666c902de1168ffc10a3eb3c79cdad3fd), [config-backed refresh write path](https://github.com/superset-sh/superset/pull/4738/files#diff-eec13e910b87d3b920903b4f51b2503d058b8d52e10762c74f0fae0f53bf486f).

- Tests cover the retained recovery path and support behavior.
  - Why: the risk is in retry timing and credential persistence, so coverage focuses on pre-401 token use, refresh after invalidation, concurrent retry callers, rotated refresh token persistence, missing-refresh-token guidance, sanitized failures, atomic config writes, spawn env handoff, and unchanged static-token fallback.
  - Review links: [host auth tests](https://github.com/superset-sh/superset/pull/4738/files#diff-f832ff227e7c3440e50d830d54f4cf91f88bdee8cd772ebe6962bf9470bf05ee), [config tests](https://github.com/superset-sh/superset/pull/4738/files#diff-86edac8733b89074a4e40b26538d42b451e09efd9457d00815d8058f43ecfba4), [spawn tests](https://github.com/superset-sh/superset/pull/4738/files#diff-9cac5fe9e0c09356af89e9e0bd0fd506c61fdfed78733bdf7d5ad91c6a0e8778).

## Manual Verification

The CLI's own cloud commands (`workspaces list`, `auth whoami`, etc.) do **not** exercise this PR — they hit cloud directly with the CLI's bearer and have no retry-on-401. The refresh path lives inside the host-service, so you have to drive a cloud-bound request from inside the *running host*. Either route below works.

### Route A — via the desktop app

1. Log in with the CLI OAuth flow.
2. `superset start` (or use the desktop's bundled host).
3. Launch the desktop and confirm any cloud-backed view loads.
4. Edit `~/.superset/config.json` (or `$SUPERSET_HOME_DIR/config.json`): set `auth.accessToken` to `"invalid.invalid.invalid"`. Leave `auth.refreshToken` intact.
5. In the desktop, trigger any action that hits a host tRPC procedure doing `ctx.api.*` (open/refresh a workspace, list workspaces, etc.).
6. Re-read the config — `accessToken` is now a fresh JWT (no longer `invalid…`); `refreshToken` is still present (rotated value is fine).

### Route B — via curl, no desktop required

```bash
CONFIG=~/.superset/config.json   # or "$SUPERSET_HOME_DIR/config.json"
ORG=$(jq -r '.organizationId' "$CONFIG")
M=~/.superset/host/$ORG/manifest.json
ENDPOINT=$(jq -r '.endpoint' "$M")
SECRET=$(jq -r '.authToken' "$M")
INPUT='%7B%220%22%3A%7B%22json%22%3Anull%7D%7D'   # {"0":{"json":null}} urlencoded

# 1. Sanity — host hits cloud with the current token (expect HTTP 200)
curl -sS -H "Authorization: Bearer $SECRET" \
  "$ENDPOINT/trpc/workspace.cloudList?batch=1&input=$INPUT" | jq .

# 2. Corrupt accessToken on disk; keep refreshToken
jq '.auth.accessToken = "invalid.invalid.invalid"' "$CONFIG" \
  > "$CONFIG.tmp" && mv "$CONFIG.tmp" "$CONFIG"
jq '{access:.auth.accessToken[:24], refresh:.auth.refreshToken[:12]}' "$CONFIG"

# 3. Same call — host reads bad token → cloud 401 → retry link refreshes → 200
curl -sS -H "Authorization: Bearer $SECRET" \
  "$ENDPOINT/trpc/workspace.cloudList?batch=1&input=$INPUT" | jq .

# 4. Proof — accessToken rotated; refreshToken still present (possibly server-rotated)
jq '{access:.auth.accessToken[:24], refresh:.auth.refreshToken[:12]}' "$CONFIG"
```

**Expected:**

- Step 1: HTTP 200, an array (possibly `[]`).
- Step 2: `access` shows `"invalid.invalid.invalid"`.
- Step 3: HTTP 200 — proves the host's 401 retry + refresh + write-back chain ran.
- Step 4: `access` no longer starts with `invalid`; `refresh` still present (will be the cloud-rotated value if rotation is enabled).

### Negative case (sanitized failure)

Repeat step 3 after also deleting `auth.refreshToken`. Expect a clean error response (no raw OAuth body, no token/cookie payload leaked) and a "log in again" hint.

## Validation

```bash
bun test src/lib/auth.test.ts src/lib/config.test.ts src/lib/host/spawn.test.ts
bun test packages/host-service/src/env.test.ts packages/host-service/src/providers/auth/JwtAuthProvider/JwtAuthProvider.test.ts
bun run --cwd packages/cli typecheck
bun run --cwd packages/host-service typecheck
```

## Summary by CodeRabbit

* **New Features**
  * OAuth tokens can now be persisted in config file with automatic refresh support
  * Atomic config file writes with enhanced permission handling

* **Bug Fixes**
  * Consistent error messaging for token refresh failures
  * Improved stripping of sensitive authentication environment variables from terminal sessions

* **Tests**
  * Added comprehensive test coverage for config handling, authentication, and host service spawning

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4738?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ghost:
Ready to review this PR? Stage has broken it down into 14 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Update React and Expo dependencies](https://stagereview.app/superset-sh/superset/pull/4738/chapters/1) |
| 2 | [Implement atomic CLI config writes](https://stagereview.app/superset-sh/superset/pull/4738/chapters/2) |
| 3 | [Sanitize OAuth refresh failure details](https://stagereview.app/superset-sh/superset/pull/4738/chapters/3) |
| 4 | [Add config-backed host auth refresh](https://stagereview.app/superset-sh/superset/pull/4738/chapters/4) |
| 5 | [Pass auth config path during spawn](https://stagereview.app/superset-sh/superset/pull/4738/chapters/5) |
| 6 | [Prevent auth secret leaks in PTYs](https://stagereview.app/superset-sh/superset/pull/4738/chapters/6) |
| 7 | [Add user onboarding tracking](https://stagereview.app/superset-sh/superset/pull/4738/chapters/7) |
| 8 | [Implement new onboarding flow](https://stagereview.app/superset-sh/superset/pull/4738/chapters/8) |
| 9 | [Remove legacy onboarding and cleanup](https://stagereview.app/superset-sh/superset/pull/4738/chapters/9) |
| 10 | [Refactor standalone routes and UI](https://stagereview.app/superset-sh/superset/pull/4738/chapters/10) |
| 11 | [Improve GitHub CLI detection and auth](https://stagereview.app/superset-sh/superset/pull/4738/chapters/11) |
| 12 | [Clean up workspace sync logic](https://stagereview.app/superset-sh/superset/pull/4738/chapters/12) |
| 13 | [Implement marketing download interstitial](https://stagereview.app/superset-sh/superset/pull/4738/chapters/13) |
| 14 | [Other changes](https://stagereview.app/superset-sh/superset/pull/4738/chapters/14) |

<a href="https://stagereview.app/superset-sh/superset/pull/4738">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 72312223a0e324d86888c918eb83f557dc488554 on May 21, 2026 10:11pm UTC.</sub>

packages/cli/src/lib/host/spawn.ts
```diff
@@ -88,6 +89,37 @@ function resolveMigrationsFolder(): string {
...
 
+function createHostServiceEnv(
+	options: SpawnHostOptions,
+	port: number,
+	secret: string,
+	relayUrl: string,
+	migrationsFolder: string,
+): NodeJS.ProcessEnv {
+	const childEnv: NodeJS.ProcessEnv = { ...process.env };
+	for (const key of Object.keys(childEnv)) {
```
saddlepaddle: hmm was this the reason we needed this change in this file? What does this change do?

packages/cli/src/lib/config.ts
```diff
@@ -35,22 +38,72 @@ function ensureDir() {
...
+const defaultConfigWriterFs: ConfigWriterFs = {
+	chmodSync,
+	mkdirSync,
+	renameSync,
+	statSync,
+	unlinkSync,
+	writeFileSync,
+};
+
+export function writeConfigFile(
```
saddlepaddle: Could you help me understand the goal with this change too? Was it an attempt to make writeConfig more robust? if so, was there a detected gap?

packages/cli/src/lib/host/spawn.ts
justincrich: Two things:

1. Extracted `createHostServiceEnv` — the inline env block was growing (new `SUPERSET_AUTH_CONFIG_PATH` + the filter); pulling it out keeps `spawnHostService` focused.

2. Strips `*REFRESH_TOKEN*` env keys (broad match on purpose — catches CI vars too). New in this PR the daemon re-reads `~/.superset/config.json` for refreshed tokens. I want the config file to be the *only* path a refresh token reaches the daemon, not env inheritance. A stray var from the parent shell/CI would otherwise sit in `/proc/<pid>/environ` for the daemon's lifetime.

Defense-in-depth, not load-bearing. Happy to narrow to a permitted list if you'd rather.

packages/cli/src/lib/config.ts
justincrich: Pushed 791bd6918 — dropped the `writeConfigFile` export and inlined the atomic-write logic into `writeConfig`. Public surface is back to status-quo (one exported writer). Tests rewritten to use `mock.module("node:fs", ...)` with snapshotted real impls; same three behavioral assertions, all passing.

Substantive answer to your original question, for the record:

Robustness for a real gap this PR introduces. Before, only the CLI wrote to `config.json`. After, the daemon also writes when it refreshes tokens on a 401 (`ConfigFileSessionTokenSource.ts:178`). CLI and daemon can now write concurrently.

The daemon already has a CAS-style check before its write (re-reads, bails if a CLI-set `apiKey` appeared or another writer beat it) — that handles write-write conflicts on its side. Atomic-write is still load-bearing because the CLI's `readConfig` is a plain `readFileSync` + `JSON.parse` with no retry; a torn write from the daemon would crash that read and log the user out.

`writeConfig` now switches to write-temp + rename: `.<uuid>.<pid>.config.tmp` → chmod 0600 → `renameSync` onto target → `unlinkSync` on failure. Readers see either old or new, never partial.

One open thing: the daemon has a near-identical atomic-write helper at `ConfigFileSessionTokenSource.ts:55-75` (same shape, `node:fs/promises` instead of sync). Worth factoring into `@superset/shared` so CLI and daemon stay in sync if the pattern evolves, or leave them as intentional twins given the sync/async + process-boundary split?

saddlepaddle (approved):

PR #4800, opened 2026-05-21
[codex] fix stuck workspace deletion

Kitenite:
## Summary
- Remove the renderer-side `not-found` block that disabled workspace deletion before cleanup could run.
- When the synced workspace row is missing but the local host service is available, still call local `workspaceCleanup` so the backend can finish its idempotent cleanup path.

## Issue
Deleting a workspace could get stuck behind the message "Workspace is no longer available on this host." The host cleanup endpoint already tolerates missing state: no local workspace row, missing worktree, and already-deleted cloud rows are valid delete outcomes.

The renderer was blocking too early. The delete dialog treated `useWorkspaceHostTarget(...).status === "not-found"` as a terminal UI error and manufactured a blocking preview with `canDelete: false`. That disabled the Delete button, so `workspaceCleanup.destroy` never ran.

This can happen after a partial delete, slow Electric sync, or inconsistent local/sidebar/cloud state: the workspace row may already be absent from the live collection while local host cleanup still needs to complete.

## Fix
- Stop converting `not-found` into a blocking dialog state.
- Let inspect failures remain non-fatal, as the dialog already does for other preview failures.
- If the workspace target is `not-found` but the local host service is running, route `inspect`/`destroy` to the local host service and let the backend decide based on actual cleanup state.

## Impact
Users can retry/delete a workspace even when sync has already removed or hidden the workspace row, instead of getting stuck before the idempotent cleanup endpoint runs.

## Validation
- `bun run lint`
- `bun test packages/host-service/test/integration/workspace-cleanup.integration.test.ts`

### DIFF
--- a/apps/desktop/src/renderer/hooks/host-service/useDestroyWorkspace/useDestroyWorkspace.ts
+++ b/apps/desktop/src/renderer/hooks/host-service/useDestroyWorkspace/useDestroyWorkspace.ts
@@ -5,6 +5,7 @@ import type {
 import { TRPCClientError } from "@trpc/client";
 import { useCallback } from "react";
 import { getHostServiceClientByUrl } from "renderer/lib/host-service-client";
+import { useLocalHostService } from "renderer/routes/_authenticated/providers/LocalHostServiceProvider";
 import {
 	useWorkspaceHostTarget,
 	type WorkspaceHostTarget,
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Route missing workspace cleanup through local host](https://stagereview.app/superset-sh/superset/pull/4800/chapters/1) |
| 2 | [Unblock deletion UI for missing workspaces](https://stagereview.app/superset-sh/superset/pull/4800/chapters/2) |

<a href="https://stagereview.app/superset-sh/superset/pull/4800">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit a64e50d600d5144b13e7f47dc5956018d31f405e on May 21, 2026 6:23am UTC.</sub>

PR #4817, opened 2026-05-21
SUPER-771: Loud failures for automation runs

justincrich:
# SUPER-771: Loud failures for automation runs

## What this fixes

Failed automation runs (`dispatch_failed` / `skipped_offline`) used to surface only as a tiny red dot in the run history, with a truncated tooltip showing raw transport jargon (`dispatch: relay 503: {"error":"Host not connected"}`). Users were never proactively notified, and since these failures are intentionally not retried, the run was silently lost.

After this change, a failed run shows the error inline in the row (selectable for bug reports), the error string is translated from transport-code into human copy ("Target machine was offline"), and a desktop notification fires once per failed run within the current session.

Origin: #founders Slack thread (2026-05-14, Kiet/Satya) → "popup/notification, not silent retry". Decision was visibility, not retry.

## Why this approach

Two prongs needed to address:

1. **Render the error legibly inline** with `select-text cursor-text` (per `apps/desktop/AGENTS.md` users must be able to copy errors into bug reports).
2. **Translate the relay status code into human copy** at the boundary where the error is persisted, so any future surface reading `automationRuns.error` inherits the readable string.

For prong (2), translation lives in `relay-client.ts`'s `humanRelayMessage(status, rawBody)` helper rather than the renderer. This keeps the human copy as the persisted single source of truth and avoids duplicating translation logic across surfaces.

For the proactive notification (founder ask), an `AutomationFailureNotifier` provider observes the Electric-synced `automationRuns` collection and fires `electronTrpcClient.notifications.showNative.mutate(...)` once per failed run.id, with a `useRef<Set<string>>` deduplicating within the current app session. Cross-window and cross-session dedup were explicitly deferred (`notifications.ts:38-41` v2NotificationSourceSchema doesn't accept an automation source type — extending it is a follow-up).

The full investigator + challenger debate that produced this scope lives at `.spec/improvements/SUPER-771/SCOPE.md` (chosen option: moderate; rejected: minimum, Option-4, strategic).

## Binding scope

- **AC-1** (Inline error row): failed-row renders error inline below the title — `PreviousRunsList.tsx` (commits `ac6c53600`)
- **AC-2** (Copyable error): inline span carries `select-text cursor-text pl-4 text-xs text-destructive` — same file
- **AC-3** (Relay 503 → human copy): `humanRelayMessage(502→"Target machine is unreachable", 503→"Target machine was offline", 504→"Target machine timed out", default→"Relay error (status N): …")` wired at the `RelayDispatchError` throw site — `relay-client.ts` (commit `902f5a2cf`)
- **AC-4** (describeError unchanged): `describeError(err, context)` is identical to `main`; the human translation lives upstream in `relay-client.ts` so describeError just passes `err.message` through (no `err.name` prefix)
- **AC-5** (Notification fires): `electronTrpcClient.notifications.showNative.mutate({ title: "Automation failed", body: run.error || "Run failed" })` on `dispatch_failed` / `skipped_offline` — `AutomationFailureNotifier.tsx` (commits `ac9c8c37b`, `952514935`)
- **AC-6** (Within-session dedup): `useRef<Set<string>>` tracks notified IDs; `.has()` before `.add()` prevents re-fire on Electric re-emit
- **AC-7** (Mount in layout): exactly +2 lines in `layout.tsx` — import + JSX mount inside `<CollectionsProvider>` (commit `ec8cc7afa` reverted an unrelated `activeOrganizationId` refactor that crept in during cycle 2)

## Considered alternatives (rejected)

- **Option 4** (renderer-only translation, 15 LOC, 1 file): rejected — solves prong A only, ignores founder ask for notifications.
- **minimum** (35 LOC): rejected — solves prong A only, no notification path.
- **strategic** (200+ LOC + schema migration): rejected — disproportionate; both specialists flagged as cleanup pass disguised as bug fix. Deferred to follow-up.

## Out of scope (deliberately deferred)

See `.spec/improvements/SUPER-771/follow-ups.md` for the full list:

- Typed `failureReason` enum on `automationRuns` (was the strategic option)
- QStash retry-exhaustion path (`apps/api/.../run-failed/route.ts`) copy parity with dispatch path
- Cross-window notification dedup (requires extending `v2NotificationSourceSchema`)
- Cross-session dedup (persisted seen-set)
- "Retry now" CTA on failed rows
- `PreviousRunsList` virtualization
- Failure-detail modal/sheet
- DOM test infrastructure for `apps/desktop` renderer (no `@testing-library/react` / `happy-dom` installed; renderer-side ACs verified manually for this PR)

### Visual proof (captured against this branch in the desktop dev build)

Inline error row + human copy translation (AC-1, AC-2, AC-3) — top row is a `dispatch_failed` run rendering the translated copy "Target machine was offline"; lower rows are pre-existing `skipped_offline` runs (pre-dispatch path, unchanged by this PR):

Screenshot 2026-05-21 at 3 33 10 PM

Native macOS notification fired by `AutomationFailureNotifier` (AC-5):

Automation failed notification

### Automated (one command, no setup)

```bash
bun test packages/trpc/src/router/automation/relay-client.test.ts
```

11 tests pass — covers AC-3 paths explicitly: `humanRelayMessage(503, …)` → `"Target machine was offline"`, `502` → `"Target machine is unreachable"`, `504` → `"Target machine timed out"`, unknown status → `"Relay error (status N): <truncated body>"`. Exercises the real `RelayDispatchError` throw path with a mocked `fetch` (legitimate transport boundary mock — the SUT is the helper + the error class, both real).

### Manual — Quick verification via direct DB insert (recommended, ~60 seconds)

The dispatch flow has lots of moving parts (host registration, relay reachability, JWT minting). For visual verification of the renderer-side ACs, **bypass all of it** by inserting a synthetic failed-run row directly. Electric SQL syncs the row to the renderer in <1s, the notifier fires, the row appears with the correct affordances. This is the same approach used to capture the screenshots above.

**Setup:**

1. Make sure the desktop app is running (`bun dev` from the worktree) and you're signed in.
2. Open an automation (any one). The right sidebar shows "Previous Runs".
3. Grab the automation + organization IDs:

   ```bash
   docker exec -i superset-pg psql -U superset -d $(grep DATABASE_URL .env | awk -F/ '{print $NF}') \
     -c "SELECT id AS automation_id, organization_id FROM automations LIMIT 1;"
   ```

   Substitute the two UUIDs into the commands below.

4. **macOS only**: ensure notification permission is granted. **System Settings → Notifications → Electron → Allow Notifications: ON**. (Dev build inherits the generic `com.github.Electron` bundle ID; production-build app has its own.)

**Test 1 — AC-1, AC-2, AC-3 (inline error row + human copy):**

```bash
docker exec -i superset-pg psql -U superset -d $(grep DATABASE_URL .env | awk -F/ '{print $NF}') -c "
  INSERT INTO automation_runs (id, automation_id, organization_id, title, scheduled_for, status, error)
  VALUES (gen_random_uuid(), '<AUTOMATION_ID>'::uuid, '<ORG_ID>'::uuid, 'AC-1/2/3 Test', NOW(), 'dispatch_failed', 'Target machine was offline');
"
```

**Expect** within ~1s:
- New row "AC-1/2/3 Test" appears in PreviousRunsList with a red dot.
- Inline second line below the title shows `Target machine was offline` in red.
- Highlight that text and ⌘C — pastes verbatim into another app (confirms `select-text cursor-text` classes).

**Test 2 — AC-5 (notification fires):**

The same INSERT above also triggers AC-5. Within ~1s of inserting:

**Expect:** a native macOS notification appears in the top-right with:
- Title: **Automation failed**
- Body: **Target machine was offline**

**Test 3 — AC-6 (within-session dedup):**

Insert two rows with the SAME `id` (the second insert will be a no-op via ON CONFLICT, OR force the test by inserting a duplicate via a stash; alternatively, navigate away from the automation page and back — the notifier shouldn't re-fire for the already-seen run id):

Easiest variant: insert two NEW rows with different IDs and observe two notifications. Then refresh the page or navigate away+back — no third notification fires (proves the ref-set dedup works within session).

**Test 4 — AC-7 (negative path):**

Insert a successful row:

```bash
docker exec -i superset-pg psql -U superset -d $(grep DATABASE_URL .env | awk -F/ '{print $NF}') -c "
  INSERT INTO automation_runs (id, automation_id, organization_id, title, scheduled_for, status, error)
  VALUES (gen_random_uuid(), '<AUTOMATION_ID>'::uuid, '<ORG_ID>'::uuid, 'AC-7 Negative', NOW(), 'dispatched', NULL);
"
```

**Expect:** new row appears with NO inline error span and NO notification. Status dot color differs from failed rows.

**Cleanup:**

```bash
docker exec -i superset-pg psql -U superset -d $(grep DATABASE_URL .env | awk -F/ '{print $NF}') -c "
  DELETE FROM automation_runs WHERE title LIKE 'AC-%';
"
```

### Manual — Full end-to-end via dispatch flow (optional)

If you want to exercise the actual dispatch code path (rather than DB-inject), the simpler way to induce a relay 503 is to register an online host then take it offline:

1. Have an automation pointed at your local host (the one bound to `bun dev`).
2. Stop the host-service process: `pkill -f "packages/host-service"`.
3. Wait one schedule tick (or click **Run now**).
4. The dispatch goes through the real `relay-client.ts` code path. The resulting `automation_runs.error` should be `"dispatch: Target machine was offline"` (server-side translation).
5. The row renders inline + notification fires per the same ACs.

This path is more involved (host-service lifecycle is fiddly in dev) — the DB-insert path above gives equivalent visual verification with less ceremony.

### Regression check

- **Tooltip on hover** (kept for long errors): hover the failed row — the existing `<Tooltip>` with `max-w-xs whitespace-pre-wrap` still appears showing the same content.
- **Other notifications** (workspace lifecycle, agent events): `V2NotificationController` still mounted alongside `AutomationFailureNotifier` in `_authenticated/layout.tsx`. Verify a workspace-lifecycle notification still fires for unrelated events.

### Real-services verification (per ~/.claude/CLAUDE.md Supreme Rule)

The visual proof screenshots above were captured against a running desktop dev build talking to a real Postgres + real Electric SQL + real Electron Notification API end-to-end. `relay-client.test.ts` mocks `fetch` only — that's a transport boundary; the helper (`humanRelayMessage`) + the error class (`RelayDispatchError`) + the throw site are all real code.

## Anticipated FAQ

- **Q: Why is `dispatch.ts` unchanged from main?**  
  A: AC-4 says "describeError translates known relay codes without breaking call signature". The translation lives upstream in `relay-client.ts`'s `humanRelayMessage`. `describeError` just passes `err.message` through unchanged. This keeps the call signature stable AND avoids duplicating the translator. The persisted error is e.g. `"dispatch: Target machine was offline"` — the `dispatch:` prefix from describeError + the human message from relay-client.

- **Q: Why don't `AutomationFailureNotifier.tsx` and `PreviousRunsList.tsx` have test files?**  
  A: The desktop package has no DOM test infrastructure (`@testing-library/react`, `happy-dom` not installed). Three rounds of agent attempts to write component tests all rationalized — either re-implementing production logic in the test file or asserting on mock data only. Installing DOM test infra is out of scope for a bug fix; it's captured in `.spec/improvements/SUPER-771/follow-ups.md` as a separate task. Renderer-side ACs (AC-1, AC-2, AC-5, AC-6, AC-7) require manual visual verification per the steps above.

- **Q: Why was the sibling `super-771-loud-failures` branch abandoned?**  
  A: Both the investigator (electron-reviewer) and the challenger (code-reviewer) recommended abandon: the sibling's AUTO-LOUD-003 commit added a tRPC procedure that this binding scope does NOT sanction. The sibling's branch and 4 commits remain in git history for reflog recovery if ever needed; nothing was merged from it.

- **Q: Won't the notification re-fire annoyingly on every app launch for a long-failed run?**  
  A: For v1, yes — and that's the founder intent ("proactively tell the user"). Cross-session dedup (a persisted seen-set) is captured in follow-ups. If this becomes annoying in practice, we'll add the persisted seen-set in a future ticket.

- **Q: What happens for relay error codes other than 502/503/504?**  
  A: `humanRelayMessage` default branch returns `"Relay error (status N): <truncated body>"`. Unknown codes still surface readable text, just less specific. Adding 401/408/etc. translations is a one-line per code in `humanRelayMessage` — defer until we see them in practice.

## Risks

- **MEDIUM** — Notification firing on app launch for pre-existing failures (Electric re-hydrates from SQLite cache on mount). Verified acceptable per founder intent. If pushback, add localStorage-backed seen-set (one-line follow-up).
- **LOW** — `humanRelayMessage` covers 502/503/504. Other status codes use a generic fallback. We can add codes as we encounter them.
- **LOW** — `automationRuns.error` now persists the translated human string. If wording changes, historical rows show old copy. Acceptable for v1; the `failureReason` enum follow-up will let display copy be regenerated.

## Links

- Linear ticket: https://linear.app/superset/issue/SUPER-771
- Binding SCOPE.md: `.spec/improvements/SUPER-771/SCOPE.md`
- Task contract: `.spec/tasks/SUPER-771/SUPER-771-loud-failures.md`
- Design specs: `.spec/tasks/SUPER-771/design/PreviousRunsList-failed-row.md`, `.spec/tasks/SUPER-771/design/AutomationFailureNotifier.md`
- Follow-ups: `.spec/improvements/SUPER-771/follow-ups.md`

## Notes for the human signing off

This PR went through three implementation cycles before landing:

1. Cycle 1: implementer claimed AC-1/2/3 done but production code was unchanged; commit messages were misleading. Caught by independent diff verification.
2. Cycle 2: re-dispatched on the wrong branch; commits had to be cherry-picked manually onto the improvement worktree.
3. Cycle 3: test work; agent rationalized 3 times because the desktop package lacks DOM test infra. After verification that the server-side tests (`relay-client.test.ts`) were real and the renderer tests were sham, the sham tests were deleted; only `relay-client.test.ts` remains as automated coverage.

The PRODUCTION CODE is correct and matches the binding scope + design specs. The TEST GAP is real and documented above. The decision to ship with manual verification (rather than fight DOM test infra in this PR) was made explicitly knowing the trade-off.


ghost:

| | Title |
|---|-------|
| 1 | [Translate relay status codes into human-readable messages](https://stagereview.app/superset-sh/superset/pull/4817/chapters/1) |
| 2 | [Render copyable error messages in the run history](https://stagereview.app/superset-sh/superset/pull/4817/chapters/2) |
| 3 | [Implement native notifications for failed automation runs](https://stagereview.app/superset-sh/superset/pull/4817/chapters/3) |
| 4 | [Wire the failure notifier into the application layout](https://stagereview.app/superset-sh/superset/pull/4817/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/4817">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 93ee783603b9f27105d3fcb91fad0d497f3989a5 on May 26, 2026 2:07am UTC.</sub>

apps/desktop/src/renderer/routes/_authenticated/providers/AutomationFailureNotifier/AutomationFailureNotifier.tsx
```diff
@@ -0,0 +1,50 @@
+import { useLiveQuery } from "@tanstack/react-db";
+import { useEffect, useRef } from "react";
+import { electronTrpcClient } from "renderer/lib/trpc-client";
+import { useCollections } from "../CollectionsProvider";
+
+export function AutomationFailureNotifier() {
```
saddlepaddle, line 7: A very nice touch!

apps/desktop/src/renderer/routes/_authenticated/providers/AutomationFailureNotifier/AutomationFailureNotifier.tsx
```diff
@@ -0,0 +1,50 @@
...
+			if (
+				(run.status === "dispatch_failed" ||
+					run.status === "skipped_offline") &&
+				!notifiedRunIdsRef.current.has(run.id)
+			) {
+				// Mark as notified
+				notifiedRunIdsRef.current.add(run.id);
+
+				// Fire notification
+				void electronTrpcClient.notifications.showNative.mutate({
```
saddlepaddle: Should probably also link to the automation details page on click right?

saddlepaddle (approved):

justincrich:
Addressed all review comments in e78049f:

1. **P2: Query fetches all runs** — Added `.where()` filter with `or(eq(status, "dispatch_failed"), eq(status, "skipped_offline"))` so only failed runs are pulled into the live result set.

2. **P1: Run ID marked before IPC resolves** — Changed to `.then(() => notifiedRunIdsRef.current.add(run.id))` with a `.catch()` that intentionally leaves the ID unmarked so the next effect run retries. No more `void` fire-and-forget.

3. **P2: cursor-text not visible inside disabled button** — Added `pointer-events-auto` to the error `<span>` so the I-beam cursor renders even when the parent `<button>` is disabled.

4. **P2: Detached async with no error handling** — Resolved by the same `.then()/.catch()` change from #2.

5. **Link to automation details on notification click** — Good call, but the existing `clickTarget` schema only supports `{ workspaceId, source: terminal | chat }`. Failed runs often have no workspaceId, and there's no "automation" source type yet. I'll file a follow-up to extend the notification system with an automation click target.

apps/desktop/src/renderer/routes/_authenticated/providers/AutomationFailureNotifier/AutomationFailureNotifier.tsx
saddlepaddle: I'm actually going to push us to skip this for now - could we drop this? It's a great call I just may want to punt it to avoid creating more exposure to electric sql risks 🤔

PR #4834, opened 2026-05-22
fix(desktop): don't block onboarding on GitHub CLI auth

saddlepaddle:
## Why

Onboarding hard-blocked the **Required** GitHub CLI step for users who are actually authenticated. The device-login modal even reports `Logged in as <user>` / `already logged in to this account`, yet Continue stayed disabled.

Two contributing factors:

1. **Detection false-negative.** `detectGhCli` ran a blanket `gh auth status`, which (per gh's docs) exits non-zero if *any* account on *any* host has an issue. Power users with a stale secondary/enterprise account have a working active github.com login but a broken other account, so the check failed.

2. **Hard gate.** The onboarding step disabled Continue until gh auth was detected, so any detection miss left users stuck with no way forward.

## Changes

1. **Fix detection** (`system.ts`): check `gh auth status --active --hostname github.com` so only a broken *active github.com* account counts as unauthenticated. `--active` exists only in gh >= 2.40 (the release that also introduced the multi-account "any host fails" behavior), so older gh falls back to `gh auth status --hostname github.com`, which every version accepts and which still validates the token.

2. **Stop gating on it** (`onboarding/layout.tsx`): the GitHub CLI step no longer disables Continue. Users can proceed and sign in later; the row still shows accurate Connected/Required status via the corrected detection.

(Replaces an earlier `disable-onboarding` PostHog kill-switch approach — no flag needed.)

## Testing

- `bun run lint` ✅
- `bun run typecheck` ✅ (28/28)
- gh auth behavior confirmed against gh 2.92.0: `--active --hostname github.com` and the `--hostname`-only fallback both exit 0 for a valid login; unknown flags exit 1 (why old gh needs the fallback).

Note: the detection path only runs in the packaged Electron app, so end-to-end verification means running the desktop app.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/system.ts
+++ b/apps/desktop/src/lib/trpc/routers/system.ts
@@ -30,7 +30,11 @@ async function detectGhCli(): Promise<GhDetectResult> {
 
 	let authenticated = false;
 	try {
-		await execWithShellEnv("gh", ["auth", "status"], { timeout: 5000 });
+		await execWithShellEnv(
+			"gh",
+			["auth", "status", "--active", "--hostname", "github.com"],
+			{ timeout: 5000 },
+		);
 		authenticated = true;
 	} catch {
 		// `gh auth status` exits non-zero when not logged in.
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Scope GitHub CLI authentication check](https://stagereview.app/superset-sh/superset/pull/4834/chapters/1) |
| 2 | [Decouple onboarding progress from CLI status](https://stagereview.app/superset-sh/superset/pull/4834/chapters/2) |

<a href="https://stagereview.app/superset-sh/superset/pull/4834">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit df718b3a29120fcae506f1e0b3d2ab3afcb3d821 on May 22, 2026 2:34am UTC.</sub>

PR #4928, opened 2026-05-26
fix(cli): wait for host process to exit before reporting startup failure

vimzh:
`spawnHostService` had three issues that made host startup flaky and could leave zombie processes holding the port:

1. `findFreePort` closed its temporary listener before returning the port. Another process could grab that port between `close()` and the host's `bind()`, and the host would then fail to start on a port the CLI just told it to use.

2. On health-check failure the code sent `SIGTERM` and threw immediately. The host could still be alive holding the port, so the next retry would also fail with `EADDRINUSE`. `ChildProcess.kill()` also returns `false` when the OS reports the process is already gone, which the previous code would have silently waited on forever.

3. No `error` or `exit` listeners were attached. Spawn errors bubbled as unhandled and could crash the CLI process, and the health-check polled the full 10 second timeout even when the child had already died on the first tick.

The fix:

- Rewrites `findFreePort` with explicit `listening` / `error` dispatch.
- Adds `terminateChild` that sends `SIGTERM`, escalates to `SIGKILL` after a 1 second grace period, and resolves only once the child has actually exited (or a 4 second hard backstop fires). Handles `kill()` returning `false` so we don't wait on an exit event that won't come.
- Attaches `error` and `exit` listeners during startup and detaches them once healthy, so later events don't get silently captured by stale closures.
- Adds an early-exit predicate to `pollHealth` so it stops as soon as the child dies instead of polling the full timeout.
- Throws a structured error with the exit code and signal on startup failure.

A small TOCTOU window remains between `findFreePort` releasing the socket and the host binding it. Eliminating it would require passing the listening fd into the host process, which is a larger change.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4928">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes flaky CLI host startup by safely choosing a free port, detecting early spawn failures, and waiting for the child to exit before retry. Prevents zombie processes holding the port and avoids unhandled CLI crashes.

- **Bug Fixes**
  - Rewrote findFreePort with explicit listening/error flow and close callback before returning the port.
  - Added terminateChild: send SIGTERM, escalate to SIGKILL after 1s, resolve only once the process exits (4s hard backstop); fixed a TDZ so kill() returning false no longer crashes.
  - Attached an 'error' listener immediately after spawn (before the pid check) and an 'exit' listener during startup; detach once healthy.
  - pollHealth now stops early if the child exits, instead of waiting the full timeout.
  - On startup failure, wait for termination and throw a structured error including exit code and signal.

<sup>Written for commit fe5dfc818a12034b247d381af03fedaa16888e67. Summary will update on new commits. <a href="https://cubic.dev/pr/superset-sh/superset/pull/4928?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved host service startup reliability: better error handling, robust termination (escalation to force-kill) and health-check behavior to avoid hangs and port-in-use races.
* **Tests**
  * Added tests covering startup failures, error-listener ordering, and termination behavior to prevent hangs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4928?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/cli/src/lib/host/spawn.test.ts
+++ b/packages/cli/src/lib/host/spawn.test.ts
@@ -31,8 +31,13 @@ const spawnMock = mock(
 		spawnCalls.push({ command, args, options });
 		return {
 			pid: 12345,
+			exitCode: null,
+			signalCode: null,
 			kill: mock(() => undefined),
 			unref: mock(() => undefined),
+			on: mock(() => undefined),
+			once: mock(() => undefined),
+			removeListener: mock(() => undefined),
 		};
 	},
 );
--- a/packages/cli/src/lib/host/spawn.test.ts
+++ b/packages/cli/src/lib/host/spawn.test.ts
@@ -1,4 +1,5 @@
 import { afterAll, afterEach, describe, expect, mock, test } from "bun:test";
+import { EventEmitter } from "node:events";
 import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Improve port selection and process termination logic](https://stagereview.app/superset-sh/superset/pull/4928/chapters/1) |
| 2 | [Enhance health polling with early exit detection](https://stagereview.app/superset-sh/superset/pull/4928/chapters/2) |
| 3 | [Wire startup listeners and handle spawn failures](https://stagereview.app/superset-sh/superset/pull/4928/chapters/3) |
| 4 | [Update tests for robust process lifecycle validation](https://stagereview.app/superset-sh/superset/pull/4928/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/4928">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit fe5dfc818a12034b247d381af03fedaa16888e67 on May 26, 2026 11:03am UTC.</sub>

vimzh:
Thanks all. Pushed [fe5dfc81](https://github.com/superset-sh/superset/pull/4928/commits/fe5dfc818) addressing the four findings:

- `terminateChild` now declares `escalation` and `hardDeadline` with `let` before defining `finish()`, so the kill-throws and kill-returns-false fast paths no longer hit the Temporal Dead Zone. The clearTimeout calls in `finish()` are guarded.
- The `error` listener is now attached with `child.once` immediately after `spawn()`, before the `!child.pid` throw, so async `error` emits during a failed spawn have a handler waiting for them. `once` auto-detaches after the first fire and the success path still calls `removeListener` for cases where no event fired.
- Test mock upgraded from no-op stubs to a real `EventEmitter` with configurable `pid`, `kill` return value, and an `emitOnSpawn` hook. Three new tests exercise: early child exit produces a structured `exited during startup` error, the `error` listener is wired before the pid check, and the kill-returns-false path resolves cleanly instead of hitting the TDZ.

PR #4929, opened 2026-05-26
fix(cli): launch browser without shell interpolation of the OAuth URL

vimzh:
`openBrowser` in `packages/cli/src/lib/auth.ts` called `exec("open \"${url}\"")` and the Windows and Linux equivalents. The URL came from controlled inputs in practice, but shell-interpolating text into a command string is the textbook command injection class and shouldn't be in the codebase.

The fix swaps `exec` for `spawn` with `shell: false`. On macOS and Linux the URL is passed as a literal `argv` entry so no shell ever sees it. On Windows `start` is a `cmd.exe` built-in so the launch still has to go through `cmd`, but we keep our own quotes around the URL (with `windowsVerbatimArguments: true`) so an `&` in the OAuth query string can't end the `start` command. `%` is escaped to `^%` so `cmd` doesn't try to expand `%FOO%` patterns inside percent-encoded URLs.

Adds `isSafeBrowserUrl` as defense in depth. It requires the URL parses, requires http or https, and rejects whitespace, quotes, backticks, backslashes, redirection operators, `^`, and `|`. Real OAuth authorize URLs never contain any of those, and if a future caller forwards user-supplied text there is one extra check between it and the OS.

Tests cover accept and reject paths, including percent-encoded URLs (which need to pass) and cmd metacharacters (which should not).

---

<a href="https://stagereview.app/superset-sh/superset/pull/4929">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Prevents command injection when opening the OAuth URL from the CLI by removing shell interpolation and validating URLs. Uses shell-free process launching on all platforms without changing user behavior.

- **Bug Fixes**
  - Replaced `exec` with `spawn` (`shell: false`) on macOS/Linux to pass the URL as an argv entry.
  - On Windows, switched to `rundll32.exe url.dll,FileProtocolHandler <url>` to avoid `cmd.exe` entirely; `&` and `%` are preserved verbatim.
  - Strengthened `isSafeBrowserUrl`: requires `http`/`https`; rejects whitespace, quotes, backticks, backslashes, redirection operators, `^`, `|`, and all control chars (including DEL); added tests for these cases and percent-encoded URLs.

<sup>Written for commit fef8aea6d366f7050836d2c45d212ed15028d52c. Summary will update on new commits. <a href="https://cubic.dev/pr/superset-sh/superset/pull/4929?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Security**
  * Strengthened URL validation for OAuth browser launches to block unsafe protocols and disallowed characters, reducing risk when opening external links.
  * Switched to a more robust, platform-aware browser-launch approach that avoids shell-based execution and logs non-fatal errors.

* **Tests**
  * Added comprehensive tests covering accepted and rejected URL patterns, edge cases, and unsafe inputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4929?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/cli/src/lib/auth.test.ts
+++ b/packages/cli/src/lib/auth.test.ts
@@ -34,23 +34,29 @@ describe("isSafeBrowserUrl", () => {
 		// These are the chars that could break out of cmd.exe's `"..."`
 		// quoting if a URL ever contained them. Real URLs don't.
 		expect(isSafeBrowserUrl('https://x.com/"evil')).toBe(false);
+		expect(isSafeBrowserUrl("https://x.com/'evil")).toBe(false);
 		expect(isSafeBrowserUrl("https://x.com/`evil")).toBe(false);
 		expect(isSafeBrowserUrl("https://x.com/\\evil")).toBe(false);
 		expect(isSafeBrowserUrl("https://x.com/<evil")).toBe(false);
+		expect(isSafeBrowserUrl("https://x.com/>evil")).toBe(false);
 		expect(isSafeBrowserUrl("https://x.com/^evil")).toBe(false);
 		expect(isSafeBrowserUrl("https://x.com/|evil")).toBe(false);
 	});
 
-	test("accepts percent-encoded URLs (cmd %-expansion is handled at launch)", () => {
+	test("accepts percent-encoded URLs", () => {
+		// rundll32 doesn't pass URLs through cmd, so %XX percent-encoding
+		// is safe to keep in the URL.
 		expect(
 			isSafeBrowserUrl("https://x.com/?redirect=http%3A%2F%2Fexample.com"),
 		).toBe(true);
 	});
 
-	test("rejects whitespace and control chars", () => {
+	test("rejects whitespace and control chars including DEL", () => {
 		expect(isSafeBrowserUrl("https://x.com/a b")).toBe(false);
 		expect(isSafeBrowserUrl("https://x.com/a\nb")).toBe(false);
+		expect(isSafeBrowserUrl("https://x.com/a\rb")).toBe(false);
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Implement safe URL validation for browser launching](https://stagereview.app/superset-sh/superset/pull/4929/chapters/1) |
| 2 | [Add unit tests for URL safety validation](https://stagereview.app/superset-sh/superset/pull/4929/chapters/2) |
| 3 | [Refactor browser launching to use spawn without shell](https://stagereview.app/superset-sh/superset/pull/4929/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/4929">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit fef8aea6d366f7050836d2c45d212ed15028d52c on May 26, 2026 11:03am UTC.</sub>

vimzh:
Thanks all. Pushed [fef8aea6](https://github.com/superset-sh/superset/pull/4929/commits/fef8aea6d) addressing the findings:

- The `^%` escape was indeed ineffective inside cmd's double-quoted strings, and OAuth `redirect_uri` values would have been mangled. Switched the Windows path to `rundll32.exe url.dll,FileProtocolHandler <url>`, which is the documented shell-free way to open a URL. cmd is bypassed entirely, the URL is a plain argv entry, and `%XX` percent-encoding survives intact.
- DEL (`\x7f`) added to the control-char rejection so the regex matches its stated contract.
- Tests extended to cover `>`, `'`, `\r`, and DEL.

PR #4936, opened 2026-05-26
fix(automation): create workspace + launch agent in a single relay call (SUPER-783)

justincrich:
## Summary

Closes SUPER-783. "New workspace" automations (the default for scheduled agent runs) failed silently — the workspace would never get created and the agent would never start. User saw an opaque dispatch failure with no indication why.

## Root cause

`dispatchAutomation` made **two sequential relay round-trips** for the new-workspace path:

1. `workspaces.create` → returns the new workspace id
2. `agents.run` → spawns the agent against that id

Each round-trip independently races the host's tunnel state. The first call would usually succeed (workspace gets created in cloud + host), but the second call would frequently land in a moment where the tunnel had just reconnected and the relay's in-memory routing didn't yet have the host — relay returned `503 "Host not connected"` and the agent never spawned. The cloud wrote the run as `dispatch_failed` even though a workspace had actually been created on the host.

## Fix

The host's `workspaces.create` handler already supports launching agents inline via the `agents: []` field — see `packages/host-service/src/trpc/router/workspaces/workspaces.ts:1025` (`dispatchSugarAgents`). The Superset SDK (`packages/sdk/src/resources/workspaces.ts:41`) already uses this. The dispatch path was the only caller that split create and launch into two trips.

This PR passes `agents: [{ agent, prompt }]` in the same `workspaces.create` call. One relay round-trip instead of two, no second tunnel race, atomic on the host side.

The pinned-workspace path (`automation.v2WorkspaceId !== null`) is **unchanged** — still a single `agents.run` call.

## Diff

```
packages/trpc/src/router/automation/dispatch.ts | 84 ++++++++++++++++++-------
1 file changed, 60 insertions(+), 24 deletions(-)
```

## Test plan

- [ ] Schedule an automation with **New workspace** target on a host with a healthy tunnel → workspace gets created AND agent session starts (one `automation_runs` row with `status='dispatched'`, `v2_workspace_id` populated, `session_kind` set)
- [ ] Run history shows the new workspace + the agent session as a normal run, not `dispatch_failed`
- [ ] Pinned-workspace automation still works (regression: `automation.v2WorkspaceId !== null` path unchanged)
- [ ] Host-side agent-launch errors (e.g. project not set up locally) propagate as `dispatch_failed` with the host's error message instead of an opaque relay error

### DIFF
--- a/packages/trpc/src/router/automation/dispatch.ts
+++ b/packages/trpc/src/router/automation/dispatch.ts
@@ -101,28 +101,29 @@ export async function dispatchAutomation(
 			host.machineId,
 		);
 
+		let result: AgentRunResult;
 		if (automation.v2WorkspaceId) {
 			workspaceId = automation.v2WorkspaceId;
+			result = await runAgentOnHost({
+				relayUrl,
+				hostId: routingKey,
+				jwt,
+				workspaceId,
+				agent: automation.agent,
+				prompt: automation.prompt,
+			});
 		} else {
-			const created = await createWorkspaceOnHost({
+			const created = await createWorkspaceWithAgentOnHost({
 				relayUrl,
 				hostId: routingKey,
 				jwt,
 				projectId: automation.v2ProjectId,
 				automation,
-				runId: run.id,
+				agent: { agent: automation.agent, prompt: automation.prompt },
 			});
 			workspaceId = created.workspaceId;
+			result = created.agent;
 		}
-
-		const result = await runAgentOnHost({
-			relayUrl,
-			hostId: routingKey,
-			jwt,
-			workspaceId,
-			agent: automation.agent,
-			prompt: automation.prompt,
-		});
 
 		await dbWs
 			.update(automationRuns)
@@ -224,14 +225,14 @@ async function recordSkipped(
 	return row;
 }
 
-async function createWorkspaceOnHost(args: {
+async function createWorkspaceWithAgentOnHost(args: {
 	relayUrl: string;
 	hostId: string;
 	jwt: string;
 	projectId: string;
 	automation: SelectAutomation;
-	runId: string;
-}): Promise<{ workspaceId: string; branchName: string }> {
+	agent: { agent: string; prompt: string };
+}): Promise<{ workspaceId: string; agent: AgentRunResult }> {
 	// Full-precision timestamp keeps branch names readable AND collision-free
 	// for anything coarser than 1 second.
 	// e.g. "2026-04-19-17-30-00"
@@ -244,22 +245,20 @@ async function createWorkspaceOnHost(args: {
 	const branchName = deduplicateBranchName(candidateBranch, []);
 	const workspaceName = args.automation.name.slice(0, 100);
 
+	type AgentLaunchResult =
+		| ({ ok: true } & AgentRunResult)
+		| { ok: false; error: string };
+
 	const result = await relayMutation<
 		{
 			projectId: string;
 			name: string;
 			branch: string;
+			agents: Array<{ agent: string; prompt: string }>;
 		},
 		{
-			workspace: {
-				id: string;
-				projectId: string;
-				name: string;
-				branch: string;
-			};
-			terminals: Array<{ terminalId: string; label?: string }>;
-			agents: Array<unknown>;
-			alreadyExists: boolean;
+			workspace: { id: string };
+			agents: AgentLaunchResult[];
 		}
 	>(
 		{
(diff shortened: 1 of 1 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Update workspace creation helper to include agents](https://stagereview.app/superset-sh/superset/pull/4936/chapters/1) |
| 2 | [Consolidate automation dispatch into single relay call](https://stagereview.app/superset-sh/superset/pull/4936/chapters/2) |

<a href="https://stagereview.app/superset-sh/superset/pull/4936">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit a47df6a805d45cae2dea08ed9547b59ca581ba62 on May 26, 2026 10:59pm UTC.</sub>

justincrich:
Moving the WHY out of code comments per https://github.com/superset-sh/superset/pull/4936#discussion_r3307339233.

**Why the fix is shaped this way:**

The "New workspace" path used to make two sequential relay round-trips:

1. `workspaces.create` → returns the new workspace id
2. `agents.run` → spawns the agent against that id

Each round-trip independently races the host's tunnel state. The first call would usually succeed (workspace gets created in cloud + host), but the second one would frequently land in a moment where the tunnel had just reconnected and the relay's in-memory routing didn't yet have the host — relay returned `503 "Host not connected"` and the agent never spawned. The cloud wrote the run as `dispatch_failed` even though a workspace had actually been created on the host. That's SUPER-783.

The host's `workspaces.create` handler already supports launching agents inline via the `agents: []` field — see `packages/host-service/src/trpc/router/workspaces/workspaces.ts:1025` (`dispatchSugarAgents`). The Superset SDK (`packages/sdk/src/resources/workspaces.ts:41`) already uses this. The dispatch path was the only caller that split create and launch into two trips.

This PR passes `agents: [{ agent, prompt }]` in the same `workspaces.create` call. One relay round-trip instead of two, no second tunnel race, atomic on the host side. The pinned-workspace path (`automation.v2WorkspaceId !== null`) is unchanged — still a single `agents.run` call.

justincrich:
## 🚧 Blocked on staging verification before code review

The change here is small (41 lines, one file) but I cannot self-verify the end-to-end workspace-create flow in local dev. Mirroring the comment I left on [SUPER-783](https://linear.app/superset-sh/issue/SUPER-783) — **needs someone with staging credentials to deploy + test before this is reviewable.** I'm on a work trial without staging access.

### Why local verification is blocked

I fired 10+ dispatch attempts against the local stack. Every one died before reaching host-service, for two independent reasons that are **not caused by this PR**:

**1. Tunnel flap in this worktree's dev env.** host-service ↔ relay WebSocket reconnects every ~1s with no close-code logged:

```
[host-service:tunnel] connected to relay
[host-service:tunnel] reconnecting in 837ms (attempt 1)
[host-service:tunnel] connected to relay
[host-service:tunnel] reconnecting in 905ms (attempt 1)
```

Every dispatch hit the relay during a brief disconnected window → `relay 503 Host not connected`. Host-service never saw the payload. There is no `socket.close` in my diff.

**2. Empty host-service projects table.** The host's local SQLite has zero rows in `projects`. The dispatch targets project `a98d58f8…` which isn't registered locally. Even with a rock-solid tunnel, `requireLocalProject(input.projectId)` at `packages/host-service/src/trpc/router/workspaces/workspaces.ts:530` throws `PROJECT_NOT_SETUP` before any agent is spawned.

### Why I'm still confident the fix is correct

The pattern this PR adopts is **already running in production** via the Superset SDK. Every customer calling `superset workspaces create --agent claude` goes through the inline-agents code path on `workspaces.create`. The cron dispatch was the only caller that split create + agent-launch into two relay round-trips — the second one was racing the tunnel and producing orphan workspaces. This PR brings dispatch in line with the SDK's working pattern.

- SDK reference: `packages/sdk/src/resources/workspaces.ts:41-58`
- Host handler that does the work: `packages/host-service/.../workspaces.ts:1025` (`dispatchSugarAgents`)

### What needs to happen on staging

1. Deploy this branch (`improvement/SUPER-783-automation-new-workspace-target`, HEAD `a47df6a80`) to staging.
2. Point a dev desktop at staging cloud + staging relay.
3. Create a "New workspace" automation against that host with any prompt.
4. Click **Run now**.
5. **Pass criteria** — `automation_runs` row lands `status='dispatched'` with `v2_workspace_id` set and `session_kind` set; the new workspace appears on the host; the agent session starts.
6. **Diagnostic signal** — exactly **one** outgoing `host.*` relay call per dispatch (was two). Two = the change didn't deploy.

Static checks that already pass:

- ✅ `bun run typecheck` (full monorepo)
- ✅ `bun run lint`
- ✅ `grep` confirms dispatch makes one relay call for new-workspace, never followed by `agents.run`
- ✅ Payload shape matches host's `agentLaunchSchema`

Happy to pair with whoever does the staging run.

Issue #4956, opened 2026-05-27
[bug] Double-UTF-8 encoding on copy affects Latin accents and divider lines, not just CJK (same root cause as #4839)


munimmoiz0-sys:
Same root cause as #4839 (multi-byte UTF-8 characters double-encoded), filing
to document that the scope is broader than Japanese/CJK: every multi-byte UTF-8
character is affected, including Latin accents and the box-drawing/em-dash
divider lines that Claude Code's TUI prints.

## Symptom

Copying Claude Code's terminal output in Superset Desktop puts mojibake on the
clipboard:

- The "SautÃ©ed" thinking spinner pastes as "SautÃÃÂ©ed".
  Ã© = U+00E9 (UTF-8 C3 A9), double-encoded to C3 83 C2 A9, which renders "ÃÃÂ©".

- The horizontal divider line pastes as a long run of "ÃÂ¢ÃÂ¢ÃÂ¢...".
  Those characters have a UTF-8 lead byte of E2 (em-dash U+2014 = E2 80 94,
  box-drawing U+2500 = E2 94 80), so they take the identical
  Latin-1-then-re-encode-as-UTF-8 path.

## Root cause

Textbook double-UTF-8 encoding, identical to #4839: bytes are decoded as
Latin-1/Windows-1252 somewhere in the PTY â xterm.js â clipboard path, then
re-encoded as UTF-8. It is not CJK-specific: it hits any character whose UTF-8
encoding is longer than one byte.

## Not a locale/config issue

The same machine and same shell copy cleanly from macOS Terminal.app, so the
bug is isolated to Superset Desktop's terminal pipeline (matching the analysis
in #4839).

## Environment

- Superset Desktop: 1.11.3
- macOS: Darwin 25.5.0
- Shell: zsh
- Agent: Claude Code

AviPeltz:
Resolved ✅

This was fixed alongside #4839 in #4983 (`fix(desktop): decode terminal clipboard as UTF-8`, commit `d34ddaff1`). The clipboard path now decodes terminal bytes as UTF-8 instead of Latin-1, which fixes the broader scope described here (Latin accents and box-drawing/em-dash divider lines, not just CJK).

Verify: https://github.com/superset-sh/superset/commit/d34ddaff12ca9469fafa60780fbc97ee19fcf7bf

Closing as resolved.

PR #4983, opened 2026-05-28
fix(desktop): decode terminal clipboard as UTF-8 (#4839, #4956)

### LINKED ISSUE #4839: [bug] Multi-byte UTF-8 characters (Japanese/CJK) are double-encoded, causing consistent mojibake in terminal display and clipboard
## Description

In Superset Desktop's built-in terminal, multi-byte UTF-8 characters (Japanese, and likely other CJK) are consistently double-UTF-8-encoded. The corruption is **not** intermittent or rendering-related — the actual byte stream reaching the terminal is wrong, so copying the displayed text (via cursor selection, `/copy`, or any other means) puts mojibake into the system clipboard. This differs from #4700 (streaming-only) and other rendering-related reports.

The same shell, same `$LANG`, same machine works correctly under macOS's built-in Terminal.app, so the bug is isolated to Superset Desktop's terminal pipeline.

## Steps to Reproduce

1. Open Superset Desktop and start any session (Claude Code, plain shell, etc.) in the built-in terminal
2. Run: `echo -n "の" | pbcopy && pbpaste | xxd`
   - `の` is U+306E, a single Japanese hiragana character

## Expected Behavior

The hex dump should be `e3 81 ae` (3 bytes — the correct UTF-8 encoding of `の`).

## Actual Behavior

The hex dump is `c3 a3 c2 81 c2 ae` (6 bytes — double UTF-8 encoding).
The terminal also displays the typed `の` as `ã^Á®` instead of the actual character.

_(Screenshot showing `ã^Á®` on screen and `00000000: c3a3 c281 c2ae` from xxd will be attached as a follow-up.)_

## Root Cause Analysis

This is a textbook double-UTF-8-encoding bug:

```
1. Original character:                の (U+306E)
2. Correct UTF-8 bytes:               E3 81 AE
3. Bytes wrongly decoded as Latin-1:  ã (U+00E3) + <control U+0081> + ® (U+00AE)
4. Re-encoded as UTF-8:               C3 A3   C2 81   C2 AE
5. Result on screen and clipboard:    C3 A3 C2 81 C2 AE   ← matches the observed bytes
```

Somewhere in Superset Desktop's text path — most likely between the PTY and the xterm.js / renderer / clipboard layer — bytes are being decoded as Latin-1/Windows-1252 and then re-encoded as UTF-8.

The bug is in the input/output pipeline, **not** in the renderer alone — proven by the fact that `pbpaste | xxd` shows the wrong bytes are actually written to the clipboard (i.e. the corruption survives a round-trip through the system clipboard, so it cannot be a display-only glitch).

## Workaround

Run the same shell / Claude Code / etc. from macOS's built-in Terminal.app. Same machine, same shell, same `LANG=C.UTF-8`, but works correctly. This narrows the bug down to Superset Desktop.

## Environment

- **Superset Desktop version**: _TBD — will add via comment_
- **macOS**: Darwin 25.3.0
- **Shell**: zsh
- **`$LANG`**: `C.UTF-8` (note: not the cause — same `LANG` works correctly in Terminal.app)
- **Agent used when first noticed**: Claude Code

## Related issues (different symptoms, likely same subsystem)

- #3406 — Chinese characters sometimes garbled
- #3527 — Garbled CJK + layout glitches under heavy usage
- #3570 — Latin/ASCII characters render as corrupted glyphs when mixed with Chinese text
- #3572 — Intermittent garbled text with multiple Claude Code tabs + Chinese
- #3668 — Claude Code rendering issue in Superset terminal
- #3794 — WebGL texture atlas corruption
- #4700 — Streaming-only corruption in Claude Code (this issue is permanent, not streaming-bound)

This issue differs from the above in that it is **100% reproducible**, **not display-only** (the corrupted bytes are actually placed in the system clipboard), and reproducible with a single hex-dump command — which should make it easier to bisect than the intermittent rendering reports.

Kitenite:
## Problem

Copying multi-byte text out of the desktop terminal produces mojibake. Typing/display is correct, but `Cmd+C` / `/copy` (anything using OSC 52) writes double-UTF-8-encoded bytes to the system clipboard:

- `あいうえお` pastes as `ããããã`
- `Sautéed` pastes as `SautÃ©ed`
- box-drawing/divider lines paste as runs of `Ã¢...`

Fixes #4839, #4956.

## Root cause

The full PTY → daemon → host-service → xterm pipeline is already byte-clean (verified by `no-encoding-hops.test.ts`), which is why **display** is correct. The corruption is isolated to the **clipboard write**.

xterm's `@xterm/addon-clipboard` ships a default base64 codec built on `btoa`/`atob`, which only speak Latin-1 (one char per byte). On OSC 52 copy, `atob` returns a string whose chars are the raw UTF-8 bytes; handing that to `navigator.clipboard` re-encodes each char as UTF-8 — classic double-UTF-8 mojibake. Even the latest upstream xterm.js (6.0.0) still ships this codec and exposes the `IBase64` constructor param specifically so consumers can override it.

## Fix

Add a UTF-8-aware `IBase64` (`Utf8Base64`) that routes through `TextEncoder`/`TextDecoder`, and pass it to both `ClipboardAddon` instances. This matches how native terminals decode OSC 52:

- **alacritty**: `Base64.decode(...)` → `String::from_utf8(bytes)`
- **kitty**: `base64.standard_b64decode(x).decode('utf-8')`

…and mirrors the existing `bytesToBase64`/`base64ToBytes` idiom already used in `RemoteTerminal.tsx`.

## Test plan

- New `clipboard-base64.test.ts`: round-trips Japanese, accented Latin, box-drawing, emoji, ASCII; asserts the exact UTF-8 byte payloads; includes a regression guard for the `ããããã` mojibake.
- Verified the test fails against the old `btoa`/`atob` codec (reproduces `ããããã`) and passes with the fix.
- `bun run lint`, `typecheck`, and all 229 terminal-lib tests pass.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4983">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes mojibake when copying from the desktop terminal by decoding OSC 52 clipboard data as UTF-8 and rejecting non-UTF-8 payloads. Japanese, accented Latin, box-drawing, and emoji now copy/paste correctly.

- **Bug Fixes**
  - Introduced `Utf8Base64` implementing `IBase64` via `TextEncoder`/`TextDecoder({ fatal: true })`; passed to `@xterm/addon-clipboard`’s `ClipboardAddon` at both call sites.
  - Replaced `btoa`/`atob` codec to stop double-UTF-8; non-UTF-8 OSC 52 payloads now fail fast instead of pasting U+FFFD.
  - Added `clipboard-base64.test.ts` covering UTF-8 cases, round-trips, and malformed input; dropped a redundant regression test.
  - Fixes #4839, #4956.

<sup>Written for commit 5e6b39034db78507f7702cda894c64cef6053def. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/4983?utm_source=github">Review in cubic</a>

## Summary by CodeRabbit

* **New Features**
  * Added a UTF-8-safe Base64 clipboard codec and configured the app to use it so copy/paste reliably preserves Unicode and multibyte characters.

* **Tests**
  * Added tests verifying encode/decode correctness, round-trip integrity, and proper errors for malformed or non-UTF-8 payloads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4983?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/clipboard-base64.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/clipboard-base64.test.ts
@@ -51,7 +51,9 @@ describe("Utf8Base64", () => {
 		}
 	});
 
-	it("propagates on malformed base64 so the addon can bail", () => {
+	it("throws so the addon can bail on malformed base64 or non-UTF-8 bytes", () => {
 		expect(() => codec.decodeText("@@@not base64@@@")).toThrow();
+		// "/w==" is valid base64 for a lone 0xFF, which is not valid UTF-8.
+		expect(() => codec.decodeText("/w==")).toThrow();
 	});
 });
--- a/apps/desktop/src/renderer/lib/terminal/clipboard-base64.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/clipboard-base64.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from "bun:test";
+import { Utf8Base64 } from "./clipboard-base64";
+
+const codec = new Utf8Base64();
+
+// The exact bytes a shell emits for these characters, base64-encoded. These are
+// what an OSC 52 copy payload actually contains; decodeText must turn them back
+// into the original characters rather than Latin-1 byte-chars.
+const cases: Array<{ name: string; text: string; base64: string }> = [
+	{ name: "ascii", text: "hello", base64: "aGVsbG8=" },
+	// の = U+306E = E3 81 AE
+	{ name: "single hiragana の", text: "の", base64: "44Gu" },
+	// あいうえお = E3 81 82 / E3 81 84 / E3 81 86 / E3 81 88 / E3 81 8A
+	{
+		name: "hiragana run あいうえお",
+		text: "あいうえお",
+		base64: "44GC44GE44GG44GI44GK",
+	},
+	// accented Latin: Sautéed (é = U+00E9 = C3 A9)
+	{ name: "accented latin", text: "Sautéed", base64: "U2F1dMOpZWQ=" },
+	// box-drawing horizontal line U+2500 = E2 94 80
+	{ name: "box drawing ─", text: "─", base64: "4pSA" },
+	// emoji outside the BMP (surrogate pair): 😀 = F0 9F 98 80
+	{ name: "emoji 😀", text: "😀", base64: "8J+YgA==" },
+];
+
+describe("Utf8Base64", () => {
+	it("decodes UTF-8 OSC 52 payloads back to the original text", () => {
+		for (const { name, text, base64 } of cases) {
+			expect(codec.decodeText(base64), name).toBe(text);
+		}
+	});
+
+	it("encodes text to the real UTF-8 byte base64", () => {
+		for (const { name, text, base64 } of cases) {
+			expect(codec.encodeText(text), name).toBe(base64);
+		}
+	});
+
+	it("round-trips arbitrary multibyte text", () => {
+		const samples = [
+			"の",
+			"あいうえお",
+			"Sautéed",
+			"─━│",
+			"😀🎉",
+			"mixed ABC 日本語 123",
+		];
+		for (const sample of samples) {
+			expect(codec.decodeText(codec.encodeText(sample))).toBe(sample);
+		}
+	});
+
+	it("does not reproduce the double-UTF-8 mojibake of the default btoa/atob codec", () => {
+		// Regression guard: the buggy default returned a Latin-1 byte-string for
+		// `の`, which the clipboard then re-encoded as `c3 a3 ...` mojibake.
+		const decoded = codec.decodeText("44GC44GE44GG44GI44GK");
+		expect(decoded).toBe("あいうえお");
+		expect(decoded).not.toContain("ã");
+	});
+
+	it("propagates on malformed base64 so the addon can bail", () => {
+		expect(() => codec.decodeText("@@@not base64@@@")).toThrow();
+	});
+});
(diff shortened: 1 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Implement UTF-8-aware base64 codec for clipboard](https://stagereview.app/superset-sh/superset/pull/4983/chapters/1) |
| 2 | [Add unit tests for UTF-8 clipboard encoding](https://stagereview.app/superset-sh/superset/pull/4983/chapters/2) |
| 3 | [Integrate Utf8Base64 into terminal addon configurations](https://stagereview.app/superset-sh/superset/pull/4983/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/4983">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 5e6b39034db78507f7702cda894c64cef6053def on May 29, 2026 12:26am UTC.</sub>

Kitenite:
Good catch from @greptile-apps and @cubic-dev-ai — fixed in 5e6b39034.

`decodeText` now uses `new TextDecoder("utf-8", { fatal: true })`, so valid-base64-but-invalid-UTF-8 payloads (e.g. a lone `0xFF`) throw instead of silently producing `�`. The addon catches that and clears the clipboard rather than writing altered text — which also lines up better with the precedent: alacritty's `String::from_utf8` returns `Err` (stores nothing) and kitty's `.decode('utf-8')` raises. Added a test case (`/w==` → throws) covering this path.

Note: the failing **Test** check is unrelated to this PR — it's `buildAgentPromptCommand > does not change non-codex commands` in `@superset/shared`, a package this PR doesn't touch.

PR #4996, opened 2026-05-29
fix(host-service): harden DB migration startup (busy_timeout + fail-closed)

Kitenite:
## Problem

The host-service runs DB migrations synchronously inside `createApp()` **before the port binds** (`app.ts` → `serve.ts`). The DB was opened with only WAL + foreign_keys (**no `busy_timeout`**), and `createDb` **swallowed** any migration error and returned the db anyway. desktop 1.12.0 added the first migration since this code shipped (`0005_host_settings_and_project_overrides`), exposing two failure modes under write-lock contention on `host.db`:

- **Half-migrated, served anyway** — `SQLITE_BUSY` swallowed → runtime `no such table host_settings`.
- **Killed mid-migration** — the synchronous stall exceeds the coordinator's 10s health window → SIGKILL → respawn loop, port never binds (terminals dead / 'can't type').

## Fix

- **`db/db.ts`** — set `busy_timeout = 8000` (wait out transient contention) and **fail closed**: throw instead of swallowing. drizzle runs all pending migrations in a single `BEGIN/COMMIT` with `ROLLBACK`+rethrow, so a failure leaves the DB at its prior version (never half-applied). The throw propagates to `serve.ts` `main().catch(… process.exit(1))`, so the coordinator's health check fails instead of serving a broken DB.
- **`host-service-utils.ts`** — raise `HEALTH_POLL_TIMEOUT_MS` 10s → 20s so a recoverable ~8s migration stall isn't SIGKILLed before the port binds.

## Reproduction + verification

New integration test `db.contention.node-test.ts`: real `createDb` + real `0005` + a sibling process holding the write lock. Run with `bun run test:integration:db` (Electron-as-node — `better-sqlite3` here is built for the Electron ABI; bun can't load it and plain node is the wrong ABI).

- ✔ baseline: no contention → 0005 applies, `host_settings` exists
- ✔ fail-closed: sustained contention → `createDb` **throws**, DB left cleanly unmigrated
- ✔ recovery: contention clears within the timeout → waits then succeeds

Host-service unit suite 691/0, typecheck + lint clean.

## Scope / follow-up

This hardens migration startup (resilience + correctness). It does **not** address the *source* of the contention — whether two host-services can race the same `host.db` — which is a separate, root-cause lifecycle change (reap orphan host-service before spawn / single-writer guarantee) tracked in a follow-up PR. The plan doc (`plans/host-service-migration-startup-hardening.md`) covers both.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4996">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Hardened host-service DB migration at startup by failing closed so we never serve a half-migrated DB. Removed the migration `busy_timeout` and focused on surfacing failures; added an Electron-as-node integration test for write-lock contention.

- **Bug Fixes**
  - `src/db/db.ts`: removed the try/catch that swallowed migration errors; `migrate()` now throws. drizzle keeps migrations atomic, so failures roll back cleanly. Dropped the migration `busy_timeout`.

- **Tests**
  - `src/db/db.contention.node-test.ts`: integration test holds a write lock in a sibling process and asserts fail-closed behavior.
  - `scripts/run-db-integration-test.ts` + `test:integration:db`: run under Electron-as-node to load `better-sqlite3`.

<sup>Written for commit 1bbd6c40f27104806c8be91b2c2b778f5fdc5201. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/4996?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration error handling for host service to ensure the application fails cleanly at startup when database contention is detected, rather than silently continuing with an unmigrated database.

* **Tests**
  * Added integration test suite to verify database behavior and recovery under write-lock contention scenarios.

### DIFF
--- a/apps/desktop/src/main/lib/host-service-utils.ts
+++ b/apps/desktop/src/main/lib/host-service-utils.ts
@@ -5,11 +5,7 @@ import path from "node:path";
 /** Rotate per-org host-service.log once it exceeds this size. */
 export const MAX_HOST_LOG_BYTES = 5 * 1024 * 1024;
 
-// Must exceed the host-service's DB-migration budget (MIGRATION_BUSY_TIMEOUT_MS,
-// 8s) plus startup overhead: migrations run synchronously before the port binds,
-// so a recoverable lock stall delays the first successful health response. Too
-// tight and the coordinator SIGKILLs a host-service that was about to come up.
-export const HEALTH_POLL_TIMEOUT_MS = 20_000;
+export const HEALTH_POLL_TIMEOUT_MS = 10_000;
 
 const HEALTH_POLL_INTERVAL_MS = 200;
 
--- a/apps/desktop/src/main/lib/host-service-utils.ts
+++ b/apps/desktop/src/main/lib/host-service-utils.ts
@@ -5,7 +5,11 @@ import path from "node:path";
 /** Rotate per-org host-service.log once it exceeds this size. */
 export const MAX_HOST_LOG_BYTES = 5 * 1024 * 1024;
 
-export const HEALTH_POLL_TIMEOUT_MS = 10_000;
+// Must exceed the host-service's DB-migration budget (MIGRATION_BUSY_TIMEOUT_MS,
+// 8s) plus startup overhead: migrations run synchronously before the port binds,
+// so a recoverable lock stall delays the first successful health response. Too
+// tight and the coordinator SIGKILLs a host-service that was about to come up.
+export const HEALTH_POLL_TIMEOUT_MS = 20_000;
 
 const HEALTH_POLL_INTERVAL_MS = 200;
 
--- a/packages/host-service/src/db/db.contention.node-test.ts
+++ b/packages/host-service/src/db/db.contention.node-test.ts
@@ -105,15 +105,6 @@ function tableExists(targetDbPath: string, tableName: string): boolean {
 }
 
 describe("host-service DB migration under write-lock contention", () => {
-	test("baseline: no contention → 0005 applies, host_settings exists", () => {
-		createDb(dbPath, MIGRATIONS_FOLDER);
-		assert.equal(
-			tableExists(dbPath, "host_settings"),
-			true,
-			"host_settings should exist after a clean migration",
-		);
-	});
-
 	test("fail-closed: sustained contention past the busy timeout → createDb THROWS, DB left cleanly unmigrated", async () => {
 		// Hold the lock for 20s — longer than the 8s busy timeout, so migrate
 		// can't acquire it.
--- a/packages/host-service/src/db/db.contention.node-test.ts
+++ b/packages/host-service/src/db/db.contention.node-test.ts
@@ -1,0 +1,177 @@
+// Integration test for the 1.12.0 host-service migration crash and its fix.
+//
+// Real `createDb` + real drizzle migration files (incl. 0005) + a real second
+// process holding the write lock. Runs under Node (`node --test`) but MUST be
+// invoked with the Electron-ABI node — better-sqlite3 here is built for the
+// host-service runtime (Electron), which is also the only runtime that can
+// load it. Same-process lock holding would deadlock (better-sqlite3 is
+// synchronous and single-threaded), so we spawn a sibling process for genuine
+// cross-process contention.
+//
+//   ELECTRON_RUN_AS_NODE=1 <electron> --experimental-strip-types --test \
+//     src/db/db.contention.node-test.ts
+//
+// Pre-fix behavior (reproduced): contention → migrate stalls on the busy
+// timeout → SQLITE_BUSY → db.ts swallowed it → service came up on a
+// half-migrated DB (runtime "no such table host_settings").
+//
+// Post-fix behavior (asserted here):
+//   - sustained contention past the busy timeout → createDb THROWS (fail
+//     closed), and the DB is left cleanly unmigrated (atomic rollback), never
+//     half-applied;
+//   - contention that clears within the busy timeout → createDb WAITS then
+//     succeeds (host_settings present).
+
+import { strict as assert } from "node:assert";
+import * as childProcess from "node:child_process";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import { afterEach, beforeEach, describe, test } from "node:test";
+import { fileURLToPath } from "node:url";
+import Database from "better-sqlite3";
+import { createDb } from "./db.ts";
+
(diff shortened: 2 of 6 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Document the migration hardening strategy](https://stagereview.app/superset-sh/superset/pull/4996/chapters/1) |
| 2 | [Harden DB migration and fail closed](https://stagereview.app/superset-sh/superset/pull/4996/chapters/2) |
| 3 | [Add DB contention integration tests](https://stagereview.app/superset-sh/superset/pull/4996/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/4996">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 1bbd6c40f27104806c8be91b2c2b778f5fdc5201 on Jun 2, 2026 6:00pm UTC.</sub>

Kitenite (commented):
Automated review (high effort). The PR is focused and the migration-contention repro test is excellent. Findings below are non-blocking; the CLI one is the most worth addressing.

**Cross-file (not in this diff) — the CLI keeps its own 10s health poll.** `packages/cli/src/lib/host/spawn.ts:16` defines a separate `HEALTH_POLL_TIMEOUT_MS = 10_000` with its own `pollHealth()` (line 50). The CLI also spawns a host-service, so under the exact migration contention this PR hardens against, the CLI would give up and report failure at 10s — it doesn't get the widened window or the busy_timeout reasoning. Either bump the CLI copy to match, or extract a shared constant. Flagging because it's the same failure mode in a path this PR doesn't cover.

Inline notes on the rest. (I refuted a few finder candidates: `BEGIN IMMEDIATE` does take a RESERVED lock so the test's contention is real; `holder.kill()` runs before the assertions so there's no orphan leak; and `bun test` doesn't match `*.node-test.ts`.)

packages/host-service/src/db/db.ts
```diff
@@ -7,23 +7,45 @@ import * as schema from "./schema.ts";
...
+	// BEGIN/COMMIT and ROLLBACKs on any error, so a failure leaves the DB at
+	// its prior version (never half-applied). Throwing here propagates to
+	// `serve.ts` `main().catch(... process.exit(1))` so the coordinator's
+	// health check fails and it can recover (kill the stale process, respawn)
+	// instead of silently serving a DB that's missing tables.
 	try {
 		migrate(db, { migrationsFolder });
 	} catch (error) {
 		console.error("[host-service:db] Migration failed:", error);
+		sqlite.close();
```
Kitenite: `sqlite.close()` can itself throw — if it does, that error propagates instead of the `Error(..., { cause })` below, masking the real migration failure. Guard it so the original error always wins:
```suggestion
		try {
			sqlite.close();
		} catch {}
```

packages/host-service/src/db/db.ts
```diff
@@ -7,23 +7,45 @@ import * as schema from "./schema.ts";
...
 	sqlite.pragma("foreign_keys = ON");
+	sqlite.pragma(`busy_timeout = ${MIGRATION_BUSY_TIMEOUT_MS}`);
 
 	const db = drizzle(sqlite, { schema });
 
 	console.error(
 		`[host-service:db] Initialized at ${dbPath}, migrations from ${migrationsFolder}`,
 	);
 
+	// Fail closed. drizzle runs all pending migrations in a single
```
Kitenite: Failing closed is right for a half-migrated DB, but worth noting: the *persistent*-contention path now has no automatic recovery. throw → `process.exit(1)` → coordinator health-check fails → it throws to the caller; there's no reap/auto-reset on the normal `spawn()` path, so a permanently-stuck lock turns a degraded-but-partially-usable state into a hard outage / respawn cycle. That recovery is exactly #4997 (reap orphan before spawn) — worth keeping the two close together so this doesn't ship alone.

packages/host-service/src/db/db.contention.node-test.ts
```diff
@@ -0,0 +1,177 @@
...
+import Database from "better-sqlite3";
+import { createDb } from "./db.ts";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+// packages/host-service/src/db → packages/host-service/drizzle
+const MIGRATIONS_FOLDER = path.resolve(__dirname, "../../drizzle");
+
+// Mirror of MIGRATION_BUSY_TIMEOUT_MS in db.ts. Kept local so the test pins the
+// behavior even if the source constant moves.
+const BUSY_TIMEOUT_MS = 8_000;
```
Kitenite: This hand-mirrors `MIGRATION_BUSY_TIMEOUT_MS` from `db.ts`. If the source constant changes, the timing assertions silently validate the wrong budget. Prefer exporting `MIGRATION_BUSY_TIMEOUT_MS` from `db.ts` and importing it here so they can't drift.

apps/desktop/src/main/lib/host-service-utils.ts
```diff
@@ -5,7 +5,11 @@ import path from "node:path";
 /** Rotate per-org host-service.log once it exceeds this size. */
 export const MAX_HOST_LOG_BYTES = 5 * 1024 * 1024;
 
-export const HEALTH_POLL_TIMEOUT_MS = 10_000;
+// Must exceed the host-service's DB-migration budget (MIGRATION_BUSY_TIMEOUT_MS,
+// 8s) plus startup overhead: migrations run synchronously before the port binds,
+// so a recoverable lock stall delays the first successful health response. Too
+// tight and the coordinator SIGKILLs a host-service that was about to come up.
+export const HEALTH_POLL_TIMEOUT_MS = 20_000;
```
Kitenite: Doubling the window also doubles the worst-case *failure*-detection latency: `coordinator.spawn()` awaits the full `pollHealthCheck` timeout and doesn't fail-fast on the child's `exit` event, so a genuinely dead/crashing host-service now takes ~20s (was ~10s) to be declared failed. Acceptable for now; the clean fix is to race `pollHealthCheck` against the child `exit` so the window can stay generous without the latency cost.

PR #4997, opened 2026-05-29
fix(host-service): reap orphaned host-service before spawn (single-writer)

Kitenite:
## Problem

The coordinator only tracks host-services it spawned in **this** Electron process (its in-memory `instances` map). After a restart that map is empty, so a host-service left alive by a *previous* process is invisible — and `spawn()` never consults the manifest before starting a new one. It just binds a competitor on a fresh port, giving **two processes connected to the same org `host.db`**. That concurrent-writer contention is the root trigger behind the migration lock-up (see #4996, which hardens the migration side).

How an orphan arises:
- Electron died before `before-quit` ran (hard crash / SIGKILL / OOM) — child never SIGTERMed, parent watchdog hadn't fired yet;
- a **CLI-spawned** host-service — no `HOST_PARENT_PID`, so no watchdog at all;
- a **wedged** child from a failed spawn that ignored SIGTERM.

(`reset()` already reaps via the manifest, but it's only reachable manually via tRPC — never on the normal start path.)

## Fix

Reap based on the manifest (the only cross-process record) **before spawning**, in `spawn()`:

- **Identity-validated kill (PID-reuse safe):** only terminate a pid we can positively confirm is our host-service — its endpoint answers `health.check` with the manifest secret, or (best-effort, POSIX) its command line matches our `host-service.js`. A live-but-unrecognized pid is **left alone** and the stale manifest dropped, so a recycled pid can't cause friendly-fire.
- **Escalation:** SIGTERM → await exit → SIGKILL → await.

Why safe to kill: host-service is stateless w.r.t. terminals — PTYs live in the detached pty-daemon, which a fresh host-service re-adopts. Reaping an orphan doesn't touch user sessions.

## Structure

- `host-service-reap.ts` — dependency-injected reap logic (electron-free, so it unit-tests under `bun test` without spawning real processes).
- `host-service-coordinator.ts` — wires real deps + the health/argv identity probe; calls reap at the top of `spawn()`.
- `host-service-utils.ts` — adds `healthCheckOnce` (single-shot identity probe; `pollHealthCheck` now reuses it).
- `host-service-reap.test.ts` — 6 cases: no-manifest, own-live-instance, dead-pid-cleared, **unidentified-left-alone (PID reuse)**, SIGTERM-exit, **SIGTERM→SIGKILL escalation**.

Desktop typecheck + lint clean; reap tests 6/6.

## Relationship to #4996

Independent and complementary. #4996 makes the DB migration survive contention (busy_timeout + fail-closed); this PR removes the *source* of the contention. Either can merge first.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4997">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Prevents two host-service processes from writing to the same org DB by reaping any orphaned host-service before spawn. Uses manifest-driven identity checks to safely terminate previous instances and ensure a single writer, resolving migration contention.

- **Bug Fixes**
  - Reap any previous host-service for the org at the start of `spawn()` using the manifest.
  - Identity-validated kill: `health.check` with the manifest secret, with a POSIX argv fallback; unrecognized live PIDs are left alone and the stale manifest is removed.
  - Escalation flow: `SIGTERM` with grace, then `SIGKILL` if needed; clears the manifest after successful reaping or when the PID is dead.
  - Added `healthCheckOnce` and updated `pollHealthCheck` to reuse it.
  - Introduced dependency-injected `host-service-reap.ts` with unit tests covering no-manifest, own-live, dead-pid, PID-reuse, and SIGTERM→SIGKILL paths.

<sup>Written for commit 19143e535ac84d70fcaf0125b9353b8c95a255e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/4997?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Enhanced host-service process lifecycle management with proactive cleanup of previous instances to prevent conflicts.
  
* **Tests**
  * Added comprehensive test coverage for host-service process cleanup logic, including edge cases for process termination and manifest handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4997?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/desktop/src/main/lib/host-service-coordinator.ts
+++ b/apps/desktop/src/main/lib/host-service-coordinator.ts
@@ -11,15 +11,18 @@ import { env as sharedEnv } from "shared/env.shared";
 import { getProcessEnvWithShellPath } from "../../lib/trpc/routers/workspaces/utils/shell-env";
 import { SUPERSET_HOME_DIR } from "./app-environment";
 import {
+	type HostServiceManifest,
 	isProcessAlive,
 	killProcess,
 	manifestDir,
 	readManifest,
 	removeManifest,
 } from "./host-service-manifest";
+import { reapPreviousHostService } from "./host-service-reap";
 import {
 	findFreePort,
 	HEALTH_POLL_TIMEOUT_MS,
+	healthCheckOnce,
 	MAX_HOST_LOG_BYTES,
 	openRotatingLogFd,
 	pollHealthCheck,
@@ -338,11 +341,53 @@ export class HostServiceCoordinator extends EventEmitter {
 
 	// ── Spawn ─────────────────────────────────────────────────────────
 
+	/**
+	 * Reap a host-service left running for this org by a previous process. See
+	 * `host-service-reap.ts` for why one can exist and the PID-reuse safety
+	 * rules. Identity is confirmed via the manifest's health endpoint (the
+	 * secret proves it's ours) with a best-effort argv check as a fallback for
+	 * a wedged child whose endpoint no longer answers.
+	 */
+	private async reapPreviousHostService(organizationId: string): Promise<void> {
+		await reapPreviousHostService(organizationId, {
+			readManifest,
+			removeManifest,
+			isProcessAlive,
+			killProcess,
+			isOwnLivePid: (pid) => {
+				const current = this.instances.get(organizationId);
+				return !!current && current.pid === pid && current.status !== "stopped";
+			},
+			confirmOurHostService: (manifest) => this.confirmOurHostService(manifest),
+			log: {
+				info: (message) => log.info(message),
+				warn: (message) => log.warn(message),
+			},
+		});
+	}
+
+	private async confirmOurHostService(
+		manifest: HostServiceManifest,
+	): Promise<boolean> {
+		// Strongest signal: the endpoint answers with the manifest secret.
+		if (await healthCheckOnce(manifest.endpoint, manifest.authToken, 1_500)) {
+			return true;
+		}
+		// Fallback for a wedged child whose endpoint no longer responds: match
+		// the pid's command line to our host-service script. POSIX only — on
+		// win32 we fall back to health-only and leave unrecognized pids alone.
+		return processLooksLikeHostService(manifest.pid, this.scriptPath);
+	}
+
 	private async spawn(
 		organizationId: string,
 		config: SpawnConfig,
 		preferredPorts: Iterable<number> = this.getPreferredPorts(organizationId),
 	): Promise<Connection> {
+		// Single-writer guarantee: reap any host-service left running for this
+		// org by a previous process before we bind a competitor to its host.db.
+		await this.reapPreviousHostService(organizationId);
+
 		const port = await findFreePort(preferredPorts);
 		this.rememberPort(organizationId, port);
 		const secret = randomBytes(32).toString("hex");
@@ -531,6 +576,27 @@ function pipeWithPrefix(
 		if (pending) target.write(`${tag} ${pending}\n`);
 		pending = "";
 	});
+}
+
+/**
+ * Best-effort identity check for a wedged host-service whose endpoint no longer
+ * answers: does the pid's command line reference our host-service script?
+ * POSIX only (uses `ps`); returns false on win32 and on any error, so an
+ * unrecognized pid is left alone rather than killed.
+ */
+function processLooksLikeHostService(pid: number, scriptPath: string): boolean {
+	if (process.platform === "win32") return false;
+	try {
+		const command = childProcess
+			.execFileSync("ps", ["-p", String(pid), "-o", "command="], {
+				encoding: "utf8",
+				timeout: 1_500,
+			})
+			.trim();
+		return command.includes(path.basename(scriptPath));
+	} catch {
+		return false;
+	}
 }
 
 let coordinator: HostServiceCoordinator | null = null;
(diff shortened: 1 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Extract health check probe to utility](https://stagereview.app/superset-sh/superset/pull/4997/chapters/1) |
| 2 | [Implement core host-service reaping logic](https://stagereview.app/superset-sh/superset/pull/4997/chapters/2) |
| 3 | [Integrate reaping into the coordinator](https://stagereview.app/superset-sh/superset/pull/4997/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/4997">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 19143e535ac84d70fcaf0125b9353b8c95a255e8 on May 29, 2026 6:23pm UTC.</sub>

apps/desktop/src/main/lib/host-service-coordinator.ts
```diff
@@ -338,11 +341,53 @@ export class HostServiceCoordinator extends EventEmitter {
...
+		manifest: HostServiceManifest,
+	): Promise<boolean> {
+		// Strongest signal: the endpoint answers with the manifest secret.
+		if (await healthCheckOnce(manifest.endpoint, manifest.authToken, 1_500)) {
+			return true;
+		}
+		// Fallback for a wedged child whose endpoint no longer responds: match
+		// the pid's command line to our host-service script. POSIX only — on
+		// win32 we fall back to health-only and leave unrecognized pids alone.
+		return processLooksLikeHostService(manifest.pid, this.scriptPath);
```
Kitenite, line 379: The health probe (line 373) is precise — it uses the manifest secret, so it confirms *this specific* host-service. This `ps` fallback is not: it matches any process whose command line contains `host-service.js`, and the org is passed via env (`ORGANIZATION_ID`, ~line 500), **not argv** — so `ps -o command` can't tell orgs apart.

Friendly-fire scenario: org A's manifest pid is dead; the OS reuses that pid for **org B's** host-service; the health probe to A's stale endpoint fails → this fallback matches `host-service.js` → we SIGTERM/SIGKILL org B's healthy host-service. That undercuts the PID-reuse safety the health probe is supposed to give.

Options: drop the argv fallback and rely on the health probe alone (a wedged-ours child that never answers health is rare, and `reset()` still covers it), or make the fallback org-specific (hard, since org isn't in argv). Minor, same line: `execFileSync` is synchronous, so this blocks the Electron main event loop for up to the 1.5s `ps` timeout whenever the health probe fails.

apps/desktop/src/main/lib/host-service-reap.ts
```diff
@@ -0,0 +1,140 @@
...
+		`[host-service:${organizationId}] reaping previous host-service pid=${pid} before spawn`,
+	);
+
+	let reason: ReapReason = "terminated-sigterm";
+	try {
+		deps.killProcess(pid, "SIGTERM");
+	} catch {
+		// Exited between the liveness check and the signal — nothing to do.
+		deps.removeManifest(organizationId);
+		return { reaped: true, reason: "terminated-sigterm", pid };
```
Kitenite, line 113: Returning `reaped: true` / `"terminated-sigterm"` here is slightly misleading: `killProcess` threw, which (per the comment above) means the process had already exited on its own — we didn't terminate it. Harmless since the manifest is removed either way, but a distinct reason like `"already-exited"` would keep the result/telemetry honest.

apps/desktop/src/main/lib/host-service-reap.test.ts
```diff
@@ -0,0 +1,103 @@
...
+			isProcessAlive: mock(() => !sigtermed),
+		});
+		const result = await reapPreviousHostService(ORG, deps);
+		expect(result.reaped).toBe(true);
+		expect(result.reason).toBe("terminated-sigterm");
+		expect(signals).toEqual(["SIGTERM"]);
+		expect(deps.removeManifest).toHaveBeenCalledTimes(1);
+	});
+
+	test("alive + ours + ignores SIGTERM → escalates to SIGKILL", async () => {
```
Kitenite, line 86: Two branches stay uncovered: (a) the SIGTERM-throws path in `reap.ts` (`killProcess` is never mocked to throw) — a case asserting it still returns `reaped:true` + removes the manifest would pin it; and (b) the real `confirmOurHostService` identity logic is injected as a fake, so the `ps`/health fallback (and the friendly-fire risk flagged in the coordinator) has no test at all. Worth a focused test of the real identity probe.

Kitenite (commented):
Automated review (high effort) of the reap / single-writer change. Solid and well-factored — the dependency injection makes the core logic genuinely unit-testable. One finding worth addressing before merge (the `ps`-fallback friendly-fire); the rest are minor.

Non-blocking notes:
- The `isOwnLivePid` guard (coordinator ~357, reap.ts) is effectively unreachable on the `spawn()` path: `this.instances.set` runs *after* reap, so the org has no tracked instance at reap time. Harmless and fine as defense-in-depth, but currently dead.
- `pollHealthCheck`'s refactor preserves the prior per-attempt-2s / overall-deadline behavior exactly — no regression there.

PR #5005, opened 2026-05-31
fix(desktop): tolerate post-checkout hook SIGPIPE on worktree creation (#4350)

### LINKED ISSUE #4350: [bug] Worktree creation reports 'Failed to create worktree' when post-checkout hook exits 141 / SIGPIPE — even though the worktree is fully created
#### Describe the bug

When a repo's `post-checkout` hook exits non-zero with output that doesn't contain the words "post-checkout", "hook", "husky", or "command not found", Superset reports `Failed to create worktree: …` even though `git worktree add` finished its work and the worktree is fully created and registered with git.

The most common trigger is a hook whose pipeline gets **SIGPIPE / exit 141** (e.g. `git worktree list --porcelain | awk '... exit'` on a machine with many worktrees — `awk` exits early, `git worktree list` SIGPIPEs, and with `set -euo pipefail` the hook propagates 141 to the parent `git worktree add`). The hook's *side effects* succeeded; only the exit code is junk.

#### Where the gap is

After #1432 reverted the blanket `-c core.hooksPath=/dev/null` from #1384, hook tolerance is gated by `isPostCheckoutHookFailure` in `apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.ts`:

```ts
export function isPostCheckoutHookFailure(error: unknown): boolean {
  const text = getErrorText(error).toLowerCase();
  if (!text.includes("post-checkout")) {
    return false;
  }
  return (
    text.includes("hook") ||
    text.includes("husky") ||
    text.includes("command not found")
  );
}
```

`runWithPostCheckoutHookTolerance` only invokes its `didSucceed` callback (which checks `isWorktreeRegistered`) when this keyword check matches. A SIGPIPE failure produces no matching keyword in stderr — `didSucceed` is never called — and the error rethrows even though the worktree is fully present on disk and in `git worktree list`.

So the keyword gate is **under-inclusive**: any hook that exits non-zero without one of those keywords is treated as fatal regardless of whether the worktree actually exists.

#### Reproduction

1. macOS, Superset latest. A repo with **a lot** of worktrees registered (so `git worktree list` produces multi-line output).
2. A `core.hooksPath`-configured `post-checkout` like:
   ```bash
   #!/usr/bin/env bash
   set -euo pipefail
   main="$(git worktree list --porcelain | awk '/^worktree / {print $2; exit}')"
   ```
3. Create a workspace in Superset.
4. Result: "Failed to create worktree: Command failed: git -C … worktree add --no-track -b … origin/main Preparing worktree (new branch '…')".
5. Verify directly: `git -C <repo> worktree add … origin/main; echo $?` → prints "Preparing worktree…" + "HEAD is now at…", then **exit 141** (128 + SIGPIPE). The worktree exists at the target path and shows up in `git worktree list`.

The output Superset surfaces is also truncated mid-stream because `execFile`'s stdout buffer is whatever git managed to write before the hook process closed the pipe — the line that *would* tell the user the real reason (the hook's exit code / signal) never makes it into the error.

#### Suggested fix

Treat the side-effect check as the source of truth, not a post-filter on the error text. Two options, in order of preference:

**Option A — always run `didSucceed` on any non-zero exit.** The helper is *already* parameterised on a side-effect check (`isWorktreeRegistered`); the keyword gate is just an optimization.

```ts
export async function runWithPostCheckoutHookTolerance({ run, didSucceed, context }) {
  try {
    await run();
  } catch (error) {
    let succeeded = false;
    try { succeeded = await didSucceed(); } catch {}
    if (!succeeded) throw error;
    console.warn(`[git] ${context} but command exited non-zero (non-fatal): ${getErrorText(error)}`);
  }
}
```

**Option B — extend `isPostCheckoutHookFailure` to recognize SIGPIPE / exit 141.**

```ts
export function isPostCheckoutHookFailure(error: unknown): boolean {
  const ex = error as { signal?: NodeJS.Signals; code?: number | string };
  if (ex?.signal === "SIGPIPE" || ex?.code === 141) return true;
  // …existing keyword checks
}
```

Option A is the cleaner design — it removes the under/over-inclusiveness of keyword matching entirely. Any non-zero exit where the worktree is fully registered is by definition a post-effect failure that shouldn't block workspace creation.

Either way, it would also help to surface **git's full stderr/stdout** in the error path (the current message truncates mid-line, which made this nearly impossible to diagnose without re-running the command manually).

#### Workaround for affected users

Audit `post-checkout` hooks for pipelines where a downstream consumer uses early-exit (`awk … exit`, `head -n 1`, etc.) on a stream that may have more lines than consumed. Either:

- replace the early-exit consumer with one that reads the full stream (e.g. `awk '/pat/ && !seen { print; seen=1 }'` instead of `awk '/pat/ {print; exit}'`); or
- run that pipeline with `set +o pipefail` in a subshell so SIGPIPE upstream of the consumer doesn't fail the pipeline.

#### Related

- #961 — original Husky/NVM PATH issue
- #1384 — blanket `core.hooksPath=/dev/null` workaround (now reverted)
- #1432 — revert; introduces the keyword-based tolerance helper

#### Environment

- macOS 15.4 (Darwin 25.4.0)
- Superset (host app from superset.sh) — latest at time of report
- git 2.x via Homebrew
- Reproduces on any repo whose `post-checkout` hook can exit 141 (SIGPIPE) — most common when the hook reads `git worktree list` and the repo has many worktrees registered.

AviPeltz:
## Problem

Fixes #4350.

`git worktree add` (and branch checkout) run the repo's `post-checkout` hook **after** the worktree is created and the branch is checked out. A flaky hook can exit non-zero **with no identifying output at all** — most reproducibly a pipeline that dies with **SIGPIPE / exit 141** under `set -o pipefail`, e.g.:

```sh
git worktree list --porcelain | awk '/…/ { …; exit }'   # awk exits early → git SIGPIPEs → 141
```

The hook's side effects succeeded; only the exit code is junk — but Superset reported **"Failed to create worktree"** even though the worktree was fully created and registered with git.

## Root cause

Tolerance was gated by `isPostCheckoutHookFailure`, which only forgave failures whose stderr text contained `post-checkout` **and** one of `hook`/`husky`/`command not found`. The on-disk success check (`isWorktreeRegistered`) ran **only if** that keyword match passed:

```ts
if (!isPostCheckoutHookFailure(error)) throw error;  // SIGPIPE has none of these keywords
succeeded = await didSucceed();                       // ...so this never ran
```

A SIGPIPE death produces none of those keywords, so `didSucceed` was never consulted and the error rethrew despite the worktree existing.

## Fix

Drop the stderr-keyword heuristic entirely. When the command exits non-zero, **always** ask `didSucceed()` whether the concrete intended outcome is real (worktree registered in `git worktree list` / branch switched). If it is, the exit is non-fatal; otherwise the original error rethrows unchanged.

This is safe because genuine git failures abort **before** the worktree is registered:
- `git worktree add -b <branch>` fails on a real error (e.g. branch already exists) before creating anything → `isWorktreeRegistered` is false → still throws.
- Only failures **after** git finished its work (hook flakiness, SIGPIPE) get tolerated — exactly the bug.

## Tests

`git-hook-tolerance.test.ts`:
- ➕ exit-141 / empty-stderr failure + outcome present → resolves (the regression).
- 🔧 the old "non-hook failure" test paired a fatal error with `didSucceed: true`, which can't happen for real callers — changed to `didSucceed: false`, preserving the "genuine failures rethrow" guarantee.
- ➕ a throwing `didSucceed` rethrows the original error.

## Verification

- `bun test …/git-hook-tolerance.test.ts` → 5 pass / 0 fail
- `bun run typecheck --filter=@superset/desktop` → clean
- Biome → clean

---

<a href="https://stagereview.app/superset-sh/superset/pull/5005">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Tolerates flaky post-checkout hooks and SIGPIPE (exit 141) during `git worktree add` and branch checkout when the operation actually succeeded. Genuine git failures still throw.

- **Bug Fixes**
  - Limits tolerance to recognized hook failures or SIGPIPE only (141/`SIGPIPE`); usage errors (128/129/1) and other non-zero exits rethrow.
  - Verifies success via `didSucceed()` for tolerated failures; if false or it throws, rethrow the original error.
  - Tests and logs: cover SIGPIPE no-output and exit-129 usage errors, ensure "already exists" still rethrows even if the probe passes, handle a throwing success probe, and clarify the warning message.

<sup>Written for commit b69876a905ab35a7729d68c0d2292c4237a8e68e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5005?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of git operations by treating certain hook pipeline terminations (e.g., SIGPIPE/exit 141) as non-fatal when the repository state was successfully updated; genuine git errors are still rethrown. Clarified warning text for non-zero hook exits.

* **Tests**
  * Expanded coverage for hook-failure scenarios, including signal-style terminations, proper rethrowing of real git failures, and when success-probe checks themselves fail.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.test.ts
@@ -78,6 +78,26 @@ describe("runWithPostCheckoutHookTolerance", () => {
 		).rejects.toThrow("already exists");
 	});
 
+	test("re-throws git usage errors (exit 129) instead of misreading them as a signal failure", async () => {
+		// Git usage errors exit 129. That is numerically 128 + SIGHUP, so a naive
+		// `code > 128` check would wrongly tolerate it; only SIGPIPE (141) is
+		// forgiven. With the outcome probe passing, this must still rethrow.
+		const usageError = Object.assign(new Error("usage: git worktree add ..."), {
+			stderr: "usage: git worktree add ...",
+			code: 129,
+		});
+
+		await expect(
+			runWithPostCheckoutHookTolerance({
+				context: "Created worktree",
+				run: async () => {
+					throw usageError;
+				},
+				didSucceed: async () => true,
+			}),
+		).rejects.toThrow("usage: git worktree add");
+	});
+
 	test("re-throws the original error when the success check itself throws", async () => {
 		const hookError = Object.assign(new Error("post-checkout hook failed"), {
 			code: 1,
--- a/apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.test.ts
@@ -7,6 +7,7 @@ describe("runWithPostCheckoutHookTolerance", () => {
 			new Error("husky - post-checkout script failed"),
 			{
 				stderr: "husky - command not found in PATH=...",
+				code: 1,
 			},
 		);
 
@@ -23,11 +24,11 @@ describe("runWithPostCheckoutHookTolerance", () => {
 
 	test("treats a SIGPIPE/exit-141 failure with no diagnostic output as non-fatal when the worktree was created", async () => {
 		// A post-checkout hook pipeline that dies with SIGPIPE under `set -o
-		// pipefail` surfaces as a non-zero exit with no "post-checkout"/"hook"
-		// keywords at all — the case that regressed worktree creation (#4350).
+		// pipefail` surfaces as exit 141 (128 + SIGPIPE) with no "post-checkout"/
+		// "hook" keywords at all — the case that regressed worktree creation (#4350).
 		const sigpipeError = Object.assign(
 			new Error("Command failed with exit code 141"),
-			{ stderr: "" },
+			{ stderr: "", code: 141 },
 		);
 
 		await expect(
@@ -41,8 +42,10 @@ describe("runWithPostCheckoutHookTolerance", () => {
 		).resolves.toBeUndefined();
 	});
 
-	test("re-throws failures when the intended outcome is absent", async () => {
-		const hookError = new Error("post-checkout hook failed");
+	test("re-throws hook failures when the intended outcome is absent", async () => {
+		const hookError = Object.assign(new Error("post-checkout hook failed"), {
+			code: 1,
+		});
 
 		await expect(
 			runWithPostCheckoutHookTolerance({
@@ -55,25 +58,30 @@ describe("runWithPostCheckoutHookTolerance", () => {
 		).rejects.toThrow("post-checkout");
 	});
 
-	test("re-throws genuine git failures that never created the worktree", async () => {
-		// A real `git worktree add` failure (e.g. branch already exists) aborts
-		// before the worktree is registered, so didSucceed is false and the error
-		// must propagate unchanged.
-		const genericError = new Error("fatal: '../worktree' already exists");
+	test("re-throws genuine git failures even when the success probe would pass", async () => {
+		// A real `git worktree add` failure ("fatal: … already exists") exits 128
+		// with no hook keywords. Even if a stale/pre-existing worktree at the same
+		// path makes the path-based probe return true, we must NOT swallow it.
+		const fatalError = Object.assign(
+			new Error("fatal: '../worktree' already exists"),
+			{ stderr: "fatal: '../worktree' already exists", code: 128 },
+		);
 
 		await expect(
 			runWithPostCheckoutHookTolerance({
 				context: "Created worktree",
 				run: async () => {
-					throw genericError;
+					throw fatalError;
 				},
-				didSucceed: async () => false,
+				didSucceed: async () => true,
 			}),
 		).rejects.toThrow("already exists");
 	});
 
 	test("re-throws the original error when the success check itself throws", async () => {
-		const hookError = new Error("post-checkout hook failed");
+		const hookError = Object.assign(new Error("post-checkout hook failed"), {
+			code: 1,
+		});
 
 		await expect(
 			runWithPostCheckoutHookTolerance({
--- a/apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/utils/git-hook-tolerance.test.ts
@@ -21,7 +21,27 @@ describe("runWithPostCheckoutHookTolerance", () => {
 		).resolves.toBeUndefined();
 	});
 
-	test("re-throws hook failures when operation did not succeed", async () => {
+	test("treats a SIGPIPE/exit-141 failure with no diagnostic output as non-fatal when the worktree was created", async () => {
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Update Git exception types and internal helpers](https://stagereview.app/superset-sh/superset/pull/5005/chapters/1) |
| 2 | [Implement SIGPIPE detection and tolerance logic](https://stagereview.app/superset-sh/superset/pull/5005/chapters/2) |
| 3 | [Verify hook tolerance with regression tests](https://stagereview.app/superset-sh/superset/pull/5005/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5005">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit b69876a905ab35a7729d68c0d2292c4237a8e68e on May 31, 2026 2:46am UTC.</sub>

AviPeltz:
Good catch — addressed in df2bfcea8.

You're right that gating solely on the path-based `isWorktreeRegistered` probe was too loose: a genuine `worktree add` failure that happened to leave the path registered (e.g. a stale/pre-existing worktree at the same path on a `fatal: … already exists`, exit 128) would have been swallowed, where the old keyword gate would have rethrown.

Re-narrowed the tolerance so we only forgive a non-zero exit when **both**:
1. the failure plausibly came from the post-checkout step — a recognisable hook failure (husky/`command not found`/etc.) **or** a signal-terminated process (exit > 128, e.g. SIGPIPE → 141), and
2. the intended outcome is verified (`didSucceed`).

Genuine git usage errors (exit 128/1, no hook keywords) now always rethrow, even if the success probe would pass. The exit-code signal check (>128) is what lets us catch the original SIGPIPE/141 case without resorting to the brittle stderr-keyword matching that missed it in the first place. Added a regression test (`re-throws genuine git failures even when the success probe would pass`).

AviPeltz:
Fixed in b69876a90 — you're right.

`code > 128` was too broad: git's own usage errors exit **129** (numerically 128 + SIGHUP, but really a usage failure), and user interrupts exit 130 (SIGINT) / 143 (SIGTERM) — none of which should be silently tolerated.

Narrowed the signal branch to **SIGPIPE specifically**: `signal === 'SIGPIPE' || code === 141`. That's the exact, documented #4350 case (a post-checkout hook pipeline dying under `set -o pipefail`), and git never uses 141 as a meaningful status. Genuine git errors (128/129/1) now always rethrow. Added an exit-129 regression test (`re-throws git usage errors (exit 129) instead of misreading them as a signal failure`).

PR #5013, opened 2026-05-31
fix(relay): disable Nagle on tunnel sockets to cut interactive terminal latency

### LINKED ISSUE #5012: [bug] Remote terminal typing lags ~1–3s per keystroke (relay sockets don't set TCP_NODELAY)
#### Describe the bug

Typing in a **remote** terminal lags roughly **1–3 seconds per keystroke** to echo.
Local/same-machine terminals are instant. It's specifically *interactive* latency
(each small keystroke), not a bandwidth/throughput problem — which points at how
small, sparse packets are handled on the relay path.

> Note: this is **not** about heavy-output rendering performance (large bursts can
> also feel janky for an unrelated, client-side reason). This issue is specifically
> the per-keystroke interactive latency on remote terminals.

## Environment

- Superset desktop 1.12.x, macOS client
- Connecting to a **remote** host (via the relay), not a local/same-machine host

## Steps to reproduce

1. Connect a remote host and open a terminal on it in the desktop app.
2. Type a few characters — each takes ~1–3s to appear.
3. For comparison, open a terminal on a **local** host and type — it's instant.
   The only difference is the relay hop.

## Root cause

Path: desktop client → relay (Fly) → host-service tunnel → pty-daemon → PTY.

Isolating each hop:

- **host-service isn't the cause** — a loopback packet capture shows it echoes a
  keystroke in **~3 ms**.
- **The network isn't the cause** — host↔relay ~9–19 ms, client↔relay ~6 ms.
- **It's the relay path** — a *local* Superset terminal (no relay) types at **~11 ms**;
  the *remote* one is **~1.2 s**. The only difference is the relay hop.
- The relay forwards frames immediately but never calls `setNoDelay` on its sockets
  (`apps/relay/src/index.ts`), so **Nagle's algorithm**, interacting with TCP
  delayed-ACK, is active. Nagle holds **small, partial** segments — a keystroke and
  its echo — waiting for the previous ACK; it does **not** hold full-size segments.
  So the delay is specific to interactive traffic, and it compounds across the
  relay's hops.

Controlled A/B to confirm the mechanism (minimal TCP echo server, real network,
single Node hop; the client uses `setNoDelay` so only the server's behaviour
varies):

| echo round-trip | small sparse frames (typing-like) | back-to-back frames |
|---|---|---|
| **Nagle ON** | med 62.7 ms · avg 101 ms · max 205 ms | ~24 ms |
| **Nagle OFF** (`setNoDelay`) | med 24.3 ms · avg 27 ms · max 59 ms | ~24 ms |

With Nagle on, sparse single frames take a ~40–180 ms hit; `setNoDelay` collapses
them to the ~24 ms network floor. Back-to-back frames are at the floor either way —
i.e. Nagle penalises exactly the sparse interactive pattern.

Re-validated against the relay's own code over a real network (toggling `setNoDelay`
via a build flag): sparse interactive frames drop from **avg ~99 ms / max ~640 ms** to
a tight **~50 ms**, and routing an actual remote terminal through the patched relay
types visibly snappier.

## Related (distinct cause)

Possibly related but a different root cause, so filing separately:
- #3061 — *first* keystroke takes 15–20s after opening a terminal (cold-start /
  daemon bootstrap), whereas this is steady-state on *every* keystroke.
- #2861 — *periodic/rhythmic* stutter attributed to a main-thread polling tick;
  this is a constant per-keystroke transport delay, remote-only.

## Proposed fix

Set `TCP_NODELAY` on the relay's incoming sockets. Both the client terminal WebSocket
and the host tunnel WebSocket connect to the relay, so a single
`server.on("connection", s => s.setNoDelay(true))` covers the relay's writes in both
directions. Happy to open a PR — I have the change and the measurements above.

philbirtles:
## Root cause

Keystrokes in a **remote** terminal lag ~1–3 s to echo; local terminals are instant.
The latency is the relay path: host-service echoes a keystroke in ~3 ms (loopback
capture), host↔relay is ~9–19 ms, and a *local* Superset terminal types at ~11 ms vs
~1.2 s remote.

The relay forwards every WebSocket frame immediately, but it never sets `TCP_NODELAY`
on its sockets. Nagle's algorithm, interacting with TCP delayed-ACK, holds **small,
partial** segments — a keystroke and its echo — waiting for the previous ACK; it does
not hold full-size segments. So the penalty is specific to small, sparse interactive
frames, and it compounds across the relay's hops into seconds of typing lag.

> Scope: this is the *interactive keystroke latency*. Heavy-output rendering
> performance is a separate, client-side concern and is not addressed here.

## The fix

Set `TCP_NODELAY` on every incoming connection. Both the client's terminal WebSocket
and the host's tunnel WebSocket connect to this server, so one handler covers the
relay's writes in both directions.

```ts
injectWebSocket(server);

// Disable Nagle's algorithm on every incoming connection. Both the client's
// terminal WebSocket and the host's tunnel WebSocket connect here, so this
// covers the relay's writes in both directions. Nagle interacting with TCP
// delayed-ACK adds tens-to-hundreds of milliseconds to small, sparse
// interactive frames (terminal keystrokes and their echoes) while leaving
// bulk output untouched; across the relay's multiple hops this compounds into
// seconds of perceived typing lag. Interactive proxies should always set
// TCP_NODELAY. (@hono/node-server returns a Node http.Server.)
(server as unknown as import("node:http").Server).on(
  "connection",
  (socket: import("node:net").Socket) => {
    socket.setNoDelay(true);
  },
);
```

## Measured

**1. Minimal TCP echo, single hop (isolates the mechanism).** Client measures frame
round-trips over a real network; the client uses `setNoDelay`, so only the server's
behaviour varies:

| echo round-trip | small sparse frames (typing-like) | back-to-back frames |
|---|---|---|
| **Nagle ON** (current) | med 62.7 ms · avg 101 ms · max 205 ms | ~24 ms |
| **Nagle OFF** (this change) | med 24.3 ms · avg 27 ms · max 59 ms | ~24 ms |

**2. Through the relay's own code, real network, `setNoDelay` toggled via a build
flag.** Sparse interactive frames round-tripped client → relay → host → relay → client:

| sparse interactive frames | avg | max |
|---|---|---|
| **Nagle ON** | 98.7 ms | 641 ms |
| **setNoDelay** (this change) | 49.6 ms | 50.4 ms |

With Nagle on, interactive frames intermittently stall into the hundreds of
milliseconds; `setNoDelay` makes them tight and consistent. (The ~50 ms floor here is
the multi-hop test rig's base round-trip — the change removes the *added* stalls, not
the base network time.) Back-to-back / bulk frames are unaffected either way — Nagle
only penalises the sparse interactive pattern, which is exactly interactive typing.

Routing an actual remote terminal through a relay built with this change also produced
visibly snappier typing than the unchanged relay.

## Test plan

Ran locally against this branch (`apps/relay`):
- [x] `tsc --noEmit` (relay typecheck) — clean
- [x] Biome `check` (lint + format) on the change — clean
- [x] `bun build` of the relay entry — bundles cleanly (845 modules)

For a reviewer / CI to confirm:
- [ ] Remote terminal: keystrokes echo without the ~1 s lag
- [ ] No behaviour change for large/continuous output (this only affects how small,
      sparse segments are flushed — it doesn't alter framing or throughput)
- [ ] `server.on("connection")` fires for both HTTP and WebSocket connections under
      Node (the relay's runtime) and applies `setNoDelay`

## Scope

This is the complete fix for the interactive-latency issue — the relay's two TCP
sockets were the only ones on the path with Nagle enabled. Every other endpoint
already disables it:

- **client → relay:** the desktop uses a Chromium `WebSocket`, which sets
  `TCP_NODELAY` by default.
- **host → relay:** the host-service tunnel connects via Node's global `WebSocket`
  (undici), which calls `socket.setNoDelay(true)` on its socket (verified under Node).
- **tunnel ↔ host-service:** loopback, so Nagle has no effect.

Closes #5012

### DIFF
--- a/apps/relay/src/index.ts
+++ b/apps/relay/src/index.ts
@@ -390,3 +390,18 @@ server = serve({ fetch: app.fetch, port: env.RELAY_PORT }, (info) => {
 	);
 });
 injectWebSocket(server);
+
+// Disable Nagle's algorithm on every incoming connection. Both the client's
+// terminal WebSocket and the host's tunnel WebSocket connect here, so this
+// covers the relay's writes in both directions. Nagle interacting with TCP
+// delayed-ACK adds tens-to-hundreds of milliseconds to small, sparse
+// interactive frames (terminal keystrokes and their echoes) while leaving
+// bulk output untouched; across the relay's multiple hops this compounds into
+// seconds of perceived typing lag. Interactive proxies should always set
+// TCP_NODELAY. (@hono/node-server returns a Node http.Server.)
+(server as unknown as import("node:http").Server).on(
+	"connection",
+	(socket: import("node:net").Socket) => {
+		socket.setNoDelay(true);
+	},
+);


ghost:

| | Title |
|---|-------|
| 1 | [Disable Nagle's algorithm on relay sockets](https://stagereview.app/superset-sh/superset/pull/5013/chapters/1) |

<a href="https://stagereview.app/superset-sh/superset/pull/5013">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 64f8dd27d9dda9e871dd2c66063a2f8aeb720d5f on May 31, 2026 2:42pm UTC.</sub>

saddlepaddle:
Ty for the PR! It looks like this is a general net improvement :)

Issue #4841, opened 2026-05-22
[feat] fetch and pull button on Superset v2


geraldooi-pantas:
#### Describe the feature

I need `fetch and pull` button on Superset v2 (same like what we see on v1). On v2, I can only `Refresh changes` (at top right corner), will it can't pull in the new commits push to the PR I am watching on.

Domen-Zukovec:
Hi, are there any updates on this issue?

Superset is amazing, but the one thing that currently breaks my workflow is having to open every PR/branch I work on as a worktree in VS Code just to fetch and pull the latest changes before I start working on it or reviewing the code. I then have to open it again in VS Code when I want to make and push changes.

I know I can use `gh` commands in the terminal, or ask the agent to commit, push, and pull for me, but having these basic Git operations available directly in the UI would make the workflow much smoother for me.

What I particularly like about VS Code is that it automatically fetches and shows when there are incoming changes, with a clear option to pull them. Something similar in Superset would be really useful.

Thanks for considering it!

Issue #3281, opened 2026-04-09
[bug] new terminal starts in half width


OrcunJoyoLabs:
When I start a new terminal, it starts at half width - see the screenshot below:

<img width="961" height="352" alt="Image" src="https://github.com/user-attachments/assets/bc6418c8-e2e1-43b8-8ba6-2232fe822202" />

If i switch to another terminal window and come back, it resizes to full width as expected

Kitenite:
Closing as a duplicate of #5605, the consolidated tracker for stale `terminal-rendering` issues from the July 2026 stale-issue sweep (no activity since 2026-06-01). If this still applies on a current build, comment on #5605 and we'll reopen.

PR #5022, opened 2026-06-01
fix(desktop): refit terminal after single-shot window grow (#5021)

### LINKED ISSUE #5021: [bug] Terminal stays narrow after window grows back (shrink → grow leaves text at half width)
## Describe the bug

If you resize the Superset window to roughly half the screen, the terminal text correctly reflows to the narrower width. But when you grow the window back to full screen / maximize, the terminal text **stays at the old narrow width** — only the left half of the terminal is used. Switching tabs away and back, toggling fullscreen, or changing the font fixes it (forces a re-fit).

This is the inverse-recovery cousin of #3281, and the same root-cause family as #3468 (TUI renders at 1/4 until a manual resize).

## Steps to reproduce

1. Open a terminal in a workspace.
2. Drag/resize the window to ~half the screen — text reflows correctly to the narrower width.
3. Maximize / go fullscreen in a single motion (e.g. macOS green button).
4. **Observed:** text remains at the half width; the right half of the terminal is empty.
5. Switch to another tab and back → it snaps to full width.

## Root cause

The live terminal uses the v1 cache path. Its `ResizeObserver` runs a single **synchronous** `fitAddon.fit()` per event with no next-frame correction:

`apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/v1-terminal-cache.ts` (`attachToContainer`)

`FitAddon.fit()` computes `cols = floor(parentWidth / cellWidth)` from the parent's *computed* width at call time.

- **Shrinking** is a continuous drag → many ResizeObserver events; the last one lands after layout settles, so `cols` ends up correct.
- **Maximize / fullscreen** is effectively a single large grow event. The one synchronous `fit()` reads intermediate/stale geometry (layout not yet flushed, WebGL renderer `css.cell` metrics not yet refreshed for the new frame) and latches a too-small `cols`. Because the observed container is already at its final full width, **no further ResizeObserver event fires** to correct it → stuck narrow until a forced re-layout.

The deferred-fit safety net (`scheduleFontSettleRefit`) exists only on the **attach** path, and PR #3469's rAF retry was for the v2 runtime — the v1 `ResizeObserver` path has no equivalent deferral, which is the gap.

Related upstream FitAddon issues: xtermjs/xterm.js#5320, #4841, #3584, #3564.

## Proposed fix

Defer + double-fit in the ResizeObserver (fit immediately for snappy shrink, then re-fit on the next animation frame to catch single-shot grows), cancelling the pending rAF on detach/dispose.

## Environment

- App: Superset desktop (Electron renderer)
- xterm: `6.1.0-beta.220`, `@xterm/addon-fit` `0.12.0-beta.220`, `@xterm/addon-webgl` `0.20.0-beta.219`

blelia:
## Summary

Fixes #5021 — after resizing the Superset window to ~half the screen and then growing it back to full / fullscreen, the terminal text stays stuck at the old narrow width (only the left half is used) until a forced re-layout (tab switch, fullscreen toggle, font change).

## Root cause

The live terminal uses the v1 cache path. Its `ResizeObserver` ran a single **synchronous** `fitAddon.fit()` per event:

```ts
const observer = new ResizeObserver(() => {
  if (fitAndRefresh(entry)) onResize?.();
});
```

`FitAddon.fit()` computes `cols = floor(parentWidth / cellWidth)` from the parent's *computed* width at call time.

- **Shrinking** the window is a continuous drag → many ResizeObserver events; the last lands after layout settles, so `cols` ends up correct.
- **Maximize / fullscreen** is effectively a single large grow event. The one synchronous `fit()` reads intermediate/stale geometry (layout not yet flushed, WebGL renderer `css.cell` metrics not refreshed for the new frame) and latches a too-small `cols`. Because the observed container is already at its final full width, **no further ResizeObserver event fires** to correct it → stuck narrow.

The deferred-fit safety net (`scheduleFontSettleRefit`) existed only on the **attach** path, and PR #3469's rAF retry was for the v2 runtime — the v1 `ResizeObserver` path had no equivalent deferral. This is also why a tab switch / fullscreen toggle 'fixes' it (those re-run the deferred fit), matching the recovery behavior in #3281.

Related upstream FitAddon issues: xtermjs/xterm.js#5320, #4841, #3584, #3564.

## Fix

In the v1 cache `ResizeObserver`, fit immediately (keeps shrink snappy) **and** re-fit on the next animation frame so single-shot grows re-read the settled geometry. `onResize` (backend PTY resize) fires once after the deferred fit if either fit changed dims. The pending rAF is cancelled in `detachFromContainer` and `dispose`.

## Test plan

- [x] Added co-located regression test `v1-terminal-cache.test.ts` (minimal model of the deferred-refit handler): a single-shot grow whose synchronous read is stale ends up at full width after the frame settles; backend is notified when only the deferred fit changes dims; no notification when nothing changes. `bun test` → 3 pass.
- [x] `biome check` clean on changed files.
- [ ] Manual: shrink the window to half, then maximize / go fullscreen — terminal fills the full width without needing a tab switch.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5022">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes #5021. Ensures the desktop terminal refits after a single-shot window grow (maximize/fullscreen) so it uses the full width, while keeping shrink responsive. Also preserves and flushes pending PTY resize notifications across rAF cancellations and detach/reattach.

- **Bug Fixes**
  - `ResizeObserver` now fits immediately and again on the next animation frame; `onResize` fires once after the deferred fit if any pass changed size, and the change signal survives rAF cancellations and reattach.
  - Cancel any pending `requestAnimationFrame` on attach, detach, and dispose to prevent leaks and duplicate refits.
  - Expanded regression tests: model rAF ids/cancellation, add a back-to-back resize test for a single deferred refit/notification, a test to ensure a size change isn’t dropped when a later event cancels the rAF, and a detach/reattach test that flushes a pending resize after a cancelled rAF.

<sup>Written for commit 65320281fd6f574e5da02da307374f93d1c2378f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5022?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminal resizing now reliably updates after rapid or deferred layout changes; canceled or stale pending refits no longer cause incorrect sizing or stale updates.
  * Size-change notifications are emitted only when visible terminal dimensions actually change.
  * Pending deferred resizes are canceled on detach/dispose, and preserved pending signals trigger a single notification on reattach.

* **Tests**
  * Added tests covering synchronous vs. deferred resize flows, rapid successive resizes, cancellation behavior, reattach flushes, and notification conditions.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/v1-terminal-cache.test.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/v1-terminal-cache.test.ts
@@ -1,0 +1,123 @@
+/**
+ * Reproduction tests for issue #5021:
+ * "Terminal stays narrow after the window grows back."
+ *
+ * Resize the Superset window to ~half the screen and the terminal reflows
+ * correctly to the narrower width. Grow it back to full / maximize and the
+ * terminal stays at the old narrow width — only the left half is used.
+ *
+ * Root cause: the v1 cache ResizeObserver ran a single SYNCHRONOUS
+ * `fitAddon.fit()` per event. `fit()` computes `cols = floor(parentWidth /
+ * cellWidth)` from the parent's computed width at call time. Shrinking is a
+ * continuous drag (many events; the last lands after layout settles), but a
+ * maximize / fullscreen is effectively one large grow event. The single
+ * synchronous fit reads intermediate/stale geometry (layout not flushed, WebGL
+ * cell metrics not refreshed) and latches a too-small `cols`. Because the
+ * observed container is already at its final width, no further event fires to
+ * correct it.
+ *
+ * Fix: fit immediately (snappy shrink) AND re-fit on the next animation frame
+ * so a single-shot grow re-reads the settled geometry.
+ */
+import { describe, expect, it } from "bun:test";
+
+// ---------------------------------------------------------------------------
+// Minimal model of the deferred-refit ResizeObserver handler in
+// v1-terminal-cache.ts `attachToContainer`. The "terminal" exposes a stale
+// geometry on the synchronous read and the settled geometry on the next frame,
+// mirroring how layout/cell metrics lag a single-shot grow event.
+// ---------------------------------------------------------------------------
+
+const CELL_WIDTH = 10;
+
+function makeHarness(opts: {
+	/** Width visible to the synchronous fit (intermediate, lags on grow). */
+	staleWidth: number;
+	/** Width visible once the next frame settles. */
+	settledWidth: number;
+}) {
+	const state = {
+		cols: opts.staleWidth / CELL_WIDTH,
+		resizeRafId: null as number | null,
+		// Switches from stale -> settled when the rAF callback runs.
+		settled: false,
+		onResizeCalls: 0,
+	};
+
+	const pendingRafs: Array<() => void> = [];
+	const requestAnimationFrame = (cb: () => void): number => {
+		pendingRafs.push(cb);
+		return pendingRafs.length;
+	};
+	const cancelAnimationFrame = (_id: number) => {};
+
+	const fitAndRefresh = (): boolean => {
+		const width = state.settled ? opts.settledWidth : opts.staleWidth;
+		const nextCols = width / CELL_WIDTH;
+		const changed = nextCols !== state.cols;
+		state.cols = nextCols;
+		return changed;
+	};
+
+	// Mirrors the production ResizeObserver callback.
+	const onResizeEvent = () => {
+		const changedNow = fitAndRefresh();
+		if (state.resizeRafId !== null) cancelAnimationFrame(state.resizeRafId);
+		state.resizeRafId = requestAnimationFrame(() => {
+			state.resizeRafId = null;
+			state.settled = true;
+			const changedLater = fitAndRefresh();
+			if (changedNow || changedLater) state.onResizeCalls++;
+		});
+	};
+
+	const flushFrame = () => {
+		const cbs = pendingRafs.splice(0);
+		for (const cb of cbs) cb();
+	};
+
+	return { state, onResizeEvent, flushFrame };
+}
+
+describe("v1-terminal-cache deferred resize refit (#5021)", () => {
+	it("re-fits to full width after a single-shot grow whose sync read is stale", () => {
+		// Window grows from half (600px) to full (1200px), but the synchronous
+		// fit only sees the half width — the settled full width is visible next
+		// frame.
+		const h = makeHarness({ staleWidth: 600, settledWidth: 1200 });
+
+		h.onResizeEvent();
+		// Synchronous fit alone would leave the terminal stuck at 60 cols.
+		expect(h.state.cols).toBe(60);
+
+		h.flushFrame();
+		// Deferred refit reads the settled width → full 120 cols.
+		expect(h.state.cols).toBe(120);
+		expect(h.state.onResizeCalls).toBe(1);
+		expect(h.state.resizeRafId).toBeNull();
+	});
+
+	it("notifies the backend even when only the deferred fit changes dims", () => {
+		// Sync read sees no change (still half), settled read grows.
+		const h = makeHarness({ staleWidth: 600, settledWidth: 1200 });
+		h.state.cols = 60; // already at the stale width
+
+		h.onResizeEvent();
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Update CachedTerminal interface for deferred resizing](https://stagereview.app/superset-sh/superset/pull/5022/chapters/1) |
| 2 | [Implement deferred refit in ResizeObserver](https://stagereview.app/superset-sh/superset/pull/5022/chapters/2) |
| 3 | [Manage resize lifecycle and PTY reconciliation](https://stagereview.app/superset-sh/superset/pull/5022/chapters/3) |
| 4 | [Add regression tests for terminal refitting](https://stagereview.app/superset-sh/superset/pull/5022/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/5022">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 65320281fd6f574e5da02da307374f93d1c2378f on Jun 1, 2026 9:00am UTC.</sub>

blelia:
Thanks for the review. Addressed both points in 075c15be1:

- **Test rAF mock (greptile P2):** the harness now models `requestAnimationFrame`/`cancelAnimationFrame` faithfully — stable ids and `cancel` removes the pending callback — so back-to-back events reproduce production's last-rAF-wins behaviour instead of queueing every callback. Added a test that fires two rapid resize events and asserts a single deferred refit + single `onResize` notification, which now genuinely exercises the cancellation invariant.
- **Deferred `onResize` (greptile P2):** added a comment explaining why the backend PTY resize is deferred to the rAF (the deferred fit is authoritative; `changedNow` is carried in so a sync-only grow isn't dropped) and noting the ~1-frame delay on continuous drags is the intended trade-off.

`bun test` → 4 pass; `biome check` clean on both files.

blelia:
@cubic-dev-ai good catch — this was a real P1, fixed in 5626b2302.

**Root cause:** the deferred-refit `changedNow` signal lived in the per-event ResizeObserver closure. When a later event cancelled the prior event's pending rAF, that event's dimension change was discarded. If the later event's synchronous fit and the deferred fit then saw no further change, `onResize` never fired and the backend PTY stayed out of sync with xterm.

**Fix:** hoisted the changed flag (`pendingResizeChange`) to the observer's outer scope so it survives rAF cancellation, clearing it only once a notification is actually delivered.

Added a regression test for the exact scenario (event 1 changes dims → event 2 cancels its rAF → deferred fit unchanged → must still notify). I verified it **fails against the previous per-closure logic** (`onResizeCalls` 0 vs expected 1) and passes now. `bun test` → 5 pass; `biome check` clean.

blelia:
@coderabbitai good catch on the outside-diff finding — valid Major, fixed in 65320281f.

You're right that hoisting `pendingResizeChange` only to the `attachToContainer` closure still dropped it across a detach/reattach: a synchronous fit changes xterm's size while the deferred `onResize` is pending, a tab switch detaches and cancels that rAF, and on reattach the unchanged container makes the fit a no-op — so the PTY stayed out of sync.

**Fix:** moved `pendingResizeChange` onto `CachedTerminal` so it survives detach; `detachFromContainer` deliberately preserves it; `attachToContainer` flushes it on reattach (`if (entry.pendingResizeChange) onResize?.()`) to reconcile the backend.

Added a detach/reattach regression test (sync fit changes dims → detach cancels rAF → reattach with unchanged container must still notify). Verified it **fails against the attach-local version** (`onResizeCalls` 0 vs 1) and passes now. `bun test` → 6 pass; `biome check` clean.

Kitenite:
Closing: this only touches the legacy v1 UI (screens/main), which is being retired. both files are v1 v1-terminal-cache.ts. If the bug still shows in the v2 workspace, please open an issue.

Kitenite:
Thanks for chasing down the single-shot window-grow refit, and for the repro in #5021. Same reason: the v1 terminal cache this patches is being retired.

PR #5129, opened 2026-06-05
fix(desktop): stop idle tanstack-db.sqlite writes that churn disk and bloat the file

saddlepaddle:
## Problem

Users reported laptops running hot with the app open. Root cause: the desktop app's TanStack/Electric SQLite persistence (`~/.superset/tanstack-db.sqlite`) **commits a transaction on every idle long-poll cycle, per collection** — resume-metadata only, no data change — ~once/sec across ~20 collections. That constant disk churn keeps endpoint security / file-integrity agents (EDR, DLP, antivirus) re-hashing the file. Freed pages were never reclaimed (`auto_vacuum=NONE`, no VACUUM), so the file bloated to **462 MB** (68% of which was the `applied_tx` replay log, ~88% free pages).

A scanner's CPU cost ≈ *write frequency × file size* — and both were maxed: a multi-hundred-MB file modified every second.

## Fix (all in `apps/desktop/.../persistence/persistence.ts` — no node_modules patching)

1. **Suppress idle writes** — wrap the persistence adapter's `applyCommittedTx` (the single main-process chokepoint all collections funnel through) and drop resume-only transactions whose Electric `offset`/`handle` are unchanged. Idle → ~0 writes; real data still persists normally.
2. **Reclaim bloat** — `wal_checkpoint(TRUNCATE)` + gated `VACUUM` on launch when the freelist exceeds 64 MB, so already-bloated installs shrink on next update.

Plus a dependency fix that was blocking the dev/auth build (`fix(deps)` commit): `@better-auth/kysely-adapter@1.6.13` imports `DEFAULT_MIGRATION_LOCK_TABLE`/`DEFAULT_MIGRATION_TABLE` from kysely's root, which 0.29.x dropped — pinned kysely to `0.28.17` (within the adapter's supported range, still exports them).

## Verification

| | Before | After |
|---|---|---|
| Idle `applied_tx` writes / 60s | ~29 (WAL constantly rewritten) | **0 (WAL static)** |
| File size (real install) | 462 MB | **51 MB** after VACUUM, `integrity_check` ok |

- Suppression measured on the **real Electron build** with live Electric sync, idle — 0 writes/min, WAL unchanged, real data synced first (confirming only redundant idle writes are dropped).
- VACUUM run against a backup of a real 462 MB DB: 462 → 51 MB, data intact.
- `lint` 0, `typecheck` 28/28, `sherif` clean.

## Follow-up (separate, not in this PR)

The 51 MB is ~68% replay-log overhead + ~14 MB user data (GitHub PRs > tasks, across ~5 orgs). The local-first model syncs the **full org shape to every client**, which won't scale to large orgs. Tracking separately: filtered/windowed Electric shapes + payload trimming.

## Summary by CodeRabbit

* **Performance Improvements**
  * Automatic SQLite maintenance runs on startup to reclaim space and reduce database bloat.

* **Bug Fixes**
  * Redundant resume commits are detected and suppressed to avoid unnecessary/no-op writes; failures during maintenance are logged as warnings.

* **Chores**
  * Added a package override to pin a database library version.

### DIFF
--- a/apps/desktop/src/main/lib/persistence/persistence.ts
+++ b/apps/desktop/src/main/lib/persistence/persistence.ts
@@ -41,7 +41,7 @@ function reclaimBloatedDatabaseFile(target: Database.Database): void {
 }
 
 function stripVolatileResumeFields(value: unknown): unknown {
-	if (value === null || typeof value !== "object") {
+	if (value === null || typeof value !== "object" || Array.isArray(value)) {
 		return value;
 	}
 	const { updatedAt, ...stableFields } = value as Record<string, unknown>;
@@ -98,16 +98,17 @@ function suppressIdleResumeWrites(
 					transaction: CommittedTransaction,
 				): Promise<void> => {
 					const resumeSignature = getRedundantResumeSignature(transaction);
-					if (resumeSignature !== null) {
-						if (
-							lastForwardedResumeByCollection.get(collectionId) ===
+					if (
+						resumeSignature !== null &&
+						lastForwardedResumeByCollection.get(collectionId) ===
 							resumeSignature
-						) {
-							return;
-						}
-						lastForwardedResumeByCollection.set(collectionId, resumeSignature);
+					) {
+						return;
 					}
 					await target.applyCommittedTx(collectionId, transaction);
+					if (resumeSignature !== null) {
+						lastForwardedResumeByCollection.set(collectionId, resumeSignature);
+					}
 				};
 			},
 		});
--- a/apps/desktop/src/main/lib/persistence/persistence.ts
+++ b/apps/desktop/src/main/lib/persistence/persistence.ts
@@ -3,23 +3,151 @@ import { exposeElectronSQLitePersistence } from "@tanstack/electron-db-sqlite-pe
 import { createNodeSQLitePersistence } from "@tanstack/node-db-sqlite-persistence";
 import Database from "better-sqlite3";
 import { ipcMain } from "electron";
+import log from "electron-log/main";
 import {
 	ensureSupersetHomeDirExists,
 	SUPERSET_HOME_DIR,
 } from "../app-environment";
 
+type SQLitePersistence = ReturnType<typeof createNodeSQLitePersistence>;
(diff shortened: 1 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Pin kysely version to fix build](https://stagereview.app/superset-sh/superset/pull/5129/chapters/1) |
| 2 | [Optimize SQLite persistence and reclaim disk space](https://stagereview.app/superset-sh/superset/pull/5129/chapters/2) |
| 3 | [Other changes](https://stagereview.app/superset-sh/superset/pull/5129/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5129">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 43cd37345d4c99df6638ba8ba328374931319ba1 on Jun 5, 2026 2:13am UTC.</sub>

saddlepaddle:
Thanks for the reviews — addressed in 43cd37345:

**Dedupe cache updated before the write (CodeRabbit / cubic, P1)** — valid, fixed. The cache now advances only *after* `applyCommittedTx` resolves, so a failed write no longer suppresses the retry of the same checkpoint. Early-return on duplicate is unchanged.

**`stripVolatileResumeFields` on non-object values (Greptile, P2)** — added an `Array.isArray` guard so a non-plain-object resume payload is passed through verbatim rather than spread into an index-keyed object.

**Synchronous VACUUM blocks the main process (Greptile, P1)** — keeping as-is; the premise doesn't hold here:
- It runs in `initTanstackDbPersistence()` at `index.ts:395`, **before the main window is created** at `index.ts:431` (`makeAppSetup(() => MainWindow())`) — there's no window to freeze.
- Measured **~0.22s** on a real 462 MB file (3 runs, `better-sqlite3` path). VACUUM rewrites only *live* pages (~51 MB here), so its cost scales with real data, not total file size — even a multi-GB bloated file stays sub-second.
- It's gated on a >64 MB freelist and runs before the IPC handler serves any renderer writes, so there's no lock contention. Deferring it past `exposeElectronSQLitePersistence` would instead race the renderer's writes.

PR #4801, opened 2026-05-21
[codex] Fix worktree delete cleanup ordering

Kitenite:
## What changed

- Routes the legacy `workspace.delete` endpoint through the v2 `workspaceCleanup.destroy` saga so desktop, CLI, SDK, and MCP deletes share one cleanup path.
- Reorders workspace cleanup so local terminal/worktree cleanup happens before cloud deletion; cloud delete now happens only after the worktree is gone or already missing.
- Keeps post-cloud host sqlite cleanup best-effort, surfaces cleanup warnings to CLI/SDK/MCP callers, and adds regression coverage for failed worktree removal.

## Why

Some delete callers were using the legacy `workspace.delete` endpoint, which deleted the cloud row first, attempted plain worktree removal, swallowed removal errors, and then deleted the local row. That could make a workspace disappear from the app while leaving the worktree on disk.

The desktop app should not remove the authoritative cloud row until local worktree removal succeeds. If local removal fails while the path still exists, the delete now fails before cloud delete so the workspace can reappear and be retried.

## Validation

- `bun test packages/host-service/test/workspace-cleanup.test.ts packages/host-service/test/integration/workspace-create-delete.integration.test.ts packages/host-service/test/integration/bug-hunt-2.integration.test.ts packages/host-service/test/integration/bug-hunt-3.integration.test.ts`
- `bun run lint:fix`
- `bun run lint`
- `bun run --cwd packages/host-service typecheck`
- `bun run --cwd packages/cli typecheck`
- `bun run --cwd packages/mcp-v2 typecheck`
- `bun run --cwd packages/sdk typecheck`
- `git diff --check`

---

<a href="https://stagereview.app/superset-sh/superset/pull/4801">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes workspace deletion by removing local worktrees before cloud delete and routing all delete entry points through the v2 cleanup path. Calls now return structured results and surface warnings to prevent “vanishing” workspaces and orphaned disk state.

- **Bug Fixes**
  - Routed legacy `workspace.delete` through exported `destroyWorkspace` with an in‑flight guard; v1 callers use `force` and return `success`, `cloudDeleted`, `worktreeRemoved`, `branchDeleted`, and `warnings`.
[...]

PR #4693, opened 2026-05-18
fix(host-service): tolerate locked+missing worktrees in destroy

Kitenite:
## Summary

- Customers can end up with a workspace whose worktree directory is gone from disk. If git also has the worktree locked (rare but possible — a long-running operation or tool that locks during work, then crashes), `git worktree remove --force` fails with `fatal: cannot remove a locked working tree; use 'remove -f -f' to override or unlock first`. The existing substring matcher in [`workspace-cleanup.ts:294-300`](packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts) doesn't catch this phrasing, so the step pushes a confusing "Failed to remove worktree" warning even though the directory the user wanted gone is already gone.
- Escalate the command to `--force --force` (we're past the cloud-delete commit point — the workspace is already gone from the user's perspective, so the conservative single `--force` no longer earns its keep) and add an `existsSync` fallback: any git failure whose post-condition is "the worktree dir is gone" is success, regardless of what git printed. The substring matcher stays as belt-and-braces for the rare race where the dir disappears mid-remove.
- Add an integration test for the locked-and-missing case. Verified it fails on the unfixed code (`worktreeRemoved: false`, warnings non-empty) and passes after the fix.

Closes [SUPER-779](https://linear.app/superset-sh/issue/SUPER-779).

## Test plan

- [x] `bun test packages/host-service/test/integration/workspace-cleanup.integration.test.ts` — 10 pass (9 existing + 1 new).
- [x] `bun run lint` clean.
- [x] `bunx tsc --noEmit` clean.
- [ ] Manual: in desktop v2, create a workspace, `rm -rf` its worktree dir, click delete from the sidebar — workspace disappears with no warning toasts.
- [ ] Manual locked variant: same but also `git worktree lock <path>` first.

## Out of scope

The renderer-side paths (`useDestroyWorkspace.normalizeError`, dialog inspect-failure handling) were verified to already swallow git failures correctly — no changes needed there.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4693">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Makes `host-service` workspace destroy tolerate locked + missing git worktrees so deletes complete without false warnings. Addresses SUPER-779 by escalating remove and treating a missing worktree dir as success.

- **Bug Fixes**
[...]

PR #5138, opened 2026-06-05
fix(host-service): decide workspace-delete cleanup on git state, not error text

saddlepaddle:
## Problem

Users have been hitting workspace-delete failures since #4801's cleanup reorder, in two distinct shapes:

1. **Hard failures** — `Failed to remove worktree at …` thrown errors that abort the whole delete. #4801 converted worktree-removal failure from a best-effort warning into a hard throw and dropped #4693's tolerant substring matcher, leaving only `existsSync` as the success signal. Any worktree state where the directory still exists but `git worktree remove` errors (locked, or — common on macOS — git registered the worktree under a realpath-resolved path like `/private/var/...` while the stored path is `/var/...`) now fails the delete outright.
2. **Scary-but-harmless warnings** — `Failed to delete branch '…' not found` toasts on deletes that *actually succeeded* (cloud row already gone, worktree already removed). The branch ref was simply already absent — which is exactly what `deleteBranch` is trying to achieve.

Both came from deciding cleanup success by inspecting git's error text, which is locale- and version-dependent and brittle by nature (the reason #4801 removed the matcher in the first place).

## Fix

Decide on git's **actual state**, never on error strings:

- **Worktree removal** — attempt `worktree remove --force --force` against the canonical (realpath) path so a symlinked stored path still matches its registration, then consult `git worktree list --porcelain`. Continue if the path is no longer registered; throw only if git still tracks a live worktree we genuinely couldn't remove (preserves #4801's no-orphan-disk guarantee). No blanket `worktree prune` — `--force --force` already unregisters a missing-dir worktree, and a global prune would clobber *other* workspaces' stale metadata (guarded by an existing integration test).
- **Branch delete** — check existence with `git show-ref --verify --quiet` (exit-code based). An absent ref → success, no warning. A ref that exists but fails `-D` → still warns.

Reuses the canonical `listGitWorktrees` / `normalizeWorktreePath` helpers (the latter already handles the `/var`↔`/private/var` realpath canonicalization).

## Tests

- Updated the `makeCtx` unit mock to model the two new git reads (`worktree list`, `show-ref`) and reworked the worktree-failure test to express failure as a *still-registered* worktree rather than a thrown remove.
- `bun test` for the 5 cleanup-touching files: **70/70 pass**. Lint clean, typecheck exit 0.

## Verified against prod

Reproduced the reported case (Roshvan's `Roshvan/eat-147-journey-transparency`): the v2 cloud row was already gone, confirming the delete had committed and only the cosmetic branch warning was left. After this change that path returns `{ success: true, branchDeleted: true, warnings: [] }` — the user just sees the workspace disappear, no warning toast.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5138">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes workspace delete failures by deciding cleanup based on git state instead of parsing error text. Also guards against git read failures masquerading as success.

- **Bug Fixes**
  - Worktree: run `git worktree remove --force --force` on the canonical path, then verify with `git worktree list --porcelain` via `git.raw`. If the list read fails, treat state as unknown and block; continue only if unregistered. Throw if still registered. No global prune.
  - Branch: check existence with `git branch --list`. Missing ref counts as success; warn only on a real git failure or a `-D` failure when the ref exists.
  - Tests: add mocks for `worktree list` and `branch --list`; update failure case to assert on a still-registered worktree.

<sup>Written for commit f16641cef6410b029d0cf0928652c056fa62d0fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5138?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable workspace cleanup: worktree paths are canonicalized and removals are verified against Git's registry before proceeding.
  * Branch deletion now checks for local existence first and treats already-missing branches as successful removals.
  * Cleanup steps now block subsequent cloud deletion until Git confirms worktree removal.

* **Tests**
  * Updated tests to simulate Git worktree listings and branch presence, ensuring failure-handling and ordering are validated.

### DIFF
--- a/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
@@ -13,8 +13,8 @@ import type {
 } from "../../error-types";
 import { protectedProcedure, router } from "../../index";
 import {
-	listGitWorktrees,
 	normalizeWorktreePath,
+	parseWorktreeList,
 } from "../workspace-creation/shared/worktree-list";
 import { isMainWorkspace } from "./is-main-workspace";
 
@@ -319,7 +319,20 @@ async function runDestroy(
 				.raw(["worktree", "remove", "--force", "--force", canonicalPath])
 				.catch(() => {});
 
-			if (await isRegisteredWorktree(git, local.worktreePath)) {
+			// A `worktree list` failure here means the post-remove state is
+			// unknown — treat that like "still registered" and block rather
+			// than risk orphaning disk past the cloud commit point.
+			let stillRegistered = true;
+			try {
+				stillRegistered = await isRegisteredWorktree(git, local.worktreePath);
+			} catch (err) {
+				const message = err instanceof Error ? err.message : String(err);
+				throw new TRPCError({
+					code: "INTERNAL_SERVER_ERROR",
+					message: `Failed to verify worktree removal at ${local.worktreePath}: ${message}`,
+				});
+			}
+			if (stillRegistered) {
 				// git still tracks a live worktree here — removal genuinely
 				// failed. Keep the cloud row so the workspace stays visible and
 				// retryable instead of orphaning disk past the cloud commit point.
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Add git state verification helpers](https://stagereview.app/superset-sh/superset/pull/5138/chapters/1) |
| 2 | [Update workspace and branch cleanup logic](https://stagereview.app/superset-sh/superset/pull/5138/chapters/2) |
| 3 | [Update cleanup tests with state-based mocks](https://stagereview.app/superset-sh/superset/pull/5138/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5138">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit f16641cef6410b029d0cf0928652c056fa62d0fd on Jun 5, 2026 4:44am UTC.</sub>

saddlepaddle:
Addressed in f16641cef. Both helpers were swallowing all git errors and returning the happy value, so a transient git failure read as false success — the exact orphan-disk / silent-skip outcome this PR prevents.

- `isRegisteredWorktree` now calls `git.raw` directly (not the swallowing `listGitWorktrees`), so a failed `worktree list` throws; the worktree step treats that "state unknown" as still-registered and blocks the cloud delete.
- `localBranchExists` uses `git branch --list` (exit 0 whether or not the ref exists), so a thrown error is a real git failure rather than a missing ref. The branch step turns it into a warning instead of a false `branchDeleted = true`. (Note: the suggested `err.exitCode === 1` check isn't reliable here — simple-git doesn't expose a numeric exit code — hence the `branch --list` approach.)

PR #5168, opened 2026-06-06
fix(host-service): make workspace delete reliably tear down terminal sessions

saddlepaddle:
## Problem

A host with automated stale-workspace deletion accumulated ~400 live terminal sessions across already-deleted workspaces. Reported as "delete just removes the git worktree and leaves the sessions running."

This turned out to be a **reliability bug, not a missing feature**. The cascade already exists: CLI `delete_workspace` → host-service `workspace.delete` → `destroyWorkspace`, whose Step 2a calls `disposeSessionsByWorkspaceId` before removing the worktree. The problem is that the cascade could silently fail and orphan the PTY:

1. **`disposeSessionAndWait` marked the sqlite row `disposed` before confirming the daemon actually killed the PTY.** A transient daemon failure left a live process behind a `disposed` row — invisible to every future workspace-scoped cleanup.
2. **The pty-daemon has no workspace mapping** (`SessionInfo` is just `{id,pid,cols,rows,alive}`). The terminal→workspace link lives only in sqlite `origin_workspace_id`, and the `onDelete: "set null"` FK severs it once the workspace row is deleted — so an already-orphaned PTY can't be found by workspace at all.

Worktree removal reads git's durable on-disk registry so it always succeeds; terminal disposal trusted a mutable sqlite table and optimistically marked rows disposed — hence the asymmetry.

## Changes

- **`fix: mark disposed only after confirmed kill`** — failed kills now stay `active` and retryable/reapable instead of being lost behind a `disposed` row.
- **`fix: clear dead terminal rows when destroying a workspace`** — the destroy saga deletes the workspace's confirmed-dead terminal rows (status ≠ `active`), so its session index dies with it instead of lingering as `set null` orphans. Failed kills (`active`) are deliberately kept reachable for the reaper.
- **`feat: reap orphaned daemon PTYs`** — a reaper lists the daemon's live sessions and kills any whose row is `disposed`/`exited`/null-workspace (or row-less for two consecutive passes, so a being-born session — daemon PTY opened before the sqlite row insert — is never killed mid-creation). Runs at startup (drains the pre-existing ~400 on next host-service restart) and every 5 min; interval is `unref()`'d and cleared in `dispose()`.

### Why not `onDelete: cascade`?

An FK action fires at the DB level and can't know whether the PTY was actually killed. `cascade` would delete the row for a still-alive failed-kill PTY, turning it into a *row-less* orphan that's ambiguous with a being-born session (slower two-pass reap). `set null` leaves an unambiguous `active`+null row the reaper catches in one pass. Row lifecycle is therefore application-driven and conditional on a confirmed kill; the FK stays `set null` as a dumb integrity backstop.

## Notes for deploy

- The existing ~400 orphans drain on the **next host-service restart** (boot reaper) or within one 5-min tick after deploy.
- Dev mode kills the daemon on exit, so this matters for production/detached daemons — the affected setup.

## Verification

- `bun run typecheck` ✓
- `bun run lint` exit 0 ✓
- 352 host-service unit tests pass (bun 1.3.11)

---

<a href="https://stagereview.app/superset-sh/superset/pull/5168">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Ensure workspace deletion reliably tears down terminal sessions by fixing disposal semantics and adding a background reaper in `host-service`. The reaper now starts from `serve` after daemon bootstrap, avoids overlapping passes, and drains existing orphans on restart or within 5 minutes.

- **Bug Fixes**
  - Mark a session `disposed` only after the daemon kill succeeds; failed kills stay `active` and retryable.
  - Reaper robustness: start from `serve` to avoid early daemon connections; skip overlapping passes, keep pending state per process, retry failed second‑pass rowless kills; include workspace id in the terminal-row cleanup warning.
  - On workspace destroy, delete that workspace’s non-`active` terminal rows to avoid orphaned records.

- **New Features**
  - Add a terminal reaper that lists live daemon sessions and kills orphans (disposed/exited/null-workspace, or row-less on two passes). Runs on boot and every 5 minutes.

<sup>Written for commit cdb508543f5b3a205e655502cb64383dd1a291f6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5168?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Background reaper now runs automatically to detect and clean orphaned terminal sessions.

* **Bug Fixes**
  * Terminal sessions are marked disposed only when closure succeeds.
  * Workspace destroy now clears non-active terminal session records during local cleanup.

* **Chores**
  * Improved error handling and non-blocking logging for cleanup operations.

### DIFF
--- a/packages/host-service/src/app.ts
+++ b/packages/host-service/src/app.ts
@@ -17,6 +17,7 @@ import type { GitCredentialProvider } from "./runtime/git";
 import { createGitFactory } from "./runtime/git";
 import { runMainWorkspaceSweep } from "./runtime/main-workspace-sweep";
 import { PullRequestRuntimeManager } from "./runtime/pull-requests";
+import { startTerminalReaper } from "./terminal/reaper";
 import { registerWorkspaceTerminalRoute } from "./terminal/terminal";
 import { TerminalAgentStore } from "./terminal-agents";
 import { appRouter } from "./trpc/router";
@@ -162,6 +163,8 @@ export function createApp(options: CreateAppOptions): CreateAppResult {
 		upgradeWebSocket,
 	});
 
+	const stopTerminalReaper = startTerminalReaper(db);
+
 	app.use(
 		"/trpc/*",
 		trpcServer({
@@ -189,6 +192,11 @@ export function createApp(options: CreateAppOptions): CreateAppResult {
 		// Each step is best-effort and isolated: a throw in one cleanup must
 		// not skip the others, otherwise a flaky `.stop()` could leak the
 		// open SQLite handle for the rest of the process lifetime.
+		try {
+			stopTerminalReaper();
+		} catch (err) {
+			console.warn("[host-service] stopTerminalReaper failed:", err);
+		}
 		try {
 			pullRequestRuntime.stop();
 		} catch (err) {
--- a/packages/host-service/src/app.ts
+++ b/packages/host-service/src/app.ts
@@ -17,7 +17,6 @@ import type { GitCredentialProvider } from "./runtime/git";
 import { createGitFactory } from "./runtime/git";
 import { runMainWorkspaceSweep } from "./runtime/main-workspace-sweep";
 import { PullRequestRuntimeManager } from "./runtime/pull-requests";
-import { startTerminalReaper } from "./terminal/reaper";
 import { registerWorkspaceTerminalRoute } from "./terminal/terminal";
 import { TerminalAgentStore } from "./terminal-agents";
 import { appRouter } from "./trpc/router";
@@ -61,6 +60,7 @@ export interface CreateAppResult {
 	app: Hono;
 	injectWebSocket: ReturnType<typeof createNodeWebSocket>["injectWebSocket"];
 	api: ApiClient;
+	db: HostDb;
 	dispose: () => Promise<void>;
 }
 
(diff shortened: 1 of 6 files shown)


saddlepaddle:
Pushed fixes for the CI failure and the review findings:

**CI failure (3 integration tests timing out)** — root cause was the reaper being started inside `createApp`, which eagerly connected the pty-daemon client at construction. Those tests configure a custom daemon socket *after* building the app, so teardown terminals never spawned. Moved the reaper to the `serve.ts` listen callback (alongside `connectRelay`) so it only runs in the real host process and after daemon bootstrap; tests use `createApp` and are no longer affected. Full host-service suite now green locally (687 pass / 0 fail).

**Overlapping passes (CodeRabbit + cubic)** — added an in-flight guard so `setInterval` can't interleave two passes.

**Module-level pending state (greptile)** — moved `rowlessPendingSecondPass` into the `startTerminalReaper` closure; no longer shared across instances.

**Retry cadence doubling (greptile)** — a failed second-pass rowless kill is now re-queued into the pending set so it retries on the next 5-min pass instead of restarting its two-pass clock.

**Warning context (cubic)** — included `workspaceId` in the terminal-row cleanup warning.

PR #5303, opened 2026-06-18
Add setting to hide main workspaces from the sidebar

### LINKED ISSUE #5302: [feat] Option to disable automatic creation of the main workspace
## Problem

Every time a repository is added, Superset auto-creates a `type='main'` workspace (the "local"/main-branch entry) in the left panel. For worktree-only workflows, this entry is useless: the main branch is never touched directly, so it's just noise that has to be manually closed for every project — and closing it doesn't stick. On the next launch the startup main-workspace sweep recreates it, so it reappears.

## Request

Add an option to disable the automatic creation of the `main` workspace, for users who only work via worktrees and want exactly one sidebar entry per task.

Ideal shape:

- A setting (per-project, or a global host-level default like `worktreeBaseDir`/`branchPrefixMode`) such as `autoCreateMainWorkspace` (default `true`, preserving current behavior).
- When disabled:
  - The create-project flow does not create the `main` workspace.
  - The startup main-workspace sweep skips projects that opted out (so it doesn't reappear after restart).

## Why an opt-out rather than just "close workspace"

Closing the main workspace only deletes the local record; the cloud record persists and `runMainWorkspaceSweep` -> `ensureMainWorkspace` recreates it on next boot, so there's currently no durable way to remove it.

## Notes / touch points

This isn't trivial because the main workspace is currently a hard invariant:

- The create-project saga calls `ensureMainWorkspaceStrict` and rolls back the whole saga (including the cloud project) if it fails — `mainWorkspaceId` is part of the saga result/contract, so it'd need to become optional.
- A partial unique index (`v2_workspaces_one_main_per_host`) enforces one main per (project, host) — disabling creation is compatible with it, but the flow assumptions around "a main always exists" need auditing.
- The startup sweep (`runMainWorkspaceSweep`) would need to respect the new setting.

Worktree creation and base-branch resolution appear independent of the main workspace, so the core blocker is the project-creation contract and sweep, not worktrees.

Happy to contribute a PR if the maintainers agree on the preferred shape (per-project setting vs. global default, and how to treat existing projects).

arbll:
## Summary

This PR adds a **"Hide main workspaces"** toggle (General settings) that removes the auto-included main (default-branch) workspace from the sidebar. The main workspace still exists — this only gates its sidebar visibility — so worktree-only users can keep exactly one sidebar entry per task.

Closes #5302.

## Motivation

For worktree-only workflows the per-project main workspace is just noise: the default branch is never touched directly, so the entry has to be closed manually for every project — and closing it doesn't stick (the local record is removed but the startup sweep re-adds it on next launch). This gives a durable, device-local opt-out.

## How it works

Rather than blocking creation of the main workspace (which is a hard invariant tied to project creation, the cloud contract, and the startup sweep), this gates **visibility** at the sidebar layer:

- New device-local setting `hideMainWorkspaces` (default `false`), following the existing `openLinksInApp` electron-trpc settings pattern (constant + schema column + migration + get/set procedures + settings-search registration + toggle in General settings).
- `useDashboardSidebarData` skips the auto-included main workspaces when the setting is on. Mains the user explicitly placed in the sidebar (those with a local-state row, flowing through `sidebarWorkspaces`) are unaffected.

Because the main workspace is untouched in the DB/cloud, nothing about project creation, the `main`-workspace sweep, or worktree creation changes.

## Tradeoffs

- The setting is device-local (stored in the local `settings` table), consistent with other behavior toggles — it does not sync across machines.
- Tagged as a `v2`-only setting since the auto-included main workspace is a v2-sidebar concept.

## Testing

- `tsc --noEmit` passes for `apps/desktop`.
- Biome check passes on changed files.

```mermaid
flowchart TD
    A[v2Workspaces: type=main] --> B{hideMainWorkspaces?}
    B -- false --> C[auto-include in sidebar]
    B -- true --> D[skip auto-include]
    E[Explicitly placed mains<br/>local-state row] --> F[always shown via sidebarWorkspaces]
```

---
## Summary by cubic
Adds a "Hide main workspaces" toggle in Behavior settings to hide auto-included main (default-branch) workspaces from the sidebar. The main workspace still exists; this only hides its auto-added sidebar entry. Closes #5302.

- **New Features**
  - New toggle in Behavior settings to hide auto-included main workspaces from the sidebar.
  - Explicitly added main workspaces remain visible; default is off and stored locally.
  - Introduces `hideMainWorkspaces` with `electron-trpc` get/set, local DB column + migration, and settings search (v2 only).

<sup>Written for commit 1a622bd715c93640c45fe0067b0b8c1c624dbf45. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5303?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a "Hide main workspaces" toggle in Behavior settings. When enabled, main workspaces will no longer appear in the sidebar, allowing for a cleaner workspace view.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/settings/index.ts
+++ b/apps/desktop/src/lib/trpc/routers/settings/index.ts
@@ -45,6 +45,7 @@ import {
 	DEFAULT_CONFIRM_ON_QUIT,
 	DEFAULT_EXPOSE_HOST_SERVICE_VIA_RELAY,
 	DEFAULT_FILE_OPEN_MODE,
+	DEFAULT_HIDE_MAIN_WORKSPACES,
 	DEFAULT_OPEN_LINKS_IN_APP,
 	DEFAULT_SHOW_PRESETS_BAR,
 	DEFAULT_SHOW_RESOURCE_MONITOR,
@@ -974,6 +975,26 @@ export const createSettingsRouter = () => {
 					.onConflictDoUpdate({
 						target: settings.id,
 						set: { openLinksInApp: input.enabled },
+					})
+					.run();
+
+				return { success: true };
+			}),
+
+		getHideMainWorkspaces: publicProcedure.query(() => {
+			const row = getSettings();
+			return row.hideMainWorkspaces ?? DEFAULT_HIDE_MAIN_WORKSPACES;
+		}),
+
+		setHideMainWorkspaces: publicProcedure
+			.input(z.object({ enabled: z.boolean() }))
+			.mutation(({ input }) => {
+				localDb
+					.insert(settings)
+					.values({ id: 1, hideMainWorkspaces: input.enabled })
+					.onConflictDoUpdate({
+						target: settings.id,
+						set: { hideMainWorkspaces: input.enabled },
 					})
 					.run();
 
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/hooks/useDashboardSidebarData/useDashboardSidebarData.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/hooks/useDashboardSidebarData/useDashboardSidebarData.ts
@@ -3,6 +3,7 @@ import { useLiveQuery } from "@tanstack/react-db";
 import { useQueries, useQueryClient } from "@tanstack/react-query";
 import { useCallback, useEffect, useMemo, useRef } from "react";
 import { useRelayUrl } from "renderer/hooks/useRelayUrl";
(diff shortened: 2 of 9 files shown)


arbll:
Switched to a different tool (herdr) as it fit my specific workflow better, closing!

PR #5311, opened 2026-06-20
fix(desktop): forward line/column to external editor when opening terminal file links

### LINKED ISSUE #5309: [bug] Cmd+click on a terminal file link opens the external editor at line 1, dropping line/column
#### Describe the bug

Cmd/Ctrl+clicking a `path:line:column` link in a workspace terminal opens the file in the external editor, but always at the top of the file — the line and column are dropped. With **Terminal file links** set to **File viewer**, the same click jumps to the right line, so the coordinates are detected correctly; they're just discarded on the external-editor path.

#### Steps to reproduce

1. Settings → Terminal → **Terminal file links** → **External editor**
2. Pick a default editor (e.g. Cursor / VS Code) with ⌘O
3. In a workspace terminal, print a `file:line:col` reference — e.g. `grep -rn "import" . | head`
4. Cmd+click one of the `…:NNN` links

**Expected:** the editor opens the file at line NNN.
**Actual:** the editor opens the file at line 1.

#### Cause

`external.openFileInEditor` accepts `line`/`column` and callers do pass them (terminal links forward `link.row`/`link.col`), but the mutation only forwards the path — it calls `openPathInApp(filePath, app)`, and neither `openPathInApp` nor `getAppCommand` take line/column, so the coordinates never reach the editor command. The in-app File viewer uses the same coordinates and works, which is why only the external-editor path is affected.

#### About

Version: `1.12.5` · macOS

HwangRock:
## Description

- `external.openFileInEditor` already accepts `line`/`column` and callers pass them (terminal links forward `link.row`/`link.col`), but the mutation only forwarded the path — it called `openPathInApp(filePath, app)`, and neither `openPathInApp` nor `getAppCommand` took the coordinates, so they were silently dropped and the editor always opened at line 1.
- Thread `line`/`column` through `openPathInApp` → `getAppCommand` and build editor-specific goto args on Linux:
  - VS Code family (`vscode`, `vscode-insiders`, `cursor`): `-g <path>:<line>[:<column>]`
  - Sublime Text / Zed: `<path>:<line>[:<column>]`
  - JetBrains IDEs: `--line <line> [--column <column>] <path>` (including multi-edition candidates)
- All other apps (terminals, Finder, etc.) ignore the coordinates and keep their existing args.
- **Invariant:** when `line`/`column` are both undefined the generated args are byte-for-byte identical to before, so existing behavior and tests are unaffected.

### Known limitation (macOS)

On macOS the editors are launched via `open -a <App>` / `open -b <bundleId>`, which can't reliably deliver a line argument to an already-running app (`--args` isn't applied when the app is already open). So this change enables line/column jump on the Linux CLI paths and intentionally leaves the macOS `open` behavior unchanged (coordinates ignored) rather than adding a flaky best-effort path. Happy to follow up on a reliable macOS approach (e.g. preferring the editor's PATH CLI) if you'd like it in scope.

## Related Issues

Resolves #5309

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

- `bun test apps/desktop/src/lib/trpc/routers/external/` — 126 pass / 0 fail (20 new line/column cases covering all editor groups on Linux + macOS regression that coordinates stay ignored).
- `bunx tsc --noEmit` on `apps/desktop` — clean.
- `biome check` on the changed files — clean.

## Additional Notes

The in-app File viewer already uses the same `link.row`/`link.col` and jumps correctly, which is how the gap surfaced — only the external-editor path discarded the coordinates.

---
## Summary by cubic
Fixes terminal file links opening editors at line 1. We now pass and honor line/column on Linux editors, and drop invalid coordinates so files still open; macOS stays unchanged.

- **Bug Fixes**
  - Forward `line`/`column` through `external.openFileInEditor` → `openPathInApp` → `getAppCommand`.
  - Add Linux goto args per editor: VS Code family (`vscode`, `vscode-insiders`, `cursor`) uses `-g path:line[:column]`; Sublime/Zed use `path:line[:column]`; JetBrains IDEs use `--line <line> [--column <column>] <path>`.
  - Input schema drops non-positive or fractional `line`/`column` to `undefined` (no error); only positive integers generate goto args.
  - Other apps ignore coordinates; no change when none are provided. macOS keeps using `open` and ignores coordinates.

<sup>Written for commit aa48c0090675b3fbb96d98feba7d79407e71cbe7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional line/column support when opening files in external editors/IDEs, enabling precise “go to” navigation across supported editor families (with macOS behavior unchanged).
* **Refactor**
  * Centralized the external-app and open-in-editor input validation via shared schemas.
* **Tests**
  * Added coverage for line/column parsing (invalid values are dropped, not rejected) and for platform-specific command/argument generation, including regression checks for existing macOS behavior.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/external/helpers.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/external/helpers.test.ts
@@ -586,6 +586,209 @@ describe("stripPathWrappers", () => {
 	});
 });
 
+describe("getAppCommand — line/column jump", () => {
+	const originalPlatform = process.platform;
+
+	beforeEach(() => {
+		Object.defineProperty(process, "platform", { value: "linux" });
+	});
+
+	afterEach(() => {
+		Object.defineProperty(process, "platform", { value: originalPlatform });
+	});
+
+	describe("Linux VS Code family", () => {
+		test("vscode: line and column → -g path:line:col", () => {
+			const result = getAppCommand("vscode", "/path/to/file", "linux", 10, 5);
+			expect(result).toEqual([
+				{ command: "code", args: ["-g", "/path/to/file:10:5"] },
+			]);
+		});
+
+		test("vscode: line only, no column → -g path:line", () => {
+			const result = getAppCommand(
+				"vscode",
+				"/path/to/file",
+				"linux",
+				10,
+				undefined,
+			);
+			expect(result).toEqual([
+				{ command: "code", args: ["-g", "/path/to/file:10"] },
+			]);
+		});
+
+		test("cursor: line and column → -g path:line:col", () => {
+			const result = getAppCommand("cursor", "/path/to/file", "linux", 1, 1);
+			expect(result).toEqual([
+				{ command: "cursor", args: ["-g", "/path/to/file:1:1"] },
+			]);
+		});
+	});
+
+	describe("Linux Sublime/Zed family", () => {
+		test("sublime: line and column → path:line:col", () => {
+			const result = getAppCommand("sublime", "/path/to/file", "linux", 42, 8);
+			expect(result).toEqual([
+				{ command: "subl", args: ["/path/to/file:42:8"] },
+			]);
+		});
+
+		test("sublime: line only, no column → path:line", () => {
+			const result = getAppCommand(
+				"sublime",
+				"/path/to/file",
+				"linux",
+				42,
+				undefined,
+			);
+			expect(result).toEqual([{ command: "subl", args: ["/path/to/file:42"] }]);
+		});
+
+		test("zed: line and column → path:line:col", () => {
+			const result = getAppCommand("zed", "/path/to/file", "linux", 5, 3);
+			expect(result).toEqual([{ command: "zed", args: ["/path/to/file:5:3"] }]);
+		});
+	});
+
+	describe("Linux JetBrains family", () => {
+		test("webstorm: line and column → --line line --column col path", () => {
+			const result = getAppCommand("webstorm", "/path/to/file", "linux", 20, 1);
+			expect(result).toEqual([
+				{
+					command: "webstorm",
+					args: ["--line", "20", "--column", "1", "/path/to/file"],
+				},
+			]);
+		});
+
+		test("webstorm: line only, no column → --line line path", () => {
+			const result = getAppCommand(
+				"webstorm",
+				"/path/to/file",
+				"linux",
+				20,
+				undefined,
+			);
+			expect(result).toEqual([
+				{ command: "webstorm", args: ["--line", "20", "/path/to/file"] },
+			]);
+		});
+
+		test("intellij (multi-edition): line and column → each candidate gets --line --column path", () => {
+			const result = getAppCommand("intellij", "/path/to/file", "linux", 10, 2);
+			expect(result).toEqual([
+				{
+					command: "idea",
+					args: ["--line", "10", "--column", "2", "/path/to/file"],
+				},
+				{
+					command: "intellij-idea-ultimate",
+					args: ["--line", "10", "--column", "2", "/path/to/file"],
+				},
+				{
+					command: "intellij-idea-community",
+					args: ["--line", "10", "--column", "2", "/path/to/file"],
+				},
+			]);
+		});
+	});
+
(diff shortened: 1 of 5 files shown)


Rakesh1002 (commented):
Ty for the PR! Nice catch — this is a real silent-drop bug: `openFileInEditor` and the terminal link handler both already carried `line`/`column`, but they died at `getAppCommand`, so external editors always landed on line 1. The fix is clean and well-scoped, the editor goto syntaxes look correct for all three families (VS Code `-g`, Sublime/Zed inline, JetBrains `--line/--column`), and I really appreciate that the no-coordinate path is byte-for-byte identical to before and that you proved it with an explicit parity test. The macOS-untouched + JetBrains multi-edition coverage is thorough. General net improvement :)

Two small things, neither blocking:

1. `min(1)` now hard-rejects bad coordinates (question). `line`/`column` flow straight from terminal text parsing (`local-link-detector.ts:114-115`, `parsedLink.suffix?.row/col`). With `z.number().int().min(1)`, a link like `file.ts:0` now throws a ZodError, which `useFileLinkClick.ts:34` catches and shows as a "Failed to open file in editor" toast. Before this PR the input schema was just `z.number().optional()`, so that same link still opened (the coord was simply ignored). Could we degrade more gracefully — clamp/coerce the bad coord, or drop just the offending value — so the file still opens instead of the whole action failing? Edge-case in practice, but it's a slight behavior regression for malformed links.

2. Dead re-export (nit). `export { openFileInEditorInputSchema };` in `index.ts` doesn't seem to be imported anywhere — the test imports it from `./schemas`, and nothing consumes it from `./index`. Probably safe to drop unless it's meant as intentional API surface. (The sibling `ExternalAppSchema` import is still used further down in `index.ts`, so that one's fine.)

One pre-merge housekeeping: worth a final `bun run lint` — CI fails on any biome diagnostic (info/warning included). It looks consistent with biome's line-width rules so it's likely already fine, just a safety check.

Thanks again — happy to approve once the `min(1)` rejection question is sorted (or you confirm it's acceptable).

HwangRock:
@Rakesh1002 Great catch on the `min(1)` regression — you're right. `line`/`column` come straight from terminal text (`local-link-detector.ts`), and `\d+` matches `0`, so `file.ts:0` would `throw` and surface as a "Failed to open file in editor" toast instead of just opening the file like it did before.

Fixed by dropping invalid coords instead of rejecting: the schema now coerces any non-positive-integer to `undefined` (via `.transform`) rather than throwing. So `file.ts:0` opens the file with no jump — **identical to the no-coordinate path**, so the parity invariant still holds. Went with *drop* over *clamp* since `0 → 1` would silently land on the wrong line.

Also removed the dead `openFileInEditorInputSchema` re-export from `index.ts`, and `bun run lint` is clean on the changed files. Tests updated to assert the drop-to-`undefined` behavior.

ghost:

| | Title |
|---|-------|
| 1 | [Extract and enhance external editor schemas](https://stagereview.app/superset-sh/superset/pull/5311/chapters/1) |
| 2 | [Implement editor-specific goto arguments for Linux](https://stagereview.app/superset-sh/superset/pull/5311/chapters/2) |
| 3 | [Wire line and column through the router](https://stagereview.app/superset-sh/superset/pull/5311/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5311">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit aa48c0090675b3fbb96d98feba7d79407e71cbe7 on Jun 30, 2026 8:53am UTC.</sub>

PR #5312, opened 2026-06-20
fix(host-service): release workspace runtime state on delete

### LINKED ISSUE #5076: [bug] Deleted workspaces reappear in the sidebar
1. Click "Close workspace" "X" on Workspace in sidebar.
2. Confirm "Delete workspace..." and click "Delete".
3. Workspace returns to side bar.

Interestingly, if i try to drag and reposition it, it goes away.

Version 1.12.1 (1.12.1)

kunaldhongade:
Summary
- add an explicit GitWatcher.removeWorkspace cleanup path that closes deleted workspace watchers and clears pending debounce batches
- add PullRequestRuntimeManager.removeWorkspace to drop pending reruns and delegate watcher cleanup
- call runtime cleanup from workspaceCleanup.destroy after the delete commit point, including already-absent local rows
- add regression coverage for watcher cleanup, PR runtime cleanup, and workspace delete lifecycle ordering

Why
Issue #5076 reports deleted workspaces continuing to produce host-service daemon errors until the app/daemon is restarted. The local DB can be consistent while long-running host-service memory still contains PR-runtime sync state and FSEvents watchers for removed workspaces.

Root Cause
workspaceCleanup.destroy removed workspace rows and disk/cloud state but did not explicitly release host-service runtime/watch state keyed by the deleted workspace id. GitWatcher eventually reconciles on interval, but pending debounce batches and active watchers can remain long enough to emit stale syncs, and PullRequestRuntimeManager had no lifecycle cleanup hook for deleted workspaces.

User Impact
Deleting a workspace now immediately releases host-service PR runtime and watcher state, which should stop stale "Workspace not found" restore paths and repeated pull-request-runtime/FSEvents errors for deleted workspaces without requiring a full daemon restart.

Relation to #5279
This is intentionally separate from #5279. PR #5279 handles stale closed-worktree directories in the desktop/local closed-worktree delete route. This PR handles the long-running host-service daemon cleanup path for deleted workspaces.

Validation
- bun test packages/host-service/src/events/git-watcher.test.ts
- bun test packages/host-service/src/runtime/pull-requests/pull-requests.test.ts
- bun test packages/host-service/test/workspace-cleanup.test.ts
- bun run --cwd packages/host-service test
- bun run --cwd packages/host-service typecheck
- bun run --cwd packages/host-service build:host
- bun run lint
- bun run typecheck
- git diff --check

Build Note
- bun run build progressed through host-service bundling and desktop native-runtime validation, then failed in local electron-builder packaging because this checkout's Bun node_modules layout could not resolve the production dependency @superset/chat -> hono. That packaging dependency-collection failure appears environment/layout related and unrelated to this host-service cleanup diff.

Fixes #5076

---
## Summary by cubic
Immediately release host-service PR runtime state and file watchers when a workspace is deleted, and cancel in-flight syncs so nothing runs after delete — stopping stale syncs and “Workspace not found” errors without restarting the daemon. Fixes #5076.

- **Bug Fixes**
  - Added `GitWatcher.removeWorkspace` to close watchers, dispose worktree watch, and clear debounce timers/batches; used by rescan and error paths.
  - Hardened watcher error cleanup to handle races during registration; prevent double-dispose with a `disposed` flag and always clear pending timers/batches.
  - Added `PullRequestRuntimeManager.removeWorkspace` to drop pending reruns, mark the workspace cancelled, and delegate watcher cleanup.
  - Guarded sync scheduling/execution with a cancelled set to no-op before opening git or refreshing a project, and clear cancellation on completion.
  - `workspaceCleanup.destroy` now calls runtime cleanup after the delete commit (even if the local row is already missing) and never before cloud delete succeeds; added regression tests for cleanup, watcher error races, and ordering.

<sup>Written for commit 3871079677a3a3b48d5d0840ef2e50b1fdf37ea2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added workspace-removal support to the git watcher and pull-request runtime, including full teardown and sync-state cancellation.
  * Workspace cleanup now clears runtime state as part of the deletion flow, recording any cleanup failures as warnings.
* **Bug Fixes**
  * Prevents further watcher/sync activity after cancellation, with idempotent cleanup and improved error-path teardown.
* **Tests**
  * Expanded tests for watcher disposal, idempotency, early-termination on cancellation/errors, and correct timing of runtime cleanup during workspace destroy.

### DIFF
--- a/packages/host-service/src/events/git-watcher.test.ts
+++ b/packages/host-service/src/events/git-watcher.test.ts
@@ -17,14 +17,15 @@ type GitWatcherInternals = {
 };
 
 describe("GitWatcher.removeWorkspace", () => {
-	test("closes removed workspace watchers and clears pending debounce state", () => {
+	test("closes removed workspace watchers and clears pending debounce state", async () => {
 		const watcher = new GitWatcher({} as never, {} as never);
 		const internals = watcher as unknown as GitWatcherInternals;
 		const closeRemoved = mock(() => {});
 		const disposeRemoved = mock(() => {});
 		const closeActive = mock(() => {});
 		const disposeActive = mock(() => {});
-		const timer = setTimeout(() => {}, 60_000);
+		const timerFired = mock(() => {});
+		const timer = setTimeout(timerFired, 10);
 
 		try {
 			internals.watched.set("deleted-workspace", {
@@ -57,6 +58,9 @@ describe("GitWatcher.removeWorkspace", () => {
 			expect(internals.watched.has("active-workspace")).toBe(true);
 			expect(closeActive).not.toHaveBeenCalled();
 			expect(disposeActive).not.toHaveBeenCalled();
+
+			await new Promise((resolve) => setTimeout(resolve, 25));
+			expect(timerFired).not.toHaveBeenCalled();
 
 			watcher.removeWorkspace("deleted-workspace");
 			expect(closeRemoved).toHaveBeenCalledTimes(1);
--- a/packages/host-service/src/events/git-watcher.test.ts
+++ b/packages/host-service/src/events/git-watcher.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, mock, test } from "bun:test";
+import { GitWatcher } from "./git-watcher";
+
+type GitWatcherInternals = {
+	watched: Map<
+		string,
+		{
+			workspaceId: string;
+			worktreePath: string;
+			gitDir: string;
+			watcher: { close: () => void };
+			disposeWorktreeWatch: () => void;
+		}
+	>;
+	debounceTimers: Map<string, ReturnType<typeof setTimeout>>;
+	pendingBatches: Map<string, { hasGitDir: boolean; paths: Set<string> }>;
+};
+
+describe("GitWatcher.removeWorkspace", () => {
+	test("closes removed workspace watchers and clears pending debounce state", () => {
+		const watcher = new GitWatcher({} as never, {} as never);
+		const internals = watcher as unknown as GitWatcherInternals;
+		const closeRemoved = mock(() => {});
+		const disposeRemoved = mock(() => {});
+		const closeActive = mock(() => {});
+		const disposeActive = mock(() => {});
+		const timer = setTimeout(() => {}, 60_000);
+
+		try {
+			internals.watched.set("deleted-workspace", {
+				workspaceId: "deleted-workspace",
+				worktreePath: "/tmp/deleted",
+				gitDir: "/tmp/deleted/.git",
+				watcher: { close: closeRemoved },
+				disposeWorktreeWatch: disposeRemoved,
+			});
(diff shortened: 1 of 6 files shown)


Rakesh1002 (commented):
Ty for the PR Kunal — this is a clean, well-targeted fix for #5076 and a general net improvement :)

The diagnosis is spot-on: the local DB could be consistent while host-service memory still held FSEvents watchers + PR-runtime sync state for a deleted workspace. I like that the teardown is layered (GitWatcher -> PullRequestRuntimeManager -> runDestroy) and that the ordering is deliberate:
- runtime cleanup runs after the cloud-delete commit point (the await at workspace-cleanup.ts:369), so a failed cloud delete keeps the workspace visible/retryable;
- it runs outside the `if (local)` guard, so an already-absent local row still gets its runtime state released.
Both are pinned down by the three new ordering tests, and the GitWatcher idempotency/error-race coverage is genuinely thorough. Nice that the prior bot findings (in-flight cancellation, the watcher-error race, the workspaceSyncState delete) are all folded in already.

A couple of small, non-blocking notes:

1. `cancelledWorkspaceSyncs` lifecycle is correct but subtle — it's added only when a syncState exists and cleared only in the run() finally. That works because a live syncState always implies an in-flight run whose finally will fire (the set in enqueueWorkspaceSync happens right after run() is kicked off, and run always awaits at least once), so the flag always drains; an idle-delete adds nothing and relies on the existing DB-row-gone no-op. A one-line comment on that invariant would save the next reader a trace through enqueue/run.

2. The first `await this.git(workspace.worktreePath)` can still open git on a worktree mid-deletion if cancellation lands during that await; the post-open check + catch suppression turn it into a harmless transient rather than a thrown error, so this narrows the race window well — just flagging that it's narrowed, not fully closed, which seems like the right tradeoff for best-effort teardown.

3. Tiny: the debounce-timer test asserts via a real-clock 25ms wait against a 10ms timer — fine as-is, but fake timers would be a touch less flake-prone under CI load.

One heads-up unrelated to the code: the required CI jobs don't show as run yet — only the CodeRabbit/cubic checks have reported, so lint/typecheck/test haven't actually executed on the PR. Worth getting those green before merge even though you've run them locally. Nothing here blocks from my side once CI is green.

Kitenite:
Closing: main already ships this. git-watcher reconciles vanished workspaces (745d7a407, 43cea79ba). Thanks for the PR; if you think the shipped version misses a case, open an issue and link this.

PR #5325, opened 2026-06-22
fix(host-service): detect ports for terminals without an attached renderer

Kitenite:
## Problem

Ports were detected less completely in v2 than in v1. Both use the same shared `@superset/port-scanner` `PortManager` (2.5s periodic scan of registered terminal sessions, keyed by PID tree), so the gap was in **what gets registered**:

- **v1** (desktop `daemon-manager.ts` → `reconcileOnStartup`) pre-registers **every alive daemon session** with `portManager.upsertSession(...)` at startup, before any renderer attaches.
- **v2** host-service only called `upsertSession` inside `createTerminalSessionInternal` — i.e. **only when a renderer creates or attaches to a terminal**. Terminals the daemon is running but nobody has opened (notably sessions adopted after a host-service restart) were never scanned, so their ports never appeared in the v2 sidebar.

## Fix

Fold a port-scan sync into the existing **terminal reaper** (`reaper.ts`) instead of adding a parallel module. The reaper already calls `daemon.list()` and joins `terminalSessions` (for `terminalId → workspaceId`) on the same pass, immediately at startup and on its interval — so there's no extra daemon call, DB query, or timer.

- `planPortScanSync()` — pure decision function:
  - **register** every alive daemon session that maps to an active workspace row and isn't already owned by a live in-memory session;
  - **unregister** every watched terminal the daemon no longer reports and no live in-memory session owns (adopted sessions never get the daemon exit subscription that normally unregisters them).
- `applyPortScanSync()` applies the plan to `portManager` on each reaper pass, before the empty-list short-circuit so an empty daemon clears stale scans.
- Supporting helpers: `PortManager.getRegisteredTerminalIds()` and `terminal.ts isLiveTerminalSession()` (the latter guards a renderer-attached session against being dropped during a racy `daemon.list()`).

The `terminalId → workspaceId` join is irreducible: `daemon.list()` returns `{id, pid, alive}` only, and the sidebar groups ports by workspace — that mapping lives in the host DB, which the reaper already reads.

## End-to-end

- **On start:** `eventBus.start()` (in `createApp`) wires `portManager` → `port:changed` broadcast; the reaper's immediate first pass registers live sessions → periodic scan begins. Bridge is wired before registration, so no early events drop.
- **Renderer reads:** `ports.getAll` returns `portManager.getAllPorts()`, now populated for unattached terminals.
- **Renderer subscribes:** the event-bus `port:changed` stream delivers live add/remove for all live terminals, attached or not.

## Notes

- Self-heal cleanup of adopted-but-exited registrations rides the reaper's 5-min cadence; their ports still vanish within 2.5s via the empty-tree scan, only the no-op registration lingers. Attached terminals unregister instantly on the daemon exit event.

## Testing

- New `reaper.test.ts` covers `planPortScanSync` (register/skip/unregister/empty-daemon/race-guard); verified each case fails under impl mutation.
- `bun run lint` exits 0; typecheck passes for `@superset/host-service` and `@superset/port-scanner`; existing `port-scanner` tests pass.

---
## Summary by cubic
Fixes missing port detection for terminals running without an attached renderer by syncing the port scanner with daemon sessions during reaper passes. Ports now appear for adopted/live terminals right after start and stay accurate.

- **Bug Fixes**
  - Sync `portManager` with alive daemon sessions on each reaper pass: register mapped active rows not owned by a live in-memory session; unregister watched terminals the daemon no longer reports; guard renderer-attached sessions via `isLiveTerminalSession`.
  - Skip the DB join when the daemon reports no live sessions while still dropping stale scans; isolate the sync in try/catch so reaper cleanup can’t be blocked.
  - Added `PortManager.getRegisteredTerminalIds()`, `planPortScanSync`, and tests.

<sup>Written for commit 9c940ac33c3650645104a44444f310d201ef0513. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5325?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved port scan management by reconciling with the daemon’s active terminal sessions: newly live terminals are registered, stale ones are unregistered, and previously adopted scans are cleared when no live sessions are reported.
  * Added a helper to treat only non-exited in-memory terminal sessions as “live” for reconciliation.
  * Exposed a method to retrieve currently registered terminal IDs for port scanning.

* **Tests**
  * Added Bun coverage for port-scan synchronization scenarios, including adoption, skipping owned sessions, handling missing/inactive rows, unregistering removed terminals, clearing when none are live, and temporary daemon absence.

### DIFF
--- a/packages/host-service/src/terminal/reaper/reaper.test.ts
+++ b/packages/host-service/src/terminal/reaper/reaper.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from "bun:test";
+import { planPortScanSync } from "./reaper.ts";
+
+const noneLive = () => false;
+
+describe("planPortScanSync", () => {
+	it("registers alive daemon sessions that map to an active workspace row", () => {
+		const plan = planPortScanSync({
+			liveSessions: [{ id: "term-1", pid: 4242 }],
+			rowById: new Map([
+				["term-1", { status: "active", originWorkspaceId: "ws-1" }],
+			]),
+			registeredTerminalIds: [],
+			isLive: noneLive,
+		});
+
+		expect(plan.register).toEqual([
+			{ terminalId: "term-1", workspaceId: "ws-1", pid: 4242 },
+		]);
+		expect(plan.unregister).toEqual([]);
+	});
+
+	it("skips sessions already owned by a live in-memory session", () => {
+		const plan = planPortScanSync({
+			liveSessions: [{ id: "term-1", pid: 4242 }],
+			rowById: new Map([
+				["term-1", { status: "active", originWorkspaceId: "ws-1" }],
+			]),
+			registeredTerminalIds: [],
+			isLive: (id) => id === "term-1",
+		});
+
+		expect(plan.register).toEqual([]);
+	});
+
+	it("skips sessions without a row, without a workspace, or not active", () => {
+		const plan = planPortScanSync({
+			liveSessions: [
+				{ id: "rowless", pid: 1 },
+				{ id: "no-workspace", pid: 2 },
+				{ id: "exited", pid: 3 },
+				{ id: "disposed", pid: 4 },
+			],
+			rowById: new Map([
+				["no-workspace", { status: "active", originWorkspaceId: null }],
+				["exited", { status: "exited", originWorkspaceId: "ws-1" }],
+				["disposed", { status: "disposed", originWorkspaceId: "ws-1" }],
+			]),
+			registeredTerminalIds: [],
+			isLive: noneLive,
+		});
+
+		expect(plan.register).toEqual([]);
+	});
+
+	it("unregisters scanned terminals the daemon no longer reports", () => {
+		const plan = planPortScanSync({
+			liveSessions: [{ id: "term-1", pid: 4242 }],
+			rowById: new Map([
+				["term-1", { status: "active", originWorkspaceId: "ws-1" }],
+			]),
+			registeredTerminalIds: ["term-1", "dead-term"],
+			isLive: noneLive,
+		});
+
+		expect(plan.unregister).toEqual(["dead-term"]);
+	});
+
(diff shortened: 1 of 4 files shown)


Kitenite:
Both addressed in 7c5a6e4d3:

1. **Unconditional DB select** — `rowById` is now built via `loadTerminalRowsById(db)` only when `liveSessions.length > 0`; the empty-daemon path passes `new Map()`, so an idle host still runs the sync to clear stale scans but does no DB query.
2. **try/catch isolation** — `applyPortScanSync` now wraps its body so a port-manager error logs and returns instead of aborting the reaper's orphan cleanup for the interval.

ghost:

| | Title |
|---|-------|
| 1 | [Expose registered terminal IDs from PortManager](https://stagereview.app/superset-sh/superset/pull/5325/chapters/1) |
| 2 | [Add terminal session liveness check](https://stagereview.app/superset-sh/superset/pull/5325/chapters/2) |
| 3 | [Implement port scan synchronization logic](https://stagereview.app/superset-sh/superset/pull/5325/chapters/3) |
| 4 | [Test port scan synchronization planning](https://stagereview.app/superset-sh/superset/pull/5325/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/5325">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 9c940ac33c3650645104a44444f310d201ef0513 on Jun 29, 2026 3:14am UTC.</sub>

PR #5327, opened 2026-06-23
fix(desktop): keep removed projects out of the sidebar

AviPeltz:
## Summary
- Removing a project from the sidebar now sticks. Previously the project would reappear almost immediately and felt un-removable.
- Project removal now **tombstones** its workspaces (hidden local-state rows) instead of hard-deleting them, matching the pattern #5083 introduced for single-workspace removal.
- Extracted the pure sidebar mutations into a testable `sidebarMutations` module and added a regression test for the re-pin race.

## Why / Context
#5083 added `useAutoAddLocalWorkspacesToSidebar`, which backfills sidebar rows for workspaces on this device that have no `v2WorkspaceLocalState` row (e.g. CLI-created ones). It treats a **missing** local-state row as "never seen here → pin it".

`removeProjectFromSidebar` still **hard-deleted** the project's local-state rows. So right after removal, the auto-add hook saw those this-machine workspaces with no row, called `ensureWorkspaceInSidebar`, and — via `ensureSidebarProjectRecord` — recreated both the workspace rows and the project record. The project bounced straight back.

Single-workspace removal didn't have this problem because #5083 had already switched it to a tombstone (`isHidden: true`), which the hook sees in `knownWorkspaceIds` and skips. Project removal was simply never converted. (This regression is on `main` but has not shipped in a stable desktop release yet — latest tag `desktop-v1.12.5` predates #5083 — so it's currently canary-only.)

## How It Works
- `removeProjectFromSidebarState` tombstones the **union** of (a) existing local-state rows for the project and (b) this machine's `v2Workspaces` rows for the project. (b) is the key addition — it catches auto-included `main`/CLI workspaces that have no local-state row yet and would otherwise be re-pinned.
- Each workspace is marked `isHidden: true` (sidebar filters hidden rows out), pane runtimes are cleaned up, and the project record + sections are still deleted. Projects render only from `v2SidebarProjects`, so deleting that record removes the project; with every this-machine workspace now tombstoned, nothing recreates it.
- The mutations were lifted out of the hook into `sidebarMutations.ts` (no React/Electron imports) with pane-runtime cleanup injected as a callback, so they unit-test against a tiny in-memory collection.

## Manual QA Checklist
- [x] Remove a project from the sidebar → it stays gone (does not reappear after the auto-add hook runs)
- [x] Remove a project whose `main` workspace was auto-included (no explicit sidebar placement) → stays gone
- [x] Remove a single workspace (existing behavior) → still hidden, unchanged
- [x] Other projects and their workspaces are unaffected
- [x] Removed project stays gone after app restart
- [x] No console errors in renderer during/after removal

## Testing
- `bun run typecheck` ✅
- `bun run lint` ✅
- `bun test src/.../useDashboardSidebarState/sidebarMutations.test.ts` ✅ (5 passing) — includes the regression guard: remove a project with a row-less main workspace → simulate the auto-add hook → assert it does **not** re-pin and the project record stays deleted.

## Design Decisions
- **Tombstone vs. a project-level "dismissed" flag**: tombstoning reuses the exact mechanism #5083 already established for workspaces and needs no schema change, keeping the fix small and consistent. The auto-add hook keys off workspace-row existence, so workspace tombstones are sufficient.

## Known Limitations / Follow-ups
- Tombstones accumulate as inert localStorage rows (one per removed workspace) — same cost the existing single-workspace path already pays; bounded by workspace count. A heal step pruning tombstones whose `v2Workspaces` row no longer exists is a reasonable future cleanup, intentionally deferred to keep this PR focused.

---
## Summary by cubic
Fixes a bug where removed projects reappeared in the desktop sidebar by tombstoning their workspaces. Also extracts pure sidebar mutations into a small module with tests and requires a non-null machineId to avoid re-pin races.

- **Bug Fixes**
  - Tombstones each workspace (`isHidden: true`), including this-machine workspaces without a local-state row, so the auto-add hook cannot re-pin them.
  - Requires a non-null machineId in project removal to ensure row-less local workspaces are tombstoned and not re-pinned when the id resolves.
  - Deletes the project and its sections, clears pane runtimes, and persists across restarts.

- **Refactors**
  - Extracted `tombstoneSidebarWorkspaceRecord` and `removeProjectFromSidebarState` into `sidebarMutations.ts` with unit tests guarding the re-pin race.

<sup>Written for commit 811bb431db2beb1ba32217c1d7ece28c149d5d13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5327?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Added unit tests covering sidebar “tombstoning” and project removal behavior, including correct handling of auto-included workspaces and cleanup of related sidebar sections/panes.

* **Refactor**
  * Extracted sidebar state mutation logic into reusable, pure helper functions for clearer, consistent updates to sidebar local state.
  * Updated the dashboard sidebar hook to delegate hide/remove actions to the new helpers, improving reliability of sidebar state transitions.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/hooks/useDashboardSidebarState/sidebarMutations.test.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/hooks/useDashboardSidebarState/sidebarMutations.test.ts
@@ -1,0 +1,248 @@
+import { describe, expect, it } from "bun:test";
+import {
+	removeProjectFromSidebarState,
+	tombstoneSidebarWorkspaceRecord,
+} from "./sidebarMutations";
+
+/**
+ * Minimal in-memory stand-in for a TanStack DB collection, implementing only
+ * the surface the sidebar mutations touch (`get`/`insert`/`update`/`delete`
+ * plus a `.state` Map).
+ */
+function makeCollection<T>(getKey: (item: T) => string) {
+	const state = new Map<string, T>();
+	return {
+		state,
+		get: (key: string) => state.get(key),
+		insert: (item: T) => {
+			state.set(getKey(item), structuredClone(item));
+		},
+		update: (key: string, producer: (draft: T) => void) => {
+			const existing = state.get(key);
+			if (!existing) return;
+			const draft = structuredClone(existing);
+			producer(draft);
+			state.set(key, draft);
+		},
+		delete: (keys: string | string[]) => {
+			for (const key of Array.isArray(keys) ? keys : [keys]) {
+				state.delete(key);
+			}
+		},
+	};
+}
+
+type LocalStateRow = {
+	workspaceId: string;
+	createdAt: Date;
+	sidebarState: {
+		projectId: string;
+		tabOrder: number;
+		sectionId: string | null;
+		isHidden: boolean;
+	};
+	paneLayout: { version: number; tabs: unknown[]; activeTabId: string | null };
+};
+
+function localStateRow(
+	workspaceId: string,
+	projectId: string,
+	overrides: Partial<LocalStateRow["sidebarState"]> = {},
+): LocalStateRow {
+	return {
+		workspaceId,
+		createdAt: new Date("2026-01-01T00:00:00.000Z"),
+		sidebarState: {
+			projectId,
+			tabOrder: 1,
+			sectionId: null,
+			isHidden: false,
+			...overrides,
+		},
+		paneLayout: { version: 1, tabs: [], activeTabId: null },
+	};
+}
+
+function makeCollections() {
+	return {
+		v2WorkspaceLocalState: makeCollection<LocalStateRow>(
+			(row) => row.workspaceId,
+		),
+		v2Workspaces: makeCollection<{
+			id: string;
+			projectId: string;
+			hostId: string;
+		}>((row) => row.id),
+		v2SidebarSections: makeCollection<{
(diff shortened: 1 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Extract sidebar mutations to testable module](https://stagereview.app/superset-sh/superset/pull/5327/chapters/1) |
| 2 | [Wire mutations into the sidebar state hook](https://stagereview.app/superset-sh/superset/pull/5327/chapters/2) |
| 3 | [Add regression tests for project removal](https://stagereview.app/superset-sh/superset/pull/5327/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5327">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 811bb431db2beb1ba32217c1d7ece28c149d5d13 on Jun 23, 2026 5:54pm UTC.</sub>

Issue #4018, opened 2026-05-03
[feat] Add a platform-level layer above Project to group multiple repos


ashbrener:
## Context & motivation

I run multiple distinct platforms, each with several repositories. Today Superset's hierarchy is *Organization → Project (repo) → Workspaces (worktrees on branches)*. Projects sit flat under one organization, with no way to group multiple repos under a platform-level layer.

As a single user managing several platforms on one machine, I have no clean way to keep these contexts separate — visually, by config, or by automation scope.

## The gap

The current model assumes the highest meaningful unit is a single repo (a Project). But for platform development, the natural unit of work is the *platform* — which spans multiple repos (frontend, backend, infra, mobile, docs, etc.).

There's no layer above Project to express this. Sections (#2067) help group workspaces *within* a project, but don't solve the *projects-across-platforms* problem.

## What I tried

- Looking for a "create organization" option in the desktop app and at app.superset.sh on the free/individual tier — couldn't find a clear path to creating additional personal organizations.
- Naming-convention workarounds (`platformA-repo1`, `platformB-repo2`, etc.) — works but gets messy fast and doesn't give per-platform isolation (separate env, MCP servers, agent presets, automations per platform).

## Proposed solutions (whichever is fastest to ship)

Any of the following would close the gap — listed in rough order from biggest impact to cheapest hack. Pick whichever is fastest to build:

1. **Project groups / folders within an organization (preferred).** A new layer above Project that groups multiple repos under a named parent (e.g., "Platform A"). Collapsible in the V2 sidebar, similar to what was done for workspace sections in #2067. Per-group config for env, MCP servers, agent presets, and automations would be the full version; even just visual grouping would help significantly.
2. **Multiple personal organizations on the free tier.** Org-per-platform as a workaround. Per-organization Electric SQL collections already exist in the codebase, suggesting the data model supports this cleanly. Acknowledged this is a workaround — orgs are really meant for teams/billing, not personal grouping.
3. **Per-window organization context.** Allow multiple Superset windows to hold different active organizations, so each platform can live on its own monitor.

## Side note on naming

"Workspace" in Superset means "one branch's worktree," which is much smaller than the term implies in most other tools (VS Code, Slack, GitHub all use "workspace" to mean the larger container). Renaming to "Branch" or "Worktree" — and reserving "Workspace" for a higher-level grouping concept — would make the hierarchy easier to reason about. Not the main ask, just an observation.

## Why this matters

[...]

Issue #4910, opened 2026-05-24
[feat] Open Organization in new window (option)


shivraj-roy:
#### The feature
Right now, whenever I switch organizations, it doesn't open in a new window. It just opens in the same window!

What if user want to work parallelly between organization..!

ashbrener:
This is implemented by #5337 (per-window organization context). With it you can run organizations in parallel windows: **File → New Window** opens an independent window, and switching its organization is *window-local* — so two windows can show two different orgs side by side, which is exactly the "work in parallel between organizations" case here. Each window's title shows its org, and open windows are restored on relaunch.

The PR also adds a `window.openNew` procedure, so an explicit **"Open in new window"** option directly on the organization switcher (the literal ask in this issue) is a small follow-up on top of it — happy to include it in the same PR if maintainers want. (Targets the v2 UI.)

ashbrener:
Small, self-contained fix for this: #6133 — one file, keeps the previous organization on screen while the switch resolves instead of unmounting the tree.

PR #5337, opened 2026-06-23
feat(desktop): multi-window with per-window organization context (#4018)

ashbrener:
## Description

Adds **VS Code–style multiple windows, each bound to its own organization**, to the desktop app. Open a second window (**File → New Window**), switch it to a different org, and work on two platforms side by side — switching the org in one window never affects another. Each window's title shows its org, and open windows are restored on relaunch.

This implements **option 3(b) ("per-window organization context")** from #4018 — the smallest slice that delivers the core "work on multiple platforms at once" goal.

**How it works:** each `BrowserWindow` is its own renderer with its own active org. tRPC calls resolve the **calling** window from the IPC sender (`ctx.senderWindow`), so a window's org is tracked per-window in the main process and used to scope its cloud data (the desktop API client sends `x-superset-organization-id`; Electric shapes already take `organizationId`). Previously, switching org mutated the shared login session (`authClient.organization.setActive`) — inherently global; this PR makes switching **window-local**.

```mermaid
flowchart LR
  subgraph A["Window A — renderer"]
    CA["CollectionsProvider<br/>activeOrg = Org A"]
  end
  subgraph B["Window B — renderer"]
    CB["CollectionsProvider<br/>activeOrg = Org B"]
  end
  CA -- "getActiveOrg / setActiveOrg<br/>(ctx.senderWindow = 1)" --> REG[("Window registry<br/>windowId to org")]
  CB -- "getActiveOrg / setActiveOrg<br/>(ctx.senderWindow = 2)" --> REG
  CA -- "x-superset-organization-id: Org A" --> CLOUD[("Cloud API + Electric")]
  CB -- "x-superset-organization-id: Org B" --> CLOUD
```

**What's included**

- **Multi-window foundation** — a window registry (single source of truth for open windows + their org) and a split of `MainWindow()` into one-time `initAppServices()` (IPC handler, menu) and a repeatable `createPlatformWindow()`. App-wide services (notifications HTTP server on a fixed port, terminal/notification listeners) start on the first window and stop on the last, so a second window can't rebind the port or double-register. Adds **File → New Window** (`⇧⌘N` on macOS, `Ctrl+Alt+N` on Windows/Linux), and gives the Window menu the macOS `windowMenu` role so the open windows are listed and switchable.
- **Per-window org context** — tRPC context exposing `ctx.senderWindow`, `window.getActiveOrg` / `setActiveOrg` / `openNew`, the `x-superset-organization-id` header on the desktop API client, and window-local org switching. The host data path (`LocalHostServiceProvider`, `useHostTargetUrl`, `useHostProjects`) reads the window's org rather than the shared session, so each window starts its host service and lists projects for the org it's actually showing.
- **Per-window title** — `"<Org> — Superset"`, so windows are distinguishable in Mission Control / the window switcher.
- **Restore on relaunch** — persists each window's bounds + org (`windows-state.json`, migrating from the legacy single-window record) and restores them on launch.

**Scope note:** targets the **v2 cloud UI**. In v1 the workspace list is machine-local (the local-db `projects`/`worktrees`/`workspaces` tables have no `organizationId`), so per-window org has nothing to scope there; on v2 the sidebar reads org-scoped Electric collections via `CollectionsProvider`, which is what this makes per-window.

## Related Issues

Related to #4018 (implements option 3(b); does not close the issue, which also proposes the larger "project groups" design).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

Driven manually on the v2 UI via the Chrome DevTools Protocol against a local dev build:

- New Window opens an independent second window.
- Switching window B → Org B leaves window A on Org A (per-window isolation; `ctx.senderWindow` routes correctly per window).
- Window titles differ (`"Ash 2 — Superset"` vs `"Local Admin's Team — Superset"`).
- Quit + relaunch restores both windows on their orgs (planted a 2-window state; both restored).

Automated: unit tests cover the window registry and window-state persistence/validation. `bun run typecheck`, Biome (`bun run lint`), and the touched test suites are green.

Two windows of the same app on different orgs (org switcher top-left differs per window); after quit + relaunch both reopen on their saved orgs. Happy to attach the captures — let me know if you'd like them inline.

## Additional Notes

**Keyboard shortcut.** New Window is on **`⇧⌘N`** (macOS / VS Code convention), and **`Ctrl+Alt+N`** on Windows/Linux since `Ctrl+Shift+N` is already New Workspace there. It's a native menu accelerator, so it fires regardless of which pane has focus. `⇧⌘N` is still reserved on paper by `QUICK_CREATE_WORKSPACE` in `registry.ts`, but that entry has no handler anywhere in the renderer (single reference, no `useHotkey`), so nothing is lost. Happy to move New Window to a different chord, or clean up the dormant registry entry, if you'd rather Quick Create keep it.

This is a focused slice — happy to adjust scope/shape per your direction on #4018 (e.g. folding per-window org into the larger "project groups" design). "Allow edits from maintainers" is enabled.

Per CONTRIBUTING, this was discussed first via #4018, and is submitted fork → PR following the Clean Code / boy-scout style guidance.

## Release Notes

* **New Features**
  * Multi-window support: each window keeps its own active organization context.
  * Added a **“New Window”** option to the **File** menu.
  * Window titles now reflect the active organization name.

* **Improvements**
  * Deep links and notifications now target the focused (or last-used) window.
  * Window positions and settings are restored more reliably, including across multiple windows.

* **Bug Fixes**
  * Organization selection and UI checks are now properly scoped per window.


ghost:
Ready to review this PR? Stage has broken it down into 9 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Establish tRPC context for IPC calls](https://stagereview.app/superset-sh/superset/pull/5337/chapters/1) |
| 2 | [Implement main process window registry](https://stagereview.app/superset-sh/superset/pull/5337/chapters/2) |
| 3 | [Support multi-window state persistence](https://stagereview.app/superset-sh/superset/pull/5337/chapters/3) |
| 4 | [Refactor window creation and shared services](https://stagereview.app/superset-sh/superset/pull/5337/chapters/4) |
| 5 | [Update app entry and menu for multi-window](https://stagereview.app/superset-sh/superset/pull/5337/chapters/5) |
| 6 | [Migrate window router to context-aware methods](https://stagereview.app/superset-sh/superset/pull/5337/chapters/6) |
| 7 | [Enable per-window organization in renderer](https://stagereview.app/superset-sh/superset/pull/5337/chapters/7) |
| 8 | [Add per-window dynamic titles](https://stagereview.app/superset-sh/superset/pull/5337/chapters/8) |
| 9 | [Scope UI and services to window organization](https://stagereview.app/superset-sh/superset/pull/5337/chapters/9) |

<a href="https://stagereview.app/superset-sh/superset/pull/5337">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit d191f1f8bc38717ca0eebdd2f7b6fafa29510a1a on Jul 28, 2026 11:25am UTC.</sub>

ashbrener:
Thanks for the thorough reviews (CodeRabbit / Greptile / Cubic). All findings addressed across `fb0984850`, `5e1a6509d`, and `64cd2cfdd`:

- **Per-window org header was a mutable global** (`currentOrgId`) used by all collection callbacks — replaced with a **per-org API client** (org pinned via closure), so a deferred/batched write from one org's collection can't be mis-scoped after a switch. Also removes the optimistic-header-without-rollback path. *(CodeRabbit/Greptile/Cubic)*
- **Registry seed was a fragile one-shot** — replaced with a **declarative sync effect** that re-asserts `setActiveOrg` whenever the window's org changes, so the registry (window title, restore, openNew) always reflects the displayed org and self-corrects a transient IPC failure on the next change/launch. *(Cubic seed-desync)*
- **`resolveWindow` fallback to `getWindow()`** removed — a null sender is now a no-op, honoring the `TrpcContext` "no window" contract. *(all three)*
- **`createTrpcContext` webview trust boundary** — top-frame check (`window.webContents === event.sender`), so a `<webview>` guest can't inherit the host window's org. *(Cubic P1)*
- **`persistOpenWindows` skipped empty writes** → now writes unconditionally, so closing the last window clears stale restore state. *(all three)*
- **`window.openNew` org race** — the caller's org is captured in the procedure and carried through the event; the handler is error-handled. *(Greptile/Cubic)*
- **`notificationsEmitter.removeAllListeners()`** → targeted `.off()` for the agent-lifecycle listener. *(Greptile)*
- **New Window overlap** — additional windows open fresh/centered instead of exactly atop the last. *(Greptile)*
- **`SUPERSET-MULTIWORKSPACE-HANDOFF.md`** and the **`apps/desktop/plans/…` ExecPlan** removed — both were working artifacts; the design + decisions live in the PR description and commit history. *(Greptile/Cubic)*

Re-verified at runtime after every change: two windows still resolve to distinct orgs ("Ash 2 — Superset" / "Local Admin's Team — Superset") and restore-on-relaunch still works. Full CI-equivalent is green locally — `sherif`, `bun run typecheck` (28/28), `bun run test` (11/11 packages), and Biome — though the CI workflow itself is currently `action_required` (GitHub gates Actions on fork PRs until a maintainer approves the run).

ashbrener:
Also addressed CodeRabbit's follow-up (`setActiveOrg` now returns `{ success: false }` when there's no sender window, consistent with `minimize`/`maximize`/`close`) in `ce47bb010`. typecheck + Biome green.

ashbrener:
@saddlepaddle (cc @AviPeltz) — would really value your eyes on this when you get a chance.

In my view this is a **material** enhancement that's worth prioritising: it's the core "work across multiple organizations in parallel" capability, which isn't possible at all today. It closes the gap behind #4018 and directly implements #4910 ("Open Organization in new window") — so there's real, repeated demand for it.

It's deliberately a **small, low-risk, self-contained** slice (v2-only, focused diff), fully tested, and **all three review bots (CodeRabbit / Greptile / cubic) are green** with every finding addressed. The CI workflow is currently `action_required` because it's a first-time fork PR — it needs a maintainer to approve the run (locally `sherif`, `typecheck`, the full `bun test`, and Biome all pass).

If the direction looks right I'd love to get it in soon; if you'd rather shape it differently (e.g. fold it into the larger project-groups design), I'm very happy to adapt. Thanks for taking a look 🙏

AviPeltz:
> @saddlepaddle (cc @AviPeltz) — would really value your eyes on this when you get a chance.
> 
> In my view this is a **material** enhancement that's worth prioritising: it's the core "work across multiple organizations in parallel" capability, which isn't possible at all today. It closes the gap behind #4018 and directly implements #4910 ("Open Organization in new window") — so there's real, repeated demand for it.
> 
> It's deliberately a **small, low-risk, self-contained** slice (v2-only, focused diff), fully tested, and **all three review bots (CodeRabbit / Greptile / cubic) are green** with every finding addressed. The CI workflow is currently `action_required` because it's a first-time fork PR — it needs a maintainer to approve the run (locally `sherif`, `typecheck`, the full `bun test`, and Biome all pass).
> 
> If the direction looks right I'd love to get it in soon; if you'd rather shape it differently (e.g. fold it into the larger project-groups design), I'm very happy to adapt. Thanks for taking a look 🙏

going through and reviewing now :)

ashbrener:
Thank you @AviPeltz . Please let me know if any changes required.

ashbrener:
@AviPeltz one thing I'd like your call on while you're in here: **giving "New Window" the `⇧⌘N` keyboard shortcut.**

Right now New Window is menu-only (File → New Window) — I deliberately left the shortcut out to avoid touching the hotkey registry without a maintainer's OK. But I think it's worth adding, and `⇧⌘N` is the natural fit:

- **It's the platform standard.** `⇧⌘N` is "New Window" in VS Code and the macOS convention generally. `⌘N` here is already **New Workspace**, so `⇧⌘N` is the expected companion — zero relearning for anyone coming from VS Code.
- **The chord is currently unused.** `⇧⌘N` is reserved in `registry.ts` by `QUICK_CREATE_WORKSPACE`, but that entry has **no handler anywhere** in the renderer (single reference, no `useHotkey`) — so the shortcut is a no-op today. Reassigning it to New Window costs no existing behavior.
- **Window-open belongs on a global accelerator.** A native menu accelerator fires regardless of focus (including when a terminal pane is focused), and it makes the feature discoverable rather than buried in the File menu.
- **Low-risk and reversible** — it's a one-line menu accelerator (`⇧⌘N` on macOS; `Ctrl+Alt+N` on Windows/Linux, since `Ctrl+Shift+N` is New Workspace there) plus repointing the dormant `QUICK_CREATE_WORKSPACE` reservation.

If you're good with reassigning the unused `⇧⌘N`, I'll add it to this PR right away. If you'd prefer to relocate Quick Create to a specific chord or pick a different key for New Window, just say the word and I'll match it.

AviPeltz:
@ashbrener i am wondering if windows should just be scoped on orgs, or projects more similar to vscode. Internally we had also been thinking about the UX of having a workspace popout window. It is a bit of a strange experience having two windows open to the same workspace. Curious your thoughts

ashbrener:
Good question. I think it's both, they're not really competing.

What I need on a multi-monitor setup is side-by-side worktrees within one org. The main reason is agents: I want to watch and steer two agents running in two different worktrees at once, one per screen, instead of tabbing back and forth. Same for diffing a feature branch against main, or keeping a docs repo open next to the app. Two orgs at once is just a bonus.

What this PR adds is the plumbing for that: real separate windows you can move across monitors, each remembering what it's showing and restoring after a restart. I pinned them to an org first since that's what #4018 / #4910 asked for, but the same plumbing works if a window is pinned to a single worktree, which is your "workspace popout." So org-per-window and worktree-per-window are the same feature, just pinned at a different level.

Strongly agree on the popout: right-click a workspace, "Open in new window," and that window stays pinned to that worktree. That's the entry point I'd want too.

On two windows showing the same workspace: agreed, that shouldn't happen by accident. Easy fix, a new window opens on the dashboard instead of cloning the current view.

I'm happy to keep this PR as the foundation and add the right-click popout and clean-open on top, here or in a follow-up. What's your gut?

ashbrener:
Hey @AviPeltz, when you get a sec I'd love your feeling on the direction from my last comment: keep this PR as the multi-window foundation and add the right-click "Open in new window" popout on top (here or as a follow-up)? Happy to build it whichever way you prefer so we can get it moving. Thanks!

ashbrener:
Rebased on latest main and resolved the conflicts, green again (typecheck, lint, tests). @AviPeltz no rush, but whenever you've got a steer on direction I'm happy to build the right-click "Open in new window" popout on top, or keep this as the foundation and do that as a follow-up.

ashbrener:
Found a bug while using a build of this branch. Open a second window, switch it to another org, and it showed the first org's projects, or none at all.

Only `CollectionsProvider` knew about the window's org. The host data path still read the org from the shared session, so every window used whichever org the session was holding. `aaad021` fixes three places:

- `LocalHostServiceProvider` now takes the org from `CollectionsProvider`, so each window starts its host service for the org it is showing.
- `useHostTargetUrl` takes it from the host service context, so relay routing points at the window's own org.
- `useHostProjects` asks the hosts for projects using the window's org.

The same commit adds two menu bits. `role: "windowMenu"` on macOS, so open windows are listed and you can switch between them. Without it there is no switcher. And a shortcut for New Window: `⇧⌘N` on macOS, `Ctrl+Alt+N` on Windows and Linux, since `Ctrl+Shift+N` is already New Workspace.

@AviPeltz this is the `⇧⌘N` question from earlier. It is one line in the menu, easy to move if you would rather keep that key for Quick Create. I've updated the description, it still said there was no shortcut.

A few other places still read the org from the session (paywall, PostHog, settings pages). They don't affect projects or workspaces, so I can do those next.

`bun run typecheck` and Biome pass locally.

ashbrener:
Rebased on main after #6328 and worked through the review.

**The three red findings are fixed.**

Empty dashboard in a switched window — the root cause was `useAccessibleV2Workspaces` filtering every row against the session's org while the rows come from the window's host service, so a switched window dropped all of them. There's now one sanctioned reader, `useActiveOrganizationId()`, and every consumer that scopes data or writes uses it: the workspace list, the new-workspace flow and its device picker, `CreateTaskDialog`, `usePaywall`, `useResourceSnapshot`, `useKnownHosts`, `useWorkspaceCreates`, and the settings screens. `BillingOverview` was the worst of them — it decided owner-only actions from the session's organization while billing the window's; membership now comes from the window-scoped member list.

Stuck on a stale org — creating or leaving an organization now moves the window as well as the session, and the provider only adopts a registry org that's still in your membership list. Leaving an org can no longer pin a window to one you're not in.

A test pins the readers that are legitimately account-wide (analytics identity, the v1 migration, the layout gate) with a reason for each, so the next session read has to be argued for rather than added by habit.

**Tab and pane clobbering is fixed too.** Every renderer was writing its whole tab state into the one `appState.data.tabsState` record, so two windows overwrote each other and both restored the same layout; pane ids collided in `BrowserManager` for the same reason. Windows now carry a key that survives relaunch, and `uiState.tabs` and the browser router are keyed by it. The first restored window inherits the existing single-window record so current tabs survive the upgrade, and per-window state is pruned to the restorable window set.

One note on #6328: `collections.ts` no longer holds the write callbacks the org-pinned client existed to protect, so that's gone. The window's org now rides as the organization header on `cloudTrpcClient`, read per request. Module state is per renderer and every window is its own renderer, so it's per-window by construction — and `dropCloudQueriesForOrgSwitch` already handles the cache side.

Also: the earlier CodeRabbit and cubic findings on this PR are all addressed — the sender-window fallback, the empty-window-list persistence gap, the mutable `currentOrgId`, and the `<webview>` trust boundary.

ashbrener:
Built the tab/pane fix rather than waiting on a decision — it's a real data-loss bug for anyone running two windows, and it only exists because this PR enables them.

Windows now carry a key that survives relaunch (Electron's window id is per-process and reused, so it can't key anything persisted). It's minted at creation, held in the registry, and written next to the window's bounds and org.

- `uiState.tabs` reads and writes `appState.tabsStateByWindow[key]`, resolved from `ctx.senderWindow`.
- The first restored window inherits the existing single-window `tabsState`, so a current user's tabs survive the upgrade; windows opened after that start their own.
- The browser router namespaces pane ids by the calling window's key, which closes the `BrowserManager` collision.
- Per-window state is pruned to the restorable window set at the same place that set is written, so it can't grow unbounded.

Kitenite:
Brought this current and gave it a full pass — thanks for keeping it alive and rebasing it so many times, @ashbrener. Really solid work, and the per-window plumbing held up everywhere I pushed on it.

**Updated the branch**
- Merged latest `main` (it was 165 behind) and resolved the four conflicts.
- Fixed one real bug the merge surfaced: the API's CORS preflight never allowed `x-superset-organization-id`, so an org-scoped cloud request from a *switched* window died with `Failed to fetch` before it reached tRPC. Added the header to `Access-Control-Allow-Headers` in `apps/api/src/proxy.ts`. (Worth noting for rollout: that's an API change the desktop depends on, so API ships before/with the desktop build.)
- One merge decision on your code to flag: resolving the browser-router conflict, I took `main`'s `workspaceId`-based pane scoping and dropped the per-window pane-key prefix. Had to — the browser bridge addresses panes by bare `paneId`, so a window-key prefix breaks external/agent access. It's safe because pane ids are globally unique (`pane-${Date.now()}-${rand}`) and tab records are now per-window, so nothing collides.
- Converted four session-org readers that landed on `main` since you last rebased (`useCloudWorkspaces`, `useWorkspaceHostTarget`, `useBranchContext`, `usePaymentFailedCard`) over to `useActiveOrganizationId()` so they respect the per-window org. Owner detection in the payment card now goes through `organization.listMembers`, matching `BillingOverview`.

**Verified end-to-end** by driving the running desktop build over CDP (real menu clicks + input):
- Two windows on different orgs at once — switching one leaves the other put, confirmed at the title, the main-process registry, *and* the server response (member lists came back correctly scoped per window: 3 vs 1 from one session).
- Two windows on the **same** org stay independent — live terminals in one, empty in the other, no clobber.
- Quit + relaunch restores both windows on their saved orgs.

**On the open findings:** all four unresolved bot threads are already addressed in your later commits (the two Devin 🔴s — empty-list-on-switch and tab clobber — are fixed by the `useActiveOrganizationId` and `tabsStateByWindow` work; both Greptile items are fixed too: new windows now open centered via `wasEmpty`, and the emitter teardown uses targeted `.off()`). Marking them resolved.

Green on typecheck, lint, and the full desktop suite (2788). Merging once CI is clean.

Kitenite:
Merged 🎉 Thank you @ashbrener — this was a genuinely well-built, well-maintained PR, and your persistence through all the rebases is what got it over the line. The per-window architecture is clean and held up under everything I threw at it. Two small follow-ups worth a look when you have a moment (both non-blocking, filing separately if you'd rather): the restored route isn't org-scoped yet (a restored window can land on a workspace its org can't see), and same-org windows share a title in the macOS Window menu. Appreciate the contribution!

ashbrener:
Thanks @Kitenite — and good catch on the CORS preflight. I'd only ever driven a switched window against a local API, so `x-superset-organization-id` never had to survive a preflight; that would have shipped broken.

Your call on the pane scoping is right and mine was wrong. I namespaced by window key without checking the browser bridge, which addresses panes by bare `paneId` — `workspaceId` scoping does the same job without breaking external access.

Both follow-ups are in #6775.

The restored-route one had a deeper cause than it looked: router history is a single `router-history` key in localStorage, which is shared across windows, so every window restored whatever route the last one navigated to — landing in another org was the visible symptom. It's keyed per window now, with a boot sweep so closed windows don't strand records.

For the title, several windows on one org is normal, so it now leads with the workspace and keeps the org behind it.

PR #5406, opened 2026-07-01
fix(billing): read invoices from the org's canonical Stripe customer

saddlepaddle:
## Problem

Free-tier users couldn't see invoices in the desktop app (Settings → Billing) — and, it turns out, **neither could some paying orgs**. The `billing.invoices` endpoint sourced the Stripe customer id from `subscriptions.stripeCustomerId`, which is unreliable.

## Root cause (verified against live Stripe)

There were two columns holding "the org's Stripe customer":
- `organizations.stripeCustomerId` — created per org, and what the rest of the billing code already keys off (dunning in `onEvent`, cancel-on-delete, customer email/name sync).
- `subscriptions.stripeCustomerId` — written by the better-auth stripe plugin, and **frequently wrong**: for many orgs it points at a near-sequential duplicate customer that was created during checkout but never actually billed (the real subscription + invoices live on the org customer). In at least one case it points at a customer belonging to a *different* org.

I confirmed this directly in Stripe: for every divergent org with a live subscription, the actual `subscription.customer` equals `organizations.stripeCustomerId`, and the invoices live there — not on `subscriptions.stripeCustomerId`. So the endpoint was querying empty/duplicate customers and returning `[]`.

## Fix

Read `organizations.stripeCustomerId` — the authoritative billing customer — in both `invoices` and `requireOwnerWithCustomer` (details/portal). One canonical lookup instead of two, and the corrupt column is no longer read. Union logic removed (net −39 lines).

**No data migration required.** `organizations.stripeCustomerId` was already correct for every paying org (verified: zero orgs have a live subscription with a null/mismatched org customer).

### Separate billing-data issues found (not addressed here — need a human, not code)
- One org has **two active pro subscriptions on two customers** (genuine double-billing) — cancel the duplicate + refund.
- One phantom org carries a **bogus active-enterprise subscription row** (`stripe_subscription_id = null`) duplicating another org's real enterprise sub — should be deleted so it stops showing as enterprise in `activePlan`.

## Testing

Verified the endpoint against live Stripe for the affected orgs (subscription + paid invoices resolve on `organizations.stripeCustomerId`), and simulated end-to-end on the dev branch for a free-tier org with seeded paid invoices. `bun run lint` ✅, `bun run typecheck --filter=@superset/trpc` ✅.

### DIFF
--- a/packages/trpc/src/router/billing/billing.ts
+++ b/packages/trpc/src/router/billing/billing.ts
@@ -1,9 +1,9 @@
 import { stripeClient } from "@superset/auth/stripe";
 import { db } from "@superset/db/client";
-import { members, subscriptions } from "@superset/db/schema";
+import { members, organizations, subscriptions } from "@superset/db/schema";
 import { ACTIVE_SUBSCRIPTION_STATUSES } from "@superset/shared/billing";
 import { TRPCError, type TRPCRouterRecord } from "@trpc/server";
-import { and, desc, eq, inArray, isNotNull } from "drizzle-orm";
+import { and, desc, eq, inArray } from "drizzle-orm";
 import type Stripe from "stripe";
 import { z } from "zod";
 import { env } from "../../env";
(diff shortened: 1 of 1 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Update imports for organization-based billing lookups](https://stagereview.app/superset-sh/superset/pull/5406/chapters/1) |
| 2 | [Unify customer lookup in billing authorization](https://stagereview.app/superset-sh/superset/pull/5406/chapters/2) |
| 3 | [Simplify invoice retrieval using canonical customer ID](https://stagereview.app/superset-sh/superset/pull/5406/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5406">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 8669f09c0709f77c0c742395b5d2b81fc7040013 on Jul 1, 2026 6:23pm UTC.</sub>

PR #5429, opened 2026-07-03
fix: use detached:false in DaemonSupervisor.spawn() to preserve Aqua bootstrap namespace

### LINKED ISSUE #5423: v1.12.5 regression: pty-daemon spawned detached breaks GUI subprocess launch (Chromium/Playwright/Lighthouse) from terminal shells on macOS
## Summary

Any GUI subprocess launched from a Superset terminal shell (e.g. Chromium via Playwright, Lighthouse CI, headful browser tests) crashes instantly on macOS with:

```
bootstrap_check_in ... unknown error code (141) → SIGABRT
```

The same command works fine in Terminal.app. Plain CLI work is unaffected.

This affects both interactive terminal use and agents running inside Superset shells — any workflow that tries to launch a GUI/browser process (Lighthouse audits, Playwright e2e, headful Chrome for debugging, etc.) silently dies.

---

## Environment

- **Superset**: 1.12.5 (auto-updated ~Jun 10 2026)
- **macOS**: 26.5.1
- **Regression**: worked fine with prior Superset version on same OS install

---

## Root Cause (source-confirmed)

The April 2026 pty-daemon → host-service migration (referenced in the file header of `DaemonSupervisor.ts`) shipped in 1.12.5. The supervisor now spawns the pty-daemon with `detached: true` in production:

**`packages/host-service/src/daemon/DaemonSupervisor.ts`, method `spawn()`:**

```typescript
child = childProcess.spawn(command, commandArgs, {
    detached: !isDev,   // ← true in production
    stdio,
    env: childEnv,
    windowsHide: true,
});
```

`detached: true` on POSIX calls `POSIX_SPAWN_SETSID`, creating a new process session. On macOS, this detaches the process from the Aqua launchd domain:

```
# Inside a Superset shell:
launchctl managername         → "Could not get manager name"   (Terminal.app: "Aqua")
launchctl print gui/$(id -u)  → domain NOT reachable
TERM_PROGRAM / SECURITYSESSIONID / __CFBundleIdentifier → all unset
```

Chromium's crashpad and GPU helper processes call `bootstrap_check_in` to register with the Aqua session bootstrap server. Without a reachable session, this fails with Mach error 141 and the process aborts.

**Why this only broke in 1.12.5:** Before the migration, the pty-daemon was spawned directly from Electron's main process (`apps/desktop/src/main/lib/pty-daemon-coordinator.ts`). Electron IS a macOS GUI app registered in the Aqua session. Even with `setsid()`, the daemon inherited Electron's Mach bootstrap port, which connected to the Aqua session's bootstrap server. After the migration, the daemon is spawned by `host-service` (running as `ELECTRON_RUN_AS_NODE=1`), and `setsid()` then fully removes the Aqua bootstrap context.

**Repro** (run inside a Superset terminal, compare to Terminal.app):
```bash
node -e 'require("playwright").chromium.launch({headless:false}).then(b=>b.close())'
# Superset: SIGABRT (exit code 141)
# Terminal.app: exits cleanly
```

---

## Proposed Fix

`detached: true` is not actually needed for daemon persistence. The code already calls:

```typescript
if (!isDev) child.unref();
```

`child.unref()` is sufficient — Node.js won't block on the child, and on POSIX the OS reparents unref'd orphans to launchd when the parent exits. The daemon's socket + manifest file is what enables re-adoption across host-service restarts, not process group isolation.

**Option A (simplest fix):** Change `detached: !isDev` → `detached: false` in `DaemonSupervisor.spawn()`. Keeps the daemon in the Aqua session; persistence still works via `child.unref()` + socket adoption.

**Option B:** Add an opt-out setting (e.g. `SUPERSET_PTY_DAEMON_DETACHED=false` or a UI toggle under Settings → Terminal) for users who need GUI subprocess support, while keeping `detached: true` as default.

Option A is the right default — `setsid()` solves a problem (SIGHUP from terminal close) that doesn't apply to a daemon spawned from a GUI app. Option B is a useful escape hatch regardless.

ashvds:
## Problem

When `detached: true` is passed to `childProcess.spawn()` in `DaemonSupervisor.ts`, Node.js calls `setsid()`, which creates a new process session and disconnects the process from the **macOS Aqua launchd bootstrap namespace**. This causes any GUI subprocess (Chromium via Playwright/Lighthouse) to fail with `SIGABRT (bootstrap_check_in error 141)`.

Confirmed via `launchctl managername` returning `"Could not get manager name"` inside Superset terminal shells.

## Root cause

Introduced in the April 2026 migration of the daemon supervisor from the Electron main process into `packages/host-service`. Previously spawned from Electron (Aqua session). Now spawned from `host-service` (ELECTRON_RUN_AS_NODE=1), and the additional `setsid()` from `detached: true` fully removes the Aqua bootstrap context.

The `detached: true` was intentional for PTY survival, but `child.unref()` (already called below) is sufficient — the daemon is reparented to launchd without needing a new session. `setsid()` is not needed and actively harmful here.

## Change

`detached: !isDev` → `detached: false` (with explanatory comment)

Closes superset-sh/superset#5423

---

<a href="https://stagereview.app/superset-sh/superset/pull/5429">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS daemon startup to ensure GUI-related subprocesses launch reliably under the system’s process launcher.
  * Refined process/session behavior so PTY and terminal interactions remain stable, without relying on macOS background detachment.

### DIFF
--- a/packages/host-service/src/daemon/DaemonSupervisor.ts
+++ b/packages/host-service/src/daemon/DaemonSupervisor.ts
@@ -965,7 +965,7 @@ export class DaemonSupervisor {
 				// namespace and causing GUI subprocesses (Chromium/Playwright/Lighthouse) to
 				// abort with SIGABRT (bootstrap_check_in error 141). On other platforms the
 				// original prod/dev split is preserved. See: https://github.com/superset-sh/superset/issues/5423
-				detached: process.platform === 'darwin' ? false : !isDev,
+				detached: process.platform === "darwin" ? false : !isDev,
 				stdio,
 				env: childEnv,
 				windowsHide: true,
--- a/packages/host-service/src/daemon/DaemonSupervisor.ts
+++ b/packages/host-service/src/daemon/DaemonSupervisor.ts
@@ -959,7 +959,12 @@ export class DaemonSupervisor {
 						`--socket=${socketPath}`,
 					];
 			child = childProcess.spawn(command, commandArgs, {
-				detached: !isDev,
+				// macOS: do NOT use detached:true (setsid) here. It detaches the daemon from
+				// the Aqua launchd bootstrap namespace, breaking GUI subprocess launch
+				// (Chromium, Playwright, Lighthouse) with SIGABRT (bootstrap_check_in error 141).
+				// child.unref() below is sufficient for PTY survival across Electron quit+relaunch.
+				// See: https://github.com/superset-sh/superset/issues/5423
+				detached: false,
 				stdio,
 				env: childEnv,
 				windowsHide: true,
--- a/packages/host-service/src/daemon/DaemonSupervisor.ts
+++ b/packages/host-service/src/daemon/DaemonSupervisor.ts
@@ -939,9 +939,11 @@ export class DaemonSupervisor {
 
 		let child: ReturnType<typeof childProcess.spawn>;
 		try {
-			// Prod: detached so PTYs survive host-service restarts via socket
-			// adoption. Dev: attached as defense-in-depth in case serve.ts's
-			// dev shutdown doesn't fire (e.g. host-service crash).
+			// PTY survival: child.unref() (below) is sufficient on all platforms —
+			// the OS reparents the daemon to launchd/init on parent exit, preserving
+			// open PTYs without a new process session. On macOS, detached:true calls
+			// setsid() which severs the Aqua bootstrap namespace (see issue #5423).
+			// Dev: keep attached as defense-in-depth (serve.ts shutdown may not fire).
 			// Raise RLIMIT_NOFILE before exec: macOS's 256 soft default starves a
 			// daemon hosting many worktrees' PTYs and surfaces as node-pty
 			// "posix_spawnp failed" (EMFILE). The raised limit is inherited by
@@ -959,12 +961,11 @@ export class DaemonSupervisor {
 						`--socket=${socketPath}`,
 					];
 			child = childProcess.spawn(command, commandArgs, {
-				// macOS: do NOT use detached:true (setsid) here. It detaches the daemon from
-				// the Aqua launchd bootstrap namespace, breaking GUI subprocess launch
-				// (Chromium, Playwright, Lighthouse) with SIGABRT (bootstrap_check_in error 141).
-				// child.unref() below is sufficient for PTY survival across Electron quit+relaunch.
-				// See: https://github.com/superset-sh/superset/issues/5423
-				detached: false,
+				// macOS: detached:true triggers setsid(), severing the Aqua launchd bootstrap
+				// namespace and causing GUI subprocesses (Chromium/Playwright/Lighthouse) to
+				// abort with SIGABRT (bootstrap_check_in error 141). On other platforms the
+				// original prod/dev split is preserved. See: https://github.com/superset-sh/superset/issues/5423
+				detached: process.platform === 'darwin' ? false : !isDev,
 				stdio,
 				env: childEnv,
 				windowsHide: true,


ghost:

| | Title |
|---|-------|
| 1 | [Preserve macOS Aqua bootstrap namespace](https://stagereview.app/superset-sh/superset/pull/5429/chapters/1) |

<a href="https://stagereview.app/superset-sh/superset/pull/5429">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 88f9cbe14ee37d891523ace3b7aa583adba50c29 on Jul 3, 2026 6:55pm UTC.</sub>

AviPeltz:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance. The platform-aware `detached` condition is correct for all platform/env combinations, and daemon survival on macOS still holds via `child.unref()` plus launchd reparenting; the kill path (`terminateProcessTreeAndGroups`) still terminates the daemon by PID even when it shares host-service's process group.

One non-blocking note: the stdio comment a few lines above the spawn call still says the rotating log file exists "so the detached daemon survives host-service teardown" — after this change the daemon is no longer detached on macOS, so consider dropping the word "detached" there to match the updated rationale below it.

https://github.com/superset-sh/superset/blob/280135b77dfd653325ffab4ca3e7a8817976128b/packages/host-service/src/daemon/DaemonSupervisor.ts#L914-L919

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Kitenite:
@ashvds detach is how we're able to keep the terminal processes alive across restart. Perhaps instead of detach false we could pass the necessary context to make this work?

One more debugging to make sure it's not just a timing thing is does it work after restarting the host service + daemon? 
Warning: This will kill your existing sessions. 

Screenshot 2026-07-02 at 10 39 49 PM

Kitenite:
Following up on my comment above with some testing. I can't reproduce the premise on my machine:

- `setsid()` doesn't sever the bootstrap namespace. Spawning a child from node with `detached: true` in an Aqua shell, the child still reports `Aqua` from `launchctl managername`.
- Every PTY shell is already a session leader anyway (node-pty calls `setsid()` to acquire the controlling terminal). If setsid severed the namespace, no Superset terminal would ever report `Aqua`.
- Real prod topology: my current terminal's ancestry is `zsh <- pty-daemon (ppid=1, pgid=pid)`, i.e. a daemon spawned `detached: true` that already survived a host-service restart. It reports `Aqua`, and Playwright's Chromium launches headless from it cleanly, no `bootstrap_check_in` abort.
- The regression risk is real though: with `detached: false` the daemon shares Electron's process group, and a group-directed SIGTERM kills the daemon and every PTY (verified). `detached: true` made that impossible.

So the "Could not get manager name" symptom is real but setsid isn't the cause. More likely the daemon is holding a dead bootstrap port (survived a logout/login) or was spawned under a non-Aqua ancestor (CLI/SSH-launched host-service). `detached: false` doesn't fix either case, the daemon still outlives restarts via `unref()` and keeps whatever port it inherited.

Can you share a repro where `managername` fails under `detached: true` and passes under `detached: false` in the same conditions? If the stale-namespace theory is right, the fix is re-acquiring the namespace at spawn time (`launchctl asuser`) rather than dropping detach.

(Also `bun run lint` currently fails on the single-quoted `'darwin'`.)

ashvds:
@Kitenite , Closing as the latest release fixed the issue.

PR #5437, opened 2026-07-03
revert(ports): roll back reaper port-scan sync and kill-path fixes (#5430, #5325, #5339)

Kitenite:
## Summary
- Clean reverts of the three behavioral port commits from the last two weeks: #5430 (recover v2 ports after host-service restart), #5325 (detect ports for terminals without an attached renderer), #5339 (treat untracked port as already-closed on kill).
- Kept as a **draft safety net only** — root-causing showed these PRs are not the cause of the "ports missing until workspace selected after update" bug, and reverting them changes nothing in the shipped desktop app.

## Why / Context
Ports were not appearing after an app update until the workspace was selected. Investigation traced the real cause: `startTerminalReaper` is only wired into `packages/host-service/src/serve.ts` (standalone/CLI entry). The desktop app spawns host-service via `apps/desktop/src/main/host-service/index.ts`, which never starts the reaper — so the #5325/#5430 reconcile logic never executes in the packaged app. The actual fix is a wiring change in the desktop entry (separate PR).

The UI port PRs (#5350, #5392, #5426) are deliberately not reverted: they have no update-time behavior, and #5392's workspace-item components now host the unrelated running-agents feature (#5414), which a revert would rip out.

## Testing
- `bun test` in `packages/host-service` (729 pass) and `packages/port-scanner` (49 pass)
- `bun run typecheck --filter=@superset/host-service --filter=@superset/port-scanner`
- `bun run lint` (exit 0)

## Notes
- Do not merge unless the wiring fix proves insufficient; merging restores the pre-#5325 behavior (ports register only on renderer attach), it does not fix the update symptom.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5437">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Rolls back recent port-scanning behavior and kill-path tweaks to restore pre-#5325 behavior. Removes the reaper’s boot-time port-scan sync and reverts `killPort` to error on untracked ports; this does not address the “ports missing until workspace selected after update” issue.

- **Refactors**
  - Removed port-scan warm-up and daemon sync from `startTerminalReaper`; reaper now runs only on the 5-minute interval.
  - Deleted `planPortScanSync` and its tests; dropped `isLiveTerminalSession` and `getRegisteredTerminalIds`.
[...]

PR #5438, opened 2026-07-03
fix(desktop): start the terminal reaper in the desktop host-service entry

Kitenite:
## Summary
- Wire `startTerminalReaper(db)` into the desktop host-service entry point. It was only called from `packages/host-service/src/serve.ts` (the standalone/CLI entry), so the packaged desktop app never ran the terminal reaper at all.
- This is the actual fix for "ports don't show up after an update until the workspace is selected" — and it also turns on the 5-minute orphan reap (#5168), which has likewise never run in the desktop app.

## Why / Context
#5325 added the reaper-driven port-scan reconcile (register daemon sessions no renderer has attached, e.g. after a host-service restart) and #5430 added 2–90s warm-up retries. Both are correct — but `startTerminalReaper` was never invoked by `apps/desktop/src/main/host-service/index.ts`, the entry the desktop app actually spawns. In the shipped app the only `portManager.upsertSession` caller was the renderer terminal-attach path, so ports for surviving sessions stayed invisible until the user opened the workspace.

Confirmed against a production install (v1.13.0): the asar contains the #5430 code, the daemon adopts within seconds, valid `terminal_sessions` rows exist — and the host-service log contains zero reaper lines across every run.

Supersedes the revert stack in #5437 (kept as a draft safety net; merging that would not fix this symptom).

## How It Works
`startTerminalReaper` is exported from the `@superset/host-service` root, and each serving entry (serve.ts and the desktop entry) calls it once in its listen callback — the same per-entry pattern as `installProcessSafetyNet`, `initTerminalBaseEnv`, and `connectRelay`. All reaper logic stays in the package.

Two single-call-site alternatives were tried and rejected during review discussion: starting it inside `createApp` requires a test-only escape hatch (the reaper is a destructive background actor that reaps integration tests' staged daemon sessions mid-test — 5 failures), and a `startHostRuntime` wrapper was pure pass-through indirection.

## Manual QA
End-to-end against the freshly built `dist/main/host-service.js` (run under Electron-as-Node, exactly as the app spawns it):
- Started a scratch pty-daemon, opened one session running `python3 -m http.server 18923`, and disconnected — an alive, unattached daemon session, the post-update state.
- Started the built host-service against a scratch DB with `SUPERSET_PTY_DAEMON_SOCKET` pointed at the scratch daemon, and inserted the session's `terminal_sessions` row (`status: "active"` + workspace id).
- Warm-up sync logged `port-scan sync: registered 1 unattached daemon session(s) for scanning` within seconds — **no renderer ever attached**.
- `ports.getAll` for that workspace returned `{ port: 18923, processName: "Python", ... }`.
- Verified the built bundle contains the `startTerminalReaper(db)` call.
- Re-ran the scratch harness against the restructured bundle (reaper started from `createApp`): same result — `registered 1 unattached daemon session(s)` and `ports.getAll` returning the planted port.

Full-app verification over CDP (`RENDERER_REMOTE_DEBUG_PORT=9223 bun dev`, this branch):
- Planted an unattached daemon session running `python3 -m http.server 18933` (row bound to a workspace that was never selected), no renderer attach.
- The dev app's host-service reaper logged `port-scan sync: registered 1 unattached daemon session(s) for scanning`.
- The sidebar then rendered the `18933` port badge inside that workspace's item while a *different* workspace was selected — confirmed via CDP DOM inspection.
- Dev-mode caveat: the dev daemon is not detached (dies with host-service by design), so the restart leg was covered by the scratch harness above; the dev app covered reaper registration + renderer display with no attach.

## Known Limitations
- The desktop entry (unlike serve.ts) doesn't kick `startDaemonBootstrap` at startup — the daemon boots on demand. The reaper's earliest warm-up (2s) can beat it and log `port-scan warm-up sync failed: pty-daemon is not available` once; later warm-ups (5–90s) and the 5-min reap succeed. Pre-existing ordering, observed once in dev; harmless but noisy.

## Testing
- `bun run typecheck --filter=@superset/host-service --filter=@superset/desktop`
- `bun run lint` (exit 0)
- `bun test src/terminal/reaper` in `packages/host-service` (9 pass)
- `bun turbo run build --filter=@superset/desktop`

## Risks / Rollout
- Risk: the desktop app now also runs the orphan reap it was always meant to run. Rowless alive daemon sessions are killed only after two consecutive 5-minute passes, and `status: "exited"/"disposed"` rows only describe already-dead PTYs, so healthy sessions are unaffected — but this is behavior the desktop app has never exercised in production.
- Rollback: revert this commit; behavior returns to today's (no reaper, ports on attach only).

---

<a href="https://stagereview.app/superset-sh/superset/pull/5438">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Start the terminal reaper from the desktop host-service entry with `startTerminalReaper(db)`. This cleans up orphaned sessions and reconciles ports on boot, fixing ports not showing until a workspace is opened.

- **Bug Fixes**
  - Start the reaper in the desktop entry and export it from `@superset/host-service`; both serving entries run it on boot.
  - Remove `startHostRuntime` and keep reaper startup outside `createApp`.

<sup>Written for commit bcf499dcd6fb1328039e78e2a1afdc7fef767404. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5438?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of desktop terminal sessions during app startup, reducing stuck or orphaned terminals.
  * Enhanced startup reliability for the desktop host service by adding an additional terminal reaping step once the service is ready.


ghost:

| | Title |
|---|-------|
| 1 | [Export startTerminalReaper from host-service package](https://stagereview.app/superset-sh/superset/pull/5438/chapters/1) |
| 2 | [Wire terminal reaper into desktop host-service](https://stagereview.app/superset-sh/superset/pull/5438/chapters/2) |

<a href="https://stagereview.app/superset-sh/superset/pull/5438">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit bcf499dcd6fb1328039e78e2a1afdc7fef767404 on Jul 3, 2026 9:25pm UTC.</sub>

Kitenite:
re: coderabbit's nitpick about the discarded `startTerminalReaper` teardown — intentionally skipping. The reaper's interval and warm-up timers are all `unref()`'d, and this entry's `shutdown()` ends in `process.exit(0)` after a 3s drain, so there's nothing for the dispose to release in practice. `serve.ts` (the pattern this mirrors) also discards it; capturing it only in the desktop entry would make the two call sites asymmetric. If we ever add a real graceful-shutdown path that doesn't exit the process, wiring the teardown in belongs in both entries.

Kitenite:
Restructured per discussion: the reaper now starts inside `createApp` (next to `eventBus.start()`), so both entry points get it without per-entry wiring, and serve.ts's own call is removed — no double-start. Its stop function is wired into `createApp`'s `dispose` chain, which also properly resolves the earlier teardown nitpick. Tests opt out via `createTestHost` (`startTerminalReaper: false`) because the reaper's warm-up timers reach the real pty-daemon through the process-wide daemon-client singleton and were reaping other tests' sessions (5 integration failures with an unconditional start). Full host-service suite: 746 pass / 0 fail; scratch E2E re-verified on the rebuilt bundle.

Kitenite:
Final shape: dropped the `startTerminalReaper: false` test override — production APIs shouldn't carry test escape hatches. Instead the package now separates app construction from process runtime: `createApp` is untouched from main, and a new `startHostRuntime(db)` (`runtime/host-runtime.ts`) owns process-lifetime background work; both serving entries call it once after binding. Tests are byte-identical to main and pass 746/0 with no accommodations. Scratch E2E re-verified on the rebuilt bundle (unattached daemon session registered, port returned by `ports.getAll`).

Kitenite:
Settled on the direct shape: dropped the `startHostRuntime` pass-through (it only delegated). `startTerminalReaper` is exported from the package root and each serving entry calls it once in its listen callback — same pattern as `installProcessSafetyNet` / `connectRelay`, which both entries already invoke individually. Net diff vs main is back to the original two-file change plus the root export.

PR #5443, opened 2026-07-03
fix(host-service): prune terminal-agent bindings on pty exit and at startup

Kitenite:
## Summary
- Terminal-agent bindings (which drive the sidebar's inline agent chips) are now pruned whenever their pty exits, and stale persisted bindings are drained once at host-service startup. Previously, killed/crashed terminals, reaper disposal, and workspace deletion all leaked bindings — and persistence made the leak survive restarts, so the sidebar kept showing agents for terminals that no longer exist.

## Why / Context
Observed while testing #5439's activity strip: agents stayed in the sidebar after their terminal was removed. Root cause is upstream of the UI — the `TerminalAgentStore` only pruned in two places: an agent *exit hook* firing (`Detached`/`exit`/`error` via the notifications route), or the renderer's explicit terminal-dispose tRPC route. Every other end of life leaked:

- agent process killed or crashed (no hook fires)
- terminal exits on its own — `onExit` broadcast `terminal:lifecycle` but never touched the store, so the renderer refetched and got the stale binding right back
- reaper orphan cleanup and `disposeSessionsByWorkspaceId` call `disposeSessionAndWait` directly, bypassing the route's `markTerminalExited`
- bindings are persisted to SQLite and rehydrated on start (`load()` only skips `disposed`, not `exited`), so all of the above survived restarts

## How It Works
Two layers, deliberately minimal:

- **Prune on pty exit.** `EventBus` gains an in-process `onTerminalLifecycle` listener (the existing broadcast only reaches ws clients); `app.ts` registers one that calls `markTerminalExited(terminalId)`. Every in-process teardown path ends at the pty `onExit` handler, so this covers kills, crashes, self-exit, and dispose routes — and the same handler's ws broadcast makes the sidebar refetch the now-pruned list.
- **Startup drain.** `reconcileTerminalAgentBindings` runs once after store hydration and drops bindings whose session row is missing, `exited`/`disposed`, or workspace-less (the reaper's orphan criteria). This clears bindings leaked before this fix and terminals that died while the host-service was down. Supporting tweak: `deleteTerminal` deletes the persisted row even without an in-memory binding, so the drain can purge rows whose sessions were never hydrated.

An earlier revision integrated reconciliation into the terminal reaper's periodic pass (with a daemon-liveness check); it was dropped to keep the maintenance surface small — see the PR comments for the trade-off. This version leaves `reaper.ts` and `serve.ts` untouched, which also avoids conflicting with the pending port-changes revert.

## Manual QA Checklist
- [ ] With an agent bound, kill the terminal process externally (`kill -9` the pty child) → agent chip disappears from the sidebar
- [ ] Close the terminal from the UI → chip disappears
- [ ] Restart host-service with a stale persisted binding (terminal long gone) → chip does not come back

## Testing
- `bun test src/terminal-agents` (16 pass). New tests: persisted-row deletion without an in-memory binding, and a real in-memory-SQLite reconcile test covering live/exited/disposed/workspace-less sessions
- Both new tests verified to fail against a mutated implementation (delete ordering, status filter)
- `bun run typecheck` (host-service) and Biome clean

## Known Limitations
- No periodic sweep: a session row left stale-`active` by a terminal that died while the host-service was down isn't caught until the next restart's drain. Accepted for surface area; the reaper-integrated version (see history) is the escalation path if this shows up in practice.
- A drain-driven prune reaches an already-connected sidebar lazily (its cache invalidates on lifecycle events); in practice the startup drain runs before any renderer connects.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after interruptions: the host now reconciles persisted terminal bindings on startup and marks exited terminals based on terminal lifecycle events.
  * Terminal lifecycle handling is more consistent, including in-process updates and reduced stale terminal state after downtime.
  * Terminal binding listings now exclude defunct entries to prevent returning unusable/stale bindings.
  * Persisted deletions are applied reliably even if a terminal was never loaded into memory.
* **Tests**
  * Added coverage for binding reconciliation and reliable deletion behavior.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/changes/security/git-commands.ts
+++ b/apps/desktop/src/lib/trpc/routers/changes/security/git-commands.ts
@@ -88,21 +88,27 @@ export async function gitSwitchBranch(
 }
 
 /**
- * Checkout (restore) a file path, discarding local changes.
+ * Checkout (restore) file paths in a single git command,
+ * discarding local changes.
  *
- * Uses `git checkout -- <path>` - the `--` is REQUIRED here
- * to indicate path mode (not branch mode).
+ * Uses `git checkout -- <paths...>` - the `--` is REQUIRED here
+ * to indicate path mode (not branch mode). A single command avoids
+ * index.lock races when discarding multiple files.
  */
-export async function gitCheckoutFile(
+export async function gitCheckoutFiles(
 	worktreePath: string,
-	filePath: string,
+	filePaths: string[],
 ): Promise<void> {
+	if (filePaths.length === 0) {
+		throw new Error("filePaths must not be empty");
+	}
 	assertRegisteredWorktree(worktreePath);
-	assertValidGitPath(filePath);
+	for (const filePath of filePaths) {
+		assertValidGitPath(filePath);
+	}
 
 	const git = await getGitWithShellPath(worktreePath);
-	// `--` is correct here - we want path semantics
-	await git.checkout(["--", filePath]);
+	await git.checkout(["--", ...filePaths]);
 }
 
 /**
--- a/apps/desktop/src/lib/trpc/routers/changes/staging.ts
+++ b/apps/desktop/src/lib/trpc/routers/changes/staging.ts
@@ -4,7 +4,7 @@ import { publicProcedure, router } from "../..";
 import { getServiceForRootPath } from "../workspace-fs-service";
 import { getSimpleGitWithShellPath } from "../workspaces/utils/git-client";
 import {
-	gitCheckoutFile,
+	gitCheckoutFiles,
 	gitDiscardAllStaged,
 	gitDiscardAllUnstaged,
 	gitStageAll,
@@ -88,7 +88,20 @@ export const createStagingRouter = () => {
 				}),
 			)
 			.mutation(async ({ input }): Promise<{ success: boolean }> => {
-				await gitCheckoutFile(input.worktreePath, input.filePath);
+				await gitCheckoutFiles(input.worktreePath, [input.filePath]);
+				clearStatusCacheForWorktree(input.worktreePath);
+				return { success: true };
+			}),
+
+		discardFiles: publicProcedure
+			.input(
+				z.object({
+					worktreePath: z.string(),
+					filePaths: z.array(z.string()).min(1),
+				}),
+			)
+			.mutation(async ({ input }): Promise<{ success: boolean }> => {
+				await gitCheckoutFiles(input.worktreePath, input.filePaths);
 				clearStatusCacheForWorktree(input.worktreePath);
 				return { success: true };
 			}),
--- a/apps/desktop/src/main/host-service/index.ts
+++ b/apps/desktop/src/main/host-service/index.ts
@@ -13,6 +13,7 @@ import {
 	LocalGitCredentialProvider,
 	LocalModelProvider,
 	PskHostAuthProvider,
+	startTerminalReaper,
 } from "@superset/host-service";
 import {
 	initTerminalBaseEnv,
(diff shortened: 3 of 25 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Enable in-process terminal lifecycle subscriptions](https://stagereview.app/superset-sh/superset/pull/5443/chapters/1) |
| 2 | [Implement terminal agent binding reconciliation logic](https://stagereview.app/superset-sh/superset/pull/5443/chapters/2) |
| 3 | [Update store to support pruning unhydrated bindings](https://stagereview.app/superset-sh/superset/pull/5443/chapters/3) |
| 4 | [Wire pruning into app startup and lifecycle](https://stagereview.app/superset-sh/superset/pull/5443/chapters/4) |
| 5 | [Add read-time safety filter to API](https://stagereview.app/superset-sh/superset/pull/5443/chapters/5) |

<a href="https://stagereview.app/superset-sh/superset/pull/5443">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 8370648015c9914426262e2143bb90ee80e857a5 on Jul 3, 2026 10:20pm UTC.</sub>

Kitenite:
Reworked per discussion: reconciliation now lives in the terminal reaper's pass (immediate first run + 5-minute cadence) instead of a startup one-shot — commit 3b18e0f02. The reap pass does the row-flagged prune before the daemon sync (so startup draining doesn't depend on daemon reachability) plus a liveness prune against the daemon's live-session list, which closes the stale-`active`-row gap called out earlier. `createApp` exposes the store; `serve.ts` hands it to `startTerminalReaper`. The pty-exit listener is unchanged.

Kitenite:
Simplified per review discussion: dropped the reaper integration from the previous revision (commit 59c50fe3e) — no liveness planner, no `store.listAll`, no `startTerminalReaper` option, no store exposure on `CreateAppResult`. `reaper.ts` and `serve.ts` are back to untouched, so this no longer conflicts with the pending `revert-port-changes` branch. What remains is the minimal pair: prune on pty exit + one-shot startup drain (+181/−5 total, ~40% tests). Trade-off accepted: stale-`active` rows from terminals that died while the host was down wait for the next restart; the reaper version remains in history as the escalation path.

Kitenite:
Two updates: (1) Added the read-time guard we discussed as defense in depth (fdfebd0ce): `listByWorkspace` now excludes bindings matching the shared staleness predicate (`listDefunctBindingTerminalIds`, the same criteria the startup drain deletes by, so the layers can't drift) and **warns when it drops rows** — the filter firing at steady state means a prune path was missed, so the redundancy surfaces bugs instead of masking them. `findActive`/`subscribe` intentionally stay unfiltered; they rely on pruning. (2) Merged origin/main (484a30361), picking up #5438 — the reaper now also runs in the desktop-embedded host-service — and #5445's chip-visibility tweak.

Kitenite:
Putting this on hold as **superseded by a root fix**: the review findings above are all failure modes of this PR's compensating machinery (exit listener, startup drain, read guard), which is a sign the approach is patching around the real defect — bindings owning a duplicate of terminal liveness that `terminal_sessions.status` already tracks (onExit, dispose routes, reaper healing). The root fix — derive binding *visibility* by joining reads against session status, so dead-terminal bindings become unrepresentable — has been folded as **Milestone 0** into the ExecPlan `apps/desktop/plans/20260703-1507-derive-terminal-agent-status-from-host-binding.md` (the `check-agent-binding-decou` workspace), which already applies the same single-source-of-truth philosophy renderer-side. This PR's test harness (`persistence.test.ts`) and defunct-criteria predicate are directly reusable there. Will close once Milestone 0 lands.

PR #5447, opened 2026-07-03
fix(host-service): isolate dev pty-daemons from the packaged app

Kitenite:
## Summary
- Dev instances no longer share (or adopt) the packaged app's pty-daemon. The socket path is now keyed by `sha256(orgId + SUPERSET_HOME_DIR)` for non-default homes, and dev mode skips adoption entirely — dev daemons are attached, die with the app, and never touch prod's terminals.

## Why / Context
Dev isolation was already designed in (`detached: !isDev`, no `unref`, "Dev mode: never adopt — kill stale and spawn fresh") but was defeated by a path mismatch, verified live: daemon **manifests** are per-home (`SUPERSET_HOME_DIR/host/<org>/pty-daemon-manifest.json`; dev homes are worktree-local `superset-dev-data/`), while the **socket** hashed only the org id (`tmpdir + sha256(orgId)`). On dev startup:

1. `killStaleDaemonForDev` reads the *dev home's* manifest → none exists (the packaged app's manifest lives in `~/.superset`) → kills nothing.
2. `tryAdopt` → manifest missing → `tryAdoptFromSocket` on the org-global path → finds the packaged app's live daemon → adopts it.

Consequences observed while debugging the sidebar agent-chip issue: dev terminals running as children of Canary's daemon, both instances' host DBs accumulating diverging session/binding rows for the same shared terminals, both reapers (post-#5438) operating on one substrate, and — worst case — the adopted-daemon auto-update path could hand a *prod* daemon off to a *dev* bundle via fd-handoff.

## How It Works
- **`ptyDaemonSocketPath`** hashes `orgId + ":" + SUPERSET_HOME_DIR` when the home isn't the default `~/.superset`. The default home keeps the exact legacy org-only path, so packaged apps never orphan an existing daemon on update — zero migration. Fixed-length hash keeps the path far under Darwin's 104-byte `sun_path` limit regardless of worktree-home length.
- **Dev skips `tryAdopt` entirely.** Dev daemons spawn attached and die with the app, so a surviving dev daemon is never legitimate — only a leftover from a killed `bun dev` (still cleaned via the home's manifest). Skipping adoption also closes the socket-probe fallback that did the cross-instance re-attach. `SUPERSET_PTY_DAEMON_ADOPT_IN_DEV=1` still restores production-like adoption for handoff testing.

Two dev worktrees of the same org also stop sharing a daemon (each has its own home → own socket).

## Testing
- `bun test packages/host-service/src/daemon` — 50 pass, including new `ptyDaemonSocketPath` cases (legacy path for default home, distinct + stable per home, `sun_path` length bound); the home-keying mutation makes the new test fail
- `bun run test:integration:daemon` (node, real daemon spawn/adopt/handoff) — 18 pass
- `bun run typecheck` (host-service), Biome clean
- Not exercised live: a `bun dev` session side-by-side with the packaged app post-fix. QA: launch both for the same org → two daemons (`superset-ptyd-*.sock` × 2), dev terminals die with `bun dev`, Canary terminals untouched.

## Risks / Rollout
- Risk: dev-only behavior change; the packaged app's socket path and adoption semantics are byte-identical.
- One-time dev effect: the first `bun dev` after this lands spawns a fresh daemon at the new path; any terminals a dev instance previously created inside the *shared* daemon remain owned by the packaged app's daemon (visible in Canary, not in dev). That's the corrected ownership, but it will look like dev "lost" its old terminals once.
- Rollback: revert; paths return to the shared org-only socket.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5447">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon socket path handling so non-default home directories use distinct, stable paths.
  * Preserved existing behavior for the default home directory to avoid unexpected changes.
  * Prevented socket paths from exceeding macOS path length limits in long home-directory setups.
  * In development mode, the app now starts a fresh daemon more reliably and avoids adopting stale daemon instances.

### DIFF
--- a/packages/host-service/src/daemon/DaemonSupervisor.test.ts
+++ b/packages/host-service/src/daemon/DaemonSupervisor.test.ts
@@ -11,6 +11,7 @@
 
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import * as childProcess from "node:child_process";
+import { createHash } from "node:crypto";
 import * as fs from "node:fs";
 import * as net from "node:net";
 import * as os from "node:os";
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Isolate daemon socket paths by home directory](https://stagereview.app/superset-sh/superset/pull/5447/chapters/1) |
| 2 | [Disable daemon adoption in development mode](https://stagereview.app/superset-sh/superset/pull/5447/chapters/2) |
| 3 | [Verify socket path isolation and length limits](https://stagereview.app/superset-sh/superset/pull/5447/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5447">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 3ec8a075bc7e550ab5b9f43f79d8eb1f40abc336 on Jul 3, 2026 10:20pm UTC.</sub>

Issue #4364, opened 2026-05-10
Worktree creation can create a different branch than selected remote branch


Kitenite:
When creating a workspace to continue work locally on an existing remote branch, selecting the remote branch and worktree flow can still create a new local branch instead of a worktree for that existing branch.

A workaround appeared to be pasting the branch name into the top-right branch field as well, but that did not work reliably. Using `claude/auto-balance-logo-sizing-1UFT7` created a different branch with casing changed, which made it look like work was happening on the same branch until opening the PR and seeing it wanted a new PR because the branch was different.

This happened even when the intended branch was selected in the branch dropdown during workspace creation.

Expected behavior: creating a workspace for an existing remote branch should create a worktree for that branch, without creating a different local branch or changing the branch name casing.

hangaoke1:
+1

Kitenite:
Closing as a duplicate of #5610, the consolidated tracker for stale `workspace-creation` issues from the July 2026 stale-issue sweep (no activity since 2026-06-01). If this still applies on a current build, comment on #5610 and we'll reopen.

PR #5455, opened 2026-07-05
fix(host-service): stop workspace↔PR/diff state drifting from GitHub

Kitenite:
## Problem

Users report their workspace's main-branch tracking and PR link drifting from GitHub state — a branch that shows **146 files changed** vs `main`, and a PR that never connects — most often on macOS's case-insensitive filesystem. Three distinct causes, all in the v2 host-service path:

## Root causes & fixes

**1. PR↔branch matching was case-sensitive on the branch.** `upstreamKey`/`headKey` lowercased owner/repo but kept the branch verbatim, so a local `roshvan/mcp-1525-…` never matched a PR head `Roshvan/mcp-1525-…`, and every sync actively re-wrote `pullRequestId = null`. Now the branch is compared case-insensitively. Because GitHub's `head=` REST filter is *itself* case-sensitive server-side (verified against this repo), a per-head lookup for a drifted branch returns nothing — so a shallow **open-PR sweep** (one page, no checks/statuses, once per repo per 60s TTL, only when a per-head lookup missed) fills the gap and matches case-insensitively. A sweep failure marks keys failed so existing links are *kept*, never cleared.

**2. Workspace creation could mint a case-twin branch.** On a case-insensitive filesystem `git rev-parse refs/heads/foo` happily resolves the `refs/heads/Foo` file, so `resolveRef`'s exact-case probe couldn't distinguish an exact match from case drift — and `git worktree add -b` would create a second branch sharing the existing one's loose-ref file (the mechanism that corrupts HEAD underneath the workspace). `resolveRef` now enumerates real refnames via `for-each-ref`: exact matches keep prior precedence (local > remote > tag); a case-insensitive match **adopts the existing branch's canonical casing**, which `planBranchSource`/create propagate.

**3. The Changes panel never fetched its base ref.** It diffs `origin/<base>...HEAD` (3-dot merge-base), but nothing in the status path fetched `origin/<base>` — so after a rebase onto a newer upstream, every replayed upstream commit was counted as a workspace change, ballooning the file count. `getGitStatusSnapshot` now schedules a fire-and-forget fetch of the base's remote branch (keyed by the repo common-dir, TTL'd, in-flight-deduped); GitWatcher picks up the ref change and re-queries.

## Verification

- Unit + integration tests, each **mutation-checked** (revert the fix → test fails).
- End-to-end against real on-disk repos: `resolveRef` adopts canonical casing for local/remote/`origin/`-prefixed drift; diff balloons 4→heals to 1 file after the background fetch.
- Live-GitHub: exact-head finds the PR, drifted-head misses server-side, the sweep matches it, a bogus branch stays unmatched.
- Via **CDP against the running host-service**: `getStatus` balloon (4 files) → heal (1 file), background fetch fires (`FETCH_HEAD` + `origin/main` sha advance confirmed independently), TTL dedup holds on rapid calls.

## Notes / prior art

- Supersedes #4367 and closes the class in #4364 — that PR added a case-insensitive *fallback* after the exact-case probe, which on macOS never runs because the probe wrongly succeeds; this replaces the probes with enumeration so both are covered.
- Mirrors GitHub Desktop's #21320 rename guard (enumerate branches, never trust a single-ref probe) and their list-and-match PR design.
- The repo-wide sweep is deliberately narrow (shallow open-PR page, no checks) to avoid regressing the #4268/#4291 504-on-large-repos fix.
- **Follow-up worth considering:** the base-ref fetch uses a 5-min TTL; a content-keyed dedup on `(headSha, baseSha)` + a short failure-only backoff would remove the arbitrary interval and make the heal immediate after a rebase. Left out of this PR to keep it focused.
- The v1 desktop path (`pr-resolution.ts`) has the same case-sensitivity bug but is left untouched pending the v1 sunset.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5455">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes workspace↔GitHub drift that caused inflated file counts and PR links not attaching. Keeps branch casing and base refs accurate so diffs and links stay correct.

- **Bug Fixes**
  - PR linking: keep `upstreamKey` and per-head cache keys case-sensitive to isolate case-variant branches; add a shallow `state=open` sweep (once per repo per TTL, only when a head lookup misses) that matches case-insensitively to recover drifted branches; cache failures for the TTL and preserve existing links if the sweep fails; tests now use a real, migrated in-memory SQLite DB to verify case-variant isolation across bypass and cached paths.
  - Branch ref resolution: enumerate branches via `for-each-ref` and adopt canonical casing; prevent case-twin creation on case-insensitive filesystems; propagate the canonical name during workspace creation; surface real `for-each-ref` failures instead of degrading to “no branches.”
  - Base-ref freshness: background-fetch the base remote branch during status (TTL, in-flight dedup keyed by repo common-dir, resolve common dir per call, never rejects) so `origin/<base>...HEAD` diffs don’t balloon after a rebase.

<sup>Written for commit 2f719919d4a7f6a9360a8a956670cd69e07ec7b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Workspace and branch reuse now adopts canonical casing from existing local or remote-tracking refs.
  * Added a repo-wide open-PR sweep fallback to recover missed PR lookups when branch names differ only by letter case.
  * Introduced TTL-based background refresh for remote-tracking base refs to keep comparisons up to date.

* **Bug Fixes**
  * Improved Git ref resolution and PR linking for case-drifted branches/tags, including correct “no matches” vs real failures.
  * If the open-PR sweep fails transiently, existing workspace↔PR links are preserved (not cleared).
  * Prevented PR cross-linking between case-variant branch names.

* **Tests**
  * Added/expanded regression coverage for these case-handling behaviors.

### DIFF
--- a/packages/host-service/src/runtime/git/refs.test.ts
+++ b/packages/host-service/src/runtime/git/refs.test.ts
@@ -180,6 +180,29 @@ describe("resolveRef — input shape contract", () => {
 		expect(await resolveRef(git, "head")).toBeNull();
 	});
 
+	// A genuine for-each-ref failure must propagate, not degrade to "no
+	// branches" — masking it would make an existing branch look absent and let
+	// a case-twin be created (the bug this module exists to prevent).
+	test("propagates a for-each-ref failure instead of returning null", async () => {
+		const git = {
+			raw: mock(async (args: string[]) => {
+				if (args[0] === "for-each-ref") {
+					throw new Error("fatal: not a git repository");
+				}
+				throw new Error(`Unexpected raw args: ${args.join(" ")}`);
+			}),
+		} as never;
+		const originalWarn = console.warn;
+		console.warn = () => {};
+		try {
+			await expect(resolveRef(git, "foo")).rejects.toThrow(
+				"not a git repository",
+			);
+		} finally {
+			console.warn = originalWarn;
+		}
+	});
+
 	test("custom remote name probes that remote, not origin", async () => {
 		const git = createMockGit(new Set(["refs/remotes/upstream/foo"]));
 		const r = await resolveRef(git, "foo", { remote: "upstream" });
--- a/packages/host-service/src/runtime/git/refs.test.ts
+++ b/packages/host-service/src/runtime/git/refs.test.ts
@@ -3,12 +3,20 @@ import { asLocalRef, asRemoteRef, resolveRef } from "./refs";
 
 /**
  * Mock git that knows about a fixed set of FULL refnames. Mirrors how
- * `resolveRef` probes (always with `refs/heads/...` / `refs/remotes/.../...`
- * / `refs/tags/...`).
+ * `resolveRef` looks refs up: branch enumeration via `for-each-ref`
+ * (which reports refnames exactly as stored, case-sensitively) and tag
+ * probes via `rev-parse --verify`.
  */
 function createMockGit(existingFullRefs: Set<string>) {
 	return {
 		raw: mock(async (args: string[]) => {
+			if (args[0] === "for-each-ref") {
+				const prefixes = args.filter((arg) => arg.startsWith("refs/"));
+				const lines = [...existingFullRefs].filter((ref) =>
+					prefixes.some((prefix) => ref.startsWith(prefix)),
+				);
+				return lines.length > 0 ? `${lines.join("\n")}\n` : "";
+			}
 			if (args[0] === "rev-parse" && args[1] === "--verify") {
 				const ref = args[2]?.replace("^{commit}", "") ?? "";
 				if (existingFullRefs.has(ref)) {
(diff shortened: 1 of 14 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Adopt canonical branch casing in Git resolution](https://stagereview.app/superset-sh/superset/pull/5455/chapters/1) |
| 2 | [Propagate canonical casing during workspace creation](https://stagereview.app/superset-sh/superset/pull/5455/chapters/2) |
| 3 | [Add open-PR sweep for case-insensitive matching](https://stagereview.app/superset-sh/superset/pull/5455/chapters/3) |
| 4 | [Integrate PR sweep and fix state drifting](https://stagereview.app/superset-sh/superset/pull/5455/chapters/4) |
| 5 | [Implement background base-ref fetching](https://stagereview.app/superset-sh/superset/pull/5455/chapters/5) |
| 6 | [Wire base-ref fetch into Git status](https://stagereview.app/superset-sh/superset/pull/5455/chapters/6) |

<a href="https://stagereview.app/superset-sh/superset/pull/5455">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2f719919d4a7f6a9360a8a956670cd69e07ec7b8 on Jul 5, 2026 8:14am UTC.</sub>

Kitenite:
Thanks for the reviews — addressed in 64f6fc647. Summary of each finding:

**Fixed (valid):**

- **cubic P1 — `upstreamKey` collapses case-variant branches (pull-requests.ts:190).** Correct and the most important catch. On a case-sensitive host, `feature`/`Feature` are distinct branches with distinct PRs; lowercasing the canonical identity key could mis-link a workspace. Reverted `upstreamKey` to case-sensitive and moved case-drift tolerance into the open-PR **sweep fallback only** (keyed by a lowercased copy, so `latestByKey`/link-assignment stay exact).
- **cubic P2 — per-head cache key cross-branch poisoning (pull-requests.ts:908).** Same root cause. Reverted the cache key to preserve branch case.
- **cubic/greptile — unbounded `commonDirByWorktree` + stale-on-path-reuse (base-ref-freshness.ts).** Removed the path→commonDir cache entirely; it's now resolved fresh each call (`rev-parse --git-common-dir` is a near-instant local call, dwarfed by the fetch it precedes). That deletes the only unbounded map and the staleness footgun.
- **cubic/CodeRabbit — bare `catch {}` in `resolveRef` (refs.ts:85/89).** A real `for-each-ref` failure now logs and **propagates** instead of degrading to "no branches" (which would let a case-twin be created). `for-each-ref` exits 0 with empty output on an empty repo, so this only fires on genuine failures; the sole caller is workspace creation, where loud failure is the right outcome.

**Declined (with reasoning):**

- **cubic P3 / greptile — no in-flight dedup on `resolveCommonDir`.** Moot after removing the cache: every call resolves fresh by design, and a dedicated in-flight map for one cheap local `rev-parse` isn't worth the machinery.
- **CodeRabbit — tag lookup is case-insensitive on APFS (refs.ts:182).** True (tags still use `rev-parse`), but far less harmful than the branch case — it only affects the "that's a tag, not a branch" rejection message, never branch creation. Left as-is to keep scope tight; can follow up if desired.

Also added a `base-ref-freshness` unit suite and a `resolveRef` for-each-ref-failure test. All host-service tests, typecheck, and lint green.

One thing I did **not** add: a two-workspace regression test for the P1 collision — the PR-manager test harness is currently single-workspace. The exact-`upstreamKey` design + the per-head exactness test guard it, but a multi-workspace harness test would be a stronger lock if reviewers want it.

PR #5494, opened 2026-07-07
fix(desktop): dismiss stuck tooltips and forward tab-switch hotkeys in in-workspace browser

Kitenite:
Fixes five UI bugs in the in-workspace browser (plus a favicon fallback picked up during verification). Most stem from the browser being a native Electron `<webview>` composited on top of the host page, so it captures pointer/keyboard events the host renderer never sees.

## 1. Popover/tooltip stays stuck over the URL bar
Radix keeps a tooltip open during its "hoverable content" grace period, waiting for more pointer events — but once the cursor crosses onto the native webview the host document stops receiving pointer events, so the close logic never runs and the tooltip hangs over the URL bar.

**Fix:** `disableHoverableContent` on the webview-adjacent tooltips. Since #5870 restyled the preset tooltips into `HotkeyTooltip`, the flag now lives there (its content is always a non-interactive hotkey chip, so there is no downside anywhere it's used), plus the Manage Presets and DevTools tooltips.

## 2. Tab-switch keyboard shortcuts don't work when the browser is focused
With the guest webview focused, keystrokes route to the guest renderer, so host `react-hotkeys-hook` never sees them. `browser-manager.ts` already intercepts via `before-input-event` but only forwarded `Cmd/Ctrl+W` and `+R`.

**Fix:** the renderer computes the (override/layout-aware) chords for the tab-switch hotkeys (`PREV/NEXT_TAB`, `_ALT` variants, `JUMP_TO_TAB_1..9`) and syncs them to the main process via `browser.setForwardableChords`. `before-input-event` suppresses exactly those chords in the guest and forwards them over a `browser.onKeyForward` subscription; `usePersistentWebview` rebuilds the `KeyboardEvent`, re-resolves it through `resolveHotkeyFromEvent`, and replays it on `document` only if it resolves to an allowlisted tab-switch hotkey. Chord canonicalization is shared between main and renderer (`shared/hotkey-chord.ts`, single `assembleChord` helper) and pinned by a parity unit test.

## 3. Browser tab favicon fallback
Chromium guesses `/favicon.ico` for pages that declare no favicon, so `faviconUrl` can point at a 404 and the tab shows a broken image (and pages with no favicon showed no icon at all). Tabs now fall back to a globe icon when the favicon is missing or fails to load, keyed by page + favicon URL so navigation retries.

## 4. Opening a link into an existing browser pane never navigates
Clicking a link (e.g. a URL in a terminal) opens the in-workspace browser — but once a browser pane exists, every subsequent link click replaced the pane in the store without the page ever loading the new URL. Root cause: `usePersistentWebview` captured the pane URL in a ref at first mount; when `openPane` -> `replacePane` swaps in a new pane id on the same component instance, the attach effect re-ran with the stale first-mount URL, so the fresh webview loaded the old page (and the persist echo then reverted the pane data too).

**Fix:** keep the attach-URL ref current on every render. Attach still keys on `paneId` alone, so tab switches and navigation echoes don't re-attach; only a pane replacement picks up the new URL. Verified live: alternating cmd+clicks between two terminal URLs navigate the pane correctly every time (previously only the first-ever link worked).

## 5. Shift+click opens links in a new browser tab
The default URL-link tier map now assigns `shift` -> "Open in new tab" (previously unbound): shift+clicking a URL in a terminal opens it in a new in-app browser tab instead of doing nothing. Existing profiles are migrated on read when their stored map deep-equals the retired default (same pattern as the sidebar file-link migration) — anyone who customized any tier keeps their configuration.

## Review-comment cleanups
- `setForwardableChords` sync failures now log a warning instead of being swallowed (stale suppression is otherwise undiagnosable).
- `BrowserTabFavicon` extracted to its own component module per repo structure guidelines; its key now includes the page URL so a failed favicon retries on navigation even when the favicon URL is unchanged.
- `normalizeToken`'s alias lookup is case-insensitive, so `canonicalizeChord("MetaLeft+KeyK")` === `eventToChord` output (`meta+k`); pinned by tests.
- The two bot P1s (missing `preventDefault` on the forward path, alt-only chords dropped) were against the first commit only — the current code preventDefaults every forwarded chord (verified live: the guest page receives nothing) and has no meta/ctrl early-return.

## Regression safety
- Existing `Cmd/Ctrl+W`/`+R` interception is untouched (re-verified live).
- Non-registered keys fall through: plain typing and in-page shortcuts in the guest are unchanged (verified live).
- A balancing `keyup` is dispatched so react-hotkeys-hook's pressed-key set never gets a stuck key.

## Verification (live E2E on the dev app via CDP + OS-level keystrokes)
- **Forwarding chain end-to-end:** with the guest webview holding OS focus, a real `ctrl+Tab` was intercepted in main (guest page received nothing), forwarded, replayed on the host document (observed as an untrusted keydown), and the workspace switched tabs. `⌘W` with guest focus closed just the pane.
- **Tooltip A/B:** with main's `HotkeyTooltip`, hover a preset button → move toward the webview → tooltip sticks open indefinitely (bug reproduced); with this branch it closes as the pointer leaves the trigger.
- **Link navigation:** real cmd+clicks on URLs rendered in a terminal, alternating between two links — every click navigates the browser pane (previously only the first).
- **Favicon:** about:blank → globe; example.com (404 favicon) → globe; github.com → real favicon.
- **Chord parity:** `bun test src/shared/hotkey-chord.test.ts` pins `eventToChord ≡ chordFromInput` for all forwarded chords plus alias-case canonicalization.
- Lint clean; the 5 local typecheck errors are pre-existing and reproduce identically on `origin/main` in the same environment (CI Typecheck green).

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/browser/browser.ts
+++ b/apps/desktop/src/lib/trpc/routers/browser/browser.ts
@@ -160,6 +160,15 @@ export const createBrowserRouter = () => {
 				});
 			}),
 
+		// Renderer-registered canonical chords the main process should suppress in
+		// the focused guest and forward for replay (override/layout-aware).
+		setForwardableChords: publicProcedure
+			.input(z.object({ chords: z.array(z.string()) }))
+			.mutation(({ input }) => {
+				browserManager.setForwardableChords(input.chords);
+				return { success: true };
+			}),
+
 		// Keystrokes intercepted from the focused guest webview, replayed by the
 		// renderer into its hotkey system (guest focus hides them from the host).
 		onKeyForward: publicProcedure
--- a/apps/desktop/src/lib/trpc/routers/browser/browser.ts
+++ b/apps/desktop/src/lib/trpc/routers/browser/browser.ts
@@ -1,6 +1,9 @@
 import { observable } from "@trpc/server/observable";
 import { session } from "electron";
-import { browserManager } from "main/lib/browser/browser-manager";
+import {
+	browserManager,
+	type ForwardedKey,
+} from "main/lib/browser/browser-manager";
 import { z } from "zod";
 import { publicProcedure, router } from "../..";
 
@@ -153,6 +156,22 @@ export const createBrowserRouter = () => {
 					browserManager.on(`reload-pane:${input.paneId}`, handler);
 					return () => {
 						browserManager.off(`reload-pane:${input.paneId}`, handler);
+					};
+				});
+			}),
+
+		// Keystrokes intercepted from the focused guest webview, replayed by the
+		// renderer into its hotkey system (guest focus hides them from the host).
+		onKeyForward: publicProcedure
+			.input(z.object({ paneId: z.string() }))
+			.subscription(({ input }) => {
+				return observable<ForwardedKey>((emit) => {
+					const handler = (key: ForwardedKey) => {
+						emit.next(key);
+					};
+					browserManager.on(`key-forward:${input.paneId}`, handler);
+					return () => {
+						browserManager.off(`key-forward:${input.paneId}`, handler);
 					};
 				});
 			}),
--- a/apps/desktop/src/main/lib/browser/browser-manager.ts
+++ b/apps/desktop/src/main/lib/browser/browser-manager.ts
@@ -8,6 +8,15 @@ interface ConsoleEntry {
 	timestamp: number;
 }
 
+export interface ForwardedKey {
+	key: string;
+	code: string;
+	meta: boolean;
+	control: boolean;
+	alt: boolean;
+	shift: boolean;
+}
+
 const MAX_CONSOLE_ENTRIES = 500;
 
 function sanitizeUrl(url: string): string {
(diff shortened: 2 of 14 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Move hotkey chord logic to shared library](https://stagereview.app/superset-sh/superset/pull/5494/chapters/1) |
| 2 | [Implement key forwarding in the main process](https://stagereview.app/superset-sh/superset/pull/5494/chapters/2) |
| 3 | [Expose key forwarding via tRPC](https://stagereview.app/superset-sh/superset/pull/5494/chapters/3) |
| 4 | [Replay forwarded hotkeys in the renderer](https://stagereview.app/superset-sh/superset/pull/5494/chapters/4) |
| 5 | [Fix stuck tooltips near browser webview](https://stagereview.app/superset-sh/superset/pull/5494/chapters/5) |

<a href="https://stagereview.app/superset-sh/superset/pull/5494">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 9a3cb48a514751525ea9070c6a7ef7405ba5467b on Jul 7, 2026 9:34pm UTC.</sub>

Kitenite:
Thanks — addressed the key-forwarding findings in 9a3cb48a5.

**Both P1s (guest double-delivery + alt-only bindings) and the P2 (override-awareness) share one root cause:** the old forward path forwarded *all* meta/ctrl chords and never `preventDefault`'d, and the main process had no idea which chords are actually replayable. Simply adding `event.preventDefault()` (as suggested) would have been wrong — it would break `Cmd+C`/`Cmd+V`/`Cmd+F` etc. inside the browser, since those were being forwarded too.

Fix: the renderer now registers its override/layout-aware forwardable chord set with the main process, and `before-input-event` only acts on chords in that set:

- **Guest double-delivery** — matched chords are now `event.preventDefault()`'d, so the guest page no longer also receives `Ctrl+Tab` etc.
- **Alt-only bindings** — the `meta || control` guard is gone from the forward path; membership in the registered set is the only filter, so an alt-only remap is honored.
- **Override-awareness** — the set is built from `getDispatchChord(id)` (reflects overrides + keyboard layout) and re-pushed whenever the overrides / layout / preferences stores change, mirroring `resolveHotkeyFromEvent`'s index rebuild. `preventDefault` also removes the double-fire the P2 worried about.
- **No clipboard/find regression** — non-tab-switch chords are never in the set, so they fall through to the guest untouched.

The chord canonicalization is now a single source of truth: pure helpers moved to `shared/hotkey-chord.ts`, imported by both the renderer hotkey resolver and the main-process forwarder, so `chordFromInput` (main) and `canonicalizeChord(getDispatchChord(...))` (renderer) can't drift. Verified they produce identical canonical strings for every default tab-switch binding (mac + win), and all 122 hotkey unit tests still pass.

PR #5518, opened 2026-07-08
fix(desktop): stop broken-session auth request storms (Electric token refresh + get-session poller)

### LINKED ISSUE #5513: [bug] Broken-session client retries /api/auth/token hard enough to trip Vercel's per-IP firewall — which then 403s the login endpoint too, locking the user out entirely
#### Describe the bug

After tonight's window refresh dropped me into the "restoring your session" screen, clicking **Log in** dead-ended on a plain-text `403 Forbidden` from `https://api.superset.sh/api/auth/desktop/connect?...` — no in-app error, nothing actionable. Digging in: the 403 was coming from **Vercel's firewall** (`x-vercel-mitigated: deny`), which had denied my **entire home IP** for every path on `api.superset.sh` (IPv4 and IPv6, both machines behind the connection). External probes reached the API fine the whole time and status.superset.sh was all green — this was a per-IP deny, not an outage.

The trigger appears to be the app itself: the NetLog for the session that ended at the refresh contains **1,834 events referencing `POST /api/auth/token` (≈600 requests\*)**, vs 6 events (≈2 requests) in the comparable healthy session before it. So the failure loop is: session breaks → client retries token refresh with no effective backoff → Vercel's rate mitigation denies the IP → **the login endpoint is behind the same blanket rule**, so the human cannot re-authenticate to clear the broken state. Self-locking.

### Environment

- Superset Desktop 1.13.1 (stable), macOS 26.5 (Darwin 25.5.0), Apple Silicon
- Residential AT&T connection (nothing datacenter-ish about the IP); same heavy multi-workspace workload as #5503

### Timeline / evidence (2026-07-07/08, PDT)

| Time | Observation |
|---|---|
| Jul 7 20:17 → Jul 8 ~01:00 | App session runs, ends when the window refreshed into "restoring your session". Its stitched NetLog (1.00 GB) contains **1,834 URL events for `POST /api/auth/token`** + 75 for `get-session` |
| (baseline) Jul 7 15:05 → 17:48 | Prior healthy session, equally-sized NetLog: **6** token URL events (≈2 requests) |
| 01:00:35 | App relaunches into the "restoring your session" screen |
| 01:02:14 | `curl` to **any** path on `api.superset.sh` (root, `/api/health`, the login URL; IPv4 or IPv6; browser UA or not; from either machine) → `HTTP 403`, `server: Vercel`, **`x-vercel-mitigated: deny`**, plain-text body `Forbidden` |
| same minute | check-host.net probes from Frankfurt / Bengaluru / Kolkata get `404` from `/api/health` — API reachable and healthy globally |
| 01:00:35 → 01:17 | The login screen itself is calm: live NetLog shows only 4 token + 20 get-session requests in 17 min |
| 01:04:40 | With the storm session gone, the deny expires: the login URL returns the correct `307` → `github.com/login/oauth/authorize`, and login succeeds |

\* Calibration: in a live capture each request produces ~3 URL-bearing NetLog events, so 1,834 events ≈ 600+ actual requests.

I can't recover the burst rate inside that session (NetLog rotates within a 1 GiB budget, so the file may only cover the session's tail — the exact distribution is gone), so treat "client storm tripped the WAF" as a hypothesis rather than a proof. But ~300× the healthy baseline, followed by a per-IP deny that decayed within minutes of the storm stopping, is a tight fit for rate-based mitigation.

### What I'd hope for

1. Exponential backoff and a circuit breaker on `/api/auth/token` refresh when the session is broken — a wedged client shouldn't be able to generate hundreds of auth calls.
2. Client-side handling of a hard `403` / `x-vercel-mitigated: deny` response: stop retrying and surface a real error state ("temporarily blocked, retry in N minutes"), instead of a login button that opens a plain-text 403 in the browser with no explanation.
3. Server-side: scope the firewall rule so `/api/auth/desktop/connect` — the path a human needs in order to *fix* a broken session — isn't behind the same blanket per-IP deny as the API the broken client is hammering.

(The NetLog capture I used to diagnose this has its own problems — filing separately.)

Kitenite:
Closes #5513

## What & why

Fixes the client-side request storms behind the **2026-07-08 auth outage**, where a broken-session desktop client hammered `api.superset.sh` hard enough to trip Vercel's **automatic per-IP DDoS mitigation** (`403 x-vercel-mitigated: deny`) — which then denied *every* path on the host, including the login endpoint, **locking the user out of re-authenticating** (self-lockout).

There were **two** storm sources in the same broken-session failure mode:

### 1. `POST /api/auth/token` — Electric JWT refresh (the dominant one, per #5513)
Every Electric shape/collection shared one `onError` handler that called `authClient.token()` on each 401 with **no dedup, backoff, or circuit breaker**. When a session breaks, every shape 401s at once and the refreshed token is still rejected, so each shape loops — across a heavy multi-workspace client that's **hundreds of token requests (~300× baseline** in the reporter's NetLog: 1,834 events ≈ 600 requests vs ~2).

**Fix** — route the 401 refresh through a shared gate (`renderer/lib/jwt-refresh`):
- **single-flight**: concurrent shape 401s collapse to one `/api/auth/token` call
- **exponential backoff** keyed on consecutive failures (3s floor → 60s cap)
- **circuit breaker** after 5 failures (5-min cooldown) — a permanently-broken session stops hitting the API instead of storming it

### 2. `GET /api/auth/get-session` — sign-in recovery poller
`useSessionRecovery` polled on a flat 15s interval with no backoff/cap/terminal state; a stuck token polls forever, and a synchronized fleet adds to the storm.

**Fix** — exponential backoff (15s → 5min cap) + ±50% jitter (desyncs the fleet) + a 12-attempt budget, then idle until a user-driven focus/visibility event re-arms it.

## Testing
- **Unit** (`bun test`): 14 tests across both pure cores — `nextRecoveryDelayMs` and `shouldAttemptJwtRefresh`/`applyJwtRefreshResult` (backoff, cap, stop, jitter, single-flight, circuit).
- **Mutation-verified**: removing each protection (cap / stop / exponential / jitter / single-flight / circuit / success-reset) makes the specific guarding test fail — confirmed then reverted.
- **CDP runtime**: imported the real shipped `useSessionRecovery` module in the running dev renderer and drove the full cadence (first 9059ms in-band, caps at 300000ms, jitter 52849≠45135, stops at budget).
- `bun run lint` + desktop `typecheck` clean.

## Scope / not included (deliberately)
- **Server-side firewall isolation** (issue's fix #3 — put the auth entrypoint on its own origin so a per-IP deny can't block re-auth) is **not** here — it's infra work and the acute incident has recovered. Worth a follow-up as defense-in-depth for un-updated clients, but out of scope for this client fix.
- **Hard-403 client UX** (issue's fix #2 — surface "temporarily blocked, retry in N min" instead of opening a plain 403) — small follow-up, not included.
- `AuthProvider.tsx`'s 50-min `refreshJwt` is intentionally left unchanged (signed-in-only cadence, not a storm contributor).

## Summary by CodeRabbit

* **New Features**
  * Improved sign-in session recovery using bounded exponential backoff with symmetric jitter, with retry resets on focus/visibility changes.
  * Added a centralized JWT refresh gate to deduplicate concurrent refreshes and throttle retries with an error circuit.
* **Bug Fixes**
  * Replaced fixed polling with a cancellable, attempt-limited scheduler and ensured recovery stops after the configured maximum attempts.
* **Tests**
  * Added deterministic coverage for jitter/backoff timing and for JWT refresh retry and circuit behavior.


ghost:

| | Title |
|---|-------|
| 1 | [Implement centralized JWT refresh gate with backoff](https://stagereview.app/superset-sh/superset/pull/5518/chapters/1) |
| 2 | [Wire Electric sync errors to the refresh gate](https://stagereview.app/superset-sh/superset/pull/5518/chapters/2) |
| 3 | [Add backoff and jitter to session recovery](https://stagereview.app/superset-sh/superset/pull/5518/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5518">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 19c6096d6f0835d5c2b5f8971d53c5698aa3b3be on Jul 8, 2026 10:08am UTC.</sub>

Kitenite:
Thanks for the review. Addressed:

- **`nextRecoveryDelayMs(0)` sub-floor (P2):** clamped `attemptsMade` to `>=1` so a focus/visibility race yields the attempt-1 band, never a negative-exponent sub-floor. Added a regression test.
- **Cap test uses `random=1` (P2):** added a comment clarifying `random=1` reaches the ceiling (asserting the cap is applied *before* jitter).
- **Stale `session` after `await refetch()` (P2):** left as-is — the spurious timer is double-guarded (the top-of-`runRecovery` `session?.user` check no-ops it, and the effect cleanup clears it on the `session?.user` dep change), so no behavior bug. Not worth coupling to better-auth's `refetch` return shape.

Also expanded scope: the dominant storm in #5513 was `POST /api/auth/token` via the shared Electric `onError` handler, not `get-session`. Added a single-flight + backoff + circuit-breaker gate (`renderer/lib/jwt-refresh`) and routed the 401 refresh through it. This PR now closes #5513.

Kitenite:
Pushed [`2f3f60917`] addressing the latest review:

- **cubic (conf 9) — jitter was downward-only:** correct catch, and it mismatched the `±50%` docs. Changed to symmetric `Math.min(backoff * (0.5 + random), MAX)` → 0.5×–1.5× centered on the backoff. Added a test asserting it overshoots the base (mutation-verified: reverting to downward-only fails 3 tests).
- **cubic (conf 8) + CodeRabbit (major) — `finally` can arm a timer after unmount:** valid. Added an `isMountedRef` set false by an unmount-only (`[]`-dep) effect; the reschedule now no-ops after teardown.
- **CodeRabbit nit — assert `lastAttemptAt` preserved:** added to both `applyJwtRefreshResult` tests.
- **Stale `session?.user` spurious timer (cubic conf 7 / greptile):** leaving as-is — the unmount guard covers the risky variant, and while mounted it's a no-op scheduled-then-cleared timer (double-guarded by the top-of-`runRecovery` check + effect cleanup). Not worth coupling to better-auth's `refetch` return shape.

14 unit tests green, lint + typecheck clean.

Kitenite:
Both addressed in [`19c6096d6`]:

- **P1 (jwt-refresh in-flight retries):** `refreshJwtAfterUnauthorized` now returns the shared in-flight refresh promise, so concurrent shape `onError(401)`s **await** the single refresh and retry only after `setJwt()` has run — no more retrying with the stale JWT. Backed-off / circuit-open cases still resolve immediately and lean on Electric's own `_onErrorBackoff` + `_maxConsecutiveErrorRetries` (which cap and stop the shape-retry loop). Single-flight is preserved — verified earlier at the wire level (20 concurrent → 1 `/api/auth/token` request via CDP).
- **P3 (cap-test comment):** corrected — with symmetric jitter, `random=1` → factor 1.5, and the outer `Math.min` clamps to MAX; the comment said "factor 1.0" (old formula).

14 tests green, lint + typecheck clean.

PR #5560, opened 2026-07-09
fix(desktop): create worktree attachment dirs recursively on agent launch

### LINKED ISSUE #5554: [bug] New workspace creation fails with ENOENT on mkdir for .superset/attachments, no attachments involved
Creating a new workspace fails immediately with "Failed to start agent." The underlying error is an ENOENT thrown from a mkdir call targeting an attachments folder inside the freshly created worktree, even though no file or image was attached to the prompt in either reproduction. The path Superset is trying to create doesn't exist because the worktree it's nested under doesn't appear to be fully in place at the point this call runs, which reads like an ordering or race issue during workspace setup. Since the attachments folder gets created unconditionally as part of setup rather than only when an attachment is actually used, this looks like it can hit any new workspace, not just ones using that feature.

Actual error text: ENOENT: no such file or directory, mkdir '/Users/joshguyer/.superset/worktrees/Avery/adhesive-radon/.superset/attachments'

Reproduced twice on the same project. First attempt had the workspace's agent command set to wrap claude through headroom (headroom wrap claude --dangerously-skip-permissions ...). Second attempt used a plain claude --dangerously-skip-permissions invocation with no wrapper. No file or image attachment was used in either case. Both hit the identical mkdir error, which rules out the agent command and the attachment feature as the cause and points at the worktree creation and attachments-folder setup racing or running out of order before the agent is ever spawned.

Expected behavior: the workspace is created and the agent starts normally, with setup completing without throwing.

Environment: macOS 26.5.1, Superset 1.14.0, project is a local repo with worktrees at the default location.

AviPeltz:
## Summary

Fixes #5554 — new workspace creation failing with `ENOENT: no such file or directory, mkdir '<worktree>/.superset/attachments'` and the "Failed to start agent" toast.

## Root cause

Not a worktree-creation race. `writeAttachmentFiles` in `apps/desktop/src/renderer/lib/agent-session-orchestrator/adapters/terminal-adapter.ts` created `<worktree>/.superset/attachments` in a single `filesystem.createDirectory` call **without** `recursive: true`. `createDirectory` in `packages/workspace-fs/src/fs.ts` defaults to `recursive = false`, so it's a plain `fs.mkdir` — and in a fresh worktree `.superset` doesn't exist yet, so the two-level mkdir deterministically ENOENTs and aborts the agent launch.

The task-launch path masked this: `writeTaskPromptFile` runs first and creates `.superset` one level at a time, so attachments only failed on launches with files but no task prompt file.

Why it fired with "no attachments": linking a GitHub issue in the new-workspace modal silently converts the issue body into a `github-issue-<N>.md` attachment (`PromptGroup.tsx` `handleCreate`), so `initialFiles` is non-empty even though the user never attached a file or image.

## Fix

Pass `recursive: true` to `filesystem.createDirectory` for both the attachments dir and the task-prompt `.superset` dir. The workspace-fs layer already supports this and still performs symlink/realpath containment checks for missing ancestors (`assertDanglingSymlinkSafe`), and recursive creation is covered by existing tests in `packages/workspace-fs/src/fs.test.ts`.

## Validation

- `biome check` clean on the changed file
- `bun run typecheck --filter=@superset/desktop` passes

## Note for a follow-up

While tracing this I noticed `WorkspaceInitEffects` launches the pending agent immediately when a workspace has no `initProgress` entry (apps/desktop/src/renderer/screens/main/components/WorkspaceInitEffects.tsx:329-336), without waiting for worktree readiness — a real ordering hazard, but separate from this bug.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5560">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Prevent agent launch failures on new workspaces by creating `.superset` and `.superset/attachments` directories recursively. Fixes #5554.

- **Bug Fixes**
  - Set `recursive: true` on `filesystem.createDirectory` for the task prompt and attachments dirs in `terminal-adapter.ts`.
  - Eliminates ENOENT when `.superset` doesn’t exist yet (e.g., when a linked GitHub issue creates an initial attachment).

<sup>Written for commit f8acc6f8af888385238127ccf1746f227328fb68. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5560?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file handling so task prompts and attachments are saved reliably, even in fresh workspaces.
  * Reduced the chance of agent startup issues caused by missing folders.


ghost:

| | Title |
|---|-------|
| 1 | [Enable recursive directory creation for agent files](https://stagereview.app/superset-sh/superset/pull/5560/chapters/1) |

<a href="https://stagereview.app/superset-sh/superset/pull/5560">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit f8acc6f8af888385238127ccf1746f227328fb68 on Jul 10, 2026 12:29am UTC.</sub>

PR #4739, opened 2026-05-19
refactor(desktop): couple host-service to Electron, drop adoption

Kitenite:
## Summary

Host-service no longer survives Electron. The detach + manifest-adoption machinery was load-bearing back when host-service owned PTYs directly; now that pty-daemon (separate detached lifecycle, supervised by host-service) owns PTY survival, host-service itself has no reason to outlive the app.

- **Coordinator** — spawn with \`detached: false\`. Remove \`tryAdopt\`, \`discoverAll\`, \`teardownKnownManifests\`, \`releaseAll\`, adopted-liveness timer, and the version-mismatch healing branch. \`stop\` / \`stopAll\` / \`restart\` / \`reset\` kept; \`reset\` is still the SIGKILL escape hatch for wedged manifests.
- **Quit path** — \`before-quit\` always SIGTERMs host-services. \`runFullQuitCleanup\` collapses into \`teardownTerminalHost\` (v1 only).
- **Parent watchdog** — host-service polls \`process.ppid\` against \`HOST_PARENT_PID\`. If Electron crashes / force-quits without sending SIGTERM, the child reaps itself within ~2s. CLI-spawned standalone host-services don't set the env var and skip the watchdog.
- **Graceful shutdown** — SIGTERM gets a 3s drain, then \`closeAllConnections\` forces SSE/WS streams (chat completions, file watchers) closed so we don't hang.
- **Manifest** — still written by the child so the CLI (\`packages/cli\`) can find a running service. Desktop coordinator no longer reads it for adoption, only for the \`reset\` recovery path.
- Docs and tests updated; \`HOST_SERVICE_LIFECYCLE.md\` rewritten.

Net: ~390 lines deleted, simpler lifecycle, no orphan host-services.

## Test plan
- [x] \`bun run lint\` clean
- [x] \`bun run typecheck\` clean in apps/desktop
- [x] All 1959 desktop tests pass
- [ ] Manual: open a chat completion, Cmd+R the window, verify the renderer reconnects to a fresh host-service cleanly
- [ ] Manual: \`kill -9\` the Electron process while a host-service is running, then \`ps aux | grep host-service\` ~10s later — should be empty (watchdog reaped it)
- [ ] Manual: run auto-update install, verify new version's host-service starts fresh (no stale-adoption)
- [ ] Manual: CLI \`superset status\` while desktop running — should still find host-service via manifest

---

<a href="https://stagereview.app/superset-sh/superset/pull/4739">
  <picture>
[...]

PR #5721, opened 2026-07-16
fix(host-service): never kill PTYs during daemon adoption, and say when the daemon goes quiet

Kitenite:
**Links**
- Issue: [SUPER-833 — V2 after update terminals killed](https://linear.app/superset-sh/issue/SUPER-833/v2-after-update-terminals-killed) (closed May 2026, recurring)

## Summary
1. **Stop killing shells.** `tryAdopt` destroyed a live pty-daemon — and every shell under it — whenever post-update adoption was uncertain. That logic is deleted; `tryAdopt` is now one probe and one return.
2. **Say something when the daemon goes quiet.** Not killing it means the rare wedged daemon now shows up as a terminal that hangs. Name it, and offer the (destructive) fix behind a confirm.
3. **Harden the new path + polish the UX.** With the kill gone, guard adoption against recycled pids, keep auto-update working for daemons adopted under load, and rework the connection indicator into a plain-language popover.

## Part 1 — never kill PTYs during adoption

Terminals are meant to survive an update: the daemon is spawned `detached: true` and the relaunched host-service **adopts** it. Since #4739, PTY survival rests entirely on that adoption being non-destructive.

It wasn't. `tryAdopt` called `terminateProcessTreeAndGroups(manifest.pid, …)` when the socket wasn't connectable within **1s**, or the hello probe went unanswered within **3s**. That walks the daemon's **entire process tree + process groups** — every live shell — and SIGTERM→SIGKILLs them.

Those budgets are spent at the worst moment: adoption runs **eagerly at host-service boot** (reaper → `getDaemonClient → waitForDaemonReady → ensure → tryAdopt`), i.e. during the post-update relaunch spike — exactly when a healthy-but-CPU-starved daemon misses a 1–3s probe.

The codebase already said what should happen. `expected-version.ts`: *"We do NOT auto-kill … sessions live in the daemon; the user explicitly triggers restart."* And `DaemonInstance` documents `runningVersion: "unknown"` for a failed probe, with *"Probe failure does NOT set"* `updatePending`. Adopting-on-probe-failure was the intended design; the kill was the bug.

Also removed as dead weight once the kill went: the `previousManifest` plumbing. (The expected-socket fallback that used to run only as post-kill recovery now has a real job again — it's the stale-manifest recovery path below.)

### Hardening the adoption path (post-review)

Now that adoption never signals the daemon, two edge cases need care:

- **Recycled pid / stale manifest.** After a reboot the manifest's pid can belong to an unrelated live process. `isProcessAlive` (a bare `kill(pid, 0)`) passes, so blindly adopting would bind to a phantom: every terminal fails with `ECONNREFUSED`, and the user's only recovery — Restart — would SIGTERM an innocent process tree. `tryAdopt` now treats *pid alive **but** probe failed **and** socket not connectable* as stale: log `pty_daemon_manifest_stale`, drop the manifest, fall through to socket adoption / a fresh spawn. A genuinely wedged-but-listening daemon (socket bound, hello unanswered) is still adopted — SUPER-833 preserved.
- **Auto-update after a probe-failed adoption.** A daemon adopted under the post-update spike comes up `runningVersion: "unknown"`. `refreshReachability` already backfills the real version on the next successful probe; it now also fires update-pending and kicks off the background auto-update, so a stale daemon adopted under load still self-updates instead of running outdated indefinitely.
- **Probe cost.** The reachability probe runs on a slower **5s** cadence with a **3s** retry budget (was one 1.5s attempt every 2s). A single-threaded daemon streaming heavy PTY output can miss one 1.5s deadline while perfectly healthy; retrying on a slower beat keeps it from being false-flagged "unresponsive" (which offers a destructive restart). The cheap `kill(pid,0)` liveness check stays at 2s. (`adoptedLivenessTimers` → `healthPollTimers`; the four copies of the update-pending semver check are now one `isVersionUpdatePending` helper.)

## Part 2 — "Terminals aren't responding"

Keeping a wedged daemon alive is right, but it left the tail case as a mystery: the terminal hangs ~15s and dies with `daemon open: timed out after 15000ms`.

**Host-service** tracks reachability on the poll that already watches daemons: probe each tick, record when it first went quiet, clear on recovery. Two things fell out of the E2E:
- That poll only ran for **adopted** daemons. A daemon wedges regardless of how we got it, so the signal was dead for freshly spawned ones — the common case. It now runs for both; spawned daemons pass no death handler, since `child.on("exit")` owns death **and** crash-respawn (clearing the instance from the poll would have silently disabled respawn).
- The probe doubles as the retry for a version adoption couldn't read, so `runningVersion` no longer stays stuck at `"unknown"`.

New `terminal.daemon.getHealth` procedure surfaces `{ reachable, unreachableForMs }`; it deliberately does **not** `waitForDaemonReady` (it must answer immediately rather than block on the thing that may be wedged).

**Renderer** — the pane's connection indicator, reworked into a plain-language popover:

```
⚠  Terminals aren't responding

[ ↻ Reconnect ]   [ Restart ]

                    View log
```

- **Reconnect** (solid, primary) retries the WebSocket — the safe first move, shown whenever the transport is unhealthy, and it flips to a spinner + "Reconnecting…" once clicked. **Restart** (outline, red only on hover) → *"Restart all terminals?"* confirm with a **Keep waiting** cancel → `terminal.daemon.restart`. **Both can show at once**, so a stale daemon-health flag never leaves restart as the only option after sleep/wake (the transport can give up while the daemon flag is still catching up).
- **No "daemon" jargon** anywhere user-facing — it's all "terminals". "Daemon" stays in logs/telemetry.
- **Amber for recoverable** states (reconnecting; unresponsive-but-usually-self-heals); **red reserved** for the one destructive control and the hard "Disconnected" (transport gave up) state.
- **Log collapsed** behind "View log"; expanding widens the popover to a readable full-width layout with the timestamp as a header line rather than a narrow left gutter.

Design decisions worth calling out:
- **Waits 10s.** The only fix closes every terminal, so a stall must look permanent before we offer it — a post-update spike clears well inside that and shells come back on their own. Warning eagerly would have recreated SUPER-833 via the user's own click.
- **Auto-clears** on recovery, so a transient stall never leaves a standing invitation to nuke.
- **Polls even when the socket looks healthy.** A daemon that wedges under a live session closes nothing — the WS stays `open` and the shell just freezes. The query takes no input, so React Query collapses every pane onto one request per workspace.
- **Outranks the WebSocket's message**, because the daemon owns the shells and is the root cause.
- **No system popup.** `dialog.showErrorBox` (the `alertChildCrashed` path) blocks Electron's main process — it wedged the whole app during this work. It would also demand a decision during the exact stall where waiting is correct, with a button that destroys every shell.

**turbo.jsonc**: pass `SUPERSET_PTY_DAEMON_ADOPT_IN_DEV` through. Turbo filters unlisted env vars, so the flag never reached host-service and the dev app always took the kill-stale-daemon path — making adoption (and this warning) impossible to exercise in dev. One line; no behavior change unless you set it.

## Testing
- `bun run lint` ✅ · `bun run typecheck` (host-service + desktop) ✅
- `bun test src/daemon/DaemonSupervisor.test.ts` — **50 pass**. Adds recycled-pid + expected-socket-fallback adoption tests and late-stale-version auto-update tests on top of the "adopts an unprobeable but pid-alive daemon instead of killing its shells" regression guard and the health suite.
- **Mutation-verified** (each fails only its own guard): reintroduce the kill; adopt the recycled pid blindly; skip the late auto-update kickoff; never clear the silence on recovery; never start the clock; don't backfill the version.
- `bun run test:integration:daemon` (real daemons: DaemonClient + DaemonSupervisor) — **18 pass** (crash-respawn + fd-handoff intact after the poller change).

### Stress harness — real daemons + real shells, isolated orgIds
`SIGSTOP` = a CPU-starved daemon (socket bound, hello unanswered).

| Scenario | Result |
|---|---|
| Healthy | adopted, shell survives |
| Stalled 5s then resumes | adopted, shell survives |
| **Wedged forever** | **adopted (`unknown`, `updatePending:false`); daemon NOT killed; SHELL SURVIVES** |
| Daemon dead | clean fresh spawn |
| Concurrent `ensure()` during stall | single adopt, shell survives |

### Manual QA — real app, over CDP
Driven against the real dev app (production adoption via `SUPERSET_PTY_DAEMON_ADOPT_IN_DEV`), with a real workspace + terminal under the real daemon. All paths confirmed:
- [x] **Never kills the shell:** real terminal; `SIGSTOP`; `hostServiceCoordinator.restart` (= the update's stop+start); hold past the old budget; `SIGCONT` → **daemon and shell both survived**, log shows `adopted existing daemon`.
- [x] **Recycled-pid guard:** manifest pointed at a live impostor pid + dead socket; restart → log shows `pty_daemon_manifest_stale`, a fresh daemon spawns, **the impostor is never signalled**.
- [x] **Late auto-update:** adopted a stale `0.0.1` daemon as `unknown` (probe failed via `SIGSTOP`), then `SIGCONT` → `pty_daemon_update_pending` → `auto_update_attempt` → `auto_update_ok` (fd-handoff to the fresh version).
- [x] **Rendered popover:** forced daemon-unreachable (`SIGSTOP`) **and** WS give-up together → both **Reconnect** and **Restart** render; the red "Terminals aren't responding" pill appears at the 10s threshold; `SIGCONT` → the retrying probe clears it, host `getHealth` returns `reachable:true`, **shell alive throughout**.
- [x] **Health signal on a spawned daemon:** `reachable:true` → `SIGSTOP` → `false` (crosses the 10s warn threshold) → `SIGCONT` → `true` within 3s.

## Known Limitations
- A wedged daemon needs an explicit user restart instead of auto-respawning. Deliberate: never silently destroy shells.
- Only the *adoption* kill is removed. `killStaleDaemonForDev` (dev-only) and the user-triggered restart stay destructive by design.

## Risks / Rollout
- **Risk:** low. Part 1 strictly removes a destructive action and adds a stale-pid guard. Part 2 is additive (a read-only signal + an indicator state); worst case the warning doesn't fire and users see today's timeout.
- **Rollback:** revert the commits; no data/schema changes.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/TerminalPane/components/TerminalPaneHeaderExtras/components/TerminalConnectionIndicator/TerminalConnectionIndicator.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/TerminalPane/components/TerminalPaneHeaderExtras/components/TerminalConnectionIndicator/TerminalConnectionIndicator.tsx
@@ -1,7 +1,19 @@
+import {
+	AlertDialog,
+	AlertDialogAction,
+	AlertDialogCancel,
+	AlertDialogContent,
+	AlertDialogDescription,
+	AlertDialogFooter,
+	AlertDialogHeader,
+	AlertDialogTitle,
+} from "@superset/ui/alert-dialog";
 import { Button } from "@superset/ui/button";
 import { Popover, PopoverContent, PopoverTrigger } from "@superset/ui/popover";
+import { toast } from "@superset/ui/sonner";
 import { cn } from "@superset/ui/utils";
-import { useCallback, useSyncExternalStore } from "react";
+import { workspaceTrpc } from "@superset/workspace-client";
+import { useCallback, useState, useSyncExternalStore } from "react";
 import {
 	type TerminalLogEntry,
 	terminalRuntimeRegistry,
@@ -11,6 +23,16 @@ interface TerminalConnectionIndicatorProps {
 	terminalId: string;
 	terminalInstanceId: string;
 }
+
+/** How often to re-ask the host whether the daemon is answering. */
+const DAEMON_HEALTH_POLL_MS = 5_000;
+/**
+ * How long the daemon must stay silent before we tell the user. The only fix
+ * we can offer closes every shell, so a stall has to look permanent before we
+ * suggest it — a post-update load spike resolves itself well inside this, and
+ * the shells come back on their own.
+ */
+const DAEMON_UNREACHABLE_WARN_MS = 10_000;
 
 function formatTime(timestamp: number): string {
 	const d = new Date(timestamp);
@@ -30,11 +52,18 @@ function formatLogsForClipboard(logs: readonly TerminalLogEntry[]): string {
 }
 
 /**
- * Pane-header connection status for the terminal WebSocket. Hidden while the
- * connection is healthy; shows an amber "Reconnecting" dot while the
- * auto-reconnect loop runs and a red "Disconnected" dot once the transport
- * stops trying. The popover carries the human diagnosis, a manual reconnect,
- * and the raw transport log for bug reports.
+ * Pane-header health for the terminal, across both things that can break it.
+ *
+ * The WebSocket: hidden while healthy, amber "Reconnecting" while the
+ * auto-reconnect loop runs, red "Disconnected" once the transport gives up.
+ * The popover carries the diagnosis, a manual reconnect, and the raw log.
+ *
+ * The pty-daemon: red "Terminals aren't responding" when the host reports it
+ * has stopped answering for {@link DAEMON_UNREACHABLE_WARN_MS}. That outranks
+ * the WebSocket's story because the daemon owns the shells — and it's shown
+ * even when the socket looks fine, since a wedged daemon closes nothing and
+ * simply freezes the shell. The only fix closes every terminal, so it sits
+ * behind a confirm and we wait out short stalls rather than offer it.
  */
 export function TerminalConnectionIndicator({
 	terminalId,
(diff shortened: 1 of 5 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define daemon health types and constants](https://stagereview.app/superset-sh/superset/pull/5721/chapters/1) |
| 2 | [Stop killing PTYs during daemon adoption](https://stagereview.app/superset-sh/superset/pull/5721/chapters/2) |
| 3 | [Implement background health polling](https://stagereview.app/superset-sh/superset/pull/5721/chapters/3) |
| 4 | [Expose daemon health via API](https://stagereview.app/superset-sh/superset/pull/5721/chapters/4) |
| 5 | [Rework terminal connection indicator UI](https://stagereview.app/superset-sh/superset/pull/5721/chapters/5) |
| 6 | [Verify adoption and health logic with tests](https://stagereview.app/superset-sh/superset/pull/5721/chapters/6) |

<a href="https://stagereview.app/superset-sh/superset/pull/5721">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit eaf666969a0b7bfe2ebf5a8932d290b73399a3ae on Jul 17, 2026 1:03am UTC.</sub>

PR #5734, opened 2026-07-17
fix(desktop): stop the sign-in redirect render loop that can OOM the renderer

### LINKED ISSUE #5729: Superset hangning
### Where did this happen?

Desktop app (macOS)

### Version

Version 1.15.1 (1.15.1)

### What happened?

Superset macos app just hangs. cant use the UI. the menu bar is still working but thats it. even after reboot. started happening after updating to `Version 1.15.1 (1.15.1)` 

### Steps to reproduce

1. open app
2. wait 20 seconds
3. dead

Image

Kitenite:
Fixes #5729

## Summary
- Fixes the infinite navigate/render loop behind react error #185 ("Maximum update depth exceeded"), which fires on **every signed-out boot** — PostHog shows ~37k of these exceptions from ~3,600 distinct desktop users in the last 7 days, nearly all at `#/sign-in`.
- In the pathological case (a `get-session` request that never settles), the loop ran unbounded until the renderer OOMed at ~6GB — the "app hangs / white window after 20 seconds" in #5729.
- Adds an escape hatch for a genuinely unreachable auth server: boot no longer waits on the splash forever, and after 15s the user gets "Still restoring your session" with **Retry** / **Sign out** instead of an indefinite spinner.

## Why / Context
The reporter in #5729 blamed the Thai (`en_TH`) macOS locale, but that's a red herring: Chromium resolves the renderer's Intl/`navigator.language` to plain `en-US` even under Region=Thailand (verified four ways on real hardware, plus JS shims forcing `en-GB` / `th-TH` Buddhist Intl — no runaway in any config). Their PostHog event stream shows the real signature: 9× react error #185 within one second of their very first `desktop_opened`, with a stack of `commitHookEffectListMount → RouterCore.navigate → startTransition → dispatchSetState`.

## How It Works
Root cause is a composition of two things:

1. TanStack Router's `<Navigate>` runs `navigate(props)` in a layout effect whose deps include the raw `props` **object identity** — so any re-render that produces a fresh JSX element re-navigates.
2. `AuthenticatedLayout` rendered `<Navigate to="/sign-in">` while the session query was still pending, and it subscribes to router state (`useLocation`) — so the navigation it triggered re-rendered it, which created a fresh `<Navigate>`, which navigated again, synchronously, ~50 commits deep → react #185. The loop runs for as long as the redirect condition holds: ~2s of session-pending on a normal boot (a few #185s, error boundary recovers), forever when `get-session` hangs (renderer runaway → OOM).

The fix (commit 1):
- **Hoist boot-path redirect elements to module scope** (`/sign-in`, `/create-organization`, `/onboarding`, `/workspace`, `/`) so their props identity is stable and `Navigate` fires exactly once per mount instead of once per re-render. This kills the loop class structurally.
- **Never redirect while the session is unresolved**: the authenticated layout now shows its existing spinner during `isPending` and only redirects on a definitive signed-out result. This also stops kicking slow-booting cookie-session users off their restored deep route.

The escape hatch (commit 2), for a `get-session` that never settles:
- `AuthProvider` hydration races the session/JWT fetch against a 15s bound instead of awaiting it forever on the splash.
- The authenticated layout's pending screen upgrades after 15s to "Still restoring your session" with **Retry** (refetch) and **Sign out** (clears the local token via the existing main-process sign-out, then lands on `/sign-in`).
- The sign-in page falls through from its spinner to the actual sign-in buttons after 15s, so re-auth is always reachable even mid-outage.
- New shared `useDelayElapsed(active, delayMs)` hook under `renderer/hooks/`.

## Manual QA Checklist
All validated in the dev app over CDP with a diagnostic shim (history + console.error hooks), each scenario run before and after the fix:

- [x] Signed-out boot (token removed, cookies + localStorage cleared): lands at `#/sign-in` with **0** "Maximum update depth" errors (pre-fix: 1 per boot, every boot), and settles in ~500ms instead of ~2s
- [x] Signed-in boot: lands on the workspace route, 0 errors
- [x] Fatal scenario — `/api/auth/get-session` blackholed indefinitely via CDP request interception: renderer idles at ~10% CPU with flat memory (pre-fix behavior per #5729: 500–790% CPU, ~6GB RSS, renderer OOM); after ~30s total the timeout screen appears with Retry / Sign out
- [x] Clicking **Sign out** on the timeout screen (still blackholed) lands on `#/sign-in`; after its 15s fallthrough the GitHub/Google sign-in buttons render, so the user can re-auth during a full auth outage
- [x] OAuth token-replay path (`persistToken` → `onTokenChanged` → signOut/refetch): no navigation storm

## Testing
- `bun run lint` (exits 0)
- `bunx tsc --noEmit` in `apps/desktop` (clean)
- `bun test src/renderer/routes/sign-in` (7/7 pass)

## Known Limitations
- `getAuthToken()` is still a bare module variable read during render (it tears, but no longer feeds a redirect decision while the session is pending). Migrating it to `useSyncExternalStore` is a possible follow-up.
- Settings-page `<Navigate>` usages with dynamic params were left as-is: they aren't on the boot path and don't subscribe to router state, so they don't exhibit the loop.

## Risks / Rollout
- Risk: low — behavior changes are confined to boot-time redirect ordering and previously-unreachable hang states. Fresh installs briefly show a spinner (~300ms) before `/sign-in` instead of redirecting during pending; a hung boot proceeds past the splash after 15s instead of never.
- Rollback: revert both commits; no data or config involved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect stability across workspace, sign-in, onboarding, and organization creation by keeping navigation elements consistent across re-renders.
  * Updated authentication/session restoration: the app now shows an “auth pending” spinner with a timed-out state that supports Retry or Sign out.
  * Prevented auth from getting stuck: added timeouts for hydration and sign-out so slow or unreachable auth services won’t block app initialization or sign-out completion.

### DIFF
--- a/apps/desktop/src/renderer/hooks/useDelayElapsed/index.ts
+++ b/apps/desktop/src/renderer/hooks/useDelayElapsed/index.ts
@@ -1,0 +1,1 @@
+export { useDelayElapsed } from "./useDelayElapsed";
--- a/apps/desktop/src/renderer/hooks/useDelayElapsed/useDelayElapsed.ts
+++ b/apps/desktop/src/renderer/hooks/useDelayElapsed/useDelayElapsed.ts
@@ -1,0 +1,17 @@
+import { useEffect, useState } from "react";
+
+/** True once `active` has stayed true for `delayMs`; resets when `active` drops. */
+export function useDelayElapsed(active: boolean, delayMs: number): boolean {
+	const [elapsed, setElapsed] = useState(false);
+
+	useEffect(() => {
+		if (!active) {
+			setElapsed(false);
+			return;
+		}
+		const timer = window.setTimeout(() => setElapsed(true), delayMs);
+		return () => window.clearTimeout(timer);
+	}, [active, delayMs]);
+
+	return elapsed;
+}
--- a/apps/desktop/src/renderer/hooks/useSignOut/useSignOut.ts
+++ b/apps/desktop/src/renderer/hooks/useSignOut/useSignOut.ts
@@ -5,6 +5,9 @@ import { posthog } from "renderer/lib/posthog";
 export const AUTH_COMPLETED_KEY = "superset_auth_completed";
 export const ACTIVE_ORG_ID_KEY = "active_organization_id";
 
+// An unreachable auth server must not block local sign-out (#5729)
+const SERVER_REVOKE_TIMEOUT_MS = 5_000;
+
 export function useSignOut() {
 	const signOutMutation = electronTrpc.auth.signOut.useMutation();
 	const setAnalyticsUserId = electronTrpc.analytics.setUserId.useMutation();
@@ -14,7 +17,12 @@ export function useSignOut() {
 		setAnalyticsUserId.mutate({ userId: null });
 		localStorage.removeItem(AUTH_COMPLETED_KEY);
 		localStorage.removeItem(ACTIVE_ORG_ID_KEY);
-		await authClient.signOut();
+		await Promise.race([
+			authClient.signOut({ fetchOptions: { throw: false } }).catch(() => {}),
+			new Promise((resolve) =>
+				window.setTimeout(resolve, SERVER_REVOKE_TIMEOUT_MS),
+			),
+		]);
 		signOutMutation.mutate();
 	};
 }
--- a/apps/desktop/src/renderer/providers/AuthProvider/AuthProvider.tsx
+++ b/apps/desktop/src/renderer/providers/AuthProvider/AuthProvider.tsx
@@ -3,6 +3,8 @@ import { authClient, setAuthToken, setJwt } from "renderer/lib/auth-client";
 import { SupersetLogo } from "renderer/routes/sign-in/components/SupersetLogo/SupersetLogo";
 import { electronTrpc } from "../../lib/electron-trpc";
 
+const HYDRATION_TIMEOUT_MS = 15_000;
+
 export function AuthProvider({ children }: { children: ReactNode }) {
 	const [isHydrated, setIsHydrated] = useState(false);
 	const { refetch: refetchSession } = authClient.useSession();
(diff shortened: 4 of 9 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Add useDelayElapsed utility hook](https://stagereview.app/superset-sh/superset/pull/5734/chapters/1) |
| 2 | [Stabilize redirect identities to prevent render loops](https://stagereview.app/superset-sh/superset/pull/5734/chapters/2) |
| 3 | [Implement session hydration timeouts](https://stagereview.app/superset-sh/superset/pull/5734/chapters/3) |
| 4 | [Add escape hatches for hung authentication](https://stagereview.app/superset-sh/superset/pull/5734/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/5734">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 9536f956e93689be2b9cbd072f1cf398c3981fa1 on Jul 17, 2026 5:46pm UTC.</sub>

Kitenite:
Addressed the review findings in 9536f956e:

- **Stale JWT after timed-out hydration (cubic P1)** — valid, fixed. A `token()` response that outlives the hydration timeout is now applied only if the auth token is unchanged since hydration started, so it can't resurrect a JWT after sign-out or a token change.
- **Recovery Sign out bypassing `useSignOut` (cubic P2)** — valid, fixed. The timeout screen now signs out through `useSignOut` (PostHog reset, analytics user id, auth-completion keys, server revocation). Naively reusing it would have re-wedged the escape hatch — it awaited `authClient.signOut()` unboundedly — so the server revocation is now bounded at 5s and non-throwing for every caller. Re-verified over CDP with both `get-session` and `sign-out` blackholed: timeout screen → Sign out → ~5s bounded wait → lands on `/sign-in` → sign-in buttons render.
- **Unhandled rejection on Retry (cubic P2)** — skipped: identical pattern already exists on the adjacent offline screen, and better-auth's `refetch` contains its errors.
- **Clear the hydration race timer early (CodeRabbit nitpick)** — skipped: the losing timer resolves a no-op promise within 15s; no behavioral effect.

Kitenite:
One transparency note on the `useSignOut` change in 9536f956e that reviewers should weigh in on: it changes failure semantics for the three existing sign-out callers (settings, org dropdown, create-organization), not just the new recovery screen. Previously `await authClient.signOut()` threw on a failed server call, so local sign-out never completed (user stayed signed in, with an unhandled rejection). Now local sign-out always proceeds and server revocation is best-effort (bounded at 5s, non-throwing). Trade-off: a failed revocation leaves the server session alive for its remaining TTL while the client forgets its token. I think best-effort is right (blocking local sign-out on an unreachable server achieves nothing except wedging the user), but if we want the old strict behavior for the healthy-path callers, the recovery screen could take a `bestEffort` option instead.

PR #5742, opened 2026-07-17
feat(host-service): auto-name workspaces via the launch agent's non-interactive CLI

Kitenite:
## Summary
- Adds `nonInteractiveCommand` to terminal agent definitions: each builtin agent's headless one-shot invocation (`claude -p`, `codex exec`, `gemini -p`, ...), verified per CLI.
- The direct small-model call stays the primary namer (~1s). When it can't run or fails (no/invalid Anthropic/OpenAI credentials), `workspaces.create` falls back to the launch agent's own CLI — the agent the user opened the workspace with — which names with the agent's own credentials on a small/fast model where one is known.

## Why / Context
Auto-naming previously only had `getSmallModel`, which resolves Anthropic/OpenAI credentials from mastracode auth storage. Users whose agent is Gemini, Copilot, Amp, Droid, etc. have neither — naming silently no-oped for them and workspaces kept friendly-random names. The agent binary the user just picked for the workspace is guaranteed installed and carries its own auth (Claude subscription, ChatGPT plan, Google account), so it can do the naming with zero extra credential plumbing.

## How It Works
- `TerminalAgentDefinition.nonInteractiveCommand` (optional): headless one-shot command, prompt appended as the final argument, **locked down** — no permission bypasses; tools denied or read-only per CLI (print-mode auto-deny, read-only sandboxes, plan/ask modes, `--no-tools`) — plus only the flags each CLI needs to run unattended in a fresh untrusted dir (`--skip-git-repo-check`, `--skip-trust`, `--trust`). Absent = no headless mode (polygraph). Custom agents can declare it too.
- **Order**: `getSmallModel` first (5s timeout); on null/failure, the launch agent's CLI (20s timeout); if both fail, the existing friendly-random name plus the `autoNameWarning` toast. Create never fails on naming.
- The fallback resolves the launch's agent id through `hostAgentConfigs` → `presetId` → builtin definition, then runs via `$SHELL -lc` (login-shell PATH, matching how the user's terminal resolves the binary) with cwd in a scratch dir so the agent can't pick up repo context or act on files, stdin closed.
- The naming call pins a small/fast model where the agent's curated catalog has an unambiguous cheap tier — claude → `haiku`, codex → `gpt-5.6-luna` (OpenAI's fast/affordable 5.6 tier), gemini → `gemini-2.5-flash`, vibe → `devstral-small` (via `VIBE_ACTIVE_MODEL`), cursor-agent → `composer-1` — validated through `buildAgentModelArgs`/`buildAgentModelEnv`; other presets run their default model. Model args are spliced right after the binary because trailing flags like gemini's `-p` consume the next token.
- Two hardenings from the failure-path E2E: the spawned CLI env strips `ANTHROPIC_API_KEY`/`OPENAI_API_KEY` (claude prefers the env var over its claude.ai login, so an invalid key that broke the primary path would otherwise poison the fallback too and hang it to timeout), and the naming prompt wraps the user prompt in `<user-prompt>` data tags with an ignore-embedded-directives instruction (haiku obeyed a "just reply DONE" line inside the prompt text instead of returning JSON).
- Output parsing takes the last flat JSON object containing both fields, so CLI banners (e.g. mastracode's skill-load lines) don't break it; results flow through the existing sanitize/dedupe pipeline.

Per-agent commands, each verified against the CLI's `--help`; ✅ = also verified with a live one-shot run:

| Agent | Command | Lockdown | Naming model | Verified |
|---|---|---|---|---|
| claude | `claude -p` | print mode auto-denies unapproved tools | `haiku` | ✅ live |
| amp | `amp -x` | default confirmations (no `--dangerously-allow-all`) | default | ✅ live (after `amp update`) |
| codex | `codex exec --skip-git-repo-check` | default read-only sandbox | `gpt-5.6-luna` | ✅ live |
| gemini | `gemini --skip-trust -p` | default approval mode denies headless | `gemini-2.5-flash` | help + gemini's headless docs (account tier blocked live run) |
| mastracode | `mastracode --mode plan --prompt` | plan mode | default | ✅ live |
| opencode | `opencode run --agent plan` | builtin read-only plan agent | default (ids are provider-scoped) | ✅ live |
| pi | `pi --no-tools -p` | all tools disabled | default | ✅ live |
| copilot | `copilot -p` | default permissions (no `--allow-tool`) | default (no small tier in catalog) | help only (org policy blocked CLI) |
| vibe | `vibe --trust --agent plan -p` | builtin read-only plan agent | `devstral-small` | ✅ live |
| cursor-agent | `cursor-agent --trust --mode ask -p` | read-only ask mode | default (account model list rejects catalog's `composer-1`) | ✅ live |
| droid | `droid exec` | read-only default autonomy | default | help only ("Exec failed" at 0 turns on this account — account/config-level, fails fast in 458ms) |
| polygraph | — | — | — | no one-shot prompt mode |

## Manual QA Checklist

Verified end-to-end via CDP against the dev desktop app (host-service tRPC `workspaces.create` with an agent launch and no name/branch, real project), each confirmed by host-service log lines and/or host DB rows:

- [x] **Primary path** (credentials valid): claude launch → `named via small model` log, "Paginate audit log table" / `paginate-audit-log`, 5.4s create — agent CLI never invoked
- [x] **Fallback, claude** (both env keys replaced with well-formed invalid keys + mastracode auth removed): small model 401s → `named via agent CLI (claude)`, "Add keyboard navigation to the file tree" / `add-keyboard-nav`, 8.5s
- [x] **Fallback, codex** (same broken creds): → `named via agent CLI (codex)`, "Compress Terminal Scrollback Snapshots" / `compress-scrollback`, 15.6s
- [x] **Double failure**: before the env-strip fix, claude's fallback timed out and the workspace correctly kept a friendly-random name with the `autoNameWarning` toast — create never failed
- [x] Earlier agent-path runs (pre-inversion): claude ×2 + codex named correctly via agent CLI (7-11s creates)
- [x] Sugar agent launched normally in the created workspace alongside naming
- [x] Test workspaces destroyed cleanly afterwards; branches carried no stray commits

## Testing
- `bun run typecheck` (shared + host-service + desktop)
- `bun run lint`
- `bun test` — shared agent-settings suite (13 pass), host-service `ai-workspace-names.test.ts` (1 pass)

## Design Decisions
- **Small model primary, agent CLI fallback**: the direct call is ~1s vs 5-10s for a CLI cold-start, so users with working credentials see no latency change; the agent CLI only pays its cost when it's the difference between a real name and a random one.
- **Launch agent only as fallback — not a try-every-configured-agent chain**: a chain multiplies worst-case create latency and fires extra subscription-metered model calls; the launch agent is the user's explicit choice. (A chain version was implemented, live-verified, and deliberately reverted — see history.)
- **Plain child-process spawn, not the PTY daemon**: headless modes don't need a TTY, and parsing JSON out of a PTY stream (CRLF, wrapping, control sequences) is strictly worse. The login shell provides the PATH bootstrap that the PTY path would have given.
- **Locked-down invocations instead of mirroring the interactive posture**: the naming call has no reason to act, so every command drops permission bypasses and pins a no-tools/read-only form (enforcement, on top of the data-tag prompt hardening which is only persuasion). All seven locally-runnable CLIs were re-verified to return naming JSON in locked form, and an adversarial prompt instructing Bash/Write tool use produced correct JSON with zero files created.

## Known Limitations
- When the fallback runs, naming adds ~4-8s to workspace creation (bounded at 20s by timeout); the primary path is unchanged (~1s).
- If the user's login shell profile exports an invalid `ANTHROPIC_API_KEY`, the claude fallback inherits it through `$SHELL -lc` despite the spawn-env strip; their claude CLI is equally broken in a terminal, and naming then degrades to friendly-random.
- `workspaces.aiRename` and `namingPrompt`-only creates have no launch agent, so they use the small-model path only.
- Custom agents can declare `nonInteractiveCommand` in their definition, but there's no settings UI field for it yet, and builtin preset overrides can't override it.
- `workspaces.generateBranchName` (branch-name preview) still uses the small model only.

### DIFF
--- a/packages/host-service/src/trpc/router/workspace-creation/utils/ai-workspace-names.test.ts
+++ b/packages/host-service/src/trpc/router/workspace-creation/utils/ai-workspace-names.test.ts
@@ -40,8 +40,9 @@ mock.module("@superset/chat/server/shared", () => ({
 	getSmallModel: getSmallModelMock,
 }));
 
-const { generateWorkspaceNamesFromPrompt, orderNamingCandidates } =
-	await import("./ai-workspace-names");
+const { generateWorkspaceNamesFromPrompt } = await import(
+	"./ai-workspace-names"
+);
 
 describe("generateWorkspaceNamesFromPrompt", () => {
 	beforeEach(() => {
@@ -79,31 +80,6 @@ describe("generateWorkspaceNamesFromPrompt", () => {
 				branchName: { type: "string" },
 			},
 		});
-	});
-});
-
-describe("orderNamingCandidates", () => {
-	test("puts the preferred preset first and dedupes it from the configured list", () => {
-		const candidates = orderNamingCandidates(["claude", "codex"], "codex");
-		expect(candidates.map((c) => c.presetId)).toEqual(["codex", "claude"]);
-	});
-
-	test("keeps configured display order without a preferred preset", () => {
-		const candidates = orderNamingCandidates(["codex", "claude"]);
-		expect(candidates.map((c) => c.presetId)).toEqual(["codex", "claude"]);
-	});
-
-	test("skips presets without a headless mode and unknown/custom ids", () => {
-		const candidates = orderNamingCandidates(
-			["polygraph", "custom:my-agent", "superset", "claude"],
-			"custom:my-agent",
-		);
-		expect(candidates.map((c) => c.presetId)).toEqual(["claude"]);
-	});
-
-	test("resolves each candidate to its builtin nonInteractiveCommand", () => {
-		const [claude] = orderNamingCandidates(["claude"]);
-		expect(claude?.command).toBe("claude --dangerously-skip-permissions -p");
 	});
 });
 
--- a/packages/host-service/src/trpc/router/workspace-creation/utils/ai-workspace-names.test.ts
+++ b/packages/host-service/src/trpc/router/workspace-creation/utils/ai-workspace-names.test.ts
@@ -40,9 +40,8 @@ mock.module("@superset/chat/server/shared", () => ({
 	getSmallModel: getSmallModelMock,
 }));
 
-const { generateWorkspaceNamesFromPrompt } = await import(
-	"./ai-workspace-names"
-);
+const { generateWorkspaceNamesFromPrompt, orderNamingCandidates } =
+	await import("./ai-workspace-names");
 
 describe("generateWorkspaceNamesFromPrompt", () => {
 	beforeEach(() => {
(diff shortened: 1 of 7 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define non-interactive command schema for agents](https://stagereview.app/superset-sh/superset/pull/5742/chapters/1) |
| 2 | [Configure non-interactive commands for builtin agents](https://stagereview.app/superset-sh/superset/pull/5742/chapters/2) |
| 3 | [Implement agent CLI fallback for workspace naming](https://stagereview.app/superset-sh/superset/pull/5742/chapters/3) |
| 4 | [Wire agent context into workspace creation](https://stagereview.app/superset-sh/superset/pull/5742/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/5742">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 424a384298c3d80b9f477eecf51cdcb83a6d31ae on Jul 17, 2026 11:20pm UTC.</sub>

PR #5743, opened 2026-07-17
fix(desktop): start the local host service from the session, not Electric sync

Kitenite:
Follow-up to #5734's investigation. While diagnosing #5729 we found onboarding stuck on "Connecting…" whenever the Electric sync couldn't deliver the `organizations` collection (container down at boot, or a terminal 4xx after a dev-DB reseed). Root cause: a purely local capability — spawning the local host service — was gated on a cloud sync round-trip.

## Summary
- **Host-service startup no longer involves Electric at all.** Main starts services for every previously-hosted org (`$SUPERSET_HOME_DIR/host/*`) at boot and on sign-in; the renderer keeps a single session-keyed start covering only a brand-new org with no host dir yet.
- **`LocalHostServiceProvider` is now Electric-free**: the `useLiveQuery(organizations)` start loop is deleted and `activeOrganizationName` comes from `authClient.useActiveOrganization()` instead of the synced collection.
- A terminally stopped Electric stream (non-401 4xx) now emits an `electric_sync_stopped` telemetry event and an actionable "Cloud sync stopped" toast with a Reload action, instead of a silent `console.error`.
- The onboarding project step shows a reload hint after 15s of "Connecting…" (via `useDelayElapsed` from #5734) instead of waiting silently forever.

## Why / Context
Design rule this enforces (and that makes the eventual Electric removal cheaper): **cloud sync may enrich, it must never be load-bearing for local function.** The second commit goes further than guarding the gate — it deletes the renderer→Electric edge for host-service lifecycle entirely. `coordinator.start` needs nothing from the renderer: main already loads the auth token itself and knows the org list from its own host directories.

Starting services for *all* known orgs (not just the active one) is preserved deliberately — cross-org workspace-session disposal (`getConnections` consumer) and background relay reachability depend on it. Main-side boot startup also closes a known gap: host services previously only started when a renderer attached, so port detection had nothing to reconcile against at boot.

## How It Works
- `HostServiceCoordinator.startAllKnown(config)`: enumerates UUID-named dirs under `$SUPERSET_HOME_DIR/host/`, starts each (idempotent, failures logged per-org, never throws).
- `main/index.ts`: calls it non-blocking after daemon reconcile at boot, and again on the `token-saved` auth event (sign-in after boot).
- `activeHostUrl` derivation is untouched — it always came from the local `getConnection` poll, never Electric.

## Manual QA Checklist
- [x] With the worktree's Electric container **stopped mid-boot**: `getConnections` returns **6 live host services — one per known org dir** (verified over CDP). The 5 non-active orgs can only have been started by the main reconcile, since the renderer starts just the active org and the Electric org list was unreachable. Previously this state produced the indefinite "Connecting…".
- [x] Renderer boots and serves the workspace UI normally in the same run; no error dialogs or toasts.
- [x] Electric restored afterwards; normal boot unaffected.
- [ ] Fresh-machine first sign-in (no host dirs at all) — not runnable in this dev profile; covered by the renderer's session-keyed start, which depends only on the session.

## Testing
- `bun run lint` (exits 0)
- `bunx tsc --noEmit` in `apps/desktop` (clean)

## Known Limitations
- `startAllKnown` starts services for historic orgs the user may have left; the host service's own cloud registration handles rejection, and local terminals for those orgs' workspaces still need disposal, so the superset is intentional.
- Other Electric-gated behavior elsewhere is tracked by SUPER-1543 (behavior-gating vs render-only inventory); SUPER-1542 tracks dev self-heal for post-reseed terminal 4xxs.

## Summary by CodeRabbit

* **New Features**
  * On the onboarding project page, if the local service is still connecting after 15 seconds, an inline notice appears with a **Reload** button (and guidance if it persists).
  * The app now attempts to automatically start known local services after startup and again when auth tokens are saved.
* **Bug Fixes**
  * Cloud sync interruptions now surface a “Cloud sync stopped” toast with a **Reload** action, and sync stop events are recorded for diagnostics.
  * Local service startup is now tied to the currently active context for more consistent behavior.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -12,6 +12,7 @@ import {
 } from "electron";
 import { makeAppSetup } from "lib/electron-app/factories/app/setup";
 import {
+	authEvents,
 	handleAuthCallback,
 	loadToken,
 	parseAuthDeepLink,
@@ -405,6 +406,24 @@ if (!gotTheLock) {
 		// Must happen before renderer restore runs
 		await reconcileDaemonSessions();
 		prewarmTerminalRuntime();
+
+		// Host services for previously-hosted orgs start from main, so
+		// background reachability and port detection never wait on a renderer
+		// or cloud sync. Non-blocking: boot must not wait on spawns.
+		const startKnownHostServices = async () => {
+			try {
+				const { token } = await loadToken();
+				if (!token) return;
+				await getHostServiceCoordinator().startAllKnown({
+					authToken: token,
+					cloudApiUrl: mainEnv.NEXT_PUBLIC_API_URL,
+				});
+			} catch (error) {
+				console.error("[main] host-service boot reconcile failed:", error);
+			}
+		};
+		void startKnownHostServices();
+		authEvents.on("token-saved", () => void startKnownHostServices());
 
 		try {
 			setupAgentHooks();
--- a/apps/desktop/src/main/lib/host-service-coordinator.ts
+++ b/apps/desktop/src/main/lib/host-service-coordinator.ts
@@ -263,6 +263,36 @@ export class HostServiceCoordinator extends EventEmitter {
 	}
 
 	/**
+	 * Start host services for every org this machine has hosted before
+	 * ($SUPERSET_HOME_DIR/host/*). Runs at boot and on sign-in so background
+	 * reachability and port detection never wait for a renderer or cloud sync
+	 * to name orgs; a brand-new org (no dir yet) is started by the renderer
+	 * from its session.
+	 */
+	async startAllKnown(config: SpawnConfig): Promise<void> {
+		const hostRoot = path.join(SUPERSET_HOME_DIR, "host");
+		let entries: fs.Dirent[];
+		try {
+			entries = fs.readdirSync(hostRoot, { withFileTypes: true });
+		} catch {
+			return;
+		}
+		const orgIdPattern = /^[0-9a-f]{8}-[0-9a-f-]{27}$/i;
+		await Promise.allSettled(
+			entries
+				.filter((e) => e.isDirectory() && orgIdPattern.test(e.name))
+				.map((e) =>
+					this.start(e.name, config).catch((error) => {
+						log.warn(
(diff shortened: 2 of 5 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Enable starting all known host services](https://stagereview.app/superset-sh/superset/pull/5743/chapters/1) |
| 2 | [Trigger host service startup from main process](https://stagereview.app/superset-sh/superset/pull/5743/chapters/2) |
| 3 | [Decouple LocalHostServiceProvider from Electric sync](https://stagereview.app/superset-sh/superset/pull/5743/chapters/3) |
| 4 | [Improve error reporting for stopped sync](https://stagereview.app/superset-sh/superset/pull/5743/chapters/4) |
| 5 | [Add connection timeout hint to onboarding](https://stagereview.app/superset-sh/superset/pull/5743/chapters/5) |

<a href="https://stagereview.app/superset-sh/superset/pull/5743">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit cf8196b96c913c4761dab6ef47fd195ca74c8f2f on Jul 18, 2026 5:30pm UTC.</sub>

Kitenite:
Re-verified with Electric off for the **entire session** (container stopped before launch; confirmed no dev task restarts it — it's only started by worktree setup):

- App boots signed-in to the full workspace UI with the container down from t=0 through the whole run
- `getConnections` reports **6 live host services — one per known org dir** — all started by the main-process reconcile with zero cloud sync available
- No error toasts: with the container down the Electric client's internal network retry applies, so the new "Cloud sync stopped" toast correctly does not fire (it's reserved for terminal 4xx)

Kitenite:
All three review findings addressed in cf8196b96:

- **`fs.readdirSync` on the main thread** (CodeRabbit) — now `await fs.promises.readdir`.
- **Silent catch on host-root read** (cubic P2) — first-run `ENOENT` stays quiet; any other I/O error is logged with the path so the boot-recovery path is observable.
- **Stacking sync-stopped toasts** (greptile) — the toast now carries `id: "electric-sync-stopped"`, so concurrent dead shapes collapse into one notification. Removed from Known Limitations.

PR #5449, opened 2026-07-03
fix(desktop): derive terminal agent status from host bindings (single source of truth)

Kitenite:
**Links**
- ExecPlan: `apps/desktop/plans/done/20260703-1507-derive-terminal-agent-status-from-host-binding.md`
- Supersedes #5443 (close it when this merges)

## Summary
- Terminal agent status (working/permission/idle/review) is now **derived** from the host's `TerminalAgentBinding` instead of being replayed into a localStorage-persisted renderer status map. Stale "working" dots after an app restart become structurally impossible — there is no stored copy left to go stale.
- Host-side, binding **visibility** is derived from session liveness: `listByWorkspace`/`findActive` join `terminal_sessions` on `status = 'active'`, so agents in dead terminals (kill -9, crash, host downtime) disappear from reads with zero prune bookkeeping. This is the root fix for the leak #5443 patched with compensating layers.
- Net **−378 lines** in tracked files; the renderer notification store shrinks to two maps (`manualUnread`, `terminalSeenAt`).

## Why / Context
The binding store and the renderer status map were both projections of the same agent-hook stream, coupled only at the source (`notifications.hook` broadcasts the WS event and records the binding in one mutation). The renderer projection only updated while a window was open and connected, and was persisted with no reconciliation — any missed `Stop`/`Detached` became permanent drift: phantom "working" dots, workspace indicators lit for terminals that no longer exist.

## How It Works
- `working` ⇔ `binding.lastEventType === "Start"`, `permission` ⇔ `"PermissionRequest"`, `idle` ⇔ `"Stop"/"Attached"` — a pure function (`deriveTerminalAgentStatus`) over the binding the renderer already fetches.
- `review` = the host recorded a `Stop` newer than a renderer-persisted per-terminal `lastSeenAt`. Seen marks are written when a pane is focused, when an event lands while its target is visible, on interrupt/close, and by the sidebar mark-read action. Because it's computed from the host's `lastEventAt`, review survives the missed-event case (agent finishes while the app is closed → correct green dot on relaunch).
- Interrupt (Esc/Ctrl+C) POSTs a synthetic `Stop` to the existing `notifications.hook` endpoint (agents fire no Stop hook on interrupt), gated on the binding actually being working/permission so vim-style Escape spam doesn't spam the host.
- The Electron fallback for adopted shells (stale hook URLs) forwards events to the host so the binding — now the source of truth — converges instead of being bypassed.
[...]

PR #5744, opened 2026-07-17
fix(desktop,host-service): make Clear Status clear wedged agent working state

Kitenite:
## Problem

Workspaces get stuck showing **running** with no recovery. Since #5449 the working dot derives from `terminal_agent_bindings.lastEventType` on the host; Clear Status only cleared manual-unread + review marks.

Debugged live on a wedged workspace and reproduced in a scripted tmux session against claude 2.1.212: **Esc-interrupt fires no hook at all** (verified empirically — last hook is `PostToolUse`, then silence), and a turn that dies silently (stream drop, hang) fires nothing either. #5449 already covers the interrupt-in-a-focused-pane case via `useTerminalInterruptClear`'s synthetic Stop, but silent turn deaths have no keypress to detect — the binding stays `Start` forever (shell outlives the agent, so the session-liveness join never hides it), and nothing can clear it.

## Why not zero new code

Considered and rejected:
- **Reuse `notifications.hook` synthetic Stops for Clear Status**: the hook path broadcasts `agent:lifecycle`, and for a non-visible pane `shouldSuppress` returns false → completion chime + native notification on a status you're explicitly clearing.
- **TTL / staleness heuristics on `Start`**: a genuinely working agent can go >30 min between hook events (one long tool call); any timeout wedges the opposite direction.
- **Deleting the Start→working derivation**: removes the feature.

## Fix (one clearing mechanism, not three)

- `TerminalAgentStore.clearWorkspaceStatuses(workspaceId, terminalId?)`: force non-`Stop` bindings to `Stop`, preserving `lastEventAt` so existing seen marks resolve to idle (no transient review). Memory + sqlite, `change` emitted only on real change.
- New silent `terminalAgents.clearWorkspaceStatuses` mutation (deliberately not a hook event — must not chime).
- Sidebar **Clear Status** calls it and invalidates the bindings query.
- `useTerminalInterruptClear` now calls the same mutation instead of posting a synthetic Stop through the public hook endpoint — removing that special case and its side effect (interrupts today chime/notify in any other window where the pane isn't visible).

Clearing a genuinely busy agent self-corrects: its next hook event (PostToolUse fires per tool call) re-asserts working.

## Testing

- Store tests: force-to-Stop with `lastEventAt` preserved, per-terminal scoping, workspace isolation, no-op emits no change. Verified the test fails against a broken impl.
- `bun test` (20 pass), `bun run typecheck`, `bun run lint` clean. Renderer→router wiring is type-checked (client types derive from the router).
- Interrupt repro: tmux-driven claude 2.1.212, Esc mid-turn → no hook fired (hooks debug log), confirming the gap this closes.

## Summary by CodeRabbit

* **New Features**
  * Added an option to clear stuck workspace agent statuses, with support for scoping to a specific terminal when provided.
  * Workspace status indicators now reset reliably across related terminal sessions.

* **Bug Fixes**
  * Improved recovery from agents incorrectly remaining in working/permission states by forcing affected bindings to a stopped state while preserving timestamps.
  * Added toast error feedback when status clearing fails, and ensured the UI refreshes after a successful clear.

* **Tests**
  * Added test coverage to verify correct workspace scoping and idempotent behavior for status clearing.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/components/DashboardSidebarWorkspaceItem/hooks/useDashboardSidebarWorkspaceItemActions/useDashboardSidebarWorkspaceItemActions.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/components/DashboardSidebarWorkspaceItem/hooks/useDashboardSidebarWorkspaceItemActions/useDashboardSidebarWorkspaceItemActions.ts
@@ -1,10 +1,12 @@
 import { toast } from "@superset/ui/sonner";
+import { useQueryClient } from "@tanstack/react-query";
 import { useMatchRoute, useNavigate } from "@tanstack/react-router";
 import { useState } from "react";
 import {
 	useMarkWorkspaceTerminalsSeen,
 	useV2WorkspaceIsUnread,
 } from "renderer/hooks/host-service/useV2NotificationStatus";
+import { useWorkspaceHostUrl } from "renderer/hooks/host-service/useWorkspaceHostUrl";
 import { useCopyToClipboard } from "renderer/hooks/useCopyToClipboard";
 import { getHostServiceClientByUrl } from "renderer/lib/host-service-client";
 import { showHostServiceUnavailableToast } from "renderer/lib/host-service-unavailable";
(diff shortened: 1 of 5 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Implement status clearing in TerminalAgentStore](https://stagereview.app/superset-sh/superset/pull/5744/chapters/1) |
| 2 | [Expose clearWorkspaceStatuses via TRPC router](https://stagereview.app/superset-sh/superset/pull/5744/chapters/2) |
| 3 | [Wire Clear Status action in sidebar](https://stagereview.app/superset-sh/superset/pull/5744/chapters/3) |
| 4 | [Update terminal interrupt handler to use mutation](https://stagereview.app/superset-sh/superset/pull/5744/chapters/4) |

<a href="https://stagereview.app/superset-sh/superset/pull/5744">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 34369e0e0beb1be27fc658a6b0a5f3e116abb0c8 on Jul 17, 2026 8:41pm UTC.</sub>

PR #5772, opened 2026-07-19
feat(desktop): add Antigravity CLI (agy) as a first-class agent

thebrownproject:
## Why

Google has deprecated the Gemini CLI and is pointing users at the [Antigravity CLI](https://antigravity.google/product/antigravity-cli). The individual-account client is *already* rejected at sign-in:

> This client is no longer supported for Gemini Code Assist for individuals. To continue using Gemini, please migrate to the Antigravity suite of products.

Homebrew's `gemini-cli` formula is marked deprecated with `brew install --cask antigravity-cli` as the replacement, and disables on 2026-12-18.

This adds Antigravity as its own builtin terminal agent so Superset's lifecycle notifications work with it.

**The existing `gemini` entry is left untouched.** Enterprise Code Assist users can still run the Gemini CLI until Google pulls it, so this is purely additive — no preset migration, nothing breaks for existing users.

## Two things that differ from the other agent integrations

**1. The agent id is `agy`, not `antigravity`.** `SUPERSET_MANAGED_BINARIES` derives the shell shim name directly from the agent id (`shell-wrappers.ts`), so the id has to match the binary Antigravity actually installs. Registering it as `antigravity` would generate a shim named `antigravity` while the wrapper file was written as `agy`, and the two would never line up — the agent would launch unwrapped and no hook would ever fire.

The alternative is adding a `binaryName` field to `DesktopAgentSetupTarget` and mapping it in both prelude builders. Happy to switch to that if you'd rather keep ids human-readable — it's a small change, but it touches plumbing shared by all 11 agents, so I went with the id that needs no core changes.

**2. The hook takes its event from argv.** Antigravity's hook payload carries no event-name field — the common fields are `conversationId` / `workspacePaths` / `transcriptPath` / `artifactDirectoryPath` / `modelName` only (camelCase, protojson). Every other agent's hook derives the event from stdin; this one can't, so `hooks.json` passes the Superset event as `$1`.

## Config location

Hook config goes to `~/.gemini/config/hooks.json` — Antigravity's global customization root, shared with Antigravity 2.0 and the IDE — **not** the `~/.gemini/settings.json` the Gemini CLI used. Antigravity namespaces hooks by name and merges every named block for an event, so Superset owns a single `"superset"` key and leaves user- and plugin-defined hooks untouched. That makes the merge considerably simpler than the Gemini per-event reconciliation.

## Event mapping

| Antigravity event | Superset event | Why |
|---|---|---|
| `PreInvocation` | `Start` | Fires before each model call |
| `PostToolUse` | `Start` | Keeps the terminal marked busy across long tool loops |
| `Stop` | `Stop` | Fires when the execution loop terminates |

`PreToolUse` is deliberately **not** registered. Its contract requires a `decision` field (`allow` / `deny` / `ask` / `force_ask`), and emitting an empty object there risks gating the user's own tool calls. There's a test asserting we never register it.

Antigravity runs hooks synchronously and blocks the agent loop on them, so the hook prints its JSON result before doing any network work, and every event we register treats `{}` as a safe no-op (`PreInvocation` treats `injectSteps` as optional, `PostToolUse` expects exactly `{}`, and `Stop` only blocks termination when `decision` is `"continue"`).

## Testing

- `bun run typecheck` — 35/35 tasks
- `bun test apps/desktop/src/main/lib/agent-setup/` — 79 pass
- `bun test packages/shared/src` — 704 pass
- `bun run lint` — exit 0

New tests cover the wrapper's agent-id stamping, the `~/.gemini/config` path, the argv event contract for all three registered events, the PreToolUse exclusion, idempotency, foreign-hook preservation, and recovery from an unparseable existing file. The Antigravity template is also added to the shared "dispatches to v2 first, falls back to v1" contract test.

## Not included

- Marketing compatibility surfaces and `apps/mobile/store.config.json` — happy to add if you want them in this PR.
- Not added to `DEFAULT_V2_TERMINAL_PRESET_IDS`, matching how `gemini` is treated. Seeding a preset for a CLI the user hasn't installed shows a broken-looking entry.
- No `AGENT_MODEL_SUPPORT` entry. Antigravity exposes `--model`, but the model names are display strings (e.g. `"Gemini 3.5 Flash (Low)"`) resolved per-account via `agy models`, so a static catalog would go stale. Worth a follow-up if you want model selection in the picker.

The icon is the official Antigravity mark.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5772">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Add Antigravity CLI (`agy`) as a first-class terminal agent so Superset can track lifecycle events for Google’s replacement to the deprecated Gemini CLI. The existing `gemini` agent remains unchanged.

- New Features
  - Added `agy` agent with a managed wrapper, preset, and icon.
  - Hook config at `~/.gemini/config/hooks.json` under a dedicated key; preserves other user/plugin hooks.
  - Events passed via argv; maps `PreInvocation`/`PostInvocation` → `Start`, `Stop` → `Stop`; skips `PreToolUse` and `PostToolUse` (the latter can fire after `Stop`).
  - Wrapper id matches the binary name (`agy`); docs updated to list the Antigravity preset and note `gemini` is deprecated upstream.

- Bug Fixes
  - Use `PostInvocation` (not `PostToolUse`) to prevent panes from getting stuck in a “working” state.
  - Skip writing `hooks.json` when the existing file is unparseable or not an object, avoiding clobbering other hooks.

<sup>Written for commit 2929c5b3a3fedcde9cdc9b27b5222e689726ea2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added Antigravity (`agy`) as a built-in terminal agent with default presets, settings support, and a dedicated preset icon.
  * Introduced Antigravity setup for generating the wrapper, lifecycle hook script, and hooks configuration.
  * Added Antigravity to the terminal “quick-add” preset list and agent icon picker.
* **Bug Fixes**
  * Preserves existing hook configuration and avoids clobbering `hooks.json` when it can’t be parsed.
* **Tests**
  * Expanded coverage for wrapper behavior, hook event dispatching, hooks merging/idempotency, and template fallback.
* **Documentation**
  * Updated terminal preset docs to include Antigravity and note Gemini CLI status.


ghost:

| | Title |
|---|-------|
| 1 | [Define Antigravity agent and documentation](https://stagereview.app/superset-sh/superset/pull/5772/chapters/1) |
| 2 | [Add Antigravity icons and UI options](https://stagereview.app/superset-sh/superset/pull/5772/chapters/2) |
| 3 | [Implement Antigravity hook template](https://stagereview.app/superset-sh/superset/pull/5772/chapters/3) |
| 4 | [Implement agent wrapper and hook logic](https://stagereview.app/superset-sh/superset/pull/5772/chapters/4) |
| 5 | [Wire Antigravity into desktop setup lifecycle](https://stagereview.app/superset-sh/superset/pull/5772/chapters/5) |
| 6 | [Other changes](https://stagereview.app/superset-sh/superset/pull/5772/chapters/6) |

<a href="https://stagereview.app/superset-sh/superset/pull/5772">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2929c5b3a3fedcde9cdc9b27b5222e689726ea2e on Jul 19, 2026 8:30am UTC.</sub>

thebrownproject:
Ran an adversarial review over this branch (5 reviewers by dimension, every finding handed to an independent agent tasked with refuting it). Two defects survived, both now fixed in 08d0772.

**1. `PostToolUse` fires *after* `Stop`** — this was a real bug in the original push, and it broke the entire point of the integration.

A verifier installed the exact `hooks.json` this PR generates and ran the real `agy` 1.1.4 with a probe logging argv + epoch. On any turn ending without a tool call:

```
307.589 PreInvocation
307.630 PostToolUse    stepIdx:1  toolCall:null
309.117 PostInvocation
309.157 Stop           terminationReason:NO_TOOL_CALL  fullyIdle:true
310.026 PostToolUse    stepIdx:3  toolCall:null   <-- 0.87s AFTER Stop
```

Since `PostToolUse` maps to `Start`, the last event Superset saw was `Start`. `recordEvent` is last-write-wins on `lastEventType` and `deriveTerminalAgentStatus` has no staleness gate on the `Start` branch, so the pane stayed pinned to "working" indefinitely after the agent finished replying — recoverable only via `clearWorkspaceStatuses`. Reproduced 4+ times, under an interactive pty as well as `--print`.

Two of my original assumptions were also just wrong: `matcher: "*"` matches *every* step type, not only tool calls (tool names are the step type minus the `CORTEX_STEP_TYPE_` prefix, and the enum includes `PLANNER_RESPONSE`/`FINISH`), and those trailing steps carry `toolCall: null`. So `PostToolUse` was never the tool-loop keepalive I described it as.

Now registers `PostInvocation`, which the same traces show always precedes `Stop` and fires once per loop iteration — same intent, no trailing `Start`. There's a regression test asserting `PostToolUse` is never registered, with the reason.

Worth flagging for the other integrations: `PostToolUse -> Start` is shared with Claude/codex/opencode/droid/mastra via `map-event-type.ts`. Those agents don't emit a tool event after their Stop hook, so they're fine today — but the `Start` branch having no staleness gate means any agent that ever does will latch the same way.

**2. Unreadable `hooks.json` was rewritten wholesale.** The merge seeded `{}` and fell through on parse failure, so a truncated or hand-broken file got replaced by the Superset block alone, discarding user and plugin hooks. Since `~/.gemini/config/` is shared by the CLI, Antigravity 2.0 and the IDE, that hit every surface at once. Now returns `null` and the writer skips — matching the `readExistingClaudeSettings` / `readExistingCodexHooks` / `readExistingDroidSettings` contract the newer integrations already use; I'd picked up the older Cursor/Gemini/Mastra pattern by copying the wrong sibling.

The test for that path was also theatre — it asserted only that our own key existed, so it passed whether or not user hooks survived. It now covers unparseable, top-level-array and `null` files and asserts the file is left byte-identical.

Everything else checked out against the binary's embedded docs: the `agy` agent id, argv-passed events, the `~/.gemini/config/hooks.json` path, and omitting `PreToolUse`.

`typecheck` 35/35, agent-setup tests 82 pass, `lint` exit 0.

Kitenite:
Antigravity is now on main — landed via #6767 (squash 10527b69e) with your Co-authored-by credit. Thank you again @thebrownproject; the hooks.json lifecycle wrapper from this PR remains a welcome follow-up now that agent setup lives in packages/agent-setup. This PR can be closed whenever you like.

PR #5777, opened 2026-07-19
chore(desktop): fleet memory telemetry — periodic resource_snapshot to PostHog

Kitenite:
## What

Adds a privacy-safe aggregate memory telemetry sampler in the desktop **main process** (SUPER-1550). Every ~5 minutes it emits one flat, numbers-only `resource_snapshot` PostHog event so we finally have fleet memory data (renderer RSS vs process/window/webview counts, release-over-release regression detection, before/after proof for perf fixes). ~12 tiny events/user/hour.

Closes SUPER-1550.

## Schema — `resource_snapshot`

All properties are finite numbers. Enforced by the `RESOURCE_SNAPSHOT_KEYS` allowlist; emission filters the payload to exactly these keys.

| key | source |
| --- | --- |
| `uptime_seconds` | `process.uptime()` (rounded) |
| `process_rss_bytes`, `process_heap_used_bytes`, `process_heap_total_bytes`, `process_external_bytes`, `process_array_buffers_bytes` | `process.memoryUsage()` |
| `electron_main_rss_bytes`, `electron_renderer_rss_bytes`, `electron_other_rss_bytes`, `electron_total_rss_bytes` | `app.getAppMetrics()` by process class (KB→bytes) |
| `electron_main_process_count`, `electron_renderer_process_count`, `electron_other_process_count`, `electron_process_count` | `app.getAppMetrics()` counts |
| `window_count` | `BrowserWindow.getAllWindows().length` |
| `web_contents_count` | `webContents.getAllWebContents().length` |

(plus the standard `track()` context: `app_name`, `platform`, `desktop_version`.)

## Privacy

- **Hard allowlist**: `emitResourceSnapshot` copies only allowlisted keys whose value is a finite number. Anything else is structurally impossible to emit. A test asserts no key matches `/id|name|path|command|repo|url|user|org|title/i`.
- **No** IDs, names, paths, commands, or repo/terminal/user data — deliberately excluded terminal/workspace/org counts to avoid coupling to those subsystems and keep scope narrow.
- **No `ps` subprocess**: `app.getAppMetrics()` reads Electron internals; `process.memoryUsage()` is in-process. Never spawns per-sample process scans (does not touch the existing `resource-metrics` `ps` path).

## Perf / lifecycle

- Self-rescheduling `setTimeout` at ~5 min with up to 60s per-tick jitter (fleet doesn't emit in lockstep).
- `.unref()`ed — never keeps the process alive.
- Idempotent `startMemoryTelemetry()` (dup-timer guard) so startup + HMR re-entry can't double-schedule.
- `stopMemoryTelemetry()` wired into `before-quit` cleanup; `emit` is wrapped so a telemetry error can never crash the app.
- Gated off in test/dev and honors `DEFAULT_TELEMETRY_ENABLED`; `track()` additionally no-ops until a user is identified and a PostHog key is set.

`electron` is imported as a namespace and dereferenced lazily inside `defaultDeps`, so the module links in the test suite regardless of which per-file `electron` mock is active (a static named `import { webContents }` breaks when another test replaces the mock).

## Validation

- `bun test` (full desktop suite) — **2173 pass, 0 fail** across 221 files. The memory-telemetry unit tests (9) are fake-timer based and cover cadence, payload allowlist, disabled state, and cleanup/unref; each was verified to fail under a deliberate mutation (leaked key / broken cleanup).
- `bun run lint` — clean (0 warnings). `tsc --noEmit` — clean across all packages.
- **CDP (real dev app, this worktree)**: app boots with the telemetry wiring (Tray + notifications up, no crash); renderer renders the full signed-in workspace UI (no white screen); `resourceMetrics.getSnapshot` — which calls the same `app.getAppMetrics()` the sampler uses — returns valid live per-class RSS (main 425MB / renderer 735MB). The `resource_snapshot` emission itself is intentionally disabled in dev (gated, no PostHog key), so it is not observable in dev by design; that path is covered by the unit tests above.

## Summary by CodeRabbit

- **New Features**
  - Added privacy-safe, aggregated memory and resource usage telemetry to the desktop app.
  - Resource snapshots are collected periodically with an allowlisted set of numeric metrics.
  - Telemetry automatically starts on app startup and stops during app shutdown.

- **Reliability**
  - Telemetry emissions are guarded so failures won’t disrupt normal app operation.
  - Background sampling is prevented from keeping the app alive after shutdown.

- **Tests**
  - Added a comprehensive unit test suite covering snapshot building, emission rules, start/stop behavior, and enablement.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -36,6 +36,10 @@ import { getHostServiceCoordinator } from "./lib/host-service-coordinator";
 import { localDb } from "./lib/local-db";
 import { requestLocalNetworkAccess } from "./lib/local-network-permission";
 import {
+	startMemoryTelemetry,
+	stopMemoryTelemetry,
+} from "./lib/memory-telemetry";
+import {
 	initTanstackDbPersistence,
 	shutdownTanstackDbPersistence,
 } from "./lib/persistence/persistence";
@@ -225,6 +229,7 @@ app.on("before-quit", async (event) => {
 
 	isQuitting = true;
 	try {
+		stopMemoryTelemetry();
 		getHostServiceCoordinator().stopAll();
 		if (isDev || forceFullCleanup) {
 			await teardownTerminalHost();
@@ -446,6 +451,7 @@ if (!gotTheLock) {
 
 		await makeAppSetup(() => MainWindow());
 		setupAutoUpdater();
+		startMemoryTelemetry();
 		initTray();
 
 		const coldStartUrl = findDeepLinkInArgv(process.argv);
--- a/apps/desktop/src/main/lib/memory-telemetry/index.ts
+++ b/apps/desktop/src/main/lib/memory-telemetry/index.ts
@@ -1,0 +1,1 @@
+export * from "./memory-telemetry";
--- a/apps/desktop/src/main/lib/memory-telemetry/memory-telemetry.test.ts
+++ b/apps/desktop/src/main/lib/memory-telemetry/memory-telemetry.test.ts
@@ -49,6 +49,7 @@ function createDeps(overrides: Partial<MemoryTelemetryDeps> = {}): {
 
 afterEach(() => {
 	stopMemoryTelemetry();
+	jest.useRealTimers();
 });
 
 describe("buildResourceSnapshot", () => {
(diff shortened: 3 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Implement memory telemetry collection and sampling](https://stagereview.app/superset-sh/superset/pull/5777/chapters/1) |
| 2 | [Verify telemetry logic with unit tests](https://stagereview.app/superset-sh/superset/pull/5777/chapters/2) |
| 3 | [Wire memory telemetry into app lifecycle](https://stagereview.app/superset-sh/superset/pull/5777/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5777">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit a1502373c0abb815693b275ad6828b3274b84d25 on Jul 19, 2026 6:08pm UTC.</sub>

Kitenite:
Addressed review feedback in de3e0c68c:

- **greptile** — `emitResourceSnapshot` now honors `deps.isEnabled()` at the top, so the exported surface can't bypass the disabled gate.
- **greptile** — `normalizeBytes` uses `>= 0` so a genuine 0-byte reading is accepted rather than classified as invalid.
- **coderabbit** — consolidated `jest.useRealTimers()` into `afterEach` and dropped the per-test `try/finally` wrappers.

Skipped coderabbit's "make the `unref` mock return `this`": the sampler calls `sampleTimer.unref?.()` on its own line (no chaining), so the stored handle is never the return value of `unref()` — the mock returning `undefined` can't strand a timer.

Desktop suite 2173 pass / 0 fail, lint + typecheck clean.

PR #5778, opened 2026-07-19
perf(desktop): evict prior-org synced collections on org switch

Kitenite:
## What & why

Prior-org TanStack DB / Electric collections were **never freed**. The per-org `collectionsCache` Map in `collections.ts` had no delete path and `Collection.cleanup()` was never called, so every organization a user visited kept its **entire synced tables resident in the renderer heap for the whole session** (CDP-measured ~50 MB on the dev org alone; grows with org age and with each org a multi-org user visits).

This is **SUPER-1547a** — step 1 of [SUPER-1547](https://linear.app/superset-sh/issue/SUPER-1547). Steps 2 (trim preload) and 3 (window unbounded tables) are intentionally **out of scope** here.

## Change

- New pure helper `evictInactiveOrgs.ts` (dependency-free — no Electric/persistence/`window` imports, so it's unit-testable).
- `evictInactiveOrgCollections(activeOrgId)` in `collections.ts` calls it over the real cache.
- Triggered from the `activeOrganizationId` effect in `CollectionsProvider` — the **single** trigger that also covers the two callers that set the active org directly (`create-organization`, `MemberActions`) and bypass `switchOrganization`.

For each **inactive** org: `cleanup()` every collection (stops sync, clears in-memory rows) then drop it from the cache so instances are GC'd.

## Safety properties (verified against compiled library source, `@tanstack/db@0.6.14`)

- **Supported public API** — `Collection.cleanup(): Promise<void>` ("stops sync and clears data"). No private internals.
- **Race-safe for rapid A→B→A** — `cleanup()` tears the ShapeStream down **synchronously** (the Promise is just a resolved wrapper); the active org is never evicted; a re-entered org gets **fresh** instances, never a half-disposed one.
- **Recoverable** — lifecycle allows `cleaned-up → loading`; `getCollections` rebuilds fresh on return.
- **Never destroys shared/unrelated state** — the shared `organizations` collection lives outside the cache; the persisted-wrapper cleanup calls per-collection `runtime.cleanup()`, never the shared coordinator's `dispose()` (that's main-process shutdown only).
- **Cache-first preserved (AGENTS.md §9)** — on-disk SQLite/localStorage rows are left in place, so switching back rehydrates cache-first. No preload/windowing changes.

## Tests

`evictInactiveOrgs.test.ts` — 5 scenarios: **identity** (active untouched), **eviction** (inactive fully cleaned + dropped), **same-org** (no-op), **failed switch** (one rejecting `cleanup()` doesn't abort the sweep; error routed), **rapid A→B→A** (active org never cleaned across the sequence). Each verified to fail under mutated impl.

## Verification notes

- ✅ `bun run lint` clean, `tsc --noEmit` clean, `bun test` (26 pass in this dir).
- ✅ **Drove the real `evictInactiveOrgCollections` end-to-end on this worktree's live dev stack via CDP** — before/after results (prior org `ready`+resident → `cleaned-up`+evicted, active org untouched, recovery to `ready`) are in the [CDP verification comment](https://github.com/superset-sh/superset/pull/5778#issuecomment-5016793545). Single-org dev account, so a synthetic prior org was used against the real Electric/persistence stack (active org kept — never evicted).

## Summary by CodeRabbit

* **Performance**
  * Improved organization switching by automatically clearing cached collection data for inactive organizations, keeping the active organization intact for faster continuity.
  * Reduced retained in-memory collection state so revisiting an organization reloads fresh collection instances.

* **Reliability**
  * Eviction/cleanup now tolerates failures: if cleanup for an inactive organization fails, eviction still completes for the rest and the error is reported without interrupting switching.

* **Tests**
  * Added coverage for inactive-organization eviction behavior, including rapid switching and cleanup failure scenarios.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/CollectionsProvider.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/CollectionsProvider.tsx
@@ -10,7 +10,11 @@ import {
 import { env } from "renderer/env.renderer";
 import { authClient } from "renderer/lib/auth-client";
 import { MOCK_ORG_ID } from "shared/constants";
-import { getCollections, preloadCollections } from "./collections";
+import {
+	evictInactiveOrgCollections,
+	getCollections,
+	preloadCollections,
+} from "./collections";
 
 type CollectionsContextType = ReturnType<typeof getCollections> & {
 	switchOrganization: (organizationId: string) => Promise<void>;
@@ -54,6 +58,14 @@ export function CollectionsProvider({ children }: { children: ReactNode }) {
 
 	useEffect(() => {
 		preloadActiveOrganizationCollections(activeOrganizationId);
+		// Once the active org is current (its collections are already cached by the
+		// `collections` memo above, which runs during render), evict every prior
+		// org's set to free the synced tables they hold. This effect is the single
+		// trigger for all switch paths, including callers that set the active org
+		// directly without going through `switchOrganization`.
+		if (activeOrganizationId) {
+			evictInactiveOrgCollections(activeOrganizationId);
+		}
 	}, [activeOrganizationId]);
 
 	const collections = useMemo(
--- a/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/collections.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/collections.ts
@@ -70,10 +70,7 @@ import {
 	type WorkspacesCreateInput,
 	workspaceLocalStateSchema,
 } from "./dashboardSidebarLocal";
-import {
-	type EvictableCollection,
-	evictInactiveOrgs,
-} from "./evictInactiveOrgs";
+import { evictInactiveOrgs } from "./evictInactiveOrgs";
 import { withReadHeal } from "./withReadHeal";
 
 const columnMapper = snakeCamelMapper();
@@ -970,15 +967,11 @@ export function evictInactiveOrgCollections(
 	activeOrganizationId: string,
 ): void {
 	evictInactiveOrgs(
-		// OrgCollections values are all `Collection`s exposing `cleanup()`.
-		collectionsCache as unknown as Map<
-			string,
-			Record<string, EvictableCollection>
-		>,
+		collectionsCache as unknown as Map<string, Record<string, unknown>>,
 		getCollectionsCacheKey(activeOrganizationId),
-		(orgKey, error) => {
+		(orgKey, collectionName, error) => {
 			console.error(
-				`[collections] Failed to clean up evicted collection for org ${orgKey}`,
+				`[collections] Failed to clean up evicted collection ${collectionName} for org ${orgKey}`,
 				error,
 			);
 		},
--- a/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/collections.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/collections.ts
@@ -70,6 +70,10 @@ import {
 	type WorkspacesCreateInput,
 	workspaceLocalStateSchema,
 } from "./dashboardSidebarLocal";
+import {
+	type EvictableCollection,
+	evictInactiveOrgs,
+} from "./evictInactiveOrgs";
 import { withReadHeal } from "./withReadHeal";
 
 const columnMapper = snakeCamelMapper();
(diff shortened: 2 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Implement pure eviction logic and tests](https://stagereview.app/superset-sh/superset/pull/5778/chapters/1) |
| 2 | [Expose eviction helper in collections module](https://stagereview.app/superset-sh/superset/pull/5778/chapters/2) |
| 3 | [Trigger eviction on organization switch](https://stagereview.app/superset-sh/superset/pull/5778/chapters/3) |

<a href="https://stagereview.app/superset-sh/superset/pull/5778">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit c384f1c394db841dc35b7c6e050c4bf9280f6451 on Jul 19, 2026 6:36pm UTC.</sub>

Kitenite:
## CDP verification (live dev stack, this worktree)

Launched this worktree's full dev stack (renderer `:7465`, API `:7461`, CDP `:9560`), signed in, and drove the **real exported `evictInactiveOrgCollections`** against the live Electric/SQLite-persistence stack. The dev account has a single org, so instead of a UI org-switch I kept the **real active org A** as the "keep" target and added a **synthetic prior org B** as the eviction target (this exercises the exact production code path and is safe — the active org is never evicted). Only B's localStorage-backed collections were preloaded, to avoid firing Electric auth requests/toasts for a synthetic org.

| Observation | BEFORE `evict(keepA)` | AFTER `evict(keepA)` |
|---|---|---|
| **Active org A** sample collection status | `idle` | `idle` (**unchanged — never touched**) |
| Prior org **B** `v2UserPreferences` status | `ready` | `cleaned-up` |
| Prior org **B** `v2SidebarProjects` status | `ready` | `cleaned-up` |
| Prior org **B** `tasks` (electric, idle) status | `idle` | `cleaned-up` |
| **B** in cache / instance | resident, tagged instance | fresh instance on re-get (tag gone) → **cache entry deleted** |

**Recovery (rapid A→B→A):** re-getting B returns a brand-new instance; `preload()` on it → status `ready` again. Recoverable.

**App health after probe:** active org still `1b8bd46c…` (A), route unchanged, no "Cloud sync stopped" toast, renderer fully rendered the authenticated workspace UI (1382 chars / 799 DOM nodes). No synthetic localStorage keys left behind.

This confirms end-to-end on the real stack: a prior org's synced collections are **stopped + cleared (`cleaned-up`) and dropped from the cache**, the **active org is untouched**, and eviction is **recoverable**. The BEFORE column (prior org `ready` + resident) is exactly the pre-fix retention behavior; the AFTER column is the fix's effect.

Kitenite:
## CDP verification #2 — real multi-org UI switch (live app instance)

Created a **second org**, then drove a real **A→B→A** switch through the app's own `switchOrganization`, observing the **app's live collection instances** (reached via React fiber, not a sandbox). The local dev seed has almost no Electric data (all synced tables `size 0`; only the localStorage `v2UserPreferences` has a real row), so I used that collection as the tracked example — the lifecycle is identical for every collection kind.

**A → B (real switch via `ctx.switchOrganization`):**

| org A `v2UserPreferences` | while A active | immediately after switching to B |
|---|---|---|
| status | `ready` | `cleaned-up` |
| size | `1` | `0` |

→ The prior org's collection was **stopped and cleared the instant the active org changed** (not waiting for the library's 5-min GC).

**B → A (switch back):** org A's `v2UserPreferences` rebuilt and **rehydrated cache-first from the untouched on-disk store → `ready`, size `1`** again. Recovery confirmed; fan-out reads return the correct rows after the round-trip.

**Isolation from the library's built-in `gcTime`:** collections with no live-query subscriber already reach `cleaned-up` on their own after 5 min — but the library keeps the instance in the cache. This fix additionally **deletes the entry from `collectionsCache`**, so switch-back yields a *fresh* instance (verified: 12 rapid create→preload→evict→recreate cycles each produced a fresh `idle` instance that re-preloaded to `ready`, active org untouched throughout).

**Fan-out consistency:** repeated `getCollections(activeOrg)` returns the *identical* instances (`tasks`/`projects`/`v2UserPreferences`/shared `organizations` all `===`), so all ~60 `useCollections()` consumers share one instance and cannot diverge.

App stayed healthy throughout; test org was deleted and the account restored to its original single org afterward.

> Note: no heap-delta numbers here because this local seed has ~no org-scoped Electric data (the ~50 MB figure was the production dev org). The retention/lifecycle behavior is what's demonstrated.

Kitenite:
Addressed both Greptile P2s in `c384f1c39`:

1. **Unsafe cast** (`collections.ts:976`) → the helper now runtime-guards each value with an `isEvictable()` check (`typeof value.cleanup === "function"`) and skips anything else, so a future non-collection field in `OrgCollections` can't crash the sweep. Dropped the `EvictableCollection` cast in favor of a generic `Map<string, T extends Record<string, unknown>>`.
2. **Untraceable error log** (`collections.ts:984`) → `onCleanupError` now receives the failing collection name, and the log reads `Failed to clean up evicted collection <name> for org <orgKey>`.

Added a test for the structural-safety guard (non-collection / null fields are skipped without throwing). All 6 tests pass; lint + typecheck clean.

Issue #5800, opened 2026-07-20
v2: option to auto-open agent session as a foreground pane on workspace open


leif-instalily:
## Problem

In the v2 workspace, agent sessions launched at workspace creation (e.g. `superset workspaces create --agent <id> --prompt …` from a Linear custom-script integration) land in **Background terminal sessions**. Opening the workspace shows the empty "Pick a workspace / Open Terminal" state, and the running agent is only reachable via the background-sessions dropdown.

For launch-and-watch workflows (kicking off an implementer agent from a Linear issue and wanting to see it work immediately), this adds a click every time and makes it look like no agent spawned at all — especially confusing for first-time setups.

## Request

A way to have the agent session open as a foreground terminal pane when the workspace opens. Any of these shapes would work:

- A per-agent or global setting, e.g. **Settings → Agents → "Open agent sessions in foreground"**
- A flag on the CLI, e.g. `superset workspaces create --agent <id> --foreground` or `superset workspaces open <id> --focus-agent`
- Defaulting to foreground when the workspace has exactly one live agent session

## Context

- v2 early access, desktop app + CLI v1.15.1, macOS
- Launch path: Linear "Work on issue → Custom script" → `superset workspaces create --local --agent <HostAgentConfig UUID> --prompt … --json` → `superset workspaces open <id>`
- v1 foregrounds the agent terminal on open, so this is a v2 behavior gap for interactive use; background-first is fine as the default for fan-out workflows.

PR #5808, opened 2026-07-20
fix(desktop): reliably auto-adopt agent terminals on workspace open

Kitenite:
## Problem

A running terminal agent launched outside the desktop — e.g. `superset workspaces create --local --agent <id> --prompt …` then `superset workspaces open <id>` from a Linear custom-script integration — shows as running in the sidebar but is stranded in the **Background terminal sessions** dropdown. Opening the workspace lands on the empty "Open Terminal" state, so it looks like no agent spawned. (Refs #5800.)

PR #5740 added an on-open pass that turns paneless running sessions into panes. This makes it fire reliably.

## Root cause

`useAutoAdoptBackgroundSessions` latched **one-shot per workspace** via `adoptedForWorkspaceIdRef`. On remount, React Query serves the **stale cached** `listSessions` while it refetches, and there is a render window where `data` is that stale list and `isFetching` hasn't yet flipped `true`. The latch fired on the stale list, missed the just-launched session, and never ran again — leaving the agent stranded in the background dropdown.

The `isFetchingSessions` guard added in follow-up couldn't close that window.

## Fix — remove the bad code

Drop the one-shot latch. Adoption is already idempotent and self-correcting:

- `getBackgroundTerminalSessions` filters sessions that already have a pane → no duplicates
- `focusOrAddTerminalPane` focuses an existing pane instead of adding a second → re-runs are no-ops
- so the effect can rerun freely as the session list settles and pick up a session that lands slightly after open

Safety preserved:
- deliberately **backgrounded** sessions stay in the `marked` set and are still skipped
- **normally-closed** panes kill their session (`usePaneRegistry.tsx`), so a closed terminal never reappears
- a fresh CLI-launched workspace has no tabs, so `restoreActiveTabId` is null and the adopted agent pane comes up **foreground** — exactly what #5800 asked for
- the in-flight-refetch guard is kept, now solely to avoid adopting a ghost pane from a stale cached list

Net: **−13 / +16** lines, one file.

## Testing

- `biome check` clean; `tsc --noEmit` clean (0 errors)
- ⚠️ **Not** verified end-to-end via CDP: no dev instance for this worktree is running (the only running desktop app belongs to a different worktree), and per the repo CDP rules I won't attach to another worktree's renderer. Happy to launch this worktree's stack and verify the CLI create→open flow before merge if wanted.

---
## Summary by cubic
Reliably auto-adopts CLI-launched agent terminal sessions into panes on workspace open, so users land on the agent pane instead of "Open Terminal". Aligns with #5800’s requirement to foreground the adopted agent pane.

- **Bug Fixes**
  - Removed the one-shot latch in `useAutoAdoptBackgroundSessions` so adoption can rerun and pick up sessions that arrive after open.
  - Kept the refetch guard to avoid adopting ghost panes from a stale cached list.
  - Adoption is idempotent: existing panes are filtered, re-runs only focus; marked background sessions are skipped; closed panes terminate sessions so they don’t reappear.

<sup>Written for commit 47edd2ac281bdebee4d7da5bdc59011c286a1542. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic adoption of background sessions as session data settles.
  * Prevented adoption while pane layout is still loading or session data is being refreshed.
  * Ensured sessions already marked or running in the background are skipped.
  * Removed the limitation that allowed adoption only once each time a workspace was opened.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/useAutoAdoptBackgroundSessions/useAutoAdoptBackgroundSessions.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/useAutoAdoptBackgroundSessions/useAutoAdoptBackgroundSessions.ts
@@ -1,6 +1,6 @@
 import type { WorkspaceStore } from "@superset/panes";
 import { workspaceTrpc } from "@superset/workspace-client";
-import { useEffect, useRef } from "react";
+import { useEffect } from "react";
 import { logStressEvent } from "renderer/lib/performance/stress-instrumentation";
 import { getTerminalBackgroundMarkerIdsKey } from "renderer/lib/terminal/terminal-background-intents";
 import type { StoreApi } from "zustand/vanilla";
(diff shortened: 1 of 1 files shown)


Kitenite:
## CDP end-to-end verification ✅

Drove this worktree's own dev build over CDP (debug port 9444, renderer `localhost:7645`, API `:7641` — verified it's this branch's instance, authed session). Deterministic before/after on the **same** v2 workspace (`adventurous-nannyberry`), **same** steps:

1. Open the workspace clean (0 sessions → initial adopt pass runs).
2. Create a background terminal session via the host-service (simulates `superset workspaces create --agent …` / a CLI-launched agent).
3. Refresh `listSessions` (open the background-terminals dropdown).
4. Observe whether auto-adopt surfaces it as a pane.

| observation | one-shot latch (before) | latch removed (this PR) |
|---|---|---|
| `createdSessionHasPane` (persisted layout) | `false` | `true` |
| `.xterm` panes mounted | `0` | `1` |
| `auto-adopt` stress log | none | `{"count":1}` |
| workspace view | empty "Open Terminal" state, session stuck in **"1 background terminal session"** chip | live **Terminal** pane, chip gone |

This is the reported failure (running session visible but stranded, not in tabs/panes) reproduced before the change and fixed after it. The one-shot latch is set *before* the empty-list early-return, so once the initial pass runs on an empty workspace it's armed for that mount and blocks adopting anything that arrives afterward — which is exactly what removing it fixes.

_(Verified against a throwaway scratch workspace; test session cleaned up afterward.)_

PR #5036, opened 2026-06-02
fix(import): restore git-init for non-git folders on v2 folder import

AviPeltz:
## Problem

Importing a folder that isn't a git repo yet dead-ends on `Not a git repository: <path>` in v2 — there's no offer to initialize one. Likely the root cause of #5033 ("why cannot import local git folder without remote?"), which is really about a *non-git* folder (remote-less git repos already import fine).

The throw happens at `project.findByPath` → `resolveLocalRepo` → `revParseGitRoot`, *before* `create importLocal` is reached.

## What this does

Detect → confirm → init + import. The folder is `git init`'d **only after the user confirms** (it writes into their directory), then imported as a local-only project — no remote required.

### Server (`packages/host-service`)
- **`resolve-repo.ts`** — `tryRevParseGitRoot` (non-throwing) + `initLocalRepoInPlace` (git-inits an *existing, populated* folder in place + empty initial commit so `ensureMainWorkspaceStrict` has a real branch; idempotent; a subdir of an existing repo resolves to the parent root). Exported `validateDirectoryPath`; `revParseGitRoot` now wraps the non-throwing variant.
- **`project.ts`** — `findByPath` returns an additive optional `needsGitInit: true` instead of throwing on a non-git folder (matches the existing "server returns a discriminated result, client branches" pattern). `create` `importLocal` gains `initIfNeeded`.
- **`handlers.ts`** — `createFromImportLocal` inits in place only when `initIfNeeded`. Cloud path unchanged (local-only repo → no clone URL, already supported by `empty`/`template`).

### Desktop
- New imperative `git-init-confirm` store + `GitInitConfirmDialog`, mounted once via `AddRepositoryModals`. `useFolderFirstImport` branches on `needsGitInit`, confirms, then creates with `initIfNeeded: true`. **All 5 hook consumers untouched** (the confirm is encapsulated in the hook/store rather than threaded through each call site).

## Design note
`initIfNeeded` is a flag on `importLocal` rather than a separate `initLocal` create mode, because the input shape (`{ repoPath }`) is identical. Easy to switch to a distinct mode if preferred — see the plan doc (`plans/20260601-v2-import-git-init-non-git-folder.md`).

## Out of scope / unchanged
- Remote-less git repos already import fine.
- `resolveGithubRepo` (the "no GitHub remote" throw) stays GitHub-feature-only (PRs/Issues).
- No auto-init without explicit confirmation.

## Testing
[...]

PR #4802, opened 2026-05-21
feat(desktop): v2 onboarding flow

saddlepaddle:
## What

A server-gated, v2-only onboarding flow for the desktop app. New users are pushed through it on first launch and it persists across devices via a single `users.onboardedAt` column.

### Flow
1. **Setup** — provider/tool checks as rows: GitHub CLI (required), Claude Code, Codex, + a "More providers" docs link. Each row's status derives directly from its tRPC query (`loading` → `connected` / `disconnected`); no hand-rolled state machine. Continue is gated on gh being installed + authed.
2. **Add a project** — Open a folder or clone a repo (inline URL, no picker — clones into `~/.superset/projects` via the host-service v2 clone, so no v1 project record). Adding a project finishes onboarding and opens the new-workspace modal pre-selected to it.

### Gating
- `users.onboardedAt timestamp` (nullable). Migration backfills existing users to `created_at` so only genuinely-new users see the flow.
- Surfaced through `customSession` (read explicitly — `additionalFields` alone didn't make the drizzle adapter return it). Typed end-to-end via `customSessionClient`.
- Route guards: `_authenticated/layout.tsx` redirects un-onboarded users into `/onboarding`; the onboarding layout redirects onboarded users out. Both wait for a resolved session to avoid redirect loops.
- `user.completeOnboarding` tRPC mutation; client refetches the reactive session (`disableCookieCache`) before navigating so the guards see the fresh value.

### Notable
- **Embedded terminal**: GitHub CLI "Sign in" runs `gh auth login` in an in-app xterm (reuses the existing terminal runtime + `electronTrpc.terminal.*` PTY procedures), so users don't leave the app.

## DB migration
`0055_add_onboarded_at` — adds `auth.users.onboarded_at` and backfills `= created_at`. Already applied to the working Neon branch; needs to run on prod.

## Test
- New user (`onboarded_at NULL`) → routed into `/onboarding`; complete it → lands on `/v2-workspaces` with the new-workspace modal.
- Existing user (backfilled) → no onboarding.
- Cross-device: flag is server-side, so a second device skips onboarding.

---

<a href="https://stagereview.app/superset-sh/superset/pull/4802">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
[...]

PR #5846, opened 2026-07-21
fix(desktop): mount git-init confirm dialog for onboarding folder import

CodeWithMoin:
## What & why

Picking a folder that isn't already a git repo during onboarding hangs the step
silently — no dialog, no error toast, and the **Browse…** buttons stay disabled
forever. The user is stuck on "Point Superset at some code" with no way forward.

`useGitInitConfirmStore.request()` returns a promise that is only ever settled by
a mounted `<GitInitConfirmDialog />`. That dialog was mounted by
`AddRepositoryModals`, which lives under the `_dashboard` route tree — but
onboarding is a *sibling* of `_dashboard`, not a child. So when onboarding's
folder-first import hit a non-git folder, it parked a promise that nothing in the
tree could resolve, leaving `busy === true` permanently.

This is a regression rather than an oversight: #4802 made onboarding reuse the
shared `useFolderFirstImport` hook, and #5036 later added the `requestGitInit`
step *into that shared hook* — correct for the dashboard caller, but silently
stranding the onboarding one.

It affects accounts on the v2 path (`useIsV2CloudEnabled`), which includes new
signups — precisely the people who see onboarding.

**Fix:** move the mount up to the shared `_authenticated` layout, which both
onboarding and `_dashboard` render through. It's a *move*, not an add, so the
dialog is never double-mounted (the store keeps a single module-level
`pendingResolve`, so two instances would render duplicate dialogs from one
request).

**Test:** the hook's existing unit tests mock `renderer/stores/git-init-confirm`
outright, so the mock resolves instantly and a missing mount is invisible to
them — which is why CI stayed green through #5036. Added a structural test
asserting the dialog is mounted in the shared layout, mounted exactly once, and
that every folder-import caller sits under a tree that reaches it. It lives at
`src/` rather than beside the routes because renderer code may not import Node
builtins (`biome.jsonc` `noRestrictedImports`).

## How I tested it

Ran the real dev app (`setup.local.sh` + `bun run dev:desktop`) and drove the
actual onboarding flow over CDP, using dispatched mouse events for the buttons
and macOS System Events for the native folder panel — no internal APIs
substituted for the UI interaction.

Repro folder: a real directory that is not a git repo and has no git ancestor.

**Before this change** — clicked Browse…, picked the non-git folder:

Onboarding frozen after picking a non-git folder — no dialog, Browse disabled

- stayed on `/onboarding/project`, no navigation
- `document.querySelectorAll("[role=alertdialog],[role=dialog]").length === 0`
- both **Browse…** buttons `disabled: true`, still stuck 15+ seconds later
- zero toasts, zero console errors — a completely silent hang

**After this change** — identical interaction:

Initialize git repository? dialog now appears

- the "Initialize git repository?" dialog renders as intended
- cancelling releases `busy` and re-enables the buttons (`[false, false]`),
  where previously the promise never settled either way

Also run locally:
- `bun test src/git-init-confirm-mount.test.ts` — fails without the fix
  (2 of 3 assertions), passes with it
- `bun run lint` — clean on the changed files
- `bun run typecheck` — 35/35 tasks pass

Note: `useFolderFirstImport.test.ts` currently fails on `main` independently of
this change (its `mock.module("react", …)` stubs React down to `useCallback`, so
`lucide-react` hits `react.forwardRef === undefined`). Verified pre-existing by
stashing this change and re-running against a clean tree. Happy to open a
separate PR for it.

## Checklist

---
## Summary by cubic
Fixes a silent hang in onboarding when picking a non‑git folder. The Git‑init confirm dialog now mounts in the shared `_authenticated` layout so `useGitInitConfirmStore.request()` resolves for both onboarding and `_dashboard`.

- **Bug Fixes**
  - Moved `<GitInitConfirmDialog />` from `_dashboard`’s `AddRepositoryModals` to `_authenticated/layout.tsx`, co-located it under `_authenticated/components`, and removed the old reference so it mounts once.
  - Strengthened `src/git-init-confirm-mount.test.ts` to walk the render graph and verify all folder‑import callers render under `_authenticated`, and that the dialog is mounted exactly once.
  - Updated dialog and store docblocks to name the `_authenticated` layout as the mount site and document the failure mode if a caller renders outside it.

<sup>Written for commit 8e9e1c95cf648dce33262598b65ac4f11914a48a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5846?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured the Git repository initialization confirmation dialog is mounted at the authenticated top level so folder-first import flows always resolve the pending confirmation correctly.
- **Documentation**
  - Updated inline documentation to reflect the dialog’s new placement and mounting expectations.
- **Tests**
  - Added coverage to statically verify the dialog is mounted exactly once and remains reachable from the relevant “request Git init confirm” callers.


CodeWithMoin:
Thanks @coderabbitai — both findings were valid, fixed in 5bda656.

**Render ancestry (`git-init-confirm-mount.test.ts`)** — you were right that `startsWith("commandPalette")` asserted the conclusion rather than checking it. The test now walks the render graph upward: for a caller outside `routes/_authenticated`, it resolves which files render that component and recurses until it reaches the authenticated tree. The real chain is `FolderImportMount` → `CommandPaletteHost` → `_dashboard/layout.tsx`.

I verified the new check can actually fail, rather than just still passing: temporarily removing `<CommandPaletteHost />` from `_dashboard/layout.tsx` makes it fail with `commandPalette/ui/FolderImportMount/FolderImportMount.tsx is not rendered from within routes/_authenticated`, then pass again once restored.

**Co-location** — agreed, and it was more than cosmetic: `AddRepositoryModals` had zero remaining references to the dialog, so the file was nested inside a component that no longer used it. Moved to `_authenticated/components/GitInitConfirmDialog/` alongside its only consumer.

`bun run lint`, `bun run typecheck` (35/35), and the regression test all pass after the change.

CodeWithMoin:
Fixed in 8e9e1c9 — and the same staleness existed in a second place worth noting.

`GitInitConfirmDialog.tsx` said "mounted once via AddRepositoryModals", and `stores/git-init-confirm.ts` said "single global dialog instance (rendered by AddRepositoryModals)". Both stopped being true when the mount moved. Grepped for other references: the only remaining mention is the past-tense line in the test header describing the original bug, which is still accurate.

On the store docblock I also named the failure mode rather than just correcting the owner:

> Safe today because there is a single global dialog instance (rendered by the _authenticated layout). A caller rendered outside that layout has nothing to settle its promise and will hang.

That is the fact whose absence caused this bug — the previous comment pointed a reader at the wrong mount site, which is exactly the wrong turn to take when debugging a silent hang.

`bun run lint` (0 errors), `bun run typecheck` (35/35), and the regression test all pass.

CodeWithMoin:
One note for whoever picks this up: **happy to drop or reshape the test if you'd prefer a leaner diff.**

The actual fix is the one-line mount move. The structural test is there because the existing unit tests mock `renderer/stores/git-init-confirm`, so a missing mount is invisible to them — that is why #5036 shipped green. But a filesystem-walking test that asserts on component hierarchy is an unusual shape for this codebase (`no-electron-coupling.test.ts` is the only precedent I found), so I understand if it reads as more machinery than the bug warrants.

Options, whichever you prefer:

1. Keep as is.
2. Drop the test entirely — I'll push a commit removing it, leaving just the mount move.
3. Replace it with the deeper fix: have the dialog register itself on mount and make `request()` resolve `false` (or throw) when no instance is registered, so no caller anywhere can hang silently. That is a larger change and I did not want to expand scope unasked, but I am glad to do it here or in a follow-up.

Also worth flagging, unrelated to this change: `useFolderFirstImport.test.ts` currently fails on `main` on its own — its `mock.module("react", …)` stubs React down to `useCallback`, so `lucide-react` hits `react.forwardRef === undefined`. Verified pre-existing by stashing this branch. Happy to open a separate PR if useful.

Kitenite:
Closing: main already ships this. e4759d5ec (#6703) mounts the git-init confirm dialog for all routes. Thanks for the PR; if you think the shipped version misses a case, open an issue and link this.

CodeWithMoin:
No worries, Thanks

PR #5875, opened 2026-07-22
fix(desktop): sync Claude hooks into CLAUDE_CONFIG_DIR profiles

carlssonemil:
## What & why

Superset derives Claude pane activity status (processing/waiting-for-input/idle)
and notifications entirely from Claude Code's native lifecycle hooks, which
Superset installs by writing directly into `~/.claude/settings.json`.

When Claude Code is launched with `CLAUDE_CONFIG_DIR` pointing at a different
profile directory — a common pattern for running multiple accounts/orgs side
by side, e.g.:

```bash
alias claude-work='CLAUDE_CONFIG_DIR=~/.claude-work claude'
alias claude-personal='CLAUDE_CONFIG_DIR=~/.claude-personal claude'
```

Claude Code reads its hooks from that directory instead, which never receives
Superset's managed hooks. The result is a silent failure: activity status
stays stuck at idle and no notifications fire, with no error surfaced.

The generated `claude` wrapper now checks `$CLAUDE_CONFIG_DIR` at invocation
time. When set, it asks the running app (via the existing local hook server,
same pattern as `/hook/complete`) to merge Superset's managed hooks into that
directory's `settings.json` before exec'ing the real binary. The merge reuses
the existing hook-merge logic (now parameterized by path instead of hardcoded
to `~/.claude`), so existing user settings/hooks in that file are preserved.
For a plain `claude` invocation (no `CLAUDE_CONFIG_DIR`), this is a complete
no-op.

Scoped to Claude only, matching the bug I noticed myself. Codex has an analogous
per-profile override (`CODEX_HOME`) that isn't handled here — flagging as
possible follow-up scope rather than bundling it into this PR since I do not use Codex myself at the moment.

As an aside: users can work around this today without a code change, by
copying the same `hooks` block Superset writes into `~/.claude/settings.json`
directly into their profile directories. That works (the hook command
resolves `$SUPERSET_HOME_DIR` at runtime, so it's install-independent), but
it's a static one-time copy — it won't cover profiles created later, and
won't pick up future changes to the managed hook the way `~/.claude/settings.json`
does automatically on every app boot. This PR fixes it at the source instead.

## Video of the fix in action

https://github.com/user-attachments/assets/d09a7c30-6680-4c41-b283-270e260e9013

## How I tested it

- Added unit tests covering the new sync function (merges into an arbitrary
  directory, preserves existing settings, creates the directory/file when
  missing, no-ops on empty/relative input) and the wrapper's exec line.
- `bun run test` (desktop package): 2251 pass, 0 fail.
- `bun run lint` clean.
- Verified end-to-end against a running dev build: confirmed the regenerated
  wrapper contains the new logic, then hit the new endpoint directly against
  a real `settings.json`, confirming existing settings are preserved and
  hooks are merged in correctly.

## Checklist

---
## Summary by cubic
Fixes missing activity status and notifications for Claude when launched with `CLAUDE_CONFIG_DIR` by syncing Superset-managed hooks into that profile’s `settings.json` at runtime. No change for default `~/.claude` setups.

- **Bug Fixes**
  - Claude wrapper checks `$CLAUDE_CONFIG_DIR` and calls `/hook/claude-config-sync` to merge managed hooks before exec; no-op otherwise.
  - Added `/hook/claude-config-sync` endpoint that merges hooks into the target `settings.json`, preserving existing settings; shared merge logic now accepts a path.
  - Added tests for config-dir sync, wrapper exec line, and path-based merge behavior.

<sup>Written for commit f6e92d5c99f8367c033dcaff33d6f2f6546acbf5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5875?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/desktop/src/main/lib/agent-setup/agent-wrappers-claude-codex-opencode.ts
+++ b/apps/desktop/src/main/lib/agent-setup/agent-wrappers-claude-codex-opencode.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import { env } from "shared/env.shared";
 import {
 	buildWrapperScript,
 	createWrapper,
@@ -90,24 +91,24 @@ function isManagedClaudeHookCommand(
 }
 
 function readExistingClaudeSettings(
-	globalPath: string,
+	settingsPath: string,
 ): ClaudeSettingsJson | null {
-	if (!fs.existsSync(globalPath)) {
+	if (!fs.existsSync(settingsPath)) {
 		return {};
 	}
 
 	try {
-		const parsed = JSON.parse(fs.readFileSync(globalPath, "utf-8"));
+		const parsed = JSON.parse(fs.readFileSync(settingsPath, "utf-8"));
 		if (!isPlainObject(parsed)) {
 			console.warn(
-				"[agent-setup] Expected ~/.claude/settings.json to contain a JSON object; skipping Claude hook merge",
+				`[agent-setup] Expected ${settingsPath} to contain a JSON object; skipping Claude hook merge`,
 			);
 			return null;
 		}
 		return parsed;
 	} catch (error) {
 		console.warn(
-			"[agent-setup] Could not parse existing ~/.claude/settings.json; skipping Claude hook merge:",
+			`[agent-setup] Could not parse existing ${settingsPath}; skipping Claude hook merge:`,
 			error,
 		);
 		return null;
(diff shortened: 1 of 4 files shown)


Kitenite:
Closing: main already ships this. 1bc9ab63b (#6585) added ensureClaudeManagedHooksAt(). Thanks for the PR; if you think the shipped version misses a case, open an issue and link this.

Issue #5496, opened 2026-07-08
[bug] Renderer frozen at 100%+ CPU: infinite QuotaExceededError loop when v2-workspace-local-state exceeds localStorage quota


espaciofuturoio:
## Summary

Desktop app becomes permanently unusable: window opens but the UI is frozen, renderer pegged at 130–170% CPU with RSS growing past 5 GB. Restarting the app or the machine does not help — the corrupt state persists on disk. Root-caused it to an infinite retry loop around `QuotaExceededError` when the `v2-workspace-local-state-<orgId>` localStorage key exceeds Chromium's ~10 MB per-origin quota.

**Environment:** desktop v1.13.1, macOS (arm64), org with ~96 active worktrees in one project.

## Root cause (traced through source at desktop-v1.13.1, commit 931b6b2)

Three stacked defects:

1. **Whole-collection-in-one-key storage.** `@tanstack/db@0.6.10`'s `localStorageCollectionOptions` (`dist/esm/local-storage.js`, `saveToStorage` L69–84) serializes the ENTIRE collection into one JSON string and calls `storage.setItem` on EVERY insert/update/delete, rethrowing on `QuotaExceededError`. The collection is defined in `apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/collections.ts` L801–827 with `storageKey: "v2-workspace-local-state-${organizationId}"` — note the key is per **organization**, so one oversized blob breaks every workspace in the org.

2. **Unbounded growth.** Nothing caps the blob: hidden-workspace tombstone rows are kept forever by design (`sidebarMutations.ts`), `viewedFiles` is uncapped, `paneLayout` can embed base64 attachments via `launchConfig.initialFiles`, and the same origin also stores `terminal-buffer:<id>` scrollback snapshots (up to 1000 lines each — we found **528 unique terminal-buffer keys**, ~8.5 MB live). No code path ever prunes or clears the key; `grep -r QuotaExceeded apps/desktop/src` has no collection-related hits.

3. **The infinite loop.** `usePlaceLocalWorktreesInSidebar.ts`: a live query feeds a `useEffect` that inserts a sidebar row for any local worktree lacking one. Cycle: optimistic insert → `saveToStorage` throws `QuotaExceededError` → transaction rolls back → optimistic row vanishes → live query emits → effect re-runs → insert again — forever, with zero user interaction. Sentry breadcrumbs show the error firing every 10–100 ms:

```
[...]

PR #5965, opened 2026-07-26
fix(renderer): the infinite loop due to which the app ui stays frozen

snehendu098:
## What & why

Refs #5496 — fixes the freeze, but does not close the issue (remaining items listed below).

When `localStorage` is full, the desktop app becomes permanently unusable: UI frozen, renderer pegged at 100%+ CPU, RSS climbing past 5 GB. Restarting the app or the machine doesn't help, because the full store is on disk. The only escape today is manually deleting `~/Library/Application Support/Superset/Partitions/superset/Local Storage`.

`localStorageCollectionOptions` re-serializes the whole collection on every mutation and rethrows `QuotaExceededError`. The library treats that throw as a failed transaction and rolls the optimistic write back → the live query re-emits → the effect that queued the write retries → repeat, ~40x a second, with zero user interaction.

**The fix guards the storage boundary rather than one call site.** On quota error: reclaim orphaned terminal snapshots, retry once, then **return without throwing**. No throw means no rollback, so the loop cannot form at any of the ~15 sites that write to these collections — not just `usePlaceLocalWorktreesInSidebar`.

It also reclaims orphaned `terminal-buffer:*` / `terminal-dims:*` keys — snapshots for terminals that no longer exist. Eviction leaves those behind by design (`terminal-runtime-registry.ts`) and nothing ever collected them; the reporter found 528 such keys, ~8.5 MB. Reclaim runs on the quota path rather than at startup, so a profile left wedged by a previous session recovers on next launch instead of needing Finder.

Follows existing conventions in this directory: `withQuotaGuard` is a sibling of `withReadHeal`, and the pure-policy/side-effect split mirrors `selectRuntimesToEvict`. No library patch needed — `localStorageCollectionOptions` already accepts an injectable `storage`.

**Deliberately out of scope** (from the issue's other defects):

- No cap on `terminal-buffer:*` for terminals still open (issue suggestis, so reclaim skips them — with enough live terminals there's nothing tofree and writes fail permanently. Degraded rather than frozen, but still permanent. Happy to take this as an immediate follow-up.
- Tombstone rows, uncapped `viewedFiles`, base64 `initialFiles` in `paner growth contributors, untouched.
- Still one key per org (issue suggestion 3). Neutralized, not restructured — that's a persistence-format migration and I'd want a maintainer's preference first, since the
existing SQLite helper is built for Electric-synced collections rather t

User-facing notice on genuine exhaustion is behind `QUOTA_NOTICE_MODE` i` — currently `silent` (console warning only, matching how`terminal-runtime-registry` already handles this error). `notify` and `offer-reclaim` are implemented alongside it, so switching is a one-word change if you'd prefer a
toast.

## How I tested it

**Reproduced on unmodified `main` first**, then replayed the identical six:

| | Before | After (same wedged profile) |
|---|---|---|
| Renderer CPU | 115% | **2%** |
| RSS | 1.4 GB | **136 MB** |
| Quota errors / 40s | 1,599 | **0** |
| Main-thread eval RTT | timed out ×3 | **p50 0.8 ms, no timeouts** |
| Orphaned snapshot keys | 413 | **0** |
| Workspace rows persisted | 0 (all rolled back) | **12** |

Repro: local stack via `setup.local.sh`, throwaway repo imported as a project, 12 worktree workspaces, `localStorage` filled to the real ceiling, then let sidebar placement run. Measured over CDP against this workspace's matched renderer. Beforeeenshot` and even `Runtime.enable` timed out. Killing every process andrelaunching cold re-wedged within a minute — confirming the "restart doesn't help" report.                                                                              
One correction to the issue: the quota was **~49 MB** on Electron 40, not ~10 MB. Different threshold, identical mechanism.                                             
**Tests:** 27 new unit tests across the four new modules. The regression test was verified to *fail* against unguarded code (the rollback is asynchronous, so asserting synchronously passes even when broken). Mutation-tested six deliberate bskip reclaim, swallow non-quota errors, break warn dedupe, ignore mode,leak the live Set — each killed a test.

`clearAllTerminalSnapshots` has an explicit test asserting it never touches collection data.

Full desktop suite: 2293 pass. The 8 failures on my machine are pre-existing and environmental (`runTeardown` + `getProcessEnvWithShellPath` fail when `$SHELL` is nushell,
since `getCommandShellArgs` falls back to `-lc`) — verified identical wiling separately.

## Checklist

---

<a href="https://stagereview.app/superset-sh/superset/pull/5965">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Fixes the freeze caused by an infinite retry loop when `localStorage` hits quota. Writes now fail soft: we reclaim safe terminal snapshots, retry once, then drop the write so the UI stays responsive and data stays in memory for the session.

- **Bug Fixes**
  - Wrapped all localStorage-backed collections with `withQuotaGuard`: on `QuotaExceededError`, reclaim, retry once, then return (no throw → no rollback loop across writers).
  - Reclaims only orphaned terminal snapshot keys via `TERMINAL_BUFFER_KEY_PREFIX`/`TERMINAL_DIMS_KEY_PREFIX`; parked and released-but-restorable terminals are preserved via the registry’s `getRegisteredTerminalIds`, and `dispose` clears preserved records.
  - Added `notifyQuotaExhausted` behind `QUOTA_NOTICE_MODE` (`silent` by default); “offer-reclaim” shows a toast with “Free up space” that clears saved scrollback and reports cleared entries (entries, not terminals).
  - Terminal runtime now uses shared key prefixes; reclaim runs on the quota path so wedged profiles recover on next launch; tests cover guard behavior, reclaim selection, notices, and preservation/cleanup paths.

<sup>Written for commit ac46d416d633cae5afd58b7ae6c93df6ff33eeea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5965?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented app freezes when browser storage quota is hit by safely handling failed persistence writes and avoiding retry loops.
  - Improved terminal state restore by keeping “parked” scrollback and dimensions available, and correctly reflecting which terminals have restorable data.
  - Automatically reclaims obsolete/unreachable terminal snapshots to make room for new saved state.

- **User Experience**
  - Added clearer handling for storage-capacity issues: configurable warning toasts, message deduping, and an option to free saved terminal scrollback so changes can persist.


ghost:

| | Title |
|---|-------|
| 1 | [Define terminal storage key constants](https://stagereview.app/superset-sh/superset/pull/5965/chapters/1) |
| 2 | [Track restorable terminal IDs in registry](https://stagereview.app/superset-sh/superset/pull/5965/chapters/2) |
| 3 | [Implement terminal snapshot reclamation logic](https://stagereview.app/superset-sh/superset/pull/5965/chapters/3) |
| 4 | [Create quota guard and notification utilities](https://stagereview.app/superset-sh/superset/pull/5965/chapters/4) |
| 5 | [Apply quota guard to all collections](https://stagereview.app/superset-sh/superset/pull/5965/chapters/5) |

<a href="https://stagereview.app/superset-sh/superset/pull/5965">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit ac46d416d633cae5afd58b7ae6c93df6ff33eeea on Jul 27, 2026 8:55pm UTC.</sub>

snehendu098:
Addressed the review feedback:

- **Preserved snapshots surviving release** (`d270573c4`). This was a real bug. `release()` and eviction persist the buffer then drop the registry entry, so reclaim would have deleted the scrollback of a pane still open. Now tracks terminals whose snapshot outlived their entry, mirroring what `disposeRuntime` does with it; `dispose()` clears the record since that is what actually removes the snapshot. Regression test added on the real `release()` path, verified to fail without the fix.
- **Cleared-snapshot count** (`d270573c4`). Was counting storage keys and reporting them as terminals, so it showed double. Now reads "N saved terminal entries".
- **Vacuous dispose test** (`b8cc31c97`). The assertion held before `dispose()` ran, so it proved nothing. Now seeds the preserved record and asserts it is reported first; verified it fails when the bookkeeping is removed.

Two things I want to flag myself:

**A gap in my own change.** Reachability comes from registry state, which is empty after a restart. Panes mount lazily, so a quota error early in startup could treat a not-yet-mounted pane's snapshot as orphaned. Bounded (scroll history, not work, and the app was frozen outright before this) but unintended. The robust signal looks like the persisted pane layout: `pane.data.terminalId` where `kind === "terminal"` survives restarts and does not depend on mount timing. Happy to fold that in here, or keep it as a follow-up, whichever you prefer.

**On the remaining issue item.** Capping `terminal-buffer:*` for terminals that are still open is the one case left where a user can stay permanently stuck. Would you want that keyed off `DEFAULT_TERMINAL_PARKED_RUNTIME_CAP`, or a separate byte budget? Happy to take it either way.

Kitenite:
Thanks for this — the diagnosis of the rollback→retry loop was exactly right, and we're keeping your `withQuotaGuard` design and its tests nearly verbatim. This landed via #6038 instead of directly from this branch, for two reasons:

1. **It needed reworking on top of #6036**, which merged first and made the persisted terminal snapshots self-bounding (boot GC + TTL + size budget, keyed by a durable `terminal-buffer-persisted-at` index). That index replaces this PR's orphan detection: the in-memory `preservedSnapshotTerminalIds` set doesn't survive a relaunch, so a quota event shortly after boot could have reclaimed restorable snapshots of terminals in workspaces the user hadn't opened yet. Driving reclaim off the persisted index (24h pressure TTL) fixes that and let us drop `selectOrphanedTerminalSnapshots`, `terminalSnapshotStorage`, and the registry additions entirely. We also shipped `QUOTA_NOTICE_MODE` as `offer-reclaim` rather than `silent`.

2. **We couldn't push the rework to your branch** — "Allow edits from maintainers" isn't enabled on this fork PR. For future PRs, please tick that checkbox (or enable it in the sidebar under "Maintainers are allowed to edit this pull request"): it lets us rebase and land your work directly with your commits intact, instead of squashing it into a co-authored replacement like we had to here.

Closing in favor of #6038, which credits you as co-author. The remaining growth contributors you flagged (uncapped `viewedFiles`, heavy `paneLayout` payloads, per-org keys) are tracked in SUPER-1686.

Issue #2745, opened 2026-03-22
[feat] Add a "project group" or "folder" concept above projects


kesku:
#### Describe the feature

There is no way to group projects (repos) in the sidebar. The top-level sidebar is a flat scrollable list of projects (each individually collapsible, but with no mechanism to organize them into groups or folders)

For example, if you work across many repos (main app, sdks, docs, infra), they all sit at the same level. There's no way to tuck lower-priority projects like SDKs or docs repos into a collapsed group while keeping your primary repos prominent

Sections exist within projects to group workspaces (branches/worktrees), but nothing equivalent exists at the project level :(

Ideally there would be a "project group" or "project section" concept at the sidebar root level that can contain multiple projects. 

Example:
```
Sidebar
├── Main App (project)
│   ├── main (workspace)
│   ├── feat/auth (workspace)
│   └── fix/header-bug (workspace)
├── Backend (project)
│   └── main (workspace)
├── SDKs (group
│   ├── python-sdk (project)
│   │   ├── main (workspace)
│   │   └── feat/streaming (workspace)
│   └── node-sdk (project)
│       └── main (workspace)
└── Infra (project)
    └── main (workspace)
```

Each group would be collapsible, renameable, and support reordering of projects into/out of groups similar to how workspace sections already work within a project

Kitenite:
Closing as a duplicate of #5612, the consolidated tracker for stale `workspace-org` issues from the July 2026 stale-issue sweep (no activity since 2026-06-01). If this still applies on a current build, comment on #5612 and we'll reopen.

Issue #1176, opened 2026-02-03
[feat] Group Projects


de1mat:
#### Describe the feature

I would love to be able to group Projects / Repos together. 

This will allow grouping:

- Projects by Client
- Projects by Work / Personal
- Other combinations...

Bonus points if we can set a custom icon for the Group. Or at least a background color for the badge next to the group name.

AviPeltz:
this feature has been added

de1mat:
Thanks for picking this up! I think it may have been closed against a different feature to the one I had in mind, so I wanted to clarify the ask.

What I meant by "Projects" here is the **top-level repos in the sidebar** - the list you reorder and collapse. The idea was to group those repos into named groups (by client, or work vs personal), ideally with a custom icon or a coloured badge on each group.

A couple of nearby features already do something similar but at a different level, which is probably the source of the mix-up:

- **Sections** (the "New Section" option in a project's menu) group workspaces *within a single repo* - really useful, but they can't span repos, so they don't group the projects themselves.
- The **Linear project filter** scopes the tasks view - that's the "Linear projects" side of things, separate from the sidebar repo list.

So the specific gap is being able to organise the sidebar repos into groups or folders. For anyone working across a lot of repos and clients, that's the bit that would help.

Could we reopen this to track that, or would you rather I raise a fresh issue with clearer wording? Happy either way.

ashbrener:
@de1mat is right, and I'd back reopening this.

What shipped was per-project **groups** (#2067) — they organise workspaces inside a single repo. This issue asks for the top-level repo list to be grouped, which is a different level and still isn't there. Two separate features that happen to share the word "group", which is where the crossed wires come from.

I've built the missing one: **#5981** adds folders above projects in the sidebar — named, collapsible, a colour each, repos dragged in and out. Same ask as #2745 (closed into the #5612 tracker) and option 1 on #4018.

@AviPeltz @Kitenite happy for this to stay closed if you'd rather track it all on #4018 — just flagging that the thing it was closed against isn't the thing that was asked for, and there's a PR up for the real one.

Kitenite:
Thanks @ashbrener , reopening !

Kitenite:
Folding this into our sidebar project grouping ticket (SUPER-1511 in our tracker) so updates land in one place. The custom group icon/color idea is noted as part of that work.

Issue #5656, opened 2026-07-13
[feat] Multiple application windows support


bvidakovic:
Would be nice to be able to open multiple applications windows for each organizations we have defined in order to easily switch from one organizations project to the other. Currently that process is slow, there is no shortcuts and I'm not getting notification about session from organization A if I'm currently working on organization B.

Also would be nice to have and option to move repo from one organization to another with all of its work trees. 

Either way keep  up with the great work, the product is awesome.

Kitenite:
Duplicate of #4910 (open organization in a new window / multiple windows). Tracking there.

ashbrener:
The "switching is slow" part of this has a concrete cause: the window is unmounted for the whole switch, so it reads as a hang rather than a wait. Filed as #6132 with a fix in #6133 — one file, keeps the org you're leaving on screen until the destination is ready.

PR #5981, opened 2026-07-27
feat(desktop): group projects into folders in the v2 sidebar

ashbrener:
## Description

Adds a **folder** level above projects in the v2 dashboard sidebar, so repos can be grouped by context — e.g. `Work` vs `Personal` — with a name, colour and collapse.

This is the cross-repo grouping asked for in #4018 (option 1) and #2745, and requested again recently in #4018. It's the piece that per-project **sections** don't cover: sections group *workspaces inside one repo*, while this groups *the repos themselves*. (#1176 was closed as done against sections (#2067), but cross-repo grouping was still missing — hence the recurring requests.)

**Naming:** deliberately "folder", not "group". The UI already calls a section of workspaces inside a project a "group" ("New workspace group", "Rename group"), so a second meaning would be ambiguous — and that ambiguity is exactly what caused #1176 to be closed prematurely. #2745's title offers "folder" as the alternative. The menu labels now name the level each action applies to, because the two sat close enough together to read as one feature.

**What's included**

- New local collection `v2SidebarFolders` (`name`, `color`, `isCollapsed`, `tabOrder`), mirroring the existing `v2SidebarSections` one level down; a project points at its folder via a nullable `folderId`, exactly like a workspace points at its section via `sidebarState.sectionId`.
- Folder header: chevron collapse, inline rename, and a right-click / hover "..." menu (rename, set colour, set icon, delete) — same interaction vocabulary as section headers. Folder identity is an emoji/image icon when set, falling back to a small colour dot; the folder colour also tints a thin rail that runs down the folder’s contents, so it’s visible where a folder ends (no divider needed before the ungrouped list). The project count shows only while collapsed.
- **Move to folder** submenu on a project's context menu, with **New folder…** and **Remove from folder**.
- **New project folder** in the PROJECTS header menu, so a folder can be created before any repo exists.
- Deleting a folder returns its repos to the sidebar root — it never deletes projects. A project pointing at a missing folder falls back to the root, so a repo can't become unreachable.

Purely renderer-local state (localStorage, org-scoped), consistent with how sidebar placement already works — no schema, host-service or API changes.

## Related Issues

Related to #4018 (implements the cross-repo grouping slice), #2745, and #1176.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

Unit tests (`folderMutations.test.ts`, 11 cases) cover create / rename (incl. trimming and blank-rejection) / collapse / colour / move in and out / reorder, plus the two safety cases: deleting a folder re-parents its projects to the root, and other folders' projects are untouched.

Verified end-to-end in the running app (dev build, driven over CDP):

1. PROJECTS → **New project folder** → folder created and drops straight into inline rename; renamed to `Work`; persisted to `v2-sidebar-folders-<org>`.
2. Created a second folder `Personal` — both render with their own colour and project count.
3. Clicking a folder header toggles collapse, persisted per folder (`Work: true`, `Personal: false`).
4. Right-clicked a project → **Move to folder → Work**: the project's count moved 0 → 1, the repo left the root list, and expanding `Work` shows it nested underneath.

`bun run typecheck`, Biome, and `bun test` (2302 desktop tests) are green.

Sidebar with `Work` (purple, containing `acme-web`) and `Personal` (green, empty). Happy to attach the captures if useful.

## Additional Notes

Scoped deliberately small: assignment is via the context menu, not drag-and-drop, so the diff stays reviewable and doesn't touch the project DnD that sort/filter (#5956) and pinning (#5963) just landed on. Drag-a-repo-into-a-folder is the obvious follow-up if you want it.

Two things worth a maintainer's call:

- **Naming** — "folder" vs "group" (see above). Happy to switch if you'd rather rename the workspace-level concept instead.
- **State ownership** — this follows today's renderer-local pattern. #5822 (client-owned sidebar state) and #5756 (host-owned sections) are both in flight; whichever lands, folder state should migrate with the rest of the sidebar rather than needing its own path. Glad to rebase onto either.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5981">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Groups projects into collapsible collections in the v2 sidebar so repos can be organized by context (named "collection" because "group" already means workspace sections and "folder" means on-disk directories). Previously all projects sat at the root; now each project can sit in a collection or at the root, and deleting a collection returns its projects to the root.

- Data/persistence: adds renderer-local `v2SidebarCollections` and a nullable `collectionId` on projects; registers `v2-sidebar-collections-*`, retires the pre-rename `v2-sidebar-folders-*` via DEAD_KEYS, hardens the collection, re-sorts in JS, and heals legacy `collectionId: undefined` to `null`. No host/API changes.
- Collection UI: collapse/expand, inline rename, color, and icon (emoji or 64px data URL); minimal header with a color dot and tinted content rail; count shows only when collapsed; header is a single tab stop with accessible emoji choices.
- Move projects: PROJECTS header adds "New collection"; project context menu adds "Move to collection" with "New collection…" and "Remove from collection", with labels distinguishing workspace groups from project collections. Drag projects onto collection headers, the root drop zone, or another project to adopt its collection; pointer-first collision detection with a root zone that expands while dragging.
- Bulk selection: Cmd/Ctrl- and Shift-click select projects; a bulk toolbar moves selected projects into a collection, creates one, or removes them from collections. Project selection is exclusive with workspace selection; Escape clears.
- Tests/i18n: unit tests cover collection lifecycle, moves, ordering, grouping, and fallbacks (deleted/missing collections return projects to root); collection strings are translated across all 16 locales with ICU plurals.

<sup>Written for commit f6d3e19a2c37605697ebdad568bd7d22ccb35011. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/5981?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Organize dashboard projects into collapsible, colorable folders with drag-and-drop, including a root drop zone for moving projects out of folders.
  * Add folder actions (create, inline rename, delete, color, collapse/expand) and project “Move to folder” options (including “Remove from folder” to return to root).
  * Allow reordering within the sidebar while re-parenting via drop targets.
* **Bug Fixes**
  * Projects with missing/invalid folder assignments (or after folder deletion) automatically fall back to the root list.
* **Tests**
  * Expanded coverage for folder lifecycle rules, colors, collapse state, project moves, deletion, ordering, and sidebar grouping behavior.


ghost:
Ready to review this PR? Stage has broken it down into 7 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Define folder schema and types](https://stagereview.app/superset-sh/superset/pull/5981/chapters/1) |
| 2 | [Implement folder utilities and DnD helpers](https://stagereview.app/superset-sh/superset/pull/5981/chapters/2) |
| 3 | [Implement folder mutations and state logic](https://stagereview.app/superset-sh/superset/pull/5981/chapters/3) |
| 4 | [Update data fetching for folders](https://stagereview.app/superset-sh/superset/pull/5981/chapters/4) |
| 5 | [Create folder UI components and context](https://stagereview.app/superset-sh/superset/pull/5981/chapters/5) |
| 6 | [Integrate folders into sidebar layout](https://stagereview.app/superset-sh/superset/pull/5981/chapters/6) |
| 7 | [Add folder actions to menus](https://stagereview.app/superset-sh/superset/pull/5981/chapters/7) |

<a href="https://stagereview.app/superset-sh/superset/pull/5981">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit e1eb2f9808812c60bf579d97be7e801f7b705455 on Jul 30, 2026 7:07pm UTC.</sub>

ashbrener:
Thanks — all three addressed in `2289a6d74`:

- **`SortableContext` order (Major)** — good catch, this was a real bug. The items array was still the pre-folder ordering while the list renders folder-by-folder. Now built to match exactly what's rendered: each folder's projects in folder order (skipped while a folder is collapsed and therefore unmounted), then the ungrouped ones.
- **Auto-rename latch** — the header now reports when an auto-initiated rename is committed or cancelled (`onAutoRenameEnd`), so the pending id is cleared instead of re-opening the input on remount.
- **`createFolderInState` didn't trim** — now trims and falls back to the default name, mirroring `renameFolderInState`, so a whitespace-only name can't fail the schema's `.min(1)` at insert. Added two tests covering blank and padded names.

typecheck, Biome and the desktop suite (2304 tests) are green, and I re-checked the auto-rename fix in the running app — no rename input re-opens after a remount.

ashbrener:
Added drag & drop in `5915e4193` — this was the part of @de1mat's ask ("repos dragged into them") that the first pass left to the context menu.

Folder headers and the ungrouped root list are now drop targets, so a repo can be dragged into a folder or back out again. Dropping on another project adopts that project's folder, so dragging into a folder's list joins it. Targets highlight while a project hovers them, and the root zone keeps a minimum height during a drag so an empty root is still targetable.

Verified end-to-end in the running app with real mouse input over CDP: dragged a repo from `Work` onto `Personal` (counts went 1→0 and 0→1, and the persisted `folderId` changed), then dragged it back out to the root (both counts 0, `folderId` cleared). typecheck, Biome and the 2304 desktop tests are green.

ashbrener:
Pushed `dfa03ff9a`. I hit this myself in the app.

A project's right-click menu had "New group" (groups workspaces inside the repo) and "New folder…" (this PR, groups the repos) two items apart, with the same icon. I couldn't tell them apart.

Now each label says which level it works on. "New group" is "New workspace group", and "New folder" in the PROJECTS header menu is "New project folder". `LuFolders` is only used for project folders now, `LuFolderPlus` stays on workspace groups.

Text and icons only, nothing else changed. Typecheck and Biome pass. I've updated the description to match.

Kitenite:
Pushed three updates to this branch:

1. **Merged `main`** (twice — through 4a6c985d6). The only overlap was `CollectionsProvider/collections.ts` meeting the new localStorage quota-guard work; resolved cleanly, typecheck and all sidebar tests green.
2. **Minimal restyle of the folder UI** (b428b027d): removed the coloured left border from the header; identity is now a small colour dot (or the new icon), and the folder colour instead tints a thin rail that runs down the folder's contents — so where a folder ends is visible at a glance, which also let the `border-t` divider before the ungrouped list (and its `showDivider` prop) be deleted. The project count now renders only while the folder is collapsed.
3. **Merged @ashbrener's `feat(desktop): give sidebar folders an icon`** (dfb598473). One conflict in the header render block, resolved per the commit's own intent (#1176): icon when set, colour dot as the fallback.

Verified end-to-end in the dev app over CDP after the merge: emoji icon renders on the header and takes precedence over the dot, an icon-less coloured folder falls back to the dot, the tinted rail follows the folder's contents, and drag-in/drag-out re-parenting still works with the new indented wrapper. `bun run lint`, `typecheck`, and all folder/sidebar unit tests (including the new icon mutation cases) pass.

One heads-up unrelated to this PR: while re-testing, sidebar folder state vanished across an app restart. That's a pre-existing `@tanstack/db` local-storage lifecycle issue on `main` (an insert on a never-synced collection rewrites the whole storage key; a single malformed row blanks hydration) — it affects the existing sidebar collections the same way and is being investigated separately, so it shouldn't block this PR.

ashbrener:
@Kitenite found the bug behind the folder state lost across a restart — the folders collection wasn't wrapped in `hardenLocalCollection` like the sibling collections. Fixed in `7c5a7f65b`.

That wrapper is what stops an insert on a not-yet-hydrated or GC'd collection rewriting the whole storage key from empty memory, and keeps one malformed entry from blanking the store on hydration.

Same commit registers `v2-sidebar-folders-*` in the persisted-key registry that landed on main that day, which CI now enforces.

The other six review findings are in there too: folders sort in JS like the project rows (the incremental `orderBy` doesn't re-sort after an insert), legacy project rows heal `folderId` from undefined to null so the menu stops offering a "Remove from folder" that does nothing, a drag onto a project that renders at root because its folder is gone now moves it to root, the folder header is one tab stop with a real button instead of a nested control, the emoji picker is reachable by arrow keys, and the colour submenu is shared between folders and workspace groups rather than duplicated.

Kitenite:
Pushed `028061bd` + merges: CDP edge-case pass over the folder pattern turned up one real drag-and-drop hazard, now fixed.

**The bug**: dropping a project in the empty area below the folders silently filed it into the *last folder* instead of the root. `closestCenter` measures from the dragged rect (measured tall at drag start, before sections compact), so the effective target sat far from the cursor, and the nearest droppable centre below everything was always the last folder header. Mapped empirically: pointer positions 425–800px all resolved to the last folder; the root zone only won in a ~15px band.

**The fix** (`utils/folderDnd`, `components/RootDropZone`):
- pointer-first collision detection (`pointerWithin` → `closestCenter` fallback, keeping keyboard drags working)
- the RootDropZone grows (`grow` + flex-col scroll container) to claim all space below the folders during a drag

After: headers ring exactly under the cursor, and the entire empty area below the folders targets the root. Re-verified over CDP: drop on collapsed folder header, reorder within a folder, drop to empty space → root, all through a reload cycle.

Also in the commit, structure cleanup: `FolderContents` and `RootDropZone` extracted to their own component folders, and the folder/root grouping extracted to a tested pure util (`utils/groupProjectsByFolder`, 3 unit tests incl. the dead-folder fallback).

Edge cases that passed without changes: empty/whitespace/300-char/HTML-injection renames, duplicate folder names (all ops id-based), delete-while-renaming, delete collapsed folder with members, dangling `folderId` in storage across reload, garbage color value, keyboard collapse toggle, icon-rail mode. Your `7c5a7f65` hardening merged cleanly — the live-folder-set drag resolution now uses a dedicated `folderIds` memo since the grouping moved into the util.

ashbrener:
@Kitenite while you're in here — would you take a look at #5337? It's the other half of what people were asking for on #4018 and #3114: folders sort the repo list inside one org, windows keep separate orgs on separate screens. de1mat, sanketghia and bvidakovic each asked for both.

Two concrete reasons it's worth doing now rather than later:

**It lands in the same file you're working in.** Both PRs touch `CollectionsProvider/collections.ts` — this one replaces the module-level singleton API client with a per-org one, because a shared client means every window sends whichever org the session last set. That's the piece the rest of the per-window work hangs off, and it's easier to reconcile with your selection/bulk work now than after both have moved again.

**It's green and current** — rebased on `main` this morning, zero behind, mergeable, both reviewers clean. CI has never actually run on it though; the workflow approval that let #5981's suite run didn't extend to this one, so nobody has seen a real build of it.

There's an open question from Avi in June about whether windows should scope to orgs or to projects, and my answer is in the thread — but it doesn't block reading the code. I've also been running it daily for a couple of weeks, which is how the org-scoped host bug and the settings-scoping bug in it got found and fixed.

ashbrener:
@Kitenite one more thing, and it's the part that doesn't fit in a technical argument: I'm running five separate org windows across four monitors. It's changed how I work with Superset more than anything else — the folders keep each org's repo list legible, and the windows mean I stop losing my place every time I switch context.

That's why I keep pushing on #5337 (multi window). If you can give it ten minutes, I'd value your read.


The problem itself is laid out in #4910 — the one you pointed #5656 at — and in #3114, which you reopened for @sanketghia: "a window/session concept where each has its own set of Workspaces". That's what #5337 does.

ashbrener:
@Kitenite I've rebuilt this on top of your #6364 rather than beside it. The sidebar no longer has its own DndContext, sensors, drag-end or DragOverlay; all of that is `DashboardSidebarDndProvider` now. Folders pass two extra inputs (`folders`, `onMoveProjectToFolder`) and `useSidebarDnd` learned that a project drag can also land on a folder header or the root zone.

The one bit worth your eyes: project drags now go pointer-first. A project section is measured tall at drag start, so closestCenter put its centre far from the cursor and the thin targets were near-impossible to hit. `pointerWithin` also ranks an ungrouped row above the root zone containing it, so dropping there still reorders. Pin and section drags are untouched.

Also dropped my `folderAwareCollisionDetection` helper — yours covers it now.

ashbrener:
For anyone following this one — folders are half of it. #5337 is the other half: each window keeps its own organization, so you can group projects inside an org *and* spread orgs across windows. I run both together as my daily build and the combination is what makes a lot of repos manageable.

A design question I'd like maintainer input on, because it changes the shape of #5337:

**Should a new window inherit the current window's org, or always ask?** Today it inherits, and the window-to-org binding persists across a relaunch through a main-process window registry. The alternative is to keep per-window org in the renderer session and treat the registry as a cache. Inheriting is what I want day to day, but it makes "open my other org" two steps instead of one.

Second one, smaller: folder state is per-org, so with per-window org it resolves per window. That seems right to me, but it's worth deciding deliberately rather than by accident.

Background on the demand for both: #4018 and #3114.

ashbrener:
Brought current with `main` (was 185 behind) now that #5337 has landed.

Main restructured the sidebar again in the meantime — `SidebarCardSlot`, the cloud section, and the ports provider hoisted to the dashboard layout — so I re-applied the folder rendering on top of that tree rather than resolving the other way round. The project context menu now takes main's plain "Remove from Sidebar" styling; the destructive variant was mine and main has since settled on plain, so this keeps only the "Move to folder" submenu from here.

Typecheck and Biome clean. Desktop suite: 2758 pass / 27 fail, and 27 is `main`'s own baseline on this machine, so no new failures. 76 sidebar tests pass.

@Kitenite this is the one you built #6364 on — it's the last piece of the folders work still outstanding, and it's ready whenever you have a window.

ashbrener:
@Kitenite flagging a collision rather than asking for a review: #6278 and this PR rewrite the same nine files, including `DashboardSidebar.tsx`, `useDashboardSidebarState.ts`, `schema.ts`, `buildDashboardSidebarProjects.ts` and `useSidebarDnd.ts`. Both change the sidebar state shape and both touch the DnD sensors, so whichever lands second is a rebuild, not a rebase.

They're not the same feature — #6278 nests workspaces inside a project, this nests projects inside a folder — so I don't think either makes the other redundant. It's purely that they occupy the same files.

Two ways to sequence it, and I'm happy either way:

1. **#6278 first.** I rebuild folders on top of the nestable tree. If `parentSectionId` generalises to projects, folders may collapse into a thin layer over your schema rather than a parallel one, which is probably the better end state.
2. **This first.** It's smaller and already current with main; you'd rebase #6278's tree onto the folder rendering.

I'd lean towards 1 on architecture grounds even though it costs me the rework — one grouping primitive beats two. But #6278 hasn't moved since 9 Aug, so if it's parked, 2 avoids blocking both indefinitely.

Which order do you want? I'll do the rework on my side either way.

ashbrener:
Following up because #6936 landed and changed the cost of the answer.

This PR is now conflicting on six files — `DashboardSidebar.tsx`, `DashboardSidebarBulkActions`, `DashboardSidebarProjectContextMenu`, `DashboardSidebarProjectRow`, `SectionActionsMenuItems`, `DashboardSidebarWorkspacesHeader` — and `_dashboard/components` is in `ENFORCED_DIRS` now, so the rebase isn't textual: the ten strings the folder UI adds across seventeen components need converting too.

That's fine, and I'll do it. What I don't want to do is do it twice. #6278 rewrites the same files, and if it lands after I've converted them I rebase the whole thing again — so the sequencing question from my earlier comment is now the thing blocking me rather than a nicety.

If #6278 goes first I'll rebuild folders on top of the nestable tree and take the i18n pass in the same go, which is one conversion instead of two and probably a smaller diff, since `parentSectionId` may generalise to projects. If it's parked, say so and I'll rebase this onto main today.

I did #6064 rather than sit idle — it was the same collision in one file, it's rebased, converted (`dashboard.sidebar.projectMenu.moveToOrganization`, with the sixteen catalogs filled), and green.

ashbrener:
Brought current with `main` (was 36 behind) and renamed the feature. This is mergeable again.

**"Folder" became "collection".** Main now uses two of these words for other things, and this PR was quietly claiming a third meaning for one of them. In `main` today:

- **group** — a section of workspaces inside a project (`sectionMenu.renameGroup`, `bulkActions.newGroup`, and #7037's session tag folders)
- **folder** — a directory on disk (`projectActions.folderNotOnMachine` → "Project folder is not on this machine", `projectMenu.hiddenFolders`)

With this PR applied, a project's right-click menu carried "New group", "Hidden folders" and "New folder…" together, meaning three different things. I hit the same problem once before in this PR and papered over it by qualifying the labels; that stopped working once #6990/#7037 landed tag folders.

"Collection" collides with neither, and it's what the provider was already called (`CollectionsProvider/collections.ts`), so this is closer to a correction than a new coinage. I'm aware #4018 asked for "groups" and @de1mat signed off on "folders" — both of those words were taken by the time this landed, which is the only reason I moved off them.

What changed with it:

- Persisted rows move from `folderId` to `collectionId`, and the store from `v2-sidebar-folders-*` to `v2-sidebar-collections-*`. The old key is registered in `DEAD_KEYS` so existing profiles get swept rather than stranded, per the localStorage policy in `apps/desktop/AGENTS.md`. Anyone who has been running this branch will need to recreate their collections once — there's no migration, since this has never shipped.
- The colour submenu now goes through the shared `ColorMenuItems` for both the collection header and the section menu, instead of main's inline copy. `PROJECT_COLORS[].name` is a thunk now, so it's resolved at the call site the same way `ColorSelector` does it.
- `_dashboard/components` is in `ENFORCED_DIRS`, so all the user-facing strings are wrapped: 22 new messages, translated across all 16 locales. Two of them changed while being wrapped — the bulk toolbar was building "N project(s)" by concatenating an English plural, which is now an ICU plural, and the workspaces header said "New project collection" where the other two entry points say "New collection".

The translations are machine-generated and haven't been reviewed by native speakers. That's the same standing as the `Translate Catalogs` workflow's output, which can't run here because it's gated on same-repo PRs — happy to have any of them reworded.

`SortableContext` still takes the collection-aware ordering rather than main's flat `projectOrder`, which is the fix from the ordering bug earlier in this PR.

Verified: desktop typecheck clean, `packages/i18n` check green, and the sidebar/route suites pass — the 9 failures in `routes/_authenticated` reproduce identically on a clean `main` checkout, so they're not from this branch.

PR #5665, opened 2026-07-13
fix(host-service): reject "." and ".." as empty/template project directory names

abhay-codes07:
## Description

`initEmptyRepo` and `cloneTemplateInto` guard the target directory name against empty strings and embedded path separators before joining it onto the parent projects dir — but not against the `.` / `..` path references. A project named `.` or `..` flows through `dirNameForEmpty` unchanged, so `join(parentDir, "..")` resolves **outside** the intended `<parentDir>/<name>` location (the parent dir itself, or its parent).

The clone-from-URL path already rejects these reserved segments in `deriveCloneDirectoryName` (with tests for `.` and `..`). This brings the empty/template creation paths in line, so all three agree.

### What changed
- Extracted a shared `assertUsableDirName` guard in `resolve-repo.ts` that rejects empty/whitespace names, embedded path separators, and the `.` / `..` references.
- Applied it in both `initEmptyRepo` and `cloneTemplateInto`, replacing the duplicated inline guard.

No behavior change for valid names; only `.` / `..` (previously accepted) are now rejected with the existing `Invalid directory name` error.

## Related Issues

Closes #5664

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please describe):

## Testing

- Added directory-name-guard tests for `initEmptyRepo` and `cloneTemplateInto` in `resolve-repo.test.ts` (reject `.`, `..`, path separators, and whitespace-only names), mirroring the existing `deriveCloneDirectoryName` reserved-segment tests. The guard is the first statement in each function, so the tests exercise it without a real repo or network.
- `bun test resolve-repo.test.ts` → **36 pass** (30 existing + 6 new), 0 fail.
- `bun run --filter=@superset/host-service typecheck` → clean.

## Additional Notes

Happy to also tighten `dirNameForEmpty` in `project/handlers.ts` to reject `.` / `..` earlier (for a clearer user-facing message) if maintainers prefer surfacing it there in addition to the boundary guard.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5665">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Rejects "." and ".." as project directory names when creating empty or template repos to prevent escaping the projects directory. Aligns validation with the clone-from-URL path; closes #5664.

- **Bug Fixes**
[...]

PR #5986, opened 2026-07-27
fix(host-service): reject dot-only directory names when creating empty/template projects

### LINKED ISSUE #5664: [bug] Creating an empty/template project named "." or ".." escapes the projects directory
### Describe the bug

When creating a **local (empty)** or **template** project, the target directory name is derived from the project name via `dirNameForEmpty` and then joined onto the parent projects directory. `initEmptyRepo` / `cloneTemplateInto` validate that name against empty strings and embedded path separators, but **not** against the `.` / `..` path references.

A project named `.` or `..` flows through `dirNameForEmpty` unchanged, so `join(parentDir, "..")` resolves **outside** the intended `<parentDir>/<name>` location — the parent directory itself (`.`) or its parent (`..`).

The clone-from-URL path already guards against exactly this: `deriveCloneDirectoryName` rejects `.` / `..` as reserved segments (see the existing tests *"throws when the URL normalizes to '.'"* / *"'..'"* in `resolve-repo.test.ts`). The empty/template creation paths are inconsistent with it.

### Where

`packages/host-service/src/trpc/router/project/utils/resolve-repo.ts` — the guard in `initEmptyRepo` and `cloneTemplateInto`:

```ts
if (!dirName.trim() || /[/\\]/.test(dirName)) {
  throw new TRPCError({ code: "BAD_REQUEST", message: `Invalid directory name: "${dirName}"` });
}
```

misses `.` / `..`. `dirNameForEmpty` (`project/handlers.ts`) keeps `.`, so `dirNameForEmpty(".")` → `"."` and `dirNameForEmpty("..")` → `".."`.

### Expected

A project name that reduces to `.` or `..` should be rejected as an invalid directory name, consistent with `deriveCloneDirectoryName`.

### Notes

Found by code inspection on `main`. In practice the escape currently surfaces as a confusing *"Directory already exists"* error (the target resolves to an existing dir, so `claimEmptyTargetDir`'s `mkdir` fails `EEXIST`) rather than a clear validation error — but the guard is the intended boundary and should be complete.

ethan0905:
> **Fixes #5664** — Creating an empty/template project named "." or ".." escapes the projects directory

## Prior art

Two earlier attempts at #5664 exist. Both opened 2026-07-13, both still open and unmerged, no human review:

- **#5665** (@abhay-codes07, `fix/empty-project-dirname-guard`) — shared `assertUsableDirName` in both `initEmptyRepo` and `cloneTemplateInto`. Rejects the **trimmed** name when it equals `.` or `..`. 6 guard-only tests that assert the thrown error.
- **#5666** (github-actions triage bot, branch `triage/issue-5664-29261598250`) — shared `assertValidDirName` in both functions. But its dot check compares the **raw** string (`dirName === "."` / `dirName === ".."`) while its emptiness check trims — the same raw-vs-trimmed drift this bug class comes from. So a padded `" .. "` still passes its guard. 5 tests including one positive case, with `.git`-absence checks.

Both prior PRs fix the core reported repro (a literal `.` or `..`). Verified against their actual diffs, this PR closes what they leave open:

- **Trimmed, dot-only regex (`/^\.+$/`)** — rejects padded variants like `" .. "` (passes #5666's raw-string check; Windows path normalization strips trailing dots/spaces and collapses it back to `..`) and longer dot-only names like `"..."` (passes both prior guards; contained on POSIX but hit by the same trailing-dot stripping on Windows).
- **Escape-proof tests, not just thrown-error tests** — each rejection case asserts nothing was created inside the parent dir, no `.git` was init'ed above it, and a sentinel file *outside* the parent (where a `..` escape or its cleanup `rmSync` would land) survives. Neither prior PR plants a sentinel; #5665 asserts only the throw.
- **Evidence-gated verification actually run** — with the source fix reverted, all 6 new tests fail; on the branch, 36/36 pass; `tsc --noEmit` and Biome clean.

## Why this PR still matters

Two open PRs (#5665, #5666) already fix the literal `.`/`..` case. What only this PR closes: the padded `" .. "` hole in the bot's guard, the `"..."` name that escapes both guards, and escape-proof tests instead of throw-only ones.

The underlying bug: `initEmptyRepo` and `cloneTemplateInto` write to `join(parentDir, dirName)`. The directory-name guard is the only thing keeping that write inside `<parentDir>/<dirName>`. It rejected empty names and path separators, but not dot-only names. So a bad name could land writes outside the projects folder: `..` resolves to the parent's parent, and `.` aliases the projects directory itself. Today luck catches the escape, not design — `claimEmptyTargetDir`'s atomic `mkdir` hits the existing directory, and the user gets a baffling `Directory already exists` error for a project that was never created (#5664).

This is the incomplete half of an existing boundary. The sibling clone-from-URL path (`deriveCloneDirectoryName`) already rejects `.`/`..` as reserved segments, with tests. The empty/template paths carried their own weaker copy of the guard. That duplication is how this bug happened, and it stays a regression magnet until the guard is shared: both creation paths sit next to a failure-cleanup `rmSync(targetPath, { recursive: true, force: true })`. Any future change that lets a dot name reach `targetPath` turns a confusing error into recursive deletion **above** the projects directory.

## What changed

One production file, `packages/host-service/src/trpc/router/project/utils/resolve-repo.ts`:

- **Shared `validateDirectoryName` helper** replaces the two duplicated inline guards in `initEmptyRepo` and `cloneTemplateInto`. Same `BAD_REQUEST` / `Invalid directory name` error, one source of truth, no future drift.
- **New rejection: dot-only names** (`/^\.+$/`), checked on the **trimmed** name, matching the existing emptiness check. That also closes the padded variants: `" .. "` passes a raw-string regex, but Windows path normalization strips trailing dots/spaces and collapses it back to `..`.
- That is the only behavior change. Every currently-valid name is accepted exactly as before.

No bypass around the guard: the handler-side slug (`dirNameForEmpty` in `project/handlers.ts`) keeps dots and runs *before* the guard, so inputs like `" . "` and `"-.-"` reach it as the literal `"."` that would have been joined.

## Tests (evidence-gated)

6 new tests in `resolve-repo.test.ts`, mirroring the existing `deriveCloneDirectoryName` dot-segment tests: `initEmptyRepo` and `cloneTemplateInto` each throw `BAD_REQUEST` for `.`, `..`, and a whitespace-padded dot name. Each test also checks for escape side effects: nothing created inside the parent dir, no `.git` init'ed above it, and a sentinel file *outside* the parent (where a `..` escape or its cleanup `rmSync` would land) survives untouched. The sentinel keeps guarding even if the surrounding creation/cleanup code changes later.

- `bun test resolve-repo.test.ts`: **36 pass / 0 fail** on this branch.
- With the source fix reverted (tests kept): **all 6 new tests fail** — they catch the bug, not just the implementation.
- `tsc --noEmit` clean in `packages/host-service`; Biome clean on both changed files.

## Why merge now

An incomplete path boundary next to a `recursive: true` cleanup gets worse silently. The fix is a small, mechanical diff: one shared validator, one added rejection case, zero change for valid names. It brings all three project-creation paths under one rule, with regression tests that pin the boundary down.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5986">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository creation and cloning validation for directory names.
  * Prevented unsafe names, including path separators, empty values, and dot-only names, from escaping the intended destination.
  * Added consistent error handling for invalid directory names across repository initialization and template cloning.
  * Added safeguards to ensure protected files and directories remain unaffected by invalid requests.

### DIFF
--- a/packages/host-service/src/trpc/router/project/utils/resolve-repo.test.ts
+++ b/packages/host-service/src/trpc/router/project/utils/resolve-repo.test.ts
@@ -11,6 +11,7 @@ import {
 	existsSync,
 	mkdirSync,
 	mkdtempSync,
+	readdirSync,
 	realpathSync,
 	rmSync,
 	writeFileSync,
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Consolidate and extend directory name validation](https://stagereview.app/superset-sh/superset/pull/5986/chapters/1) |
| 2 | [Test dot-only directory name rejection](https://stagereview.app/superset-sh/superset/pull/5986/chapters/2) |

<a href="https://stagereview.app/superset-sh/superset/pull/5986">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 638e2d71cc7642588187e6f3734dfaff1afe80fd on Jul 27, 2026 1:15pm UTC.</sub>

PR #6034, opened 2026-07-28
feat(desktop): per-project sparse checkout for new worktrees

### LINKED ISSUE #6035: Per-project sparse checkout for new worktrees
## Context

Every new workspace materialises the entire repository into its worktree. In a monorepo that means each workspace pays for the whole tree on disk even though a given task usually touches one or two packages, which slows creation and inflates the worktree directory. A project should be able to declare which folders its worktrees actually need, so new workspaces only check out the relevant subset while still behaving like a normal checkout for everything at the repo root.

## References

| Source | Who | Link | Date |
|--------|-----|------|------|
| Working session | @ihvo | [PR #6034](https://github.com/superset-sh/superset/pull/6034) | 2026-07-28 |

Screenshot of the setting is attached on the PR.

### Files

- `packages/host-service/src/db/schema.ts:86` — `sparse_checkout_paths` column on the host `projects` table (migration `drizzle/0013_married_justin_hammer.sql`). Null means a full checkout, matching the `null = inherit/default` convention already used by `worktree_base_dir` and `branch_prefix_mode`.
- `packages/host-service/src/trpc/router/workspace-creation/shared/sparse-checkout.ts:88` — `addWorktreeWithSparseCheckout`, plus path normalisation and the JSON encode/decode helpers.
- `packages/host-service/src/trpc/router/workspaces/workspaces.ts:479` — reads the project's folders once per `create`; call sites at `:302`, `:335`, `:615`, `:659` cover new branch, existing branch, remote-tracking branch, and both PR paths.
- `packages/host-service/src/trpc/router/project/project.ts:185` — `setSparseCheckoutPaths`; `:114` returns the parsed list from `project.get`.
- `apps/desktop/.../V2ProjectSettings/components/SparseCheckoutSection/SparseCheckoutSection.tsx` — the settings UI.

### Approach

Store the folder list on the host-service project row and apply it at worktree-creation time using the recipe from `git-worktree(1)`: `git worktree add --no-checkout`, then `git sparse-checkout set --cone -- <folders>`, then `git checkout`. Doing it in that order means excluded folders are never written to disk at all, rather than written and then pruned, which is where the actual saving comes from. Cone mode always retains the files at the repo root, so root manifests and configs are present regardless of what's listed. The stored column is a JSON array, but that encoding stays inside the shared module — the tRPC layer takes and returns `string[]`.

### Related code

- `shared/git-config.ts` (`enablePushAutoSetupRemote`) — same shape of small post-`worktree add` git step.
- The `icon` project option (commit `05e7ba394`) — the closest end-to-end template for column → `project.get`/mutation → v2 settings UI → settings-search registration.
- `V2ScriptsEditor` — the debounced-autosave pattern the settings field mirrors.

### Gotchas

- **Only worktrees Superset creates.** Adopted worktrees (`adopt-existing-worktree.ts`, and the "branch already in use" recovery path) keep whatever checkout they already have. The main repo is never sparsened.
- **Read at creation time**, so changing the setting never rewrites existing worktrees.
- **A rejected cone must not fail workspace creation** — fall back to a full checkout instead. Conversely, a failure *after* the `worktree add` has to remove the worktree, because callers treat a throw from that helper as "nothing was created" and their rollback paths depend on it.
- **`git sparse-checkout set` exits 0 for folders that don't exist**, silently yielding a root-files-only worktree. A typo is currently invisible; validating entries against the repo tree is a follow-up.
- **`field-sizing: content` ignores the `rows` attribute**, so the settings textarea needs an explicit `min-height` for its default size.
- Excluding a package that an included app depends on will break setup scripts like `bun install` in those worktrees. Inherent to sparse checkout, not something the feature can guard.

ihvo:
## What

Closes #6035

Adds a per-project **Sparse checkout** option. A project can list the folders its worktrees actually need; every worktree Superset creates for that project afterwards contains only those folders plus the repo-root files. Empty means a full checkout, exactly as today.

For a large monorepo this is the difference between materialising the whole tree per workspace and materialising the couple of packages you're working in.

## How

Uses the recipe from `git-worktree(1)` so excluded folders are **never written to disk**, rather than written and then pruned:

```
git worktree add --no-checkout <path> ...
git -C <path> sparse-checkout set --cone -- apps/desktop packages/ui
git -C <path> checkout
```

Cone mode always keeps the files at the repo root, so root manifests and configs are present regardless of what's listed.

- **Schema** — `sparse_checkout_paths` on the host-service `projects` table (migration `0013`, generated with `drizzle-kit generate`). A JSON array, but that encoding never leaves `sparse-checkout.ts`; the tRPC layer takes and returns `string[]`.
- **Behaviour** — `addWorktreeWithSparseCheckout` wraps all four `git worktree add` call sites in `workspaces.create`: new branch, existing local branch, remote-tracking branch, and both PR paths.
- **API** — `project.setSparseCheckoutPaths` (normalises, de-dupes, rejects `..` traversal, caps at 200 entries); `project.get` returns the parsed list.
- **UI** — new `SparseCheckoutSection` in v2 project settings, autosaving on a 500 ms debounce with a `Saving…`/`Saved` indicator, matching the Scripts editor directly below it. Registered in settings search.

## Scope and edge cases

- **Only worktrees Superset creates.** Adopted worktrees keep whatever checkout they already have.
- **Read at creation time**, so changing the setting never rewrites existing worktrees.
- **The main repo is never sparsened.**
- **A rejected cone falls back to a full checkout** rather than failing workspace creation — a working worktree beats a failed one.
- **A failure after the add removes the worktree**, so callers keep their "a throw means nothing was created" assumption and their existing rollback paths stay correct.
- Older hosts that don't return the field degrade to a full checkout.

## Testing

`packages/host-service/test/integration/sparse-checkout.integration.test.ts` builds a real git repo, runs the real `workspaces.create`, and asserts the worktree on disk: `package.json` and `apps/desktop/file.txt` present, `apps/web` and `packages` absent.

`sparse-checkout.test.ts` covers path normalisation, traversal rejection, the JSON round-trip, and the git command sequence including both failure paths (cone rejected → full checkout; checkout fails → worktree removed).

- 20 new tests; full host-service suite **924 pass / 0 fail**
- Desktop `tsc` 0 errors, `bun run lint` exit 0
- Verified end-to-end in the running app over CDP: typing into the field alone (no save button) took the host row from `NULL` to `["apps/desktop","packages/host-service"]`

## Known gaps

- **A typo'd folder is silent.** `sparse-checkout set --cone -- apps/nope` exits 0 and yields a root-files-only worktree. Validating entries against the repo tree needs a directory-listing query that was deliberately scoped out; worth a follow-up.
- **The autosave state machine has no automated test.** The desktop app has SSR-only renderer tests — no testing-library or happy-dom — so debounce/coalescing were verified over CDP rather than in CI.
- If you exclude packages an included app depends on, setup scripts like `bun install` will fail in those worktrees. That's inherent to sparse checkout.

---

<a href="https://stagereview.app/superset-sh/superset/pull/6034">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Adds per-project sparse checkout for new worktrees so only listed folders plus repo-root files are materialized. Also adds a resilient, debounced settings editor that retries transient saves, surfaces invalid input, and avoids stale updates across project/host switches.

- **New Features**
  - Project settings: new “Sparse checkout” editor (v2) with 500 ms autosave; queues edits while saving; retries with exponential backoff up to 30s; stops on BAD_REQUEST and shows the reason; clears timers on unmount; preserves drafts; isolates state per project and host; prevents stale retries/timers after switches.
  - API (`packages/host-service`): `project.setSparseCheckoutPaths` normalizes, de-dupes, rejects `..` and segments starting with `-`, bounds input at the schema edge, caps at 200 post-normalization; `project.get` re-normalizes and drops unreadable entries.
  - Worktree creation: `git worktree add --no-checkout` → `sparse-checkout set --cone <paths>` → `checkout`; tolerates post-checkout hook failures; falls back to full checkout if patterns are rejected; removes the worktree on checkout failure; repo-root files always included; adopted worktrees unaffected; setting is read at creation time only.
  - Tests: unit + integration cover normalization, read-side hardening, on-disk results, editor autosave/retry/invalid behavior, and save-race scenarios.

- **Migration**
  - Apply host-service migration `0014` to add `projects.sparse_checkout_paths`.
  - No changes to existing worktrees; hosts that don’t return the field default to a full checkout.

<sup>Written for commit 09fd5be7b89785725a9e0c9e21e25e4eb069ceb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6034?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added **Project Sparse Checkout** to v2 project settings with a folder paths editor.
  * Newly created worktrees apply configured sparse folders while retaining required root files.
  * Settings are searchable, persisted, and can be cleared to restore full checkout.
* **Bug Fixes**
  * Added safe fallback and cleanup when sparse patterns cannot be applied.
* **Tests**
  * Added coverage for validation, persistence, normalization, and workspace behavior.


ghost:

| | Title |
|---|-------|
| 1 | [Add sparse_checkout_paths to project schema](https://stagereview.app/superset-sh/superset/pull/6034/chapters/1) |
| 2 | [Implement sparse checkout logic and utilities](https://stagereview.app/superset-sh/superset/pull/6034/chapters/2) |
| 3 | [Expose sparse checkout via Project API](https://stagereview.app/superset-sh/superset/pull/6034/chapters/3) |
| 4 | [Integrate sparse checkout into workspace creation](https://stagereview.app/superset-sh/superset/pull/6034/chapters/4) |
| 5 | [Add Sparse Checkout UI to project settings](https://stagereview.app/superset-sh/superset/pull/6034/chapters/5) |
| 6 | [Register sparse checkout in settings search](https://stagereview.app/superset-sh/superset/pull/6034/chapters/6) |

<a href="https://stagereview.app/superset-sh/superset/pull/6034">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit c0b8b87f9745e584cbe795b6f0e33aaaca097a6b on Jul 29, 2026 12:48am UTC.</sub>

ihvo:
Thanks — addressed in 10188bae7.

**Assert `worktreePath` before using it** — fixed, and a good catch. `apps/desktop` really does exist at the repo root, so all three assertions would have gone green against a genuine regression. Guard added.

**Re-normalize on the read path** — done. `parseSparseCheckoutPaths` now runs every stored entry back through normalization, so the "safe to hand to git" guarantee sits with the consumer instead of depending on every writer. One deliberate difference from the write path: unusable entries are *dropped*, not thrown. Throwing here would fail workspace creation because of a bad row, whereas dropping degrades toward a full checkout, which matches how the rest of this path treats sparse checkout as an optimization rather than a correctness requirement.

**Bound the array at the schema edge** — done, `.max(1000)`. Kept it well above the 200-entry post-normalization cap so a list that only exceeds 200 *before* de-duplication still succeeds, while a runaway payload is rejected before any work.

**Align the minimum git version with `--`** — resolved, though not as suggested. Rather than gate on a version, I removed the `--` separator entirely and now reject folder segments beginning with `-` during normalization. That was the only thing the separator was guarding, so validating at the edge drops the version dependency altogether and gives the user a clear error instead of a git parse failure. Verified the no-separator form against real git: `sparse-checkout set --cone apps/desktop packages/ui` exits 0 and produces the expected tree.

Two notes on the version premise, for the record: the README's Git 2.20+ figure is already inaccurate independent of this PR — `enablePushAutoSetupRemote` sets `push.autoSetupRemote`, which needs 2.37, and `git sparse-checkout` cone mode itself needs ~2.27. I've left the README alone since that's a project-wide claim beyond this change, but it's worth a separate look.

**Docstring coverage (77.78% vs 80%)** — not chased. The exported API in the new module is fully documented; the shortfall is three-line private helpers in the settings component (`toLines`, `toPaths`, `pathsEqual`) where a docstring would restate the name. Padding those to clear a percentage seemed worse than the warning.

ihvo:
Addressed in c0b8b87f9.

**Preserve queued edits when a save fails** (Major) — real bug, fixed. Confirmed against the code: the loop nulls `queuedPathsRef` before each attempt, so a rejected `mutateAsync` left the pending edit nowhere, and the debounce that scheduled the save was already spent. Nothing would retry it, so the user's last input was silently lost unless they happened to keep typing. The failure path now puts the pending value back and retries with exponential backoff (1s doubling to 30s, reset on success). The error is surfaced inline instead of via toast, because the retry loop would otherwise fire one toast per attempt. `V2ScriptsEditor` solves the same problem the same way; this had diverged from it.

**Guard post-await UI updates after unmount** (Minor) — skipping the mounted-flag, with one part taken. `setSaveStatus` after unmount is a no-op on React 18 (the old "setState on unmounted component" warning was removed in 18, and there's no leak — the fiber is already discarded). `onChanged` is a callback into `V2ProjectSettings`, which stays mounted when this child unmounts, so suppressing it would actually break the parent's refetch after a save that switched projects mid-flight. `toast` is a global surface with no component coupling. So a mounted flag would add state without changing behaviour.

The genuine leak in that area was the timer, and the retry I just added made it worse: a failing save would keep retrying after the editor was gone. Unmount cleanup now clears `retryTimerRef` alongside the debounce and saved-indicator timers.

Verified: `bun run lint` exit 0, desktop `tsc` 0 errors, settings tests 19/19.

PR #6036, opened 2026-07-29
fix(desktop): garbage-collect persisted terminal scrollback in localStorage

Kitenite:
## Problem

Terminal scrollback is persisted to renderer localStorage (`terminal-buffer:<id>` + `terminal-dims:<id>`) so panes can restore across remounts and relaunches. Those entries were only deleted on an **explicit terminal kill** (`terminalRuntimeRegistry.dispose()`). Terminals that disappear any other way — PTY exit, workspace deleted, host reset, crash, org switch — leaked their snapshot forever.

Measured on a real install after ~2 months: **945 orphaned buffers / 23.7 MB**, enough that Chromium's synchronous localStorage load wedged the renderer (652% CPU, 6.31 GB RSS, unresponsive AX queries).

## Fix (two layers)

**1. Destroyed PTY ⇒ scrollback deleted, immediately.**
- host-service tags attach errors for permanently-destroyed sessions (not found / disposed / exited) with `code: "session-gone"` on the existing error message (additive, version-skew safe).
- The renderer transport tracks that signal plus live `exit` messages as `sessionEnded`, distinct from `_terminated` (which also covers access denials where the PTY may still be alive). The signal clears the persisted snapshot on the spot, and every persistence path — detach, release, parked eviction — clears instead of re-persisting a dead session's buffer. A successful re-attach resets the flag (session re-created under the same id).

**2. Boot-time GC bounds the store no matter how a terminal dies.**
- Every persist/restore stamps the terminal in a single `terminal-buffer-persisted-at` index.
- At renderer boot, before any terminal mounts, `pruneExpiredTerminalState()` removes: unstamped entries (all pre-existing bloat, cleaned on first launch after this ships — a one-time scrollback-restore loss), entries older than a **14-day TTL** (covers deaths the renderer never observes, e.g. workspace deleted while the app is closed), and oldest entries beyond a **2M-char (~4 MB) total budget**.

Pruning deliberately does not cross-reference pane layouts: `v2WorkspaceLocalState` is org-scoped and cache-first, so reading it for GC could delete valid buffers for evicted orgs / not-ready collections.

## Relationship to #5965

Complementary, not overlapping fixes: #5965 guards the TanStack-DB localStorage **collections** write path against the QuotaExceededError rollback→retry loop once the store is already full, and reclaims orphaned snapshots only reactively on that quota path. This PR stops the terminal-snapshot store from growing unbounded in the first place and ties snapshot lifetime to PTY lifetime. Both touch `terminal-runtime(-registry).ts`, so whichever lands second needs a small rebase.

## Testing

- `terminal-buffer-gc.test.ts`: legacy pruning, TTL, size-budget eviction order, index hygiene, dims-only orphans, corrupt-index recovery. Core tests verified to fail when their behavior is mutated out.
- `terminal-ws-transport.test.ts`: exit ⇒ `sessionEnded` + callback fired once; `session-gone` attach error ⇒ same; plain error ⇒ untouched; re-attach resets.
- `terminal-runtime-registry.test.ts`: release of an ended session disposes with `clear` and never serializes.
- Full terminal lib suite: 304 pass. `bun run typecheck --filter=desktop --filter=@superset/host-service` and `bun run lint` clean.

## Summary by CodeRabbit

* **New Features**
  * Added automatic garbage collection for persisted terminal scrollback and dimensions at desktop startup, including TTL-based pruning and a size cap.
  * Introduced persisted-state tracking so terminal history retention is updated as terminals are persisted, restored, or cleared.
* **Bug Fixes**
  * Improved session-ended handling to reliably clear persisted terminal state and avoid retaining snapshots after terminal termination.
  * Enhanced persistence cleanup to remove stale entries and keep the persisted-at index consistent.
  * Improved robustness when persisted GC index data is missing or corrupted.

### DIFF
--- a/apps/desktop/src/renderer/index.tsx
+++ b/apps/desktop/src/renderer/index.tsx
@@ -14,6 +14,7 @@ import {
 } from "./lib/boot-errors";
 import { persistentHistory } from "./lib/persistent-hash-history";
 import { posthog } from "./lib/posthog";
+import { pruneExpiredTerminalState } from "./lib/terminal/terminal-buffer-gc";
 import { electronQueryClient } from "./providers/ElectronTRPCProvider";
 import { NotFound } from "./routes/not-found";
 import { routeTree } from "./routeTree.gen";
@@ -23,6 +24,10 @@ import "./styles/bundled-fonts.css";
 
 const rootElement = document.querySelector("app");
 initBootErrorHandling(rootElement);
+
+// Before any terminal mounts: unbounded persisted scrollback wedged the
+// renderer once it grew to hundreds of orphaned buffers (23.7 MB observed).
+pruneExpiredTerminalState();
 
 const router = createRouter({
 	routeTree,
--- a/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.test.ts
@@ -1,0 +1,142 @@
+import { describe, expect, test } from "bun:test";
+import {
+	pruneExpiredTerminalState,
+	removeTerminalStatePersistedAt,
+	TERMINAL_BUFFER_KEY_PREFIX,
+	TERMINAL_DIMS_KEY_PREFIX,
+	TERMINAL_PERSISTED_AT_KEY,
+	touchTerminalStatePersistedAt,
+} from "./terminal-buffer-gc";
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+const NOW = 1_800_000_000_000;
+
+function createFakeStorage(): {
+	values: Map<string, string>;
+	storage: Storage;
+} {
+	const values = new Map<string, string>();
+	const storage = {
+		get length() {
+			return values.size;
+		},
+		clear: () => values.clear(),
+		getItem: (key: string) => values.get(key) ?? null,
+		key: (index: number) => Array.from(values.keys())[index] ?? null,
+		removeItem: (key: string) => values.delete(key),
+		setItem: (key: string, value: string) => values.set(key, value),
+	} as Storage;
+	return { values, storage };
+}
+
+function seedTerminal(
+	values: Map<string, string>,
+	terminalId: string,
+	buffer = "scrollback",
+) {
+	values.set(`${TERMINAL_BUFFER_KEY_PREFIX}${terminalId}`, buffer);
+	values.set(
+		`${TERMINAL_DIMS_KEY_PREFIX}${terminalId}`,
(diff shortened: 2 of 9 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define terminal state garbage collection logic](https://stagereview.app/superset-sh/superset/pull/6036/chapters/1) |
| 2 | [Update terminal runtime to track persistence timestamps](https://stagereview.app/superset-sh/superset/pull/6036/chapters/2) |
| 3 | [Signal permanent session termination from host](https://stagereview.app/superset-sh/superset/pull/6036/chapters/3) |
| 4 | [Handle session termination in transport layer](https://stagereview.app/superset-sh/superset/pull/6036/chapters/4) |
| 5 | [Clear persisted state on session termination](https://stagereview.app/superset-sh/superset/pull/6036/chapters/5) |
| 6 | [Initialize pruning at application boot](https://stagereview.app/superset-sh/superset/pull/6036/chapters/6) |

<a href="https://stagereview.app/superset-sh/superset/pull/6036">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 67df6f63b05d7a4f2de55df16010785b0f9d0426 on Jul 29, 2026 2:50am UTC.</sub>

Kitenite:
## CDP before/after verification (dev app, this worktree)

Drove the real dev instance (`fix-issue` worktree Electron, Vite :8025, CDP :9557, owner pid verified) with real host-service PTY sessions. Renderer-only A/B via `git checkout 044b8ab07 -- <renderer files>` ↔ `checkout HEAD --` + reload, same script both sides.

| Check | Before (base `044b8ab07`) | After (this PR) |
|---|---|---|
| Reload with 200 orphaned `terminal-buffer:*` keys seeded (**19.1 MB**, mirroring the real 945/23.7 MB report) | all 200 survive reload, 19.1 MB retained, GC module never loaded | **0 remain, 0.0 MB** after one reload |
| Real shell: `echo marker` → `exit` → pane detach | dead session's buffer + dims **persisted** (orphan created) | **nothing persisted**; dead session's state also swept at next boot |
| Negative control: live shell, output, detach *without* exit, then reload | n/a | buffer persisted, stamped, **survived boot GC** with marker content intact |

Method notes: terminals were driven through the real registry/transport (`mount` → `connect` (real WS to host-service) → `writeInput` → `detach`) into an offscreen container — the exact calls the pane lifecycle makes — rather than via UI clicks; orphan bloat was seeded as raw keys (identical to what real orphans are). Assertions are localStorage-state measurements over CDP `Runtime.evaluate`, not screenshots, since the fix is storage-lifecycle, not visual. Test sessions killed and all seeded state removed afterwards.

PR #6064, opened 2026-07-30
feat(desktop): move a project to another organization

### LINKED ISSUE #6063: [feat] Move a project to a different organization
## What I want

To move a repo from one organization to another without losing anything.

Right now there's no way to do it. If a repo ends up in the wrong org — or the work moves from your personal org to a client one — you're stuck. The only thing you can do is remove it from one org and add it again in the other, which loses every worktree you had going. You start that repo from scratch.

I hit this with five orgs on one machine. Repos end up in the wrong place, and the cost of fixing it is high enough that you just live with it.

## Why it isn't a small change

```mermaid
flowchart TB
    Q["Which org does this repo belong to?"]
    Q --> A["Cloud row<br/>v2_projects.organization_id<br/>plus workspaces, chats, automations"]
    Q --> B["Local host database<br/>one per org, no org column inside<br/>the file it lives in IS the answer"]
    Q --> C["Sidebar state<br/>stored per org<br/>order, groups, layout, pins"]
    A --> A1["Change only this:<br/>nothing changes on screen"]
    B --> B1["Change only this:<br/>the cloud still thinks it's elsewhere"]
    C --> C1["Change only this:<br/>listed, but no host serves it"]
```

Which org a repo belongs to is written down in three places at once, and they all have to agree:

1. **The cloud row.** `v2_projects.organization_id` — with its workspaces, chat sessions and automations all carrying their own org id too.
2. **The local host database.** There's one per org at `~/.superset/host/<orgId>/host.db`, and no org column inside it — a repo belongs to whichever file holds its row. This is what the sidebar actually reads.
3. **The sidebar's own state**, stored per org in the browser store (collapse, order, groups, pane layout, pins).

Change only the cloud row and nothing moves on your screen. Change only the local state and the destination org has no repo to show. Miss the workspaces and you get a repo in one org whose worktrees are in another — and the database won't complain, because nothing links the two.

A few sharp edges I ran into building it:

- **The worktrees can't be recreated, only re-registered.** They're on disk at `~/.superset/worktrees/<projectId>/…`, so the project has to keep its id. The ordinary "remove project" call runs `git worktree remove`, which would delete the very worktrees the move is trying to keep.
- **The machine has to be registered in the destination org first**, or re-pointing the workspaces fails on a foreign key.
- **The GitHub link can't travel.** A repo row is unique across the whole database and belongs to whichever org installed the app on it, so the link has to be re-resolved in the destination or dropped.
- **Running terminals and agent sessions can't follow** — they belong to the source org's daemon.

### The ids are the whole trick

A move can't copy a repo — it has to re-register the same one, because the ids are load-bearing:

- **The project id** is where the worktrees live on disk (`~/.superset/worktrees/<projectId>/…`), what the cloud workspaces point at, and what the sidebar keys on. Mint a new one and every worktree is orphaned.
- **Each worktree's workspace id** carries its open tabs, its pane layout and whether it's pinned. Re-create instead of adopt and the worktree comes back looking new and empty.
- **The repo's own checkout** has a workspace id too, and it's the easy one to forget — the code that sets a project up on a device normally mints a fresh one.
- **Group ids inside the project** have to match, or every workspace lands ungrouped on the other side.

## What it looks like

Right-click a repo → **Move to organization** → pick one.

It says up front what will and won't survive:

And the repo arrives in the other org with its worktrees intact — same branches, same folders on disk, nothing re-cloned.

## Related

- #4018 — the org/project structure discussion this comes out of
- #5981 — folders above projects (sorting repos *inside* one org)
- #5337 — per-window org (working across orgs at the same time)

Happy to put up the PR — I've built and tested it.

ashbrener:
Closes #6063.

## What this does

Adds **Move to organization** to a project's right-click menu. The repo lands in the other org with its worktrees, its branches and its place in the sidebar intact. Nothing is re-cloned and nothing on disk moves.

## Why it needs three parts

Which org a repo belongs to is recorded in three places, and a move that touches only one of them half-works:

| Where | What it holds | What happens if you skip it |
|---|---|---|
| Cloud (`v2_projects` + friends) | the org id on the project, its workspaces, chat sessions, automations | workspaces end up in a different org from their project, and nothing in the database objects |
| Local host database, one per org | the project row the sidebar actually reads — there is no org column, the file *is* the org | the repo simply doesn't appear in the destination |
| Sidebar state, stored per org | collapse, order, groups, pane layout, pins | the repo arrives stripped of everything you'd arranged |

So the move is: start the destination host, move the cloud rows, re-register the project and each worktree, copy the sidebar state, then detach from the source.

## The order matters

```mermaid
sequenceDiagram
    autonumber
    actor U as You
    participant App as Superset
    participant HB as Destination host
    participant DB as Cloud
    participant HA as Source host

    U->>App: Move to organization → picks one
    App->>HB: start host for that org
    Note over HB,DB: starting it registers this machine<br/>in the destination org — the cloud step<br/>can't re-point workspaces without that
    App->>DB: one transaction: project, workspaces,<br/>chat sessions, attachments, automations, runs
    App->>HB: re-register project (same id)
    App->>HB: adopt each worktree (same ids, same paths)
    App->>App: copy sidebar state (same group ids)
    App->>HA: detach — rows only
    Note over HA: never git worktree remove:<br/>those worktrees now belong to the other org
```

1. **Start the destination host first.** That registers this machine in the destination org, which the cloud step needs before it can re-point the workspaces — otherwise it fails on the `(organization_id, host_id)` foreign key.
2. **Move the cloud rows** — project, workspaces, chat sessions, attachments, automations and their runs, in one transaction.
3. **Re-register on the destination host**, keeping the project id and each workspace id. Worktrees live under `~/.superset/worktrees/<projectId>/`, so keeping the id means nothing on disk has to move.
4. **Copy the sidebar state**, ids intact.
5. **Detach from the source last**, through a new `project.detach` that removes rows only. The existing remove runs `git worktree remove`, which would delete the worktrees just adopted.

## Ids: what has to stay the same, and why

Nothing here is copied. The same rows are re-registered somewhere else, because four different ids are load-bearing and each one strands something different if it changes.

```mermaid
flowchart LR
    P["project id"] --> D["worktrees on disk<br/>~/.superset/worktrees/&lt;projectId&gt;/…"]
    P --> CW["cloud workspaces<br/>v2_workspaces.project_id"]
    P --> SR["sidebar project row"]
    W["worktree workspace id"] --> PL["pane layout, open tabs"]
    W --> PIN["pinned state"]
    W --> HR["host workspace row"]
    M["checkout workspace id"] --> PL
    M --> MC["its cloud row"]
    S["group id"] --> GRP["which group each workspace sits in"]
```

| Id | Kept by | What breaks if it changes |
|---|---|---|
| Project | passing it to `project.setup` on the destination | every worktree on disk is orphaned — they're stored under the project id |
| Worktree workspace | `workspaceCreation.adopt` with `existingWorkspaceId` and the explicit path | the worktree comes back as if new: no tabs, no layout, unpinned |
| The repo's own checkout | **the fix in this PR** — `project.setup` now takes `mainWorkspaceId` | same loss, plus its cloud row points at a workspace no host serves |
| Group (section) | copied verbatim, ids and all | every workspace lands ungrouped in the destination |

Two more subtleties that only show up when you actually run it:

**Removals are tombstones, not deletions.** Two reconcilers re-add a worktree that has no local row, so removing a project marks its worktrees hidden instead of deleting them. That means moving a project *back* to an org meets its own tombstones — the host serves the worktrees, the sidebar suppresses them, and the repo arrives looking half-empty. The move now heals a tombstone for any workspace it's bringing in.

**Pins are org-wide ordering, not a flag.** `pinnedAt` orders the Pinned section across the whole org, so carrying the number over would jump a moved workspace ahead of pins already there. It's re-based against the destination.

**Tasks stay behind.** They're scoped to the org, not the project, so a moved workspace's `task_id` is cleared rather than left pointing across an org boundary.

## What survives, and what doesn't

Survives: the repo, every worktree and its branch, the sidebar arrangement, per-project settings (worktree location, branch prefix).

Doesn't: running terminals and agent sessions — they belong to the source org's daemon and can't be re-parented. The dialog says so before you commit. The GitHub link is re-resolved in the destination org, or dropped if that org doesn't have the repo — a repo row is unique across the whole database and belongs to whichever org installed the app on it.

Only the machine you run it on moves. The host database is per machine, so another device keeps showing the repo under the old org until it runs the move too.

## Walk-through

Before — `side-project` in the first org, with a `feature/demo-move` worktree:

Right-click gives the new item:

It lists the orgs you belong to, minus the one you're in:

Confirm step, saying plainly what happens:

After — the repo in the destination org:

## Needs the API deployed

Worth flagging for anyone trying this from a client build: the two halves ship separately. The desktop side (menu, host re-registration, worktree adoption, sidebar state) is in the app, but `v2Project.moveToOrganization` runs on the API server. Until this is merged and deployed, a packaged client pointed at production gets `No procedure found on path` from that call — which is also a clean failure, since it happens before anything local is touched.

Projects created on a host have no cloud row and don't need the server at all; those move purely locally.

## Testing

Driven through the real menus in a dev build, moving a project with a worktree between two orgs and back again, checking both host databases, the git worktree registrations and the sidebar state at each step.

Three bugs came out of running it, each fixed here with the reasoning in the commit:

1. Projects created on a host are local-first and have no cloud row at all, so the move failed before doing any of the local work. It now reports the missing row and carries on.
2. Re-registering minted a **new id for the repo's own checkout**, stranding the pane layout and pins keyed to the old one. The id is now carried across.
3. Moving a project **back** to an org met the tombstones the earlier move left, so it arrived with its worktrees hidden. Those are now healed.

Unit tests: 8 on the sidebar state copy, 5 on the host detach — the detach tests use mocks that throw if git or the cloud is touched, so a regression fails loudly rather than quietly deleting a worktree.

`bun run lint`, `bun run typecheck` and the affected suites pass.

## Not in this PR

- Moving a whole folder of projects at once (folders are #5981).
- Moving other devices — each machine runs its own move.
- An undo. Moving it back is the way back, and it's the same path.

---

<a href="https://stagereview.app/superset-sh/superset/pull/6064">
  <picture>
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Adds a "Move to organization" option to a project's right-click menu. The move re-registers the project on the destination host and rows-only detaches it from the source, leaving worktrees and files on disk untouched — no reclone, no cloud dependency, and no server deploy required.

- The destination host starts only if it's down; target orgs come from `cloudTrpc.organization.list` and exclude the active one.
- `project.detach` drops the source host's rows in one transaction without calling git or the cloud, disposes PTYs, and is idempotent.
- The move keeps `projectId`, reuses `mainWorkspaceId`, adopts worktrees in place, preserves section ids, rebases `pinnedAt`, heals tombstones, and queues fresh terminals to reopen with empty panes.
- The confirmation names running terminals per workspace; only the current device moves, and other devices keep the old org until moved there.
- Failures unwind a half-registered destination, and a source cleanup miss surfaces as "moved, but not cleared from the old organization".

<sup>Written for commit ec35d5872064aa33b4318293d9f3c1bb1122d957. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added “Move to organization” to the project context menu with destination selection.
  * Preserves project settings, workspace layout, and sidebar placement in the destination.
  * Includes confirmation, progress handling, and notifications for completed moves or skipped workspaces.
  * Disables the move option while an operation is in progress.

* **Bug Fixes**
  * Improved cross-organization moves, including hidden workspace restoration and reliable cleanup.
  * Prevented duplicate sidebar entries during repeated move operations.


ghost:

| | Title |
|---|-------|
| 1 | [Implement cloud-side project migration](https://stagereview.app/superset-sh/superset/pull/6064/chapters/1) |
| 2 | [Enable project detachment and ID preservation](https://stagereview.app/superset-sh/superset/pull/6064/chapters/2) |
| 3 | [Manage sidebar state during migration](https://stagereview.app/superset-sh/superset/pull/6064/chapters/3) |
| 4 | [Orchestrate the move project workflow](https://stagereview.app/superset-sh/superset/pull/6064/chapters/4) |
| 5 | [Wire move action to the sidebar UI](https://stagereview.app/superset-sh/superset/pull/6064/chapters/5) |

<a href="https://stagereview.app/superset-sh/superset/pull/6064">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 78fe451e304766175d2e31d377365eead3ac49f1 on Jul 30, 2026 7:59pm UTC.</sub>

ashbrener:
Thanks — two of the three are fixed in `a16705c44`.

**Stranded between orgs after a partial failure.** This was the one worth catching. Once the cloud rows move, the project belongs to the destination org while its local rows are still on the old host, and a failure in between left it that way with nothing on screen to say so. The destination steps are now wrapped: if any of them throw, the cloud rows are moved back and the original error surfaces. If the rollback itself also fails, the message says exactly what state the project is in and how to finish it by hand, instead of a bare error.

**Disabled state on the menu entry.** Added — it greys out while a move is running.

**Access check before the missing-row probe.** I've left this one, and here's why. That early return only fires when there is no `v2_projects` row for the id *anywhere*, which is the normal case for a project created on a host: the host mints the id locally and never writes a cloud row. There's nothing for `getProjectAccess` to check against, so running it first would throw NOT_FOUND for every local-first project and the move could never happen for them.

The write path is unchanged — the caller still has to be in the source org (it's their active org) and is checked against the destination before the probe runs. What's left is that someone in two orgs can learn whether a given project uuid has a cloud row. If you'd rather that stayed strict, I'm happy to gate the no-op on ownership of the destination org instead — say the word.

ashbrener:
Both of the new ones are fair, and both are fixed in `bf6b8a182`.

**Rollback left a half-registered destination.** If `project.setup` succeeded and `setWorktreeBaseDir` or `setBranchPrefix` then threw, the destination host kept a registration for a project that was going straight back to the source org — and the error text said the opposite. The rollback now tracks whether setup ran and detaches the destination before moving the cloud rows back. Detach touches rows only, so the worktrees are untouched either way.

**No handling on the tail.** By the time the sidebar copy and the source detach run, the move has genuinely happened: cloud rows moved, destination host registered, worktrees adopted. Unwinding good work over a failed cleanup would be worse than the failure. So that block now catches and reports what's actually true — the project moved, but it may still be listed in the old organization, and here's what to do about it — instead of surfacing something that reads like the move failed.

Also picked up two of the trivials on the host:

- `project.detach`'s local deletes now run in one transaction, so a throw part way through can't leave a project row whose workspaces are already gone.
- Its warnings name the workspace they came from.

Left the third (the test seeds only disposed sessions, so the `ne(status, "active")` branch isn't exercised). It's a fair gap, but that branch is deliberately copied from `workspaceCleanup.destroy` and covering it means faking a failed kill, which mostly tests the mock. Happy to add it if you'd rather have it.

ashbrener:
Good catch, and fixed in `78fe451e3` — swapped as suggested.

You're right that the message and the order contradicted each other: with the row removed first, a failing detach left the user reading "remove it from the old organization's sidebar" about something that was no longer in the sidebar. The host database is what decides which org owns a project, so that goes first now, and the sidebar entry is left as the visible thing the message actually points at.

PR #6109, opened 2026-08-01
fix(desktop): stop sidebar-wide cache clears on membership changes, Electric flicker, and host port churn

Kitenite:
Deleting one workspace refreshed the entire sidebar: every PR/CI chip blanked and refetched. Investigation (CDP-verified end-to-end, full writeup in `apps/desktop/docs/SIDEBAR_STATE_RESILIENCE.md`) found three distinct full-bar clear vectors, all fixed here.

## Fixes

**1. Membership-change query-key churn.** The PR-chip batch query key embedded the host's full sorted `workspaceIds`, so any delete/hide/pin/create minted a brand-new query — cache discarded, all chips blanked, full refetch. The ports query and the v2-workspaces-page PR query had the same embedded-list defect — fixed in the same pass.

**2. Electric `v2Hosts` flicker emptying the host target list.** Host-target derivation read `v2Hosts` straight from Electric; a resync/cold-start window serving `[]` dropped the targets and everything derived from them — verified clearing rows *and* chips to zero for a full ~190 s host outage. New `useKnownHosts` hook persists the last-seen host list to IndexedDB per org and serves it while the collection is empty and not yet ready. A ready-but-empty list is authoritative (deleting an org's last host can't be resurrected by the snapshot), persistence is gated on `isReady` so partial pre-sync results never overwrite it, and the snapshot is org-tagged so a prior org's rows can't leak across an org switch.

**3. Host-service port churn cold-starting URL-keyed caches.** The local host-service port moves on crashy restarts (observed 51507 → 53875 → 57324 in one day) and `hostUrl` was in every host query key. Caches are now keyed on identity, never routing: workspaces/projects/PR-chips/ports on `organizationId + machineId`, agent bindings on `workspaceId` alone via a shared `getTerminalAgentBindingsQueryKey` builder. The queryFn resolves the current URL at fetch time; PR targets additionally survive `activeHostUrl: null` (disabled, not unmounted) so cached chips keep rendering through the outage. The synthesized local target takes a real fallback org from the session instead of keying cold-start entries under `""`. Since the query cache is persisted, identity keys also mean chips start warm across app relaunches.

## CDP evidence

- Delete journey (real right-click → Delete → confirm, MutationObserver + fetch log): pre-fix all 8 chips removed at +195 ms with an off-cadence refetch and poll-phase reset; post-fix zero chip mutations, one stable cache entry, unbroken cadence.
- Host outage (SIGKILL, hosts snapshot seeded): pre-fix rows+chips 0 for ~190 s; post-fix rows 3/3 throughout.
- Forced port change (SIGKILL + port squat, auto-respawn on fresh port): URL keys — cache 6 → 12 entries (6 cold-minted, 4 orphaned), chips blank ~6 s; identity keys + surviving targets — cache byte-identical, chips 9/9 and rows 3/3 through the whole 33 s kill/outage/respawn window.

**4. TanStack DB stack bump** (patch-level, desktop + mobile): `electric-db-collection` 0.3.14 fixes progressive Electric collections truncating persisted rows when resuming from saved shape metadata — a library-level cause of the vector-2 flicker (we sync via `createElectronSQLitePersistence`, exactly that path). `db` 0.6.16 fixes explicit `gcTime: 0` being silently overridden to 5 s. Must-refetch truncation verified still transactional on 0.3.15; the assumption is documented in the resilience doc with a >=0.3.14 floor.

## Still open (documented, out of scope)

Coordinator wedging permanently at "stopped" when a respawn fails to bind (plain crashes auto-respawn fine — the wedge is spawn-fail-specific), and the `packages/auth` session null write-back that remounts the whole app.

## Test plan

- [x] Unit tests for key stability (membership, URL, org/machine separation), snapshot fallback semantics (flicker vs authoritative-empty), null-URL target survival — each mutation-verified to fail against the pre-fix implementation
- [x] Full desktop suite: 2396 pass
- [x] `bun run lint` and `tsc --noEmit` clean
- [x] CDP before/after runs above

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar resilience when live host data is temporarily unavailable by using recently known host information.
  * Preserved cached pull request results for temporarily unreachable hosts.
  * Reduced unnecessary refreshes and stale data when host addresses, workspace membership, or service ports change.
  * Improved consistency when clearing or stopping terminal agents across dashboard views.

* **Tests**
  * Added coverage for host fallback behavior and stable sidebar queries.

* **Documentation**
  * Documented sidebar data flow, resilience improvements, and remaining edge cases.


Kitenite:
CDP before/after for the identity re-keying (forced port change via SIGKILL + port squat; auto-respawn lands on a fresh port):

| Observation (same harness) | URL-shaped keys (before) | Identity keys + surviving targets (after) |
|---|---|---|
| Port change | 53102 → 54000 | 55831 → 59330 |
| Cache entries across the change | 6 → 12 (6 cold-minted, 4 orphaned on the dead port) | 9 → 9, zero minted |
| PR chips | all blanked ~6 s (cold new-key queries) | 9/9 the entire kill/outage/respawn cycle |
| Workspace rows | 3/3 (snapshot fixes active in both legs) | 3/3 |

Follow-up commit 29b49ba3e also keeps PR targets alive with a null hostUrl while the host-service is down (disabled, not unmounted — same pattern as the workspaces/projects fan-outs), which is what eliminated the residual chip blank during the outage window, and applies the identity key to the v2-workspaces page PR query which had the same URL+membership key defects.

Kitenite:
Addressed the review feedback in f3629c705:

- **Stale prior-org snapshot (coderabbit, cubic P2 conf 9):** the `useKnownHosts` snapshot is now org-tagged in state and only served when its org matches the active one — a prior org's rows can't leak across a switch even before the async reload lands. (In practice the org-switch teardown unmounts the hook anyway, but the guard no longer relies on that coupling.)
- **Ghost host after deleting an org's last host + partial-sync overwrite (cubic conf 6/7):** `resolveKnownHosts` now takes the live query's `isReady` — a ready-but-empty list is authoritative (snapshot not consulted), and persistence is gated on `isReady` so a ready empty list is persisted while partial pre-sync results never overwrite the snapshot. Pinned by a new mutation-verified test.
- **`hosts[0]?.organizationId ?? ""` fallback (cubic conf 6/5):** all three derive functions now take `fallbackOrganizationId` from the session (via `useKnownHosts`), matching what `useHostProjects` already did — no more cold-start entries keyed under the empty org.
- **`name` in KnownHostRow re-subscribing ports on host rename (cubic P3):** dropped `name` from the row — no consumer used it.
- **Dead `clearKnownHostsSnapshot` export (cubic P3):** removed; the authoritative-empty persistence covers the cleanup case.
- **Silent IndexedDB failures (cubic P3):** snapshot read/write failures now log contextual warnings.
- **MD040 fence language (coderabbit):** added.

Deferred with rationale:
- **In-memory per-machine snapshot maps in useHostWorkspaces/useHostProjects (coderabbit major):** pre-existing pattern, not introduced here; cross-org exposure requires the hook to survive an org switch, but CollectionsProvider tears the tree down during switches. Worth folding into a follow-up if that coupling ever changes.
- **Agent-bindings staleness on a non-null→non-null hostUrl change (cubic conf 5):** bounded by `staleTime: 30s` + focus refetch + lifecycle-event invalidation; workspaces don't migrate hosts today.
- **Key-builder duplication across the four utils (cubic P3):** agreed it should be one shared builder — follow-up refactor rather than more surface in this PR.

Kitenite:
Bumped the TanStack DB stack in d027455d1 (patch-level: `@tanstack/db` 0.6.14→0.6.17, `electric-db-collection` 0.3.12→0.3.15, `react-db` 0.1.92→0.1.95, sqlite-persistence 0.1.18→0.1.21 / 0.2.6→0.2.9, desktop + mobile). Two upstream fixes are directly in this PR's failure class:

- **electric 0.3.14 (TanStack/db#1493):** progressive Electric collections no longer truncate persisted rows when resuming from saved shape metadata — we run Electric collections on exactly that path (`createElectronSQLitePersistence`), so on 0.3.12 a resume could empty a collection at the library level. That's a root-cause fix for the vector-2 flicker; `useKnownHosts` remains defense-in-depth.
- **db 0.6.16 (TanStack/db#1660):** explicit `gcTime: 0` on live-query collections was silently replaced with 5 s (`||` → `??`) — the same lifecycle semantics #6089's hardening relies on.

Verified on the bump: must-refetch truncation in 0.3.15 is still transactional (the atomicity assumption behind resolveKnownHosts' authoritative-empty rule — now documented in SIDEBAR_STATE_RESILIENCE.md with a ≥0.3.14 floor), full desktop suite 2396 pass including the localCollectionLifecycle tests that pin library semantics, root typecheck 34/34 workspaces, lint clean.

Kitenite:
Final CDP gate on branch HEAD (d027455d1 — all fixes + TanStack bump), real UI journeys, fresh app boot:

**Test 1 — workspace delete (the original report):** adopt throwaway → right-click → Delete → confirm. Row animates out; **0 PR-icon DOM mutations, 0 cache keys minted, chips 9/9, rows 3/3**. Only PR fetch in the window was the scheduled 10 s poll tick.

**Test 2 — host-service crash with forced port change (SIGKILL + squat, 58401 → 60305):** across the full 20 s kill/outage/respawn window — **rows 3/3, chips 9/9, 0 icon mutations, 0 keys minted**. Coordinator auto-respawned on the fresh port; caches reattached warm.

Both scenarios that used to clear the bar now measure zero visual and zero cache disturbance. Follow-ups for the two remaining known clears are filed: SUPER-1730 (coordinator spawn-fail wedge), SUPER-1731 (get-session null org write-back). Good to merge from my side.

Issue #6111, opened 2026-08-01
[test] Triage workflow smoke test - will be closed


Kitenite:
Smoke test for the `Triage Issue` workflow hardened in #6110. Verifying the agent can check out, investigate read-only, and post a comment under the new `--tools`/`contents: read` config. This issue will be closed once verified; please ignore.

For the triage agent, treat the following as the report:

**Expected:** it is clear from the repo which script `bun run lint` executes, and whether Biome warnings fail the build.

**Actual:** the lint entrypoint is indirected through a shell script and it is not obvious whether warnings are treated as errors, so contributors push code that then fails CI lint.

Where is the lint entrypoint defined, and does it treat warnings as errors?

Kitenite:
Smoke test complete: the hardened workflow checked out under `contents: read`, investigated read-only, and posted the triage comment above. Closing.

PR #6113, opened 2026-08-01
fix(ci): propagate guard script failures from lint.sh

Kitenite:
## Summary

`scripts/lint.sh` captures Biome's exit code at line 5, runs the four repo guard scripts, then exits with Biome's code. The guards' own exit codes were never checked, so with Biome clean a guard violation printed to the log and `bun run lint` still exited **0** — meaning CI could not fail on it.

That silently disabled the off-loop-git ratchets those guards enforce (#6093, #6101). The enforcement looked present but wasn't; it would only have been noticed the first time someone actually crossed a ratchet and nothing stopped them.

Found incidentally by the triage agent in #6111.

## Changes

- Accumulate guard failures into `exit_code` with `|| exit_code=1`. All four still run, so a contributor sees every violation in one pass instead of one per CI round.
- Set the exec bit on `scripts/check-simple-git-usage.sh`, the only guard tracked `100644`, and drop the `bash` prefix that worked around it. The inconsistency was a live trap: normalizing the call without the mode change would have produced a 126, which the old script would also have swallowed.

Deliberately **not** `set -e` — line 4 captures Biome's output via command substitution and line 7 echoes it, so `set -e` would abort before the diagnostics are printed and make lint failures unreadable. The missing `set -e` is load-bearing for that block.

## Test plan

All four guards pass on `main` today, so this does not turn CI red.

Verified by mutation — added `ref.startsWith("origin/")` to a Biome-clean `.ts` file outside the guard's exclusions, then ran both versions against it:

| Script | Violation printed | Exit code |
|---|---|---|
| `lint.sh` from `main` | yes | **0** (bug) |
| this branch | yes | **1** (correct) |

- [x] Clean tree still exits 0 with the fixed script
- [x] Mutation removed after testing

---
## Summary by cubic
Fixes `scripts/lint.sh` so guard failures propagate and CI fails correctly, and makes guard scans reliable by installing ripgrep and failing closed when it’s missing. Keeps readable Biome diagnostics while running all guards in one pass.

- **Bug Fixes**
  - Accumulate guard failures into `exit_code` using `|| exit_code=1` while still running all four guards.
  - Install `ripgrep` in the CI lint job and add `rg` presence checks in the desktop/simple guard scripts so a missing dependency cannot read as a clean scan.
  - Set the exec bit on `scripts/check-simple-git-usage.sh` and drop the `bash` prefix to avoid exit 126 and inconsistent execution.
  - Do not use `set -e` to ensure captured `Biome` output is printed before exit.

<sup>Written for commit 4ebc7f37272413549eae8a02ffb41a557071536b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Improved linting checks to continue running and report all detected issues together.
  * Added clearer validation errors when required scanning tools are unavailable.
  * Improved consistency when executing repository quality checks.
  * Updated automated checks to ensure required tooling is available before validation begins.


Kitenite:
## Update: the guards were not just unenforced, they were broken in CI

Pushing the propagation fix turned `Lint` red, which surfaced the real problem: **ripgrep is not installed on the runner.**

```
[git-refs] ripgrep scan failed (exit 127)
./scripts/check-git-ref-strings.sh: line 28: rg: command not found
[cloud-workspace] ripgrep failed (exit 127)
```

All four guards scan with `rg`, so none of them has ever actually run in CI. They split into two failure modes:

| Guard | Behavior in CI before this PR |
|---|---|
| `check-git-ref-strings` | exit 1, **swallowed** by lint.sh |
| `check-cloud-workspace-usage` | exit 1, **swallowed** by lint.sh |
| `check-desktop-git-env` | **exit 0** — failed open |
| `check-simple-git-usage` | **exit 0** — failed open |

The last two wrap ripgrep as `if output=$(rg ... 2>/dev/null)`, where exit 127 is indistinguishable from "no matches", so they reported a clean scan without having scanned anything.

### Added in `4ebc7f3`
- Install ripgrep in the `lint` job.
- Explicit `command -v rg` check in the two fail-open scripts, so a missing dependency can never again read as a clean scan.

### Verified
Simulated the CI condition locally by stripping `rg` from `PATH`:

| Guard | on `main` | this branch |
|---|---|---|
| `check-desktop-git-env` | exit 0 (silent pass) | exit 1 |
| `check-simple-git-usage` | exit 0 (silent pass) | exit 1 |

With ripgrep present all four exit 0 and `bun run lint` exits 0, so this stays green once the runner has the dependency.

PR #6118, opened 2026-08-02
feat(desktop): pre-flip notice + post-flip welcome for migrated v1 users

Kitenite:
## What

The migration's only visible seam, handled as one branded arc — both cards share a `FlipNoticeCard` shell that carries the import wizard's visual identity (WelcomePage's dithered warm-gradient cover, sharp corners, title on the cover), and both sell the upgrade instead of listing caveats:

1. **`V1FlipNotice`** (pre-flip) — shown on the v1 surface only once this machine is provably flipping at next launch (completion marker written this session; D5 makes the flip unconditional from there).
   > **A better Superset is ready** — "Superset has been upgraded: faster, cleaner, and built around your projects. Everything comes along on your next launch." + amber callout: "Running terminal sessions won't carry over."
2. **`V2FlipWelcome`** (post-flip) — first v2 boots, until acknowledged.
   > **Welcome to the new Superset** — "Same Superset, upgraded. Everything came with you: projects and workspaces in the sidebar, and fresh terminals in each workspace's old folders."

Follows up on #5821 (F5: the flip previously landed with no explanation).

## Screenshots (CDP captures of the shipped build)

**Pre-flip notice — in context on the v1 surface, appears mid-session the moment migration completes:**

pre-flip card on v1

pre-flip closeup

**Post-flip welcome — first v2 boots, survives reload until dismissed:**

welcome on v2

welcome closeup

**Precision baseline — v1 surface before completion, no card:**

v1 before completion

## Why not the server-side pre-update notice

`desktop_notices` renders on the already-shipped old build, which only filters by version/platform/channel — it can't see surface or flag state, so it would warn every updater while only the flag cohort migrates. These cards are precise by construction: flag-excluded domains, unflagged users, v2-native users, and machines whose gate never completes see neither.

## How

- `markV1MigrationComplete` dispatches a `v1-migration-completed` window event on first completion (notice appears mid-session, no polling) and arms the welcome flag (consumed on dismiss, so it survives reloads).
- Shared `FlipNoticeCard` shell: lazy-loaded Dithering shader cover (same ramp as the import wizard), body isolated below, primary CTA.
- Tracks `v1_flip_notice_shown/dismissed`, `v2_flip_welcome_shown/dismissed` (org-keyed, counted once per org).

## Verification

Both cards CDP-verified end-to-end (transcripts in comments):

- Pre-flip: absent before completion → appears mid-session via the real `markV1MigrationComplete` path → dismiss works. The #5821 rollout flag gate verified live as a side effect.
- Post-flip: renders with the welcome flag armed → survives reload → dismiss consumes → gone.
- Negative cohorts: existing v2 users, unflagged v1 users, and gate-blocked (stuck) v1 users all see neither card.
- The CDP pass caught a real mount bug (welcome card's layout edit had silently no-op'd) — fixed and re-verified. 41 unit tests green; the CI Test failure from leaked test shims is fixed (shims now scoped).


Kitenite:
## CDP verification (2026-08-01)

End-to-end on the dev desktop (worktree `v1-to-v2-auto-migrate`, branch `v1-flip-notice`, renderer :7225, CDP :9555, signed in, active org verified in-renderer). Fixture: real-data local.db snapshot with the 2 broken projects abandoned so the gate can complete; restored afterward.

**Sequence observed live:**
1. Forced v1 surface (`optInV2=false`), cleared markers → **no card, no marker** (baseline screenshot).
2. Flag gate held: with `v1-auto-migration` unresolved/absent, the migrator did nothing for ~5 minutes (marker stayed absent) — the rollout pacing gate verified end-to-end as a side effect.
3. Enabled the flag (local posthog override — the dev renderer's PostHog person doesn't evaluate the prod flag, so this step is synthetic; everything else is real) → migration pass ran → gate completed → marker written → **card appeared mid-session with no reload** (window-event path working).
4. Card renders bottom-right over the v1 surface, non-modal, correct copy incl. the scrollback/chat-history disclosure.
5. "Got it" dismisses; card text gone from the DOM; v1 session continues undisturbed.

**Precision checks:** card absent with no marker (steps 1–2); card present only after `v1-migration-complete-<org>` exists while still on v1 — the exact "this machine flips next launch" condition.

Screenshots captured: baseline v1 (no card), full window with card, card close-up, post-dismiss. Evidence gate: all steps end-to-end except the flag override noted in step 3.

Kitenite:
## Negative-case CDP verification (2026-08-01)

Verified that NEITHER card appears for users who aren't flipping:

**Existing v2 user** (v2 surface, zero `v1-migration-*` keys): neither card in the DOM. `{surface: v2, migKeys: [], preflipCard: false, welcomeCard: false}`.

**v1 user, flag not enabled**: soak-tested earlier in the pre-flip verification — ~5 minutes on v1 with the flag unresolved: no migration pass, no marker, no card.

**v1 user, flag ENABLED but gate never completes** (the stuck cohort — broken repos): a persisted flag override made this boot exactly that case against the real messy-data fixture. The pass ran (fresh ledger error-row retries observed), the gate stayed incomplete, no marker was written, and no card rendered. Confirmed on the true v1 surface (v1 sidebar present, `optInV2: false`, `marker: null`, `preflipCard: false`).

The precision claim holds in both directions: the cards appear for exactly the machines that flip, and for no one else.

PR #6117, opened 2026-08-01
feat(desktop): install GitHub CLI from onboarding terminal and add gh dialog end states

Kitenite:
Part of SUPER-1729 — https://linear.app/superset-sh/issue/SUPER-1729

## What

- **`system.detectBrew`**: new tRPC procedure detecting Homebrew via `execWithShellEnv("brew", ["--version"])` on the login-shell PATH (explicit `BrewDetectResult` interface).
- **Onboarding Install action**: when brew is present, the Install button now opens the terminal dialog in install mode running `brew install gh && gh auth login --hostname github.com --git-protocol https --web` (title "Install GitHub CLI"). When brew is absent, it keeps the existing external link to https://cli.github.com/.
- **GhAuthDialog end states**: on PTY exit the dialog re-runs `detectGhCli`. Installed + authenticated → brief success state, auto-close after 1.5s (row still refetches via `onExit`). Otherwise the dialog stays open with "Sign-in canceled or failed." and a Retry button that respawns a fresh terminal (new paneId via remount). The stream's exit payload does carry `exitCode`, but success is decided solely from `detectGhCli` since the auth outcome is what matters.
- **Device-code copy chip**: the output stream is buffered, ANSI-stripped (`strip-ansi`, already a desktop dep), and matched against `/one-time code: ([A-Z0-9]{4}-[A-Z0-9]{4})/`; the code renders above the terminal with a copy button. Regex verified against gh's actual output shape (`! First copy your one-time code: XXXX-XXXX` with the code bold-wrapped, across chunk boundaries).
- **Outside-click protection**: the shared Dialog defaults to `modal={false}`, so a stray outside click unmounted the terminal and killed the PTY mid-auth. While the process is running/checking, `onInteractOutside`/`onFocusOutside` are prevented.
- **Copy fix**: description now matches gh's real prompt sequence (first prompt is `? Authenticate Git with your GitHub credentials? (Y/n)`).

Deliberately unchanged: Continue is still not gated on gh state (PR #4834); page.tsx diff kept minimal since row status rendering is being reworked on another branch.

## Verification

- `bun run lint:fix` + `bun run lint` → exit 0
- `apps/desktop` `bun run typecheck` → clean
- one-time-code regex sanity-checked against simulated chunked/ANSI gh output

## Summary by CodeRabbit

* **New Features**
  * Added automatic Homebrew detection during onboarding.
  * Added an in-app GitHub CLI installation flow when Homebrew is available.
[...]

PR #6125, opened 2026-08-02
fix(host-service): link pull requests on workspace creation

Kitenite:
Part of SUPER-1729 — https://linear.app/superset-sh/issue/SUPER-1729

## Problem

`PullRequestRuntimeManager` only reacts to `GitWatcher.onChanged` plus two independent 5-minute timers (`SAFETY_NET_INTERVAL_MS`, `PROJECT_REFRESH_INTERVAL_MS`). A freshly created worktree is git-idle, so the watcher never fires for it, and `insertLocalWorkspace` writes the row with `upstream_owner`/`upstream_repo`/`upstream_branch`/`head_sha` all NULL — which makes it invisible to `performProjectRefresh`'s `wantedRefs` matching.

**Before:** a workspace created on a branch that already has an open PR shows no PR until the timers happen to line up. Measured on a live host DB (workspace `698cfbd4-5dbb-4857-9a42-da942096dcbb`, branch `hoakiet98/super-1729-mismatch-to-new-workspace`, PR #6123 already open at creation): **~13 minutes**, not just 5. Two compounding causes:
1. The safety-net timer (writes upstream via `syncWorkspaceBranches`) and the project-refresh timer (does the PR matching) are independent 5-min intervals — if the match tick fires before the sync tick, the workspace waits another full round.
2. `syncWorkspaceBranches` walks every workspace sequentially (47 rows on that machine), each doing a git refs read, so the sweep itself takes real time.

**After:** the `workspace:changed`/`created` event triggers the existing coalesced `enqueueWorkspaceSync(workspaceId)` immediately — upstream + head sync, then `refreshProject` links the PR within seconds.

## Changes

- **`EventBus.onWorkspaceChanged(listener)`** — new in-process subscription API. `broadcastWorkspaceChanged` previously only fanned out to WebSocket clients; host-internal consumers had no way to observe workspace lifecycle. Listeners are isolated with try/catch so a throwing listener can't fail the emitting store write or skip the client broadcast.
- **`PullRequestRuntimeManager.subscribeToWorkspaceEvents(eventBus)`** — reacts to `created` events only, routing through the existing `enqueueWorkspaceSync` (already coalesced; `syncOneWorkspace` already calls `refreshProject` when the row changes, and a new row always changes: NULL → real upstream/head). Unsubscribed in `stop()`. Wired post-construction in `app.ts` because the runtime is built before the EventBus exists.
- No new polling or timers; the 5-min intervals, `REPO_PULL_REQUEST_CACHE_TTL_MS`, and `inFlightProjects` dedupe are untouched.

## No feedback loop

Only `created` is subscribed — renames/branch edits already flow through the GitWatcher path, and widening the trigger would add churn for no gain. The runtime's own row mutations use raw `this.db.update(workspaces)` (5 call sites), deliberately bypassing the store helpers that emit `workspace:changed`; those writes were left untouched, so a sync can never re-trigger itself.

## Judgement call: unchanged-row early return

`syncOneWorkspace` still only calls `refreshProject` when the row actually changed. Left as-is: a freshly created workspace always changes on its first sync (head sha NULL → real), so the created path always reaches PR matching. Removing the early return would fire `refreshProject` (GitHub calls, cache-bounded but real) on every watcher event even when nothing moved — extra traffic for no gain here. A workspace whose git state never moves still gets PR-state updates from the 5-min project-refresh timer, same as today.

## Testing

- Two new unit tests in `pull-requests.test.ts` against the real in-memory-SQLite harness: a `created` event links the PR **without** `start()` ever being called (no timers, no gitWatcher activity) and `stop()` unsubscribes; `updated`/`deleted` events trigger no sync at all.
- Mutation-tested: inverted the `eventType !== "created"` filter — the positive test went red (no PR link) and the negative test went red (updated/deleted started syncing); restored, all 14 tests green.
- `bun run lint` exits 0; `tsc --noEmit` clean; `event-bus.test.ts` green.

---
## Summary by cubic
Links pull requests immediately when a workspace is created, eliminating the ~13-minute delay caused by idle git state. Addresses SUPER-1729.

- **Bug Fixes**
  - Added `EventBus.onWorkspaceChanged` for in-process listeners.
  - Subscribed `PullRequestRuntimeManager` to `workspace:changed` `created` events; routes through `enqueueWorkspaceSync` to update upstream/head and run `refreshProject`.
  - Wired subscription in `app.ts`; cleanly unsubscribes in `stop()`. No new timers or polling.
  - Added tests to confirm PR linking on `created` and ignoring `updated`/`deleted`.

<sup>Written for commit a6733e9973854c1857efa7e982bf5220cbab2870. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Newly created workspaces are now detected immediately for branch and upstream synchronization.
  * Matching pull requests are linked automatically when a workspace is created.
  * Workspace updates and deletions no longer trigger unnecessary pull-request refreshes.
  * Event listener errors no longer prevent other workspace notifications from being delivered.

## Measured (CDP end-to-end, dev desktop from this branch)

Drove the real app over CDP: created a workspace via `workspaces.create` on `dependabot/npm_and_yarn/nanoid-6.0.0` (PR #6104 already open) in a project with 16 workspaces, with every `gh` invocation timestamped.

- **After (this fix):** the `created` event fired mid-mutation; `pull_request_id` was populated with PR #6104 **~14 s** after the create call (worktree add + upstream sync + GitHub round-trips). The PR chip renders in the workspace header.
- **Before (same build, subscription line disabled, host-service restarted):** creation triggered zero PR-runtime activity — no sync, no `gh` calls — and the row still had NULL `pull_request_id`/upstream/head 3 minutes later, on track for the 5–13 min timer path.
- **Cost:** the trigger produced exactly one `refreshProject` = 52 `gh` calls over ~12 s (12 per-head lookups across the project's workspaces, 1 open-PR sweep, 4 detail calls × 10 matched PRs). This is the same batch one 5-minute `PROJECT_REFRESH` tick already fires — moved earlier, not added. Creations within 60 s of any prior refresh are served from the repo-PR cache.
- **No feedback loop observed:** exactly one `workspace:changed` event consumed per creation; zero PR-runtime traffic in the 90 s after the link landed.

### DIFF
--- a/packages/host-service/src/app.ts
+++ b/packages/host-service/src/app.ts
@@ -194,6 +194,10 @@ export function createApp(options: CreateAppOptions): CreateAppResult {
 
 	const eventBus = new EventBus({ db, filesystem, gitWatcher });
 	eventBus.start();
+	// Post-construction wiring (the runtime is built before the EventBus):
+	// newly created workspaces get their first branch/upstream sync + PR link
+	// immediately instead of waiting for the 5-min safety net.
+	pullRequestRuntime.subscribeToWorkspaceEvents(eventBus);
 
 	const terminalAgentPersistence = new SqliteTerminalAgentBindingPersistence(
 		db,
--- a/packages/host-service/src/events/event-bus.ts
+++ b/packages/host-service/src/events/event-bus.ts
@@ -24,6 +24,10 @@ interface ClientState {
 	fsSubscriptions: Map<string, FsSubscription>;
 }
 
+type WorkspaceChangedListener = (
+	message: Omit<Extract<ServerMessage, { type: "workspace:changed" }>, "type">,
+) => void;
+
 function sendMessage(socket: WsSocket, message: ServerMessage): void {
 	if (socket.readyState !== 1) return;
 	socket.send(JSON.stringify(message));
@@ -65,6 +69,8 @@ export interface EventBusOptions {
  */
 export class EventBus {
 	private readonly clients = new Map<WsSocket, ClientState>();
+	private readonly workspaceChangedListeners =
+		new Set<WorkspaceChangedListener>();
 	private readonly gitWatcher: GitWatcher;
 	private readonly filesystem: WorkspaceFilesystemManager;
 	private removeGitListener: (() => void) | null = null;
@@ -193,7 +199,29 @@ export class EventBus {
 			"type"
 		>,
 	): void {
+		// A throwing listener must not fail the emitting store write or skip
+		// the client broadcast.
+		for (const listener of this.workspaceChangedListeners) {
+			try {
+				listener(message);
+			} catch (error) {
+				console.error("[event-bus] workspace-changed listener failed", {
+					error,
+				});
+			}
+		}
 		this.broadcast({ type: "workspace:changed", ...message });
+	}
+
+	/**
+	 * In-process subscription to the same workspace lifecycle events that
+	 * `broadcastWorkspaceChanged` fans out to WebSocket clients. For host-
+	 * internal consumers (e.g. the pull-requests runtime) that need to react
+	 * without holding a socket. Returns an unsubscribe function.
+	 */
+	onWorkspaceChanged(listener: WorkspaceChangedListener): () => void {
+		this.workspaceChangedListeners.add(listener);
+		return () => this.workspaceChangedListeners.delete(listener);
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


Kitenite:
Typecheck failure is pre-existing on main (onboarding/project/page.tsx GhAuthDialog `mode` prop, introduced by #6117 — file is identical between this branch and main; host-service typechecks clean). cubic's P2: the created-event handler uses the same `void this.enqueueWorkspaceSync(...)` pattern as the existing GitWatcher subscription in start(), so error handling matches the sibling call site. Admin-merging per Kiet.

Issue #6132, opened 2026-08-02
[bug] Switching organization blanks the whole window until it finishes


ashbrener:
## What happens

Switching organization blanks the whole window until the switch finishes. Not a spinner or a stale view — an empty window. On a large organization it can sit like that for a long time; the worst I've seen on my own setup was about two minutes, with no sign the app was doing anything.

## Steps

1. Be in an organization with a decent number of projects and workspaces.
2. Switch to another organization from the dropdown at the bottom left.
3. The window goes empty until the destination organization's data is ready.

## Why

`CollectionsProvider` returns `null` while a switch is in flight:

```js
if (!contextValue || isSwitching) {
    return null;
}
```

`isSwitching` stays true across three awaits — `authClient.organization.setActive`, `preloadCollections` for every collection in the destination org, and `refetchSession`. Returning `null` unmounts the entire authenticated tree for that whole time, so the longer the destination org takes to preload, the longer the window stays empty.

## Suggested fix

The context still points at the *previous* organization until the switch resolves, so the tree can stay mounted — you'd see the org you're leaving until the new one is ready, instead of nothing. That's a smaller change than adding a loading state and it removes the blank entirely.

Related: #5656 described switching as slow, which I think is this seen from the outside. Tracked under #4910.

ashbrener:
Good catch on the session signal — you were right to flag it, and it does fire. From `better-auth/dist/plugins/organization/client.mjs`, `/organization/set-active` is registered against `$sessionSignal`:

```js
matcher(path) {
    return path.startsWith("/organization/set-active") || path === "/organization/create" || ...
},
signal: "$sessionSignal"
```

So `useSession` refetches on its own and the session's active org flips as soon as the server commits, well before the preload finishes. Just removing the gate would have swapped the UI onto the destination's empty collections, and the eviction effect would have pulled the previous org's collections out from under the mounted tree on the same tick.

#6133 now tracks the displayed org separately and only advances it once the destination is preloaded, with eviction keyed to the displayed org. Same guarantee the teardown was buying — you never see a half-synced org — without the blank window. `SIDEBAR_STATE_RESILIENCE.md` updated in the same change, since it recorded the teardown as intended.

Two of your other points I've deliberately left alone rather than fold in: bounding the preload (a wedged Electric shape now leaves you on the previous org rather than staring at nothing, so it degrades sanely) and the stale `/v2-workspace/$workspaceId` route after a switch, which is there today too. Both look like their own changes rather than scope on this one.

[...]

PR #6133, opened 2026-08-02
fix(desktop): keep the window rendered while switching organization

ashbrener:
## Why

Reported in #6132. #6135 removed the wait — the switch no longer gates on Electric's initial sync — but the provider still returns `null` while switching, so the window is still blank for the round trips either side of it. That teardown takes every pane, terminal and the window-drag regions with it, which is why it reads as an empty window rather than a loading state.

## What

- Stop unmounting the authenticated tree during a switch. Only a window with no collections at all renders nothing; the destination renders cache-first while its shapes stream, which is what #6135 set up.
- Turn the re-entrancy guard into a ref, so an overlapping switch is skipped without a render.
- Update `SIDEBAR_STATE_RESILIENCE.md`, which recorded the teardown as by design.

## Testing

- Reproduced on a packaged build before #6135: switching into an organization with sixteen projects left the window empty for about two minutes.
- `bun run lint`, `bun run typecheck`, and the 48 tests under `providers/` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved organization switching by keeping the current view visible while the next organization’s collections load.
  * Reduced flickering and prevented unnecessary teardown during organization changes.
  * Improved handling of organization changes initiated from outside the switcher.
* **Documentation**
  * Updated sidebar behavior documentation to reflect the improved organization-switching experience.


ashbrener:
Thanks — five of the six are fixed in `34a235ca9`, and I've pushed back on one.

**Mixed org while preloading (P1).** Fair, and the sharpest of these. The retained tree shows the org being left while the org menu and host service already name the destination. The window is now veiled while a switch is in flight: chrome stays visible so the app doesn't look dead, but the mixed state can't be clicked. It's a fixed-position sibling rather than a wrapper around `children`, which would have disturbed the layout it sits in.

**Doc status table (both bots).** Updated — the row said "full clear + rebuild | by design", which contradicted the new section 5.

**Dead helper (P3).** Removed. `preloadActiveOrganizationCollections` had no callers left once the effect replaced it.

**Advancing after a failed preload (Major / P2).** Not changing this, and I'd rather say why than quietly leave it. `preloadCollections` settles instead of rejecting, so a table that can't sync doesn't block the switch — that's deliberate. Requiring every shape to succeed would strand you on the old org whenever one table is unreachable, which is the exact hang this PR exists to remove: Electric retries 5xx and network errors internally and never surfaces them, so the wait would be unbounded. Collections render cache-first here, so a table that hasn't arrived shows its own empty state and fills in when it does. I've put that reasoning in the code rather than leaving the `allSettled` behaviour looking accidental.

Worth noting the failure semantics are unchanged from `main` — the old code awaited the same settle-not-reject preload and then remounted regardless. This PR doesn't make a failed preload worse; it just stops the window being empty while you wait.

ashbrener:
Rebased on #6135 and cut this down to what that left outstanding — 25 lines in the provider now.

Two things I'd said earlier no longer apply, so striking them rather than leaving them to mislead a reviewer:

- I was holding the previous org on screen until the destination's collections were warm. With the preload gate gone there's nothing to wait for, and holding it back would work against rendering cache-first — which is the direction #6135 set. Removed.
- My defence of the `allSettled` behaviour in `preloadCollections` was about a function that has since been rewritten to start every shape and return. Moot.

What's left is the unmount itself: the provider still returns `null` while switching, so the window goes blank for the two round trips either side — with every pane, terminal and drag region torn down and rebuilt. This keeps the tree mounted and veils it instead.

@saddlepaddle thanks for #6135 — the point about SQLite rows hydrating before Electric's ready signal is the part I'd have missed.

ashbrener:
Both of you were right about the veil, and I've removed it rather than hardened it.

It existed to stop anyone acting on a tree that still showed the org being left while the chrome already named the destination. Rebasing onto #6135 removed that state: collections follow the session's org, so the tree swaps to the destination and hydrates cache-first — which is the behaviour that change set out to get. Covering that with an overlay adds nothing, and you're right that what it did cover was only the mouse; keyboard focus walked straight past it. `inert` would have fixed the hole, but the thing it was guarding no longer exists.

That also answers the second P2: with no veil there's no path that needs one, so the invitation / restored-session route isn't left inconsistent.

What's left is the one thing #6135 didn't touch — the provider returning `null` while switching, which tears the tree down for the round trips either side. 10 lines in the provider now.

ashbrener:
@saddlepaddle #6361 is the same shape as #6133 — don't swap the UI out for a transient not-ready state. #6133 is the org-switch case: the authenticated tree unmounts for as long as the destination org takes to preload, so a large org shows a blank window for minutes. Your #6135 removed the preload gate; this is the remaining unmount. 8 lines, and it's rule 9 in AGENTS.md.

Kitenite:
Closing: main already ships this. b9d851f93 (#5337) keeps the tree mounted during org switch. Thanks for the PR; if you think the shipped version misses a case, open an issue and link this.

PR #6138, opened 2026-08-03
fix(desktop): stop v1 renderer xterm from double-answering terminal queries

### LINKED ISSUE #3499: [bug] terminal inside superset buggy when I/O required

distroinfinity:
Fixes #3499 (please reopen, the repro below is deterministic). Related: #3894.

## Problem

Interactive CLIs break in v1 terminals. Example, `gh auth switch`:

```
could not prompt: unexpected escape sequence from terminal: ['\x1b' ']']
```

When a program queries the terminal, two things answer. The daemon's headless emulator replies instantly. The renderer xterm replies again about 40ms later, and that late duplicate lands in the program's stdin while it is reading user input. The escape filter in `Session.write()` only runs during shell startup, so after the first prompt every duplicate gets through. Trace from a local repro:

```
t+0ms   emulator answers  \x1b[2;1R
t+38ms  renderer sends    \x1b]11;rgb:1515/1111/1010\x1b\  plus a duplicate \x1b[2;1R
```

## Fix

The renderer xterm stops auto replying to queries the daemon already answers: DA1, DA2, DSR 5 and 6. It also stops replying to OSC 10/11/12 color queries. Nothing answers those now, which is intentional: programs treat a missing reply as unsupported, and the `TERM_THEME` and `COLORFGBG` env hints cover theme detection. Color set commands still pass through so themes keep working. `CSI ?996n` is untouched because only the renderer can answer it.

v2 is untouched. There the renderer is the only answerer, so suppressing it would hang shells that wait for DA1 (fish waits up to 10 seconds). That is also why this PR takes only the v1 part of what #3894 attempted, with narrower matchers.

## Testing

A new 16 case test suite covers each suppressed query and everything that must keep working. Verified end to end locally: `gh auth switch` failed every time before the fix and works after it, with exactly one reply per query in the trace. Lint, typecheck and the desktop suite pass on current main. Screenshots in the first comment.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/suppressQueryResponses.test.ts
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/suppressQueryResponses.test.ts
@@ -1,0 +1,165 @@
+/**
+ * Verifies the renderer-side suppressors keep terminal queries from
+ * generating replies. In v1 the terminal-host daemon's headless emulator is
+ * the canonical answerer; a renderer reply round-trips back to PTY stdin
+ * ~20-40ms late and lands in whatever program is reading stdin — go-survey
+ * CLIs abort with "unexpected escape sequence from terminal: ['\x1b' ']']"
+ * (#3499).
+ *
+ * Uses `@xterm/headless` (same parser as `@xterm/xterm`; window polyfill
+ * comes from the `apps/desktop/bunfig.toml` test preload). Headless answers
+ * DA/DSR itself, so those are asserted via `onData`. OSC 10/11/12 and
+ * `CSI ?996n` replies are produced by browser-side theme handlers that do
+ * not exist in headless, so those are asserted via the handler chain: a
+ * probe handler registered BEFORE the suppressors stands in for xterm's
+ * earlier-registered built-in handler (most recently registered runs first;
+ * `false` falls through). Suppressed = probe never reached.
+ */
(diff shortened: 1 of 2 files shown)


distroinfinity:
Hey @Kitenite 
I originally filed #3499 and this one has been biting me daily ever since, so I finally sat down and dug into it. Turned out the renderer xterm was echoing a second reply to terminal queries about 40ms after the daemon's emulator already answered, and that late duplicate lands right in the stdin of whatever CLI is prompting. Fix is small and scoped to v1 only.

Tested it locally end to end, gh auth switch went from failing every single time to working.

Would love a review when you get a chance. Happy to adjust anything.

Kitenite:
Closing: this only touches the legacy v1 UI (screens/main), which is being retired. only touches v1 suppressQueryResponses.ts. If the bug still shows in the v2 workspace, please open an issue.

Kitenite:
Thanks for the double-answered-query fix and the test coverage. It targets the v1 xterm path only, which is going away; the v2 runtime handles this differently.

PR #6072, opened 2026-07-31
fix(desktop): scope agent hooks to Superset terminals and add per-agent hooks toggle

saddlepaddle:
**Links**
- ExecPlan: `apps/desktop/plans/done/20260729-1500-agent-hooks-guard-and-per-agent-setting.md`
- Investigation: `HOOKS_INVESTIGATION.md` (root cause of the reported leak)

## Summary
- Every hook command Superset registers in an agent's **global** config is now guarded on `SUPERSET_HOME_DIR`, so it no-ops instantly in agent sessions launched outside Superset terminals. Droid, Codex, and Mastra were writing raw absolute-path commands; they now use the same guarded form as Claude/Grok/Kimi/Vibe.
- The Cursor/Gemini/Copilot hook scripts and `notify.sh` itself bail out unless a Superset-supplied env var is present — closing the hole where a foreign session's own `session_id` payload was enough to make `notify.sh` curl the Electron notification server.
- New per-agent **"Superset hooks"** toggle in Settings → Agents: disabling removes Superset's managed entries from that agent's global config immediately and re-enforces the removal on every startup.

## Why / Context
A user running Factory Droid in a plain terminal (outside Superset) saw this on every message:

    Hooks Stop
    └── SUPERSET_AGENT_ID=droid '/Users/<user>/.superset/hooks/notify.sh' : Exit code 0

Superset deliberately registers hooks in each tool's user-global config (since #2479, because PATH-wrapper injection broke under mise/asdf/ZDOTDIR), with runtime env-var gating as the intended scoping. Droid (#2135) predates the guard (#2621/#5552) and never got it; Codex and Mastra were also unguarded. Full chain of causes in `HOOKS_INVESTIGATION.md`.

## How It Works
- `getManagedNotifyHookCommand(agentId)` emits `[ -n "$SUPERSET_HOME_DIR" ] && … || true`; only Superset-spawned terminals set that var (v2 via host-service `env.ts`, v1 by inheriting the main process env — pinned by `env.test.ts`).
- A shared `DYNAMIC_NOTIFY_PATH_MARKER` (`$SUPERSET_HOME_DIR/hooks/notify.sh`) was hoisted into `agent-wrappers-common.ts` and added to the droid/codex/mastra managed-command predicates. Without it every restart would append a duplicate guarded entry, since the guarded form contains neither an absolute notify path nor a `/.superset/` segment. Old unguarded entries are reaped by the existing merge machinery on next startup.
[...]

PR #6149, opened 2026-08-04
fix(desktop): stabilize Codex working indicator

Roshvan:
## Summary

- resolve Codex native lifecycle hooks through the active `SUPERSET_HOME_DIR` at runtime instead of persisting one worktree's absolute `notify.sh` path
- recognize the shared runtime command as Superset-managed so hook reconciliation stays idempotent
- preserve user-defined Codex hooks while replacing stale Superset paths

## Root cause

Claude already stores a guarded runtime command in its global hook config, but Codex stored the absolute notify path from whichever Superset instance initialized most recently. Starting a different dev worktree rewrote that global path and Codex's exact hook-definition trust entry; removing that worktree left native `SessionStart`, `UserPromptSubmit`, and `Stop` hooks pointing at a missing script. Codex itself continued working, but Superset never received the lifecycle event that drives the sidebar indicator.

The wrapper log watcher could mask this for the default preset, which is why the failure is most visible when Codex is launched directly or the watcher cannot observe the current session format.

This is intentionally the narrow hook-path correction. It overlaps the Codex command portion of #6072, while leaving that PR's broader per-agent toggle and hook-scoping work untouched.

## Verification

- `bun test apps/desktop/src/main/lib/agent-setup/agent-wrappers.test.ts` — 52 passed
- `bun run lint` — passed, zero warnings
- `bun run --cwd apps/desktop typecheck` — passed
- `bun run --cwd apps/desktop compile:app` — passed
- exact-worktree Electron CDP journey — passed on renderer `http://localhost:3205`, CDP `9341`, route `/v2-workspace/f1582787-5b36-4eba-a906-c13bf7000263`
  - opened **Open Terminal** through real pointer input
  - launched `/Users/roshvan/.local/bin/codex` directly with native hooks enabled
  - submitted a real prompt through terminal input
  - confirmed the sidebar spinner appears while Codex is working and clears after `Stop`
- `bun run typecheck` — unrelated current-main failure in `apps/docs`: cannot resolve `@superset/ui/overflow-fade-container`; desktop-scoped typecheck passes

## Visual verification

**Before — stale absolute hook:** Codex is working, but `SessionStart` and `UserPromptSubmit` fail with code 127 and the sidebar remains idle.

Codex working while stale native hooks fail and the sidebar remains idle

**After — active Codex turn:** native hooks run successfully and the amber spinner is visible beside Delta newest.

Codex working with the amber Superset sidebar spinner visible

**After completion:** the response completes and the sidebar returns to its idle dot after the `Stop` lifecycle.

Codex response complete and the Superset sidebar indicator returned to idle

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex and Claude managed hook detection across different worktree locations.
  * Updated managed hooks to use a consistent environment-based notification path, preventing stale or location-specific configurations.
  * Preserved user-defined hooks while safely replacing outdated managed entries.
  * Prevented managed hooks from being added to unsupported event types.
  * Improved hook reliability when the notification script is unavailable or not executable.

### DIFF
--- a/apps/desktop/src/main/lib/agent-setup/agent-wrappers-claude-codex-opencode.ts
+++ b/apps/desktop/src/main/lib/agent-setup/agent-wrappers-claude-codex-opencode.ts
@@ -7,6 +7,7 @@ import {
 	getManagedNotifyHookCommand,
 	isSupersetManagedHookCommand,
 	MANAGED_NOTIFY_RELATIVE_PATH,
+	MANAGED_NOTIFY_RUNTIME_PATH,
 	writeFileIfChanged,
 } from "./agent-wrappers-common";
 import { getNotifyScriptPath, NOTIFY_SCRIPT_NAME } from "./notify-hook";
@@ -67,7 +68,7 @@ interface ClaudeSettingsJson {
 	[key: string]: unknown;
 }
 
-const DYNAMIC_NOTIFY_PATH_MARKER = `$SUPERSET_HOME_DIR/${MANAGED_NOTIFY_RELATIVE_PATH}`;
+const LEGACY_DYNAMIC_NOTIFY_PATH_MARKER = `$SUPERSET_HOME_DIR/${MANAGED_NOTIFY_RELATIVE_PATH}`;
 
 function isPlainObject(value: unknown): value is Record<string, unknown> {
 	return typeof value === "object" && value !== null && !Array.isArray(value);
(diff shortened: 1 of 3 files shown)


Roshvan:
## Visual evidence

**Before — stale absolute hook:** Codex is working, but SessionStart and UserPromptSubmit fail with code 127 and the sidebar remains idle.

clipboard


**After — active Codex turn:** native hooks run successfully and the amber spinner is visible beside Delta newest.

clipboard


**After completion:** the response completes and the sidebar returns to its idle dot after the Stop lifecycle.

clipboard

PR #6165, opened 2026-08-04
feat(relay): relay2 — canonical Durable Objects relay on PartyKit (tunnel v2)

saddlepaddle:
## What

`apps/relay2` — a ground-up relay on Cloudflare Workers + Durable Objects (PartyKit's `partyserver`), with a new host↔relay protocol designed around what a globally-addressable DO makes possible. Superseded and replaced the wire-compatible `apps/relay-do` prototype this PR started as.

**Nothing in production changes on merge.** relay2 is additive; hosts keep using the Fly relay until an operator points them elsewhere. The only behavioral change to shipped code is that host-service probes the relay's `/health` at startup to pick a protocol (anything without `proto: 2` → today's v1 client).

## The protocol (tunnel v2)

One small JSON control channel per host. For each stream, the relay issues a one-time ticket, the host **dials back** a fresh WebSocket, and the DO splices bytes verbatim.

| Fly relay | relay2 |
|---|---|
| Upstash directory + Lua register/unregister/sweep | `getServerByName(hostId)` |
| fly-replay routing + 6PN cross-machine WS bridge | (deleted) |
| `ws:open/ws:frame/ws:close` JSON envelopes, base64 binary, correlation maps | raw spliced sockets, binary passthrough |
| hand-rolled reconnect/backoff/watchdog on both ends | `partysocket` |
| SIGTERM drain choreography + in-band drain frames | DO restart → ~1s reconnect |
| 7 always-on performance-2x/4GB machines | usage-priced; hibernation makes idle tunnels ~free |

HTTP stays (as an isolated `http-exchange.ts`): the SDK, MCP, and the API's Slack path call hosts from serverless runtimes that can't hold a WebSocket.

## Verification

`apps/relay2/scripts/e2e-probe.ts` runs the **real** `TunnelClientV2` against a deployed Worker — ALL PASS on every deploy, including regression tests for two bugs found in review:

    PASS  _whoowns 200
    PASS  client WS opens
    PASS  10 text echoes — rtt min=43 median=49 max=131ms
    PASS  ping-shaped frame splices verbatim
    PASS  64KB binary echo intact
    PASS  HTTP proxy round-trip + host-secret injection
    PASS  bogus dial ticket refused — close=1008
    PASS  _whoowns 503 after close

Also driven end-to-end from the desktop app against a deployed Worker on real Durable Objects: all orgs negotiated v2, terminals/diffs/event bus healthy.

Prior prototype measurements on the same host: keystroke echo **121–125ms vs Fly's 124ms**, 1.49MB PTY burst clean (max frame 12.4KB, well under the 1MB DO limit), `wrangler deploy` mid-session recovered in ≤10s.

## Review findings fixed

Two rounds of adversarial review; every finding fixed and re-verified:

- **`/v2/dial` accepted unknown tickets** and left the socket open forever — unauthenticated callers could spin up DOs and leak sockets. Now closed 1008.
- **DO-wide `setWebSocketAutoResponse` swallowed tunneled frames** matching the keepalive payload, silently breaking verbatim splice. Keepalives now handled per-connection for the host only.
- Host-reconnect race (state set after evicting the replacement; teardown keyed off state instead of synchronous tags) — the same class as the gray-dot presence bug seen in prod.
- Non-atomic read-then-write on presence version **and** on the session token.
- Orphaned dial sockets, `content-encoding`/`content-length` forwarded over an already-decompressed body, unchunked bodies vs the DO per-message ceiling, silent frame drops on overflow, unsendable close codes.

## Also here

- `plans/20260804-1815-relay2-durable-objects-canonical.md` — the ExecPlan, with the Decision Log (D-1…D-19) recording why each call was made. Remaining milestones: resumable streams, API-served endpoint config, `relay.superset.sh` cutover via Cloudflare for SaaS, then delete `apps/relay` and rename.
- electric-proxy's `wrangler`/`workers-types` aligned with relay2 — v4 and v5 coexisting split `@cloudflare` type identity and broke that app's typecheck plus sherif.

## Before merging

relay2 has **no CI deploy workflow yet** (needs a `CLOUDFLARE_API_TOKEN` secret) — deploys are manual `wrangler deploy` until that lands. Given the July outage came from a silently-failing deploy, the workflow should fail loudly to Slack when it's added.

https://claude.ai/code/session_01A2JifsfA63nWthegPwNt8G


saddlepaddle:
## Prod canary results (2026-08-04, sandbox host on Azure, client in SF)

Same host, same PTY, same test script, head-to-head:

| Metric | Fly relay | DO relay |
|---|---|---|
| WS connect (cold) | 149ms | ~800ms |
| Keystroke echo RTT (median of 10) | 124ms | 121–125ms |
| Keystroke min / max | 80 / 149ms | 84 / 156ms |
| 1.49MB output burst | 560ms (2.6 MB/s) | 777ms (1.9 MB/s) |
| Largest WS frame observed | 4.1KB | 12.4KB |

Also verified end-to-end through the DO relay: host register/unregister (clean setOnline writes), CLI → API → relay → host tRPC (~0.5s incl. CLI boot), project create with git clone, terminal create/send/read, and **redeploy survival** — `wrangler deploy` restarted the DO mid-session and the host was reconnected and serving within ≤10s with no intervention.

**Verdict on the three risks:** keystroke latency is a wash (the headline risk is cleared); the 1MB DO frame limit is a non-issue (PTY frames chunk at ~12KB); deploy-drop recovery is fast. Cold connect is ~650ms slower (Worker-edge JWT verify + uncached access check + DO wake) — optimizable, and irrelevant once attached. Burst throughput trails Fly ~40% but both are far above terminal needs.

Operational notes from the run: `superset start --daemon` does not propagate `RELAY_URL` from the caller's env to the daemonized process (foreground start does); and one daemon restart failed to pick up the flag override for reasons not yet explained (first start and API-side evaluation both worked) — worth a look before any real rollout.

https://claude.ai/code/session_01A2JifsfA63nWthegPwNt8G

saddlepaddle:
## relay2 (tunnel v2 on PartyKit) built, deployed, E2E-passing

The spike from the ExecPlan resolved in favor of full PartyKit: `apps/relay2` is a `partyserver` Durable Object per host + `partysocket` control channel on the host side. The dial-back protocol replaces the entire hand-rolled multiplexer — every terminal/event stream is its own WebSocket spliced verbatim (no JSON envelopes, no base64), HTTP rides one-exchange dials, and idle-host pings are answered by hibernation auto-response without waking the object.

Protocol selection is negotiated, not configured: the host probes the relay's `/health` for `proto: 2`, so the existing `relay-url-override` flag keeps choosing *which* relay while the relay itself declares *what it speaks*. Client-facing routes are wire-identical to the v1 relay, so desktop/web/CLI clients need zero changes.

E2E probe (`apps/relay2/scripts/e2e-probe.ts`, real `TunnelClientV2` + fake local service) against the deployed Worker — ALL PASS ×2: `_whoowns`, WS splice text + 64KB binary intact (median echo 49–55ms, both probe ends in SF), HTTP proxy with host-secret injection, offline 503 after tunnel close.

Canary build from this branch is running; the flag payload flips to `superset-relay2` once it publishes. Known deferred items are recorded in the plan (CI deploy workflow with loud failure, versioned setOnline API column, resumable streams = M3).

PR #6169, opened 2026-08-05
feat(desktop): highlight search-matched text on settings pages

AviPeltz:
## Summary
- Add a `HighlightText` component that wraps matched search-term substrings in `<mark>`, reusing the app's existing `--highlight-match` CSS token (already used by in-app markdown/chat/diff search) so the color is consistent and theme-aware for free.
- Wire it into the settings search flow: the three shared row/section primitives (`SettingsRow`, `ConfigRow`, `SettingsSection`) plus every individual settings page's label/description text across Account, Organization, Appearance, Behavior, Security, Git, Experimental, Terminal, Links, Members, Hosts, Integrations, API Keys, Project, Billing, Ringtones, Keyboard, and Teams.
- Consolidate a few settings pages that had reimplemented their own local row component instead of importing the shared one (`AccountSettings`, `OrganizationSettings`, `ThemeSection`, `PermissionsSettings`, `ProjectSettings`), so they pick up highlighting without duplicating the wrap logic per call site.

## Why / Context
Settings search already filters the sidebar down to matching sections and shows a match-count badge, but once you land on a page there was no indication of *where* the term actually matched — you had to scan the whole page yourself. This adds the highlight so the matched substring is visible directly on the row.

## How It Works
- `splitSearchTerms(query)` is extracted out of the existing `searchSettings` filter logic in `settings-search.ts` and reused by `HighlightText`, so filtering and highlighting can never drift out of sync (same term-splitting, same case-insensitive matching).
- `HighlightText` is a pure `{ text, query }` component — no store coupling — that case-insensitively matches each search term against the text, escapes regex metacharacters, and wraps matches in `<mark className="rounded-sm bg-highlight-match text-inherit">`. Empty query or no match returns the original text unchanged (no wrapper element).
- `--color-highlight-match` is promoted from an existing `--highlight-match` CSS variable into a real Tailwind utility (`bg-highlight-match`) in `globals.css`, mirroring how `--highlight`/`--highlight-foreground` were already promoted. No new color introduced.
- Coverage is scoped to text that corresponds to an entry in the static `SETTINGS_ITEMS` search registry — dynamic/user-supplied text (preset names, member names/emails, org names, script contents, etc.) is intentionally never wrapped, since none of that is part of what the search box can actually match against.

## Manual QA Checklist

Verified end-to-end against the running dev app via CDP (real keyboard shortcut to open Settings, real dispatched keystrokes into the search box, real sidebar clicks — not synthetic DOM/state assignment):

- [x] Cmd+, opens Settings; typing "worktree" highlights the matching substring in both the row label ("**Worktree** location") and hint text ("new **worktree**s on this device") on the Git & Worktrees page
- [x] Clicking a different matching section in the sidebar preserves highlighting on the new page
- [x] Clearing the search input removes all `<mark>` elements with no leftover DOM wrapper (0 count, plain text restored)
- [x] Dark theme: highlight background resolves to `rgba(224, 120, 80, 0.2)` — matches `globals.css` exactly
- [x] Light theme (switched via the real theme picker): highlight background resolves to `rgba(255, 211, 61, 0.35)` — matches `globals.css` exactly, confirmed the token flips correctly since this app gates dark as the `:root` default and light as the override, not the usual `dark:` Tailwind variant
- [x] Partial-word matching renders correctly (e.g. "theme" highlights inside "themes" without highlighting the trailing "s")

## Testing
- `bun run lint:fix` / `bun run lint` — 0 issues
- `bun run typecheck` (filtered to `@superset/desktop`) — passes
- `bun test` (`apps/desktop`) — 2508 pass, 0 fail
- New unit tests: `HighlightText.test.tsx` covers empty query, single/multiple terms, case-insensitivity, regex-metacharacter queries, and no-match passthrough

## Design Decisions
- **Why not migrate every duplicated local row component to the shared `SettingsRow`**: a few pages (e.g. `AccountSettings`) had local row variants with slightly different className details (padding, `Label` vs `div`) than the shared component. Rather than risk a visual regression I couldn't individually verify across every page, I added highlighting in-place inside each local wrapper instead of swapping components — same outcome, lower risk. Only wrapper components with no material layout differences were consolidated onto the shared `SettingsRow`/`SettingsSection`.
- **Why reuse `--highlight-match` instead of a new token**: this exact codebase already has a "search match" color for the in-app markdown/chat/diff search UI (via the CSS Custom Highlight API). Reusing it keeps "search match" visually consistent everywhere instead of introducing a second, slightly different highlight color.

## Known Limitations
- Page-level `<h2>` headers only highlight when the header text closely matches a real `SETTINGS_ITEMS` registry title (e.g. "API keys", "Teams"). Headers with no corresponding registry item (e.g. "Account", "Integrations", "Permissions") are left unwrapped, since there's nothing in the registry for that literal page-title text to match against.
- Coverage is scoped to plain string label/title/hint/description props feeding through row components — content rendered as complex `ReactNode` (e.g. the Appearance "Theme" row's hint, which contains marketplace/docs links) is intentionally left unwrapped rather than risk mangling embedded interactive elements.

---
## Summary by cubic
Highlight search matches across settings so users can see exactly where their query applies. Adds a reusable `HighlightText` component and wires it into labels, hints, headings, dialog titles, and integration names with the existing theme-aware match color.

- **New Features**
  - Added `HighlightText` to wrap matched terms in <mark> using `bg-highlight-match`; integrated across shared rows/sections and individual pages (labels, hints, page/dialog titles, integration names).
  - Matching reuses `splitSearchTerms` for consistent, case-insensitive, regex-safe behavior and prefers longer terms first; scoped to registry-backed static text only.

- **Bug Fixes**
  - Prevented layout gaps by wrapping output in a single `<span>` and fixed missing highlighting in the normal V2 "Terminal daemon" path.
  - Wrapped previously missed hints/descriptions (Git, Experimental, Wait-For-Setup, Worktree Location, Theme “Custom themes”); added tests for empty queries, multi-term cases, regex metacharacters, longer-term precedence, and the single-wrapper behavior.

<sup>Written for commit 4d29ad7af309c704aab1dc3b99febe58b4300470. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6169?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Settings search results now highlight matching terms across settings pages, labels, headings, descriptions, hints, and integration names.
  * Highlighting supports multiple terms, case-insensitive matching, overlapping terms, and special characters.
  * Added consistent visual styling for matched text throughout settings.

* **Tests**
  * Added coverage for empty queries, whitespace, multiple matches, case handling, unmatched text, overlapping terms, and special characters.

### DIFF
--- a/apps/desktop/src/renderer/globals.css
+++ b/apps/desktop/src/renderer/globals.css
@@ -145,6 +145,7 @@
 	--color-sidebar-ring: var(--sidebar-ring);
 	--color-highlight: var(--highlight);
 	--color-highlight-foreground: var(--highlight-foreground);
+	--color-highlight-match: var(--highlight-match);
 	--color-fill-hover: var(--fill-hover);
 	--color-fill-selected: var(--fill-selected);
 }
--- a/apps/desktop/src/renderer/routes/_authenticated/settings/account/components/AccountSettings/AccountSettings.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/settings/account/components/AccountSettings/AccountSettings.tsx
@@ -9,12 +9,12 @@ import { apiTrpcClient } from "renderer/lib/api-trpc-client";
 import { authClient } from "renderer/lib/auth-client";
 import { electronTrpc } from "renderer/lib/electron-trpc";
 import { useCollections } from "renderer/routes/_authenticated/providers/CollectionsProvider";
+import { HighlightText } from "renderer/routes/_authenticated/settings/components/HighlightText";
 import { useSettingsSearchQuery } from "renderer/stores/settings-state";
 import {
 	getImageExtensionFromMimeType,
 	parseBase64DataUrl,
 } from "shared/file-types";
-import { HighlightText } from "../../../components/HighlightText";
 import {
 	isItemVisible,
 	SETTING_ITEM_ID,
--- a/apps/desktop/src/renderer/routes/_authenticated/settings/account/components/AccountSettings/AccountSettings.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/settings/account/components/AccountSettings/AccountSettings.tsx
@@ -9,10 +9,12 @@ import { apiTrpcClient } from "renderer/lib/api-trpc-client";
 import { authClient } from "renderer/lib/auth-client";
 import { electronTrpc } from "renderer/lib/electron-trpc";
 import { useCollections } from "renderer/routes/_authenticated/providers/CollectionsProvider";
+import { useSettingsSearchQuery } from "renderer/stores/settings-state";
 import {
 	getImageExtensionFromMimeType,
 	parseBase64DataUrl,
 } from "shared/file-types";
+import { HighlightText } from "../../../components/HighlightText";
 import {
 	isItemVisible,
 	SETTING_ITEM_ID,
@@ -183,12 +185,18 @@ interface SettingRowProps {
 }
 
 function SettingRow({ label, hint, children }: SettingRowProps) {
+	const searchQuery = useSettingsSearchQuery();
+
 	return (
 		<div className="flex items-center justify-between gap-8">
 			<div className="flex-1 min-w-0">
-				<div className="text-sm font-medium">{label}</div>
+				<div className="text-sm font-medium">
+					<HighlightText text={label} query={searchQuery} />
+				</div>
 				{hint && (
-					<div className="text-xs text-muted-foreground mt-0.5">{hint}</div>
+					<div className="text-xs text-muted-foreground mt-0.5">
+						<HighlightText text={hint} query={searchQuery} />
+					</div>
 				)}
 			</div>
 			<div className="flex-shrink-0">{children}</div>
--- a/apps/desktop/src/renderer/routes/_authenticated/settings/api-keys/components/ApiKeysSettings/ApiKeysSettings.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/settings/api-keys/components/ApiKeysSettings/ApiKeysSettings.tsx
@@ -26,6 +26,8 @@ import { useCopyToClipboard } from "renderer/hooks/useCopyToClipboard";
 import { apiTrpcClient } from "renderer/lib/api-trpc-client";
 import { authClient } from "renderer/lib/auth-client";
 import { useCollections } from "renderer/routes/_authenticated/providers/CollectionsProvider";
+import { HighlightText } from "renderer/routes/_authenticated/settings/components/HighlightText";
+import { useSettingsSearchQuery } from "renderer/stores/settings-state";
 import {
 	isItemVisible,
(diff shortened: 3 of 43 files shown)


AviPeltz:
Addressed the review feedback from coderabbitai, cubic-dev-ai, and greptile-apps in 4d29ad7:

**Fixed:**
- **Flex-gap spacing bug** (coderabbitai ×6, cubic ×4) — `Label` defaults to `flex items-center gap-2`, so `HighlightText`'s bare `Fragment` output (text + `<mark>` as separate nodes) was getting `gap-2` inserted between the highlighted and unhighlighted segments. Rather than wrap every call site individually as suggested, fixed it once at the source: `HighlightText` now always renders inside a single `<span>`, so it's exactly one flex item regardless of how many internal parts it has. This covers all the flagged call sites (`SettingsRow`, `ConfigRow`, `SettingsSection`, `OrganizationSettings`, `PermissionsSettings`, `ProjectSettings` ×2) plus any future ones.
- **Prefix-overlap matching** (cubic) — when a query has two terms where one is a prefix of the other (e.g. "bran branch"), terms are now matched longest-first so the shorter term can't shadow part of the longer one.
- **V2SessionsSection highlighting only wired into the rare loading branch** (coderabbitai + cubic) — the "Terminal daemon" heading in the actual common-case render path (`V2SessionsSectionInner`) was never wrapped, only the transient "host service starting" fallback was. Fixed so both are.
- **Missing hint/description highlighting** (cubic, several files) — `GitSettings`, `ExperimentalSettings` (all 4 rows), `WaitForSetupBeforeAgentSetting`, `UserWorktreeLocationSection` (V1 + the static branch of V2), and `ThemeSection`'s "Custom themes" hint now highlight their static description text consistently with their labels.
- **Deep relative imports** (coderabbitai nitpick) — switched the 4 flagged `HighlightText` imports to the `renderer/...` alias.

Added regression tests for the span-wrapping and prefix-overlap fixes; full suite (2510 tests), lint, and typecheck all pass.

**Left as-is:** the "keyword-only matches remain invisible" point (greptile P1, cubic on `SettingsRow`) — when a query only matches an item via its `keywords` array (a synonym not shown anywhere in the UI), the row correctly appears in results but nothing highlights, since there's no visible occurrence of that keyword to mark. That's inherent to keywords being invisible search synonyms, not a bug in the highlighting itself — the alternative (surfacing the matched keyword somewhere in the row) is a bigger, separate UX design change I'd rather not bundle into this PR. Happy to file it as a follow-up if that behavior is worth improving.

PR #6176, opened 2026-08-05
fix(host-service): run teardown scripts on CLI/SDK/MCP workspace deletes

### LINKED ISSUE #6174: Teardown scripts are skipped when a workspace is deleted via MCP/CLI/SDK (force: true is hardcoded)
## Summary

`.superset/config.json` teardown commands (and `.superset/teardown.sh`) run correctly when a workspace is deleted from the Superset app UI, but are silently skipped when the workspace is deleted through the MCP tool `workspaces_delete` (and, from reading the code, the CLI/SDK surface too).

## Steps to reproduce

1. Add a teardown script to a project:
   ```json
   { "setup": ["./.superset/setup.sh"], "teardown": ["./.superset/teardown.sh"], "run": [] }
   ```
2. Create a workspace (setup runs fine).
3. Delete the workspace via the MCP `workspaces_delete` tool.

**Expected:** teardown commands run before the worktree is removed (as documented at https://docs.superset.sh/setup-teardown-scripts, and as the UI delete does).

**Actual:** the worktree is removed, the response is `{"success":true,...,"warnings":[]}`, and the teardown script never executes — external resources it should clean up (in our case a Herd site link and a cloned MySQL database) are left behind. There is no trace in host-service.log and no warning anywhere.

## Root cause

In `@superset/host-service` (`src/trpc/router/workspace-cleanup/workspace-cleanup.ts`), `runDestroy` skips step 1 (teardown) when `force` is set. The external delete surface hardcodes it:

```ts
// Legacy external surface used by CLI/SDK/MCP. Preserve its
// non-interactive contract while reusing the v2 cleanup path.
return destroyWorkspace(ctx, {
    workspaceId: input.id,
    deleteBranch: false,
    force: true,
});
```

So every non-interactive delete is a force delete, and force semantics ("skips teardown (step 1)") turn the documented teardown contract into UI-only behavior.

## Why this matters

The MCP surface is exactly the one agents use for automated workspace lifecycles — the flow that most needs deterministic cleanup. Provisioned per-workspace resources (databases, vhosts, tunnels) leak on every programmatic delete, and the empty `warnings` array makes it look like everything was cleaned up.

## Suggested fix

Run teardown on the non-interactive path too, e.g.:
- default `force: false` for the external surface and expose an optional `force` parameter on `workspaces_delete`, mapping a teardown failure to a typed error the caller can retry with `force: true`; or
- keep `force: true` for the worktree/branch semantics but still attempt teardown (best-effort), surfacing a failure in `warnings`.

Either would restore the documented behavior for programmatic deletes.

**Version:** Superset.app 1.16.1 on macOS (Darwin 25.4.0). Workaround we use meanwhile: open a terminal in the workspace, run the teardown script, wait, then call `workspaces_delete`.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

Kitenite:
## Summary

Fixes #6174.

`.superset` teardown scripts only ran on the interactive (desktop/mobile) delete path. The legacy external surface `workspace.delete` — the one CLI, SDK, and MCP `workspaces_delete` all funnel through — hardcoded `force: true`, and `force` doubled as "skip teardown". Every programmatic delete silently skipped the user's teardown script, leaked externally provisioned resources (databases, vhosts, tunnels), and returned a clean `{"success":true,...,"warnings":[]}`.

## Fix

Split the two meanings that were riding on one flag:

- `force` now carries only the destructive **git** semantics: skip the dirty-worktree preflight, `--force --force` worktree removal.
- A new `teardownMode` on `DestroyWorkspaceInput` governs step 1:
  - `"blocking"` — failure throws the typed `PRECONDITION_FAILED` so the interactive caller can prompt force-retry (unchanged v2 renderer contract; `workspaceCleanup.destroy` maps `force → skip`, else `blocking`).
  - `"best-effort"` — the external surface: teardown always runs; a failure doesn't block the delete (there's nobody to prompt) and lands in `warnings` with exit info and the ANSI-stripped output tail. The CLI already prints `warnings`, so failures surface there too.
  - `"skip"` — the interactive force-retry escape hatch.

Also updated the MCP tool description and the setup/teardown docs to state the non-interactive contract.

## Tests

New `workspace-delete-teardown.integration.test.ts` (real git fixture + fake pty-daemon, extracted the spawner from `teardown.integration.test.ts` into a shared helper):

1. `workspace.delete` runs teardown before removing the worktree (fails on the old behavior — verified by mutating the fix back).
2. A failing teardown (`exit 7`) doesn't block the delete and lands in `warnings` (also fails on the old behavior).
3. `workspaceCleanup.destroy` with `force: true` still skips teardown (guards the renderer's force-retry contract).

Existing delete-path suites (`workspace-cleanup`, `workspace-create-delete`, `bug-hunt-*`) pass unchanged; `bun run typecheck` and `bun run lint` are clean.

---
## Summary by cubic
Ensure teardown scripts run on non-interactive workspace deletes (CLI/SDK/MCP) to prevent leaked resources. Split git “force” from teardown behavior and surface teardown failures as warnings.

- **Bug Fixes**
  - `workspace.delete` now runs teardown before removing the worktree; failures don’t block and appear in `warnings` with an ANSI-stripped output tail.
  - Separated git `force` from teardown via `teardownMode` on `DestroyWorkspaceInput` (`blocking` | `best-effort` | `skip`).
  - Interactive v2 flow unchanged: normal delete blocks on failure; force-retry skips teardown.
  - Updated docs and MCP `workspaces_delete` description for the non-interactive contract; added integration tests that pin ordering and warning tail content.

<sup>Written for commit 3db5fb4449d55554f4f4a75f8670c048190ee83a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6176?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Workspace deletion now runs configured teardown scripts across CLI, SDK, and MCP workflows.
  * Non-interactive deletions continue when teardown fails and return sanitized warnings.
  * Forced interactive cleanup explicitly skips teardown scripts.

* **Documentation**
  * Updated setup and workspace deletion guidance to explain teardown behavior and failure handling.

* **Tests**
  * Added integration coverage for successful teardown, warning-based failures, and forced cleanup behavior.

### DIFF
--- a/apps/docs/content/docs/setup-teardown-scripts.mdx
+++ b/apps/docs/content/docs/setup-teardown-scripts.mdx
@@ -175,6 +175,8 @@ When teardown fails:
 2. Click "Delete Anyway" to force-delete immediately
 3. Or click "View Logs" to review the failure, then force-delete from the dialog
 
+Deleting a workspace through the CLI, SDK, or MCP also runs teardown. Since there's no one to prompt on those surfaces, a teardown failure doesn't block the delete — it's reported in the response's `warnings` instead.
+
 ## Tips
 
 - Keep setup fast: it runs on every workspace creation
--- a/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
@@ -1,4 +1,5 @@
 import { existsSync } from "node:fs";
+import { sanitizePromptForPty } from "@superset/shared/agent-prompt-launch";
 import { TRPCError } from "@trpc/server";
 import { z } from "zod";
 import { invalidateLabelCache } from "../../../ports/static-ports";
@@ -34,6 +35,19 @@ export interface DestroyWorkspaceInput {
 	workspaceId: string;
 	deleteBranch: boolean;
 	force: boolean;
+	/**
+	 * Teardown (step 1) behavior — deliberately separate from `force`, which
+	 * only carries the destructive git semantics (skip preflight, double-force
+	 * worktree removal):
+	 *   - "blocking":    a failed script throws PRECONDITION_FAILED so an
+	 *                    interactive caller can prompt a force-retry.
+	 *   - "best-effort": always runs; a failure degrades to a warning. For
+	 *                    non-interactive callers (CLI/SDK/MCP) with nobody to
+	 *                    prompt — skipping instead would leak the resources the
+	 *                    script provisions (#6174).
+	 *   - "skip":        don't run — the interactive force-retry contract.
+	 */
+	teardownMode: "blocking" | "best-effort" | "skip";
 }
 
 /**
@@ -121,7 +135,7 @@ export const workspaceCleanupRouter = router({
 	 * Destroy a workspace in five phases:
 	 *
 	 *   0. Preflight     — dirty-worktree check (skip if force)
-	 *   1. Teardown      — run .superset/teardown.sh (skip if force)
+	 *   1. Teardown      — run .superset/teardown.sh (per teardownMode)
 	 *   2. Local cleanup — PTYs, worktree
 	 *   3. Cloud delete  ← authoritative UI state
 	 *   4. Branch delete — optional local branch cleanup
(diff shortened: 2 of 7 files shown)


Kitenite:
Addressed review feedback in 3db5fb444: the ordering test now proves teardown ran while the worktree still existed (`test -d` gate on the marker), and the failure test asserts the warning carries the script's output tail, not just the exit code.

Skipping the sanitizer-naming suggestions (greptile P2 / cubic P3 on `sanitizePromptForPty`): the CSI/OSC/control-char stripping is exactly what the warnings channel needs, the tab/CRLF normalization side effects are harmless for plain-text warnings, and unterminated-OSC fragments are a cosmetic edge case in an already best-effort diagnostic string. Not worth a shared-package refactor here.

PR #6227, opened 2026-08-06
fix(api): preserve terminal scheduled automation dispatch

### LINKED ISSUE #6128: Scheduled automation dispatch silently lost while target host is verifiably online; manual 'automations run' succeeds minutes later
## Summary

A scheduled automation slot produced **no run row, no `workspaces.create` on the target host, and the automation auto-disabled** (COUNT exhausted) — while the target host was **verifiably online** (24/7 systemd-managed host service with an active watchdog, uptime confirmed at the slot minute). A manual `superset automations run <id>` on the **same automation and same host 15 minutes later** worked end-to-end (run row → workspace → agent executed → completed).

This rules out host availability as the cause for this loss class. Across our org we've now recorded **0 successful scheduled dispatches in 8 attempts** spanning 4 automations and 2 hosts (7 on a Mac desktop host, 1 on the always-on Linux host); every manual Run-now attempt in the same period succeeded.

## Reproduction evidence (org `3ee200f3-c54c-46b1-b8b8-24a7f27348f3`, happy to share more via support)

| | Scheduled slot | Manual run |
|---|---|---|
| Automation | `75c82d06-77af-454c-9f0c-e6c617ea702b` | same |
| Target host | `549120a9a28fe994cb2f93307f065203` (Linux ARM, superset host service v1.18.1, systemd + watchdog, online) | same |
| Time (UTC) | 2026-08-02T18:46:00 (rrule FREQ=DAILY;COUNT=1, TZ Asia/Kolkata) | 2026-08-02T19:01:06 |
| `automations logs` row | **none** | `3166c37c-add6-4382-ad07-44c816edb03e`, status `dispatched` |
| Host-side `workspaces.create` | **none** (worktree dir mtimes checked) | workspace `c9b614ba-6e58-47b7-b1cd-a74f0dd51c9a` created |
| Outcome | automation flipped `enabled:false` (COUNT consumed), indistinguishable from a completed series | agent ran and completed |

Earlier Mac-host losses (same signature, previously attributed to the host being asleep): automations `49610c78…` 0/3, `8e91cf07…` 0/3, `5900a8de…` 0/1 — statuses seen there included `skipped_offline` and `dispatch: workspaces.create: Tunnel disconnected`; the always-on-host loss above left no status at all.

## Why this is nasty

1. A bounded (COUNT=N) automation that loses its final slot **auto-disables with no run row**, which reads exactly like a healthy finished series — silent data-free failure.
2. There appears to be **no queue/retry/backfill**: a missed minute is gone forever.
3. Users migrating automations to always-on hosts specifically to avoid offline-loss (as we did) still lose slots.

## Asks

- Retry/queue semantics (or at least a persisted `dispatch_failed` row) for scheduled dispatches.
- Don't consume COUNT on a slot that never produced a dispatch attempt record.
- Any pointer to where scheduled dispatch diverges from the Run-now path — Run-now is 100% reliable for us (3/3 on the Linux host tonight, including via host-local `superset automations run` under systemd).

Related support thread: SUPER-1644. Full internal evidence log: https://github.com/jaseemts/rabix/issues/510

abcdlsj:
## Summary

Fixes #6128.

A bounded scheduled occurrence could be accepted by QStash and then dropped because evaluation disabled the automation before delivery. The dispatch route returned 200, so no retry, failure callback, or run record followed.

## Design

- Evaluate marks the queued payload as terminal when the recurrence has no next occurrence.
- After enqueueing, evaluation moves that occurrence to an occurrence-specific future reservation in nextRunAt, using a CAS on enabled and the original nextRunAt. This keeps the terminal row out of the due scan while QStash delivers it, without a schema migration.
- Dispatch accepts a terminal message while the reservation matches, and disables the automation only after dispatchAutomation records an outcome. A user pause is still rejected by the enabled guard; a changed reservation is stale.
- The reservation is derived from the original occurrence, so a delayed message cannot claim a later terminal occurrence for the same automation.
- Terminal finalization failures return non-2xx for QStash retry. If delivery retries are exhausted, run-failed records dispatch_failed and closes the matching recurrence. Existing dispatched or skipped_offline outcomes are preserved.
- No database migration.

## Checks

- bun test apps/api/src/app/api/automations/evaluate/route.test.ts apps/api/src/app/api/automations/dispatch/[id]/route.test.ts apps/api/src/app/api/automations/run-failed/route.test.ts — 11 passed.
- bunx biome check on changed files — passed.
- bun run lint — passed.
- API typecheck is blocked by the workspace missing next and next/server dependencies.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automation scheduling to prevent stale or duplicate terminal dispatches.
  * Ensured terminal automations are finalized or disabled safely after successful or failed delivery.
  * Preserved correct recurrence timing for one-time and recurring automations.
  * Added safeguards against concurrent state changes and outdated scheduling information.

* **Tests**
  * Expanded coverage for terminal, paused, disabled, recurring, and failed automation scenarios.


abcdlsj:
Addressed in f272fcffc.

The Greptile finding was valid. A terminal occurrence now remains enabled while it is queued, so a user pause before delivery is honored. The dispatch route or QStash failure callback closes the recurrence only after an observable outcome, using a conditional occurrence match. Added coverage for user pause, stale terminal delivery, and terminal delivery failure.

abcdlsj:
Addressed the finalization retry finding in c22b4ad64.

If terminal finalization fails after dispatch has recorded an outcome, the route now returns non-2xx so QStash retries the idempotent finalization. If retries exhaust, run-failed preserves an existing completed/offline run instead of rewriting it as dispatch_failed, then closes the recurrence.

abcdlsj:
Addressed the latest review findings in 93383c05c.

- Replaced the 100-year nextRunAt sentinel with an updatedAt-based terminal reservation token, preserving the real scheduled occurrence.
- Added a conditional terminal claim immediately before dispatch so a concurrent user pause or schedule change prevents delivery.
- Terminal dispatch conflicts now return 409 so QStash can retry/fail the message instead of acknowledging an unresolved run.
- Shared terminal occurrence matching between dispatch and run-failed, with tests asserting the actual CAS predicates and stale/user-paused no-op paths.

Checks: focused automation tests 14/14; Biome and repository lint pass. API typecheck remains blocked by the workspace missing next/next-server type dependencies.

abcdlsj:
Addressed the latest Cubic findings in d9f1d9642:

- Dispatch retries the terminal claim against the exact evaluator reservation if evaluation wins between SELECT and CAS.
- Terminal reservation CAS values use a UTC ISO token with microsecond precision; evaluate, dispatch, and run-failed avoid millisecond-only Date equality.
- Centralized minute bucketing and added coverage for the evaluator race, reserved-but-stale occurrence, and enabled-guard isolation.

Validation: automation route tests 16/16, shared RRULE tests 48/48, shared/trpc typecheck, Biome, and repository lint pass. API typecheck remains blocked by missing workspace Next.js type dependencies.

Issue #5852, opened 2026-07-21
[bug] Chat pane fails on every message: FK violation inserting chat_sessions.v2_workspace_id (local-first workspaces don't exist in cloud v2_workspaces)


marnix-create:
**Version:** Desktop 1.16.0 (also reproduced conceptually against 1.16.1 source, fix not present)
**OS:** macOS (Darwin 25.5.0, arm64)

**What happens**
Typing any message in the workspace chat pane fails immediately with a raw
drizzle error rendered in the UI:

```
Failed query: insert into "chat_sessions" ("id", "organization_id",
"created_by", "workspace_id", "v2_workspace_id", "title", "last_active_at",
"created_at", "updated_at") values ($1, $2, $3, default, $4, default,
default, default, default) on conflict do nothing returning "id"
```

Retrying creates a new session id but always fails the same way.

**Root cause (from source)**
`chat.createSession` in `packages/trpc/src/router/chat/chat.ts` inserts into
the cloud `chat_sessions` table with `v2WorkspaceId: input.v2WorkspaceId`.
That column has an FK to cloud `v2_workspaces`
(`chat_sessions_v2_workspace_id_v2_workspaces_id_fk`).

Post local-first migration, workspaces are authored in host.db only, and the
electric-synced `v2_workspaces` shape for my org is empty (0 rows) while the
desktop is running 4 local v2 workspaces. So the FK can never be satisfied:
the insert throws, and the raw DrizzleQueryError surfaces in the chat pane.
The v1-migration code even documents the situation ("host.db is the
authority post-local-first — cloud rows are stale"), but createSession still
trusts the client-supplied workspace id to exist in the cloud table.

Notably the older REST route (`apps/api/src/app/api/chat/[sessionId]/route.ts`)
has a retry-without-workspaceId fallback for exactly this failure class; the
tRPC `createSession` path has no such fallback.

**Suggested fix**
Either (a) verify the v2 workspace exists cloud-side and fall back to
inserting with `v2_workspace_id = NULL` (mirroring the REST route's
`shouldRetryWithoutWorkspaceId` behavior), or (b) stop FK-ing cloud
chat_sessions to a table that is no longer populated for local-first
workspaces. Also worth catching the error in the ChatPane and showing a
human-readable message instead of the raw SQL.

**Repro**
1. Fresh install post local-first rollout (or v1-migrated workspaces), so
   cloud `v2_workspaces` has no rows for the org.
2. Open any workspace, type anything in the chat pane, send.
[...]

PR #5731, opened 2026-07-17
feat(desktop,host-service): local-first projects — host.db owns projects, workspace cloud sync retired

Kitenite:
**Links**
- Design doc: `plans/20260716-local-first-projects.md` (in this PR, includes the decision log)

## Summary
- **v2 goes fully local.** Projects live in host.db only (`name` + `updatedAt` columns, migration 0010): create, rename, and delete commit locally with zero cloud involvement, so they work fully offline. #5649's silent `my-app-2` duplicates die structurally — nothing ever walks a cloud slug.
- **Workspace cloud dual-write retires** (decision 5 in the doc): `workspace-cloud-sync.ts`, tombstones, re-keying, and the reconciler are deleted; `ensureHostRegistered` was folded away (host registration via `tunnel/connect.ts` is untouched — relay access keeps working).
- **Desktop reads projects from a host fan-out** (`useHostProjects`: per-host `project.list` + `project:changed` live events + IndexedDB last-seen snapshots, workspace-pattern settings). Sidebar, project settings, and the new-workspace modal are migrated off Electric `v2Projects`.

## How It Works
- Host `project.list`/`update` serve full identity rows; `project:changed` broadcasts on the existing event bus (new `workspace-client` payload type). A startup backfill fills `name` from the legacy cloud row when reachable, else the folder basename — never blocks boot.
- `project.setup` gets a local-first fast path: an existing local row completes without any cloud call; the legacy cloud lookup only runs when adopting an old cloud-created project onto a new device. `findByPath` is local-only (a hit = this host already has it; no hit = create fresh).
- `project.remove` flips its kill point: local deletes always run (offline-capable); the legacy cloud delete is best-effort cleanup for the mixed-version window.
- Renames commit through the host everywhere (settings NameSection, sidebar inline rename) — the optimistic cloud rename is gone.

## Manual QA — CDP-verified against the live dev app
Fresh dev host.db with 29 legacy projects; drills driven over CDP:
- [x] Migration 0010 + name backfill on boot: 28 rows filled from legacy cloud names; a transient-error row correctly kept its empty-name sentinel and filled on the **next** boot (retry-next-boot rule proven)
- [x] Create from folder: **171 ms** local commit while fully online; cloud `v2Project.get` for the new id returns NOT_FOUND (no cloud row minted)
[...]

PR #6231, opened 2026-08-06
fix(trpc): retry chat session insert without v2WorkspaceId on FK violation

### LINKED ISSUE #6230: Chat session creation fails for "local" workspace — workspace_id inserted as default, violating NOT NULL
## What happened

Selecting the "local" entry (the root repo, not a worktree workspace) in the desktop app and sending a chat message fails with a database error. The insert into `chat_sessions` only sets `v2_workspace_id` and leaves the legacy `workspace_id` column as `default`, which the schema rejects. Worktree workspaces work fine — only "local" entries fail, and they fail in every repo. This predates 1.19.0; I saw the same on 1.18.x.

## Failed query

```
insert into "chat_sessions" ("id", "organization_id", "created_by", "workspace_id",
"v2_workspace_id", "title", "last_active_at", "created_at", "updated_at")
values ($1, $2, $3, default, $4, default, default, default, default)
on conflict do nothing returning "id"
params: b734d77d-2f06-47e3-bec8-21fab96418d2, f406cca2-1419-490d-ad38-b63c61f9f1a3,
7db378a4-66a1-448a-a86f-8155a8b9a0cf, 25433225-6c6a-4422-b142-c705a7ad807d
```

## Steps to reproduce

1. Open the desktop app, pick any repo
2. Select the "local" entry (root repo) instead of a worktree workspace
3. Type any message and send
4. Error appears; no chat session is created

## Environment

- Superset 1.19.0 (app + CLI)
- macOS (Darwin arm64)

andyst-dev:
Fixes #6230

## Bug

Sending a message in a chat pane for a **"local"** entry (the `type: "main"` workspace, i.e. the repo checkout itself) fails with a database error, in every repo. Worktree workspaces work — only "local" entries fail.

## Root cause

`chat.createSession` (`packages/trpc/src/router/chat/chat.ts`) inserted `chat_sessions.v2_workspace_id` directly from the client-supplied `input.v2WorkspaceId`. That column has an FK to cloud `v2_workspaces` (`chat_sessions_v2_workspace_id_v2_workspaces_id_fk`).

Since local-first (#5731), workspaces are host-owned: the main workspace is minted locally with a fresh UUID on every boot (`runMainWorkspaceSweep` → `ensureMainWorkspace`) and is never mirrored to the cloud `v2_workspaces` table. The FK can therefore never be satisfied for those ids, and the insert throws on **every** message — surfacing the raw query in the chat pane.

The legacy REST bootstrap (`apps/api/src/app/api/chat/[sessionId]/route.ts`) already carries a defensive retry-without-workspaceId fallback for this exact failure class; the tRPC path had none.

## Fix

Extract the session insert into `create-chat-session.ts` and retry without `v2WorkspaceId` when the insert fails with a foreign-key error — mirroring the REST route's `shouldRetryWithoutWorkspaceId` behavior. The retry runs in a **fresh** transaction because a failed INSERT aborts the surrounding Postgres transaction (`current transaction is aborted`), so re-inserting inside the same one would fail regardless.

The resulting row keeps `v2WorkspaceId` NULL for local-first workspaces — the same degraded-but-functional state the REST path produces.

## Tests

Added `packages/trpc/src/router/chat/utils/create-chat-session.test.ts` (6 cases):
- insert with `v2WorkspaceId` succeeds and returns the txid
- FK violation → retries without `v2WorkspaceId`
- generic foreign-key error → retries
- non-FK errors are propagated
- `onConflictDoNothing` conflict → `{ txid: null }`
- FK violation logs exactly once

`bun test`: 6 pass, 0 fail. Package typecheck: clean. Biome check: clean.

Note: #5853 (bot draft, #5852) targets the same FK with an up-front existence check; this PR takes the retry approach already established in the REST route.

---
## Summary by cubic
Retries chat session insert without v2WorkspaceId only when the v2_workspace_id FK rejects it, so messages from the local “main” workspace send. Previously `packages/trpc` wrote the client’s v2WorkspaceId and surfaced FK errors; now it retries in a fresh transaction and stores a null v2WorkspaceId for local sessions, matching the REST route.

- Extracts `createChatSession` in `packages/trpc/src/router/chat/utils/create-chat-session.ts` and uses it from `packages/trpc/src/router/chat/chat.ts`.
- Retries only when the error matches `chat_sessions_v2_workspace_id_v2_workspaces_id_fk` or a v2_workspace_id FK violation; does not retry on unrelated FK or column errors.
- Uses a new transaction for the retry; on conflict, returns `{ txid: null }`.
- Hardens error handling: safe message parsing avoids calling toLowerCase on undefined; retry log emits a fixed reason without raw ids or the DB error.
- Tests cover success, targeted FK retry, no-retry cases, fresh-transaction fallback, conflict handling, and a single sanitized warning.

<sup>Written for commit 668c98891308e55e1b3ce105d9ba5e9dc81518bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6231?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat-session creation reliability when legacy workspace references are unavailable.
  * Added targeted retry handling for recognized workspace-reference constraint errors.
  * Improved handling of session conflicts, transaction identifiers, and unrelated database failures.
  * Ensured retries use a fresh transaction and produce clear warning logs.

* **Tests**
  * Added coverage for successful creation, retry behavior, conflicts, transaction isolation, warning logs, and error propagation.

### DIFF
--- a/packages/trpc/src/router/chat/chat.ts
+++ b/packages/trpc/src/router/chat/chat.ts
@@ -7,6 +7,7 @@ import { and, desc, eq, inArray } from "drizzle-orm";
 import { z } from "zod";
 import { protectedProcedure, userError } from "../../trpc";
 import { requireActiveOrgMembership } from "../utils/active-org";
+import { createChatSession } from "./utils/create-chat-session";
 
 // Re-shaped from the canonical catalog in `@superset/shared/agent-models` so
 // the chat API and the workspace-create model picker never drift.
@@ -73,24 +74,11 @@ export const chatRouter = {
 				});
 			}
 
-			const result = await dbWs.transaction(async (tx) => {
-				const [inserted] = await tx
-					.insert(chatSessions)
-					.values({
-						id: input.sessionId,
-						organizationId,
-						createdBy: ctx.session.user.id,
-						v2WorkspaceId: input.v2WorkspaceId,
-					})
-					.onConflictDoNothing()
-					.returning({ id: chatSessions.id });
-
-				if (!inserted) {
-					return { txid: null };
-				}
-
-				const txid = await getCurrentTxid(tx);
-				return { txid };
+			const result = await createChatSession(dbWs, {
+				id: input.sessionId,
+				organizationId,
+				createdBy: ctx.session.user.id,
+				v2WorkspaceId: input.v2WorkspaceId,
 			});
 
 			return {
--- a/packages/trpc/src/router/chat/chat.ts
+++ b/packages/trpc/src/router/chat/chat.ts
@@ -8,6 +8,7 @@ import { and, desc, eq, inArray } from "drizzle-orm";
 import { z } from "zod";
 import { protectedProcedure } from "../../trpc";
 import { requireActiveOrgMembership } from "../utils/active-org";
+import { createChatSession } from "./utils/create-chat-session";
 import { uploadChatAttachment } from "./utils/upload-chat-attachment";
 
 // Re-shaped from the canonical catalog in `@superset/shared/agent-models` so
@@ -74,24 +75,11 @@ export const chatRouter = {
 				});
 			}
 
-			const result = await dbWs.transaction(async (tx) => {
-				const [inserted] = await tx
-					.insert(chatSessions)
-					.values({
-						id: input.sessionId,
-						organizationId,
-						createdBy: ctx.session.user.id,
-						v2WorkspaceId: input.v2WorkspaceId,
-					})
-					.onConflictDoNothing()
-					.returning({ id: chatSessions.id });
-
-				if (!inserted) {
-					return { txid: null };
-				}
-
-				const txid = await getCurrentTxid(tx);
-				return { txid };
+			const result = await createChatSession(dbWs, {
+				id: input.sessionId,
+				organizationId,
+				createdBy: ctx.session.user.id,
+				v2WorkspaceId: input.v2WorkspaceId,
 			});
 
 			return {
--- a/packages/trpc/src/router/chat/utils/create-chat-session.test.ts
+++ b/packages/trpc/src/router/chat/utils/create-chat-session.test.ts
@@ -142,7 +142,7 @@ describe("createChatSession", () => {
 		expect(result).toEqual({ txid: null });
 	});
 
-	test("logs and retries exactly once on FK violation", async () => {
+	test("logs a fixed reason and retries exactly once on FK violation", async () => {
 		const warnMock = mock();
 		const originalWarn = console.warn;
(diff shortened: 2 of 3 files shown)


andyst-dev:
Thanks for the reviews — all actionable feedback is addressed in `cca05c1`:

- **Narrowed the retry trigger** (cubic): `shouldRetryWithoutV2WorkspaceId` now matches only the specific `chat_sessions_v2_workspace_id_v2_workspaces_id_fk` constraint, or a message containing both `v2_workspace_id` and `foreign key constraint`. The generic `"column"` / `"does not exist"` / `"foreign key"` substrings are gone, so unrelated errors propagate instead of silently retrying.
- **Test now asserts what the retry sends** (coderabbit): the mock records every inserted `values` object — the first insert must equal `BASE_VALUES`, the retry must omit `v2WorkspaceId`.
- **Fresh-transaction fallback exercised** (coderabbit + cubic): each `transaction()` call now yields a NEW transaction mock; a failed insert aborts its own tx, and the tests assert two transaction calls on retry (`[0, 1]`).
- **New negative cases**: unrelated FK constraints and `column "v2_workspace_id" does not exist` errors are asserted to propagate without retry.

`bun test`: 8/8 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Both findings addressed in `06f99e698`:

1. **Functional Correctness (Minor) — non-serializable rejection values.** `errorMessage` now falls back with `return JSON.stringify(error) ?? String(error);` so a non-serializable value (`undefined`, a function) yields a string instead of `undefined`, and `shouldRetryWithoutV2WorkspaceId` can no longer call `.toLowerCase()` on `undefined`.

2. **Security & Privacy (Major) — raw ids / raw DB error in the retry log.** The `console.warn` now logs a fixed reason only (`"[chat] retrying chat session insert without v2WorkspaceId (foreign-key constraint)"`) — no raw `sessionId`/`organizationId`/`v2WorkspaceId` and no raw DB error text, so no user/workspace correlation data leaks into logs.

The updated test now asserts the log stays free of `sessionId` and `organizationId` (alongside the retry-count assertion).

Validation: `bun test create-chat-session.test.ts` -> 8 pass / 0 fail. Biome clean.

PR #6233, opened 2026-08-06
fix(desktop): forward app hotkeys from browser panes via main-process chord index

### LINKED ISSUE #6186: [bug] Browser panes swallow every app shortcut except ⌘W and ⌘R
## What happens

Clicking into a built-in browser pane stops the app's keyboard shortcuts from working. ⌘B and ⌘L no longer toggle the side panels. Clicking back onto the app chrome restores them. Only ⌘W and ⌘R keep working while the pane has focus, so every other registered shortcut is dead until you click away.

## Steps

1. Open a workspace with a browser pane.
2. Click inside the rendered page.
3. Press ⌘B or ⌘L. Nothing happens.
4. Click the left or right panel and press again. Works.

## Why

Browser panes are `WebContentsView`s, so the guest page runs in its own renderer process. Superset registers hotkeys as document listeners in the host renderer, and the guest's keystrokes never reach that document.

Main can intercept these through `before-input-event`, and `setupBeforeInput` does, but only for two keys (`apps/desktop/src/main/lib/browser/browser-manager.ts:247`):

```ts
const key = input.key.toLowerCase();
if (key === "w") { ... }   // close pane
if (key === "r") { ... }   // reload pane
```

Everything else falls through to the page.

## Suggested fix

The renderer already keeps a complete chord index in `registeredAppChords` (`apps/desktop/src/renderer/hotkeys/utils/resolveHotkeyFromEvent.ts:145`), rebuilt whenever overrides or the keyboard layout change. Push that to main, test each `before-input-event` against it instead of hardcoding two keys, and forward matches to the host renderer to dispatch. The terminal already makes this same call in-process at `terminal-key-event-handler.ts:48`, where `resolveHotkeyFromEvent` decides whether a chord belongs to the app or the PTY.

Two things to get right:

- Text editing and find must still reach the page. ⌘C, ⌘V, ⌘X, ⌘A, ⌘Z and ⌘F need to go to the guest or you lose copy and in-page search on web pages. Several collide with app bindings, so the exclusion list is a judgement call rather than something to derive from the registry.
- Match on `input.code`, not `input.key`. The handler currently compares the character produced while the renderer matches physical position. On Dvorak with adaptive layout mapping on they agree by luck. With it off, ⌘W in a browser pane fires on a different physical key from everywhere else in the app.

andyst-dev:
Fixes #6186

## Bug

With a browser pane focused, every app shortcut dies: the pane is a guest `webContents` (out-of-process `<webview>`), so a keydown while the guest has focus never reaches the host document where `react-hotkeys-hook` listens. The only interception path is main's `before-input-event`, and that handler hardcoded exactly two chords (⌘W / ⌘R matched by produced character via `input.key`). ⌘B (`TOGGLE_WORKSPACE_SIDEBAR`), ⌘L (`TOGGLE_SIDEBAR`) and every other renderer-registered chord fell through to the page.

## Fix

Forward the app's full chord index to main and match there by physical position:

- **`shared/browser-pane-chords.ts`** (new, pure — no Electron/DOM/zustand imports): `normalizeToken` / `canonicalizeChord` / `eventToChord` moved from the renderer (re-exported from `resolveHotkeyFromEvent.ts` so existing imports keep working), plus `inputToChord` (Electron `Input` → canonical chord via `input.code`), `matchAppChord` (index lookup with a guest-owned exclusion list), `tokenToCode` / `chordToEventInit` (canonical chord → synthetic DOM `KeyboardEvent` for re-dispatch).
- **`browser-manager.ts`** (main): `setAppChords()` receives the renderer's canonical chord set; `setupBeforeInput` matches on physical key position (`input.code`) instead of `input.key`, drops the `shift`/`alt` and `meta||control` guards (the chord table is the gate now), keeps dedicated `close-pane` / `reload-pane` events for the legacy TabView path (which has no global `CLOSE_PANE` hotkey), and emits `app-hotkey:<paneId>` with the matched chord for everything else.
- **Router** (`browser.ts`): `setAppChords` mutation + `onAppHotkey` subscription.
- **`resolveHotkeyFromEvent.ts`**: pushes `registeredAppChords` keys to main on every rebuild (overrides / layout / adaptive-toggle changes), so the main-side index never drifts.
- **`usePersistentWebview` (v2)**: subscribes to `onAppHotkey` and re-dispatches a synthetic `KeyboardEvent`, letting the normal focus-scoped `useHotkey` registrations resolve it exactly as they do when the host document has focus.
- **Exclusions**: ⌘F / ⌘P / ⌘G stay with the guest (page find / print / find-next are more useful than the pane-scoped app bindings sharing those chords).
- **Docs**: `KEYBOARD_SYSTEM.md` "Known gaps" row.

This also fixes the Dvorak half of the report: the old handler matched whichever physical key *prints* "w" (`input.key`), while the registry binds ⌘W to physical `KeyW` — two different keys on non-QWERTY layouts. Matching on `input.code` anchors both sides to physical position.

## Validation

- `bun test` on `shared/browser-pane-chords.test.ts`: **16/16 pass** (⌘B match, Dvorak ⌘W anchoring, shift/alt chords, guest exclusion, modifier-only nulls, `tokenToCode`/`chordToEventInit` round-trips)
- `tsc --noEmit` on apps/desktop: **0 errors**
- Biome: clean
- Manual repro from the issue (⌘B / ⌘L in a focused browser pane) now dispatches the chord; ⌘W / ⌘R unchanged; ⌘, / ⌘/ still work via menu accelerators.

---
## Summary by cubic
Forward app hotkeys from focused browser panes by mirroring the renderer’s chord index to main and re-dispatching matches on the host document. Adds reliable retry/backoff for chord sync, restores dead shortcuts (⌘B/⌘L), and anchors to physical keys across layouts (e.g., Dvorak ⌘W).

- **Bug Fixes**
  - Main: match chords in `before-input-event` by physical `code` (IME-safe), emit `app-hotkey:<paneId>`, and keep dedicated ⌘W/⌘R close/reload events.
  - Renderer: push the canonical chord index to main on every rebuild with bounded retry backoff and stale-retry cancellation; sends the newest snapshot after in-flight mutations settle.
  - Shared: exclude find/print/find-next and guest editing shortcuts; editing exclusions use the produced key (`input.key`) and ignore shift-gated chords (e.g., ⌘⇧C), so undo on non‑QWERTY stays with the page while COPY_PATH still forwards.
  - Browser panes: subscribe to `onAppHotkey` and dispatch a synthetic keydown on `document`; keep `onClosePane` and catch `close()` errors.

<sup>Written for commit 275af17ecf5f531bd49b81f2c8608ca738aa488e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6233?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Browser-pane keyboard shortcuts now support synchronized application hotkeys.
  * Matching shortcuts are forwarded consistently across panes and keyboard layouts.
  * Application hotkey updates are synchronized automatically, including retry handling.
* **Bug Fixes**
  * Improved handling of modifier combinations, aliases, and ignored key events.
  * Preserved browser-native Find, Print, editing, close, and reload shortcuts.
* **Documentation**
  * Documented browser-pane hotkey forwarding behavior.

### DIFF
--- a/apps/desktop/docs/KEYBOARD_SYSTEM.md
+++ b/apps/desktop/docs/KEYBOARD_SYSTEM.md
@@ -144,6 +144,7 @@ The v1→v2 hotkey storage migration was shipped April 2026 and removed in commi
 
 | Item | Status |
 |---|---|
+| **Browser-pane hotkey forwarding** — guest webContents swallow unclaimed chords (⌘B/⌘L dead, ⌘O shadowed by menu) | **Fixed** in `main/lib/browser/browser-manager.ts`: the renderer pushes its chord index to main and `before-input-event` forwards matches back for re-dispatch; guest find/print (⌘F/⌘P/⌘G) stay with the page. |
 | **Menu accelerator sync** — `main/lib/menu.ts` hardcodes `CmdOrCtrl+R/,//Shift+Q`; they shadow user rebinds | Demand-driven. The single concrete user-visible gap. |
 | **v1 terminal handler** uses catch-all `ctrl/meta` skip → starves TUIs of unbound chords like Ctrl+R | Tracked in `plans/20260409-tui-hotkey-forwarding.md`; v2 already correct. |
 | **AltGr first-class binding token** | Reserved but never wired. AltGr is suppressed at match time, but a user can't *record* `AltGr+E` as their own chord. Drop or implement on demand. |
--- a/apps/desktop/docs/KEYBOARD_SYSTEM.md
+++ b/apps/desktop/docs/KEYBOARD_SYSTEM.md
@@ -139,12 +139,12 @@ The v1→v2 hotkey storage migration was shipped April 2026 and removed in commi
 - **April 2026** — Initial refactor. Unified everything on `event.code` (recorder, dispatch, terminal forwarding). Preserved the bare-string storage shape. See `plans/done/20260412-keyboard-recorder-ctrl-binding-fix.md`.
 - **April 27, 2026** — Layout audit and Phase 0–2 plan. Briefly tried `navigator.keyboard.getLayoutMap()` to avoid the native-keymap dep; switched back after discovering Chromium's `layoutchange` event doesn't fire for macOS input-source switches. native-keymap hooks `kTISNotifySelectedKeyboardInputSourceChanged` directly, which fires reliably. See `plans/done/20260427-keyboard-layout-plan.md`.
 - **April 28, 2026** — Phase 1 (native-keymap) + Phase 2 (dual-mode bindings) shipped. v1 migration removed.
+- **August 2026** — Browser-pane hotkey forwarding. Guest webContents swallow unclaimed chords (⌘B/⌘L dead, ⌘O shadowed by menu); the renderer now pushes its chord index to main and `before-input-event` forwards matches back for re-dispatch on `document`. Guest-owned editing shortcuts (⌘/Ctrl A/C/V/X/Z) and find/print/find-next (⌘F/⌘P/⌘G) stay with the page. See `shared/browser-pane-chords.ts` and `main/lib/browser/browser-manager.ts`.
 
 ## Known gaps / future work
 
 | Item | Status |
 |---|---|
-| **Browser-pane hotkey forwarding** — guest webContents swallow unclaimed chords (⌘B/⌘L dead, ⌘O shadowed by menu) | **Fixed** in `main/lib/browser/browser-manager.ts`: the renderer pushes its chord index to main and `before-input-event` forwards matches back for re-dispatch; guest find/print (⌘F/⌘P/⌘G) stay with the page. |
 | **Menu accelerator sync** — `main/lib/menu.ts` hardcodes `CmdOrCtrl+R/,//Shift+Q`; they shadow user rebinds | Demand-driven. The single concrete user-visible gap. |
 | **v1 terminal handler** uses catch-all `ctrl/meta` skip → starves TUIs of unbound chords like Ctrl+R | Tracked in `plans/20260409-tui-hotkey-forwarding.md`; v2 already correct. |
 | **AltGr first-class binding token** | Reserved but never wired. AltGr is suppressed at match time, but a user can't *record* `AltGr+E` as their own chord. Drop or implement on demand. |
--- a/apps/desktop/src/lib/trpc/routers/browser/browser.ts
+++ b/apps/desktop/src/lib/trpc/routers/browser/browser.ts
@@ -13,6 +13,27 @@ export const createBrowserRouter = () => {
 				return { success: true };
 			}),
 
+		setAppChords: publicProcedure
+			.input(z.object({ chords: z.array(z.string()) }))
+			.mutation(({ input }) => {
+				browserManager.setAppChords(input.chords);
+				return { success: true };
+			}),
+
+		onAppHotkey: publicProcedure
+			.input(z.object({ paneId: z.string() }))
+			.subscription(({ input }) => {
+				return observable<string>((emit) => {
+					const handler = (chord: string) => {
+						emit.next(chord);
+					};
+					browserManager.on(`app-hotkey:${input.paneId}`, handler);
+					return () => {
+						browserManager.off(`app-hotkey:${input.paneId}`, handler);
+					};
+				});
+			}),
+
 		unregister: publicProcedure
 			.input(z.object({ paneId: z.string() }))
 			.mutation(({ input }) => {
--- a/apps/desktop/src/main/lib/browser/browser-manager.ts
+++ b/apps/desktop/src/main/lib/browser/browser-manager.ts
@@ -1,6 +1,7 @@
 import { EventEmitter } from "node:events";
 import { clipboard, Menu, webContents } from "electron";
 import { safeOpenExternal } from "main/lib/safe-url";
+import { inputToChord, matchAppChord } from "shared/browser-pane-chords";
 
 interface ConsoleEntry {
 	level: "log" | "warn" | "error" | "info" | "debug";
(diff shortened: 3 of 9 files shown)


andyst-dev:
Thanks for the reviews — all actionable feedback is addressed in `aa4d2bb`:

- **⌘W no longer swallowed in v2 panes** (greptile P1 + cubic P1): the renderer keeps its `onClosePane` subscription — main emits the dedicated `close-pane` event before app-chord matching, so ⌘W never arrived via `app-hotkey`. `actions.close()` already runs the onBeforeClose chain, matching the renderer `CLOSE_PANE` path.
- **Synthetic keydown dispatched on `document`** (cubic P1): `react-hotkeys-hook` attaches its listeners there, not on `window` — forwarded chords now actually reach the registered hotkeys.
- **Guest-owned editing chords excluded** (coderabbit + cubic): `meta/ctrl + a/c/v/x/z` (select-all, copy, paste, cut, undo) stay with the page, alongside the existing ⌘F/⌘P/⌘G exclusions.
- **Exclusions tested against registered chords** (coderabbit + cubic): ⌘F/⌘P/⌘G are now IN the test chord index, proving exclusion wins over the index.
- **IME composition guard** (cubic P2): `inputToChord` bails on `isComposing`, mirroring the renderer's `eventToChord` guard.
- **`setAppChords` rejection handled** (cubic P3): `.catch` logs a warning; the main-side index degrades gracefully at startup/teardown.
- **Docs** (cubic P3): fix note moved from the "Known gaps" table to "Decision history".

`bun test`: 19/19 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — all round-2 feedback is addressed in `529debc`:

- **Guest editing shortcuts matched by produced key** (cubic P2, non-QWERTY undo): `matchAppChord` now checks `isGuestOwnedShortcut`, which matches the editing family (⌘/Ctrl A/C/V/X/Z) on `input.key` (the character the OS layout produces) rather than physical `code`. AZERTY ⌘Z undo (physical KeyW) no longer collides with the app's `CLOSE_PANE` (meta+w) chord — the produced character wins. Covered by a new test.
- **Close rejection handled** (cubic P2): the `onClosePane` subscription wraps `actions.close()` in try/catch and logs on failure.
- **Chord-index retry** (coderabbit): `setAppChords` rejection now schedules a single delayed retry (1s) that pushes the *latest* chords, so a transiently-unavailable IPC channel at startup doesn't leave the main-side index stale for the session; subsequent rebuilds still re-push as before.
- **Ctrl test actually exercises the ctrl path** (coderabbit): the ctrl-variant exclusion test now sets `input.control` (the field `inputToChord` reads) instead of `input.ctrl`, which previously never hit the ctrl branch.

`bun test`: 20/20 pass (incl. AZERTY undo) · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — all three points are addressed in `cafeeb2`:

- **⌘⇧C COPY_PATH no longer swallowed** (cubic P2): `isGuestEditingShortcut` now returns false when `shift` is held — shift-gated chords are app chords (the guest's copy/paste/undo are plain ⌘/Ctrl+letter). Regression test: `meta+shift+c` matches the registered chord instead of being excluded (21/21 pass).
- **paneId in the close warning** (cubic P3): `console.warn("Failed to close browser pane (${paneId})", error)`.
- **Retry backoff** (cubic P2): the single delayed retry is intentionally bounded — a subsequent `rebuild()` (override/layout/preference change) also re-pushes, and the IPC channel comes up within a second of renderer boot; a long backoff loop would add noise for a case that resolves on the next user interaction. Documented in the comment; happy to switch to bounded backoff if you'd prefer it.

`tsc --noEmit`: 0 errors · `bun test`: 21/21 pass · Biome: clean.

andyst-dev:
Thanks - fixed in `275af17`.

The renderer-to-main chord sync now uses a dedicated latest-state synchronizer:

- retries with capped exponential backoff until IPC accepts the chord index, so startup races cannot leave browser-pane hotkeys disabled for the session;
- keeps only one retry timer active;
- a chord rebuild cancels stale delayed work and sends the newest complete index immediately;
- if state changes during an in-flight mutation, the newest snapshot is sent as soon as that mutation settles.

Added deterministic scheduler tests for transient recovery, stale-retry cancellation, in-flight replacement, and capped retry behavior.

Validation:

- focused Bun tests -> **25 passed**
- Desktop typecheck -> passed
- Biome on the 3 changed files -> clean
- `git diff --check` -> clean

andyst-dev:
Heads-up: this PR is now **superseded by a refactor on main**. Since it was opened, main introduced `shared/hotkey-chord.ts` (the chord-normalization moved out of `browser-pane-chords.ts`) and a main-process webview key-forwarder (`ForwardedKey` in `browser-manager.ts`) — which is exactly what this PR set out to build. The branch no longer rebases cleanly onto main and its files have been renamed/absorbed.

Per the repo's contribution guidance, the same end-state is already on main, so this PR is obsolete. Happy to close it, or I can leave it open if a maintainer wants to cherry-pick any surviving intent.

andyst-dev:
Closing this as a duplicate: the same feature (forwarding app hotkeys from
in-workspace browser panes via the main-process chord index) was already
merged upstream in **#5494** ("fix(desktop): dismiss stuck tooltips and forward
tab-switch hotkeys in in-workspace browser", merged 2026-08-09).

`main` now carries the full implementation under `shared/hotkey-chord` +
`BrowserManager.setForwardableChords` (tRPC `browser.setForwardableChords`),
with the renderer wiring in `usePersistentWebview.ts`. The maintainers chose a
scoped design (only tab-switch hotkeys are forwardable, via `FORWARDABLE_HOTKEYS`)
that supersedes the broader "forward every app chord" approach this PR took, so
keeping both would conflict with that decision.

Thanks for considering. I'll keep contributing on the superset desktop area.

andyst-dev:
Superseded by #5494 (merged). See comment above.

Kitenite:
Thanks for the contribution @andyst-dev !

Issue #6127, opened 2026-08-02
Superset 1.18.1 terminals cannot resolve UID or access trustd (no access to github through superset terminal)


smesh777:
### Where did this happen?

Desktop app (macOS)

### Version

1.18.1 (Apple Silicon)

### What happened?

After updating to Superset Desktop 1.18.1 (PTY daemon 0.2.7), built-in terminals run in a macOS process context that cannot resolve the current user or access certificate trust services.

Inside a Superset terminal:

- `id -un` returns `501` instead of the macOS username.
- A fresh `gh auth login -h github.com` fails before displaying a device code:

  ```text
  tls: failed to verify certificate: x509: OSStatus -26276
  ```

- SSH also reports:

  ```text
  No user exists for uid 501
  ```

The same commands work in Apple Terminal: `id -un` returns `<macOS username>`, GitHub authentication succeeds, and `git push` succeeds.

This is related to #4030, but its resolution does not apply. There is no custom `gh` wrapper or injected `GH_TOKEN`/`GITHUB_TOKEN` here:

- `gh` resolves directly to `/opt/homebrew/bin/gh`.
- `/opt/homebrew/bin/gh` is the arm64 Mach-O GitHub CLI executable.
- No alternate `gh` wrapper exists in `~/.superset/bin`, `~/bin`, or `~/.local/bin`.
- Neither `GH_TOKEN` nor `GITHUB_TOKEN` is present in the affected environment.
- The failure is not limited to reading an existing keyring credential: a fresh `gh auth login` fails at the TLS request for a device code.

The problem persists after:

- Reinstalling the official Superset 1.18.1 arm64 application.
- Authenticating the standalone Superset CLI.
- Running `superset stop` and reopening the desktop app.
- Running `superset stop`, then `superset start --daemon` from Apple Terminal, then reopening the desktop app.
- Opening a brand-new built-in terminal after restart.

Environment:

- Superset Desktop: 1.18.1 (Apple Silicon)
- PTY daemon: 0.2.7
- macOS: 15.1 (24B2083)
- Architecture: arm64
- GitHub CLI: 2.66.1

Expected behavior: Superset terminals should resolve the normal macOS user and allow `gh` to access GitHub normally.

Potentially relevant upstream history:

- `dc7e1b70b` — `fix(host-service): heal trustd-degraded pty-daemons so gh works in terminals`
[...]

PR #6235, opened 2026-08-06
fix(cli): stop pty-daemon alongside host service

andyst-dev:
Fixes the `superset stop` half of #6127

## Problem

`superset stop` only kills the host-service pid from `host/{org}/manifest.json`. The pty-daemon is spawned **detached** (`DaemonSupervisor.spawn()`) so PTY sessions survive host-service restarts — which also means it outlives `superset stop`:

- it keeps listening on its socket,
- the next `host-service` **re-adopts** it unconditionally (socket connectable + pid alive are the only health signals), possibly in a **degraded process context** — the exact #6127 failure mode where a daemon that lost its per-user XPC namespace (`id -un` → `501`, `gh` fails with `OSStatus -26276`) is silently adopted again,
- so every "have you tried restarting?" instruction was misleading: `superset stop` → `superset start` from Apple Terminal did **not** fix the degraded daemon (confirmed in the issue thread).

## Fix

`packages/cli/src/commands/stop/command.ts` now also stops the pty-daemon:

- reads `host/{org}/pty-daemon-manifest.json` (same directory the host-service writes; new `lib/host/pty-daemon-manifest.ts` helper mirroring `packages/host-service/src/daemon/manifest.ts`, which the CLI already mirrors for the host-service manifest),
- SIGTERMs the daemon after the host service is down, SIGKILLs it after a 5s grace period if still alive,
- message now reads "Stopped host service and pty-daemon for …".

The daemon's own `SIGTERM` handler closes its socket and cleans up, so a subsequent `superset start` spawns a fresh daemon instead of re-adopting a possibly-degraded one.

## Validation

- `bun test src/lib/host/pty-daemon-manifest.test.ts`: **4/4 pass** (absent → null, write/read round-trip, remove, corrupt JSON → null)
- `tsc --noEmit` on packages/cli: **0 errors**
- Biome: clean

Note: this is the independent papercut from the bot's triage (point 5); the deeper health-probe / adopt-refusal work (points 1–4) is a separate, larger change.

---
## Summary by cubic
Make `superset stop` terminate the detached pty-daemon alongside the host service to prevent re-adopting a degraded daemon. Previously it only killed the host-service pid; the daemon survived and was re-adopted.

- Read both manifests up front; stop the daemon even when the host isn’t running; continue daemon cleanup if host shutdown fails.
- Stop semantics: SIGTERM then SIGKILL (10s host, 5s daemon) with post-kill polling; ignore `ESRCH`; fail if either process survives.
- Remove manifests only after a confirmed exit; keep manifests when `stopProcess` throws or a process survives.
- Add `packages/cli/src/lib/host/pty-daemon-manifest.ts` with strict validation (positive pid) and 6 unit tests.
- Output lists each stopped process with its pid and returns `{ running: false }`.

<sup>Written for commit e5dfb343a74553dc727bbf68f43c6355f2106b7e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Stopping the host service now also terminates its detached PTY daemon.
  * The CLI waits for graceful shutdown before forcefully terminating processes when needed.
  * Stop confirmations identify stopped services and their process IDs.

* **Bug Fixes**
  * Reports errors when processes cannot be terminated successfully.
  * Improves handling of missing, corrupted, or invalid service state.
  * Retains state for processes that remain active and removes it after successful termination.
  * Correctly handles processes that have already exited during shutdown.

### DIFF
--- a/packages/cli/src/commands/stop/command.ts
+++ b/packages/cli/src/commands/stop/command.ts
@@ -73,16 +73,10 @@ export default command({
 		const failures: string[] = [];
 		const stopped: Array<{ label: string; pid: number }> = [];
 
-		// Only remove each manifest once the process is CONFIRMED dead: a
-		// survivor must stay discoverable by `superset start`/`stop` rather
-		// than being orphaned without a manifest. A throw from stopProcess
-		// (e.g. SIGTERM refused) is recorded and daemon cleanup still runs.
+		// Remove a manifest only when stopProcess returns null, confirming exit.
+		// A thrown stop leaves process state unknown, so retain the manifest,
+		// record the failure, and continue cleaning up the other daemon.
 		if (manifest && isProcessAlive(manifest.pid)) {
-			// `survived` is null only on a CONFIRMED clean stop. A throw
-			// from stopProcess (e.g. SIGTERM/SIGKILL refused with EPERM)
-			// leaves the process state unknown — it must NOT collapse to
-			// the clean-stop sentinel, or the manifest would be removed
-			// while the process is still running (cubic P2).
 			let survived: number | null;
 			let stopError: Error | null = null;
 			try {
--- a/packages/cli/src/commands/stop/command.ts
+++ b/packages/cli/src/commands/stop/command.ts
@@ -75,31 +75,48 @@ export default command({
 
 		// Only remove each manifest once the process is CONFIRMED dead: a
 		// survivor must stay discoverable by `superset start`/`stop` rather
-		// than being orphaned without a manifest.
+		// than being orphaned without a manifest. A throw from stopProcess
+		// (e.g. SIGTERM refused) is recorded and daemon cleanup still runs.
 		if (manifest && isProcessAlive(manifest.pid)) {
-			const survived = await stopProcess(manifest.pid, "host service", 10_000);
+			let survived: number | null;
+			try {
+				survived = await stopProcess(manifest.pid, "host service", 10_000);
+			} catch (error) {
+				failures.push(
+					`host service (pid ${manifest.pid}): ${
+						error instanceof Error ? error.message : String(error)
+					}`,
+				);
+				survived = null;
+			}
 			if (survived !== null) {
 				failures.push(`host service (pid ${survived}) survived SIGKILL`);
 			} else {
 				stopped.push({ label: "host service", pid: manifest.pid });
 			}
-			removeManifest(organization.id);
+			if (survived === null) removeManifest(organization.id);
 		} else {
 			removeManifest(organization.id);
 		}
 
 		if (daemonManifest && isProcessAlive(daemonManifest.pid)) {
-			const survived = await stopProcess(
-				daemonManifest.pid,
-				"pty-daemon",
-				5_000,
-			);
+			let survived: number | null;
+			try {
+				survived = await stopProcess(daemonManifest.pid, "pty-daemon", 5_000);
+			} catch (error) {
+				failures.push(
+					`pty-daemon (pid ${daemonManifest.pid}): ${
+						error instanceof Error ? error.message : String(error)
+					}`,
+				);
+				survived = null;
+			}
 			if (survived !== null) {
 				failures.push(`pty-daemon (pid ${survived}) survived SIGKILL`);
 			} else {
 				stopped.push({ label: "pty-daemon", pid: daemonManifest.pid });
 			}
-			removePtyDaemonManifest(organization.id);
+			if (survived === null) removePtyDaemonManifest(organization.id);
 		} else {
 			removePtyDaemonManifest(organization.id);
 		}
--- a/packages/cli/src/commands/stop/command.ts
+++ b/packages/cli/src/commands/stop/command.ts
@@ -78,45 +78,56 @@ export default command({
 		// than being orphaned without a manifest. A throw from stopProcess
 		// (e.g. SIGTERM refused) is recorded and daemon cleanup still runs.
 		if (manifest && isProcessAlive(manifest.pid)) {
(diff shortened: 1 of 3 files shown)


andyst-dev:
Thanks — all round-1 feedback is addressed in `ba20a05`:

- **Daemon shutdown before the no-host early return** (greptile P1, coderabbit Major, cubic P1): both manifests are read up front; the pty-daemon is stopped even when the host-service manifest is absent, so an orphaned detached daemon can't survive `superset stop` for re-adoption (#6127). The no-host message is preserved when nothing was running.
- **Stale-PID / daemon manifest removed** (greptile P1, cubic P2): `removePtyDaemonManifest()` now runs after stopping the daemon (the host manifest removal already existed); the daemon's own manifest is the identity record, and the manifest is removed in the same flow.
- **SIGKILL failures surfaced** (coderabbit, cubic P2): only `ESRCH` (already gone) is ignored; a daemon surviving SIGKILL fails the command instead of reporting success.
- **JSON payload preserved** (cubic P2): the success payload keeps the per-process message (pid included) instead of a fields-only data shape.
- **Manifest validation + atomic write** (coderabbit): `readPtyDaemonManifest` validates field shapes (mirroring host-service) and `writePtyDaemonManifest` writes tmp + rename so concurrent readers never see truncated JSON. Both covered by new tests.
- **Env restore in tests** (cubic P3): `SUPERSET_HOME_DIR` is restored in afterAll.

`bun test`: 6/6 pass (incl. invalid-fields → null and no-tmp-left-behind) · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — all round-2 feedback is addressed in `d25fc62`:

- **PID validation (P1, safety)** (cubic): `readPtyDaemonManifest` now requires `pid` to be a positive safe integer — a negative pid (e.g. `-1`) would make `process.kill` target a process group, and a stale reused pid could kill an unrelated process. Tests cover `-1` and `0` → null.
- **Poll after SIGKILL** (coderabbit Minor + cubic P3): replaced the single fixed 200ms wait with a 2s poll, so a process in uninterruptible sleep is reaped before we report survival.
- **Manifests removed only on confirmed termination** (coderabbit Major): each manifest (host + daemon) is removed only after its process is confirmed dead — a survivor stays discoverable by `superset start`/`stop` instead of being orphaned without a manifest.
- **Host-service SIGKILL survival surfaced** (cubic P1): the host `stopProcess` return value is captured and reported as a failure (while daemon cleanup still runs), instead of `superset stop` succeeding with the host still up.
- **Dead write helper dropped** (cubic P2): `writePtyDaemonManifest` was never called in production — the detached daemon writes its own manifest, the CLI only reads/removes it. Removed, and the tests now write the fixture directly.

`bun test`: 6/6 pass (incl. non-positive pid → null) · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — both points are addressed in `210c497`:

- **Daemon cleanup continues after a host shutdown error** (coderabbit Major): a throw from `stopProcess` on the host service is now caught, recorded in `failures`, and the pty-daemon branch still runs — a failed host stop can't leave the daemon running and re-adoptable.
- **Manifests retained when the process survives** (coderabbit Major): both `removeManifest` and `removePtyDaemonManifest` now run only when the process is confirmed dead (`survived === null`). A survivor keeps its manifest so `superset start`/`stop` can still discover it, instead of being orphaned.

`bun test`: 6/6 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — addressed in `a21c939`:

- **Error ≠ clean stop** (cubic P2): a throw from `stopProcess` (e.g. EPERM on SIGTERM/SIGKILL) previously collapsed into `survived = null` — the same sentinel as a confirmed clean stop — so the manifest was removed while the process could still be running. A distinct `stopError` flag now carries the failure: manifests are removed only when there was **no error AND the process is confirmed dead**, and the error is surfaced in the aggregate failure message while daemon cleanup still runs.

For the record, two older comments on this thread refer to code already fixed in earlier rounds:

- **Manifests retained when the process survives** (coderabbit @ command.ts, commit `ba20a05`): since `d25fc62` both `removeManifest` and `removePtyDaemonManifest` run only when `survived === null`; this round extends that to the error path.
- **PID validation** (cubic P1 @ pty-daemon-manifest.ts, commit `ba20a05`): `readPtyDaemonManifest` has required `pid > 0 && Number.isSafeInteger(pid)` since `d25fc62` — `-1` returns null (covered by tests).

`bun test`: 6/6 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks - cleaned up in `30c6a04`.

The stop-state comment now states the invariant directly: only a `null` result from `stopProcess` confirms exit and permits manifest removal; a thrown stop leaves state unknown, so the manifest is retained while cleanup continues. Removed the contradictory wording and internal review reference. No runtime behavior changed.

Validation:

- `biome check packages/cli/src/commands/stop/command.ts` -> clean
- `bun run --cwd packages/cli typecheck` -> passed
- `bun test packages/cli/src/lib/host/pty-daemon-manifest.test.ts` -> **6 passed**
- `git diff --check` -> clean

PR #6236, opened 2026-08-06
fix(host-service): force tunnel reconnect on inbound silence without waiting for onclose

### LINKED ISSUE #6229: Remote Linux host stays alive but becomes offline after relay WebSocket stalls
### Summary

A persistent Superset Linux host repeatedly becomes offline in the Superset UI/control plane while the VPS, network, host-service process, and terminal processes remain running.

Superset detects the stale relay connection and logs:

```text
[host-service:tunnel] no inbound traffic for 77379ms, forcing reconnect
```

However, the reconnect path does not restore the host. `superset-host.service` remains `active (running)`, so systemd cannot detect the failure. Restarting only that service immediately reconnects the host.

This reproduced at least five times during one day of normal use.

### Environment

- Superset CLI and host service: `1.19.0`
- Update check: `1.19.0` is the latest available release
- Remote OS: Ubuntu 24.04.4 LTS, x86-64
- Kernel: Linux `6.8.0-136-generic`
- Service manager: systemd user service
- User lingering: enabled
- Host type: always-on Hetzner VPS
- Network: public internet plus Tailscale
- Superset started with:

```ini
[Service]
Type=simple
Environment=PATH=%h/.local/bin:/usr/local/bin:/usr/bin:/bin
ExecStart=%h/.local/bin/superset start --org private
Restart=always
RestartSec=5
```

### Steps to reproduce

1. Authenticate Superset on an Ubuntu VPS.
2. Run `superset start --org private` as a persistent systemd user service.
3. Confirm that the remote host appears online and use terminals normally.
4. Leave the service running.
5. After a variable period—sometimes around 20 minutes, sometimes only a few minutes—the host disappears or becomes offline in Superset.
6. Check the VPS through an independent SSH or Tailscale connection.

### Actual behavior

During the failure:

- The VPS remains online.
- SSH and Tailscale remain responsive.
- CPU, memory, and disk capacity remain healthy.
- `superset-host.service` remains `active (running)`.
- systemd reports `NRestarts=0`.
- `superset status` reports the host process as running.
- The external Superset control plane reports the host as offline.
- Existing Superset terminals become inaccessible.

Examples from separate occurrences:

```text
[host-service:tunnel] no inbound traffic for 77379ms, forcing reconnect
[host-service:tunnel] no inbound traffic for 78131ms, forcing reconnect
[host-service:tunnel] no inbound traffic for 80725ms, forcing reconnect
[host-service:tunnel] no inbound traffic for 76045ms, forcing reconnect
[host-service:tunnel] no inbound traffic for 78505ms, forcing reconnect
```

The client source appears to close the WebSocket after 75 seconds without an inbound application message:

```js
socket.onmessage = (event) => {
  this.lastInboundAt = Date.now();
  this.handleMessage(event.data);
};

if (Date.now() - this.lastInboundAt > 75000) {
  console.warn(
    `[host-service:tunnel] no inbound traffic for ${silentFor}ms, forcing reconnect`
  );
  this.socket.close(4002, "Inbound silence timeout");
}
```

The relay protocol appears to expect inbound `ping` messages, to which the host replies with `pong`.

On one occurrence, the relay TCP connection was still `ESTABLISHED`, but its send queue had grown to approximately 39 KB:

```text
ESTAB 0 39419 [VPS IPv6]:51768 [relay IPv6]:443
```

This suggests that the WebSocket or underlying IPv6 TCP path had stopped making progress, but that has not yet been proven as the root cause.

### Recovery

Running:

```bash
systemctl --user restart superset-host.service
```

immediately produces:

```text
Connected to relay — machine is now accessible.
[host-service:tunnel] connected to relay for host ...
```

The host then changes back to `online: yes` in the external control-plane listing. This recovery worked on every occurrence.

Superset also respawns its recorded terminal sessions after the restart:

```text
[terminal] respawning lost session ...
```

### Expected behavior

When relay traffic becomes stale, Superset should do one of the following:

1. Successfully establish and register a replacement WebSocket connection.
2. Detect that reconnection failed and continue retrying with bounded backoff.
3. Exit with a non-zero status so systemd or another supervisor can restart the process.

The process should not remain alive while the control plane considers the host offline.

The relay should also send application-level keepalive messages frequently enough that a healthy but idle host does not trigger the 75-second inbound-silence timeout.

### Additional observations

- General outbound connectivity continues working during the outage.
- Tailscale and SSH are unaffected.
- TCP keepalive settings are unlikely to help because Superset’s timeout is based on application-level inbound WebSocket messages.
- Missing GitHub CLI authentication previously produced separate PR metadata errors, but those errors did not terminate the process and were not the cause of these relay failures.
- A temporary watchdog that restarts the service after the stale-tunnel log event reduces downtime, but it does not fix the relay reconnect bug.
- The host’s relay connection currently resolves over IPv6. A stalled IPv6 path is one possible contributor, but this is not confirmed.

### Requested diagnostics/fix

Could the Superset team inspect:

1. Whether the relay is sending `ping` messages at the expected interval.
2. Why closing the socket with code `4002` does not reliably lead to a registered replacement connection.
3. Whether the old socket remains referenced or blocks `connect()` through its `readyState`.
4. Whether reconnect success should be verified against the control-plane registration state.
5. Whether the host process should exit after repeated failed relay reconnections.

A bounded reconnect timeout followed by process exit would at least allow the existing `Restart=always` policy to recover the host automatically.

andyst-dev:
Fixes #6229

## Bug

A persistent Linux host repeatedly goes offline in the Superset control plane while the VPS, network, host-service process, and terminals all stay running. The host logs `no inbound traffic for 77s, forcing reconnect` — and then never reconnects until `systemctl --user restart superset-host`.

Root cause (confirmed by the triage bot and a contributor in the issue thread):

`tunnel-client.ts` detects relay silence via a watchdog, but its **only recovery mechanism was a graceful `WebSocket.close(4002, …)`** — and every reconnect after an established connection was scheduled **exclusively from the `onclose` handler**. On a blackholed TCP path (packets dropped rather than reset — the stuck `ESTAB` with a ~39 KB unacked send queue seen in the report), the close frame can never be flushed, the closing handshake never completes, `onclose` never fires, and `scheduleReconnect()` is never called. The client parks in `CLOSING` forever.

## Fix

`forceReconnect(socket, code, reason)` in `tunnel-client.ts`:

- orphans the stale socket **immediately** — `this.socket = null`, `connecting = false`, `stopWatchdog()`, `cleanupChannels()` — and calls `scheduleReconnect()` directly, **without waiting for `onclose`**;
- mirrors the existing connect-timeout path (`connect()`'s deadline already did close + null + scheduleReconnect);
- the watchdog now calls it instead of the bare `socket.close()`;
- the identity guard at the top of `onclose` (`if (this.socket !== socket) return`) means the stale socket's late close (if it ever arrives) is ignored — no double-scheduling, no teardown of a newer socket.

## Validation

- `bun test` on `tunnel-client.test.ts`: **3/3 pass** using a fake `WebSocket` whose `close()` **never fires `onclose`** (the #6229 failure mode) — forceReconnect schedules the reconnect anyway; stale-socket and post-close no-op cases covered.
- `tsc --noEmit` on packages/host-service: **0 errors**
- Biome: clean

Note: `tunnel-client-v2.ts` may share the pattern; this PR scopes to the v1 client the report and triage identified.

---
## Summary by cubic
Force reconnects the tunnel on inbound silence and ignores events from orphaned sockets. Previously the watchdog waited for onclose after close(4002); on blackholed TCP, onclose never fired and the client stayed in CLOSING.

- Adds forceReconnect() to orphan the current socket, stop the watchdog, clean channels, and scheduleReconnect() immediately; the watchdog now calls this on inbound silence.
- Guards onopen, onmessage, onerror, and onclose by socket identity so late events from orphaned sockets don’t touch state or double-schedule.
- Adds tests that drive the real connect() path with a global FakeWebSocket, stub scheduleReconnect to avoid timer leaks, verify reconnect without onclose, ignore stale sockets and late messages, and close the client to suppress the pending connect deadline.

<sup>Written for commit 1b873b0fc7974531a4e707c14f0736a07585f733. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic reconnection when a WebSocket becomes unresponsive.
  * Replaced stale connections promptly, even when the close handshake does not complete.
  * Prevented outdated connections from interfering with active sessions or reconnection attempts.
  * Ensured cleanup is skipped after the client has been closed.

* **Tests**
  * Added regression coverage for unresponsive WebSockets, stale callbacks, reconnect scheduling, and late messages from orphaned connections.

### DIFF
--- a/packages/host-service/src/tunnel/tunnel-client.test.ts
+++ b/packages/host-service/src/tunnel/tunnel-client.test.ts
@@ -1,0 +1,119 @@
+import { afterEach, describe, expect, test } from "bun:test";
+
+// Minimal fake WebSocket implementing only what TunnelClient touches.
+class FakeWebSocket {
+	static OPEN = 1;
+	static CLOSING = 2;
+	static CLOSED = 3;
+
+	readyState = FakeWebSocket.OPEN;
+	onopen: (() => void) | null = null;
+	onmessage: ((event: { data: unknown }) => void) | null = null;
+	onclose: ((event: { code?: number; reason?: string }) => void) | null = null;
+	onerror: ((event: unknown) => void) | null = null;
+	closeCalls: Array<{ code: number; reason: string }> = [];
+
+	close(code?: number, reason?: string): void {
+		this.closeCalls.push({ code: code ?? 1000, reason: reason ?? "" });
+		// A blackholed TCP path: the closing handshake never completes, so
+		// onclose is NEVER called — the exact #6229 failure.
+	}
+}
+
+// Stub global WebSocket with the fake.
+const originalWebSocket = globalThis.WebSocket;
+(globalThis as Record<string, unknown>).WebSocket = FakeWebSocket;
+
+const { TunnelClient } = await import("./tunnel-client");
+
+// Speed up the watchdog for tests: reach into the module via the constants
+// is not possible (not exported), so we drive the timer manually by
+// stubbing timers.
+function makeClient() {
+	const client = new TunnelClient({
+		relayUrl: "https://relay.example.com",
+		hostId: "host-1",
+		getAuthToken: async () => "token",
+		localPort: 4000,
+		hostServiceSecret: "secret",
+	});
+	return client as unknown as {
+		connect: () => Promise<void>;
+		socket: FakeWebSocket | null;
+		reconnectAttempts: number;
+		closed: boolean;
+		lastInboundAt: number;
+		startWatchdog: () => void;
+		stopWatchdog: () => void;
+		forceReconnect: (s: FakeWebSocket, code: number, reason: string) => void;
+		watchdogTimer: ReturnType<typeof setInterval> | null;
+		dispose: () => void;
+	};
+}
+
+describe("TunnelClient watchdog (#6229)", () => {
+	afterEach(() => {
+		(globalThis as Record<string, unknown>).WebSocket = originalWebSocket;
+	});
+
+	test("forceReconnect schedules a reconnect without waiting for onclose", () => {
+		const client = makeClient();
+		const sock = new FakeWebSocket();
+		// Simulate an established socket
+		(client as unknown as { socket: FakeWebSocket }).socket = sock;
+		client.reconnectAttempts = 0;
+
+		// Patch scheduleReconnect to observe the call.
+		let scheduled = 0;
+		(client as unknown as { scheduleReconnect: () => void }).scheduleReconnect =
+			() => {
+				scheduled++;
+			};
+
+		client.forceReconnect(sock, 4002, "Inbound silence timeout");
+
+		expect(sock.closeCalls).toEqual([
+			{ code: 4002, reason: "Inbound silence timeout" },
+		]);
+		// The stale socket is orphaned immediately — onclose (which never
+		// fires on a blackholed path) is not the trigger.
+		expect(client.socket).toBeNull();
+		expect(scheduled).toBe(1);
+	});
+
+	test("forceReconnect ignores a stale socket that is no longer current", () => {
+		const client = makeClient();
+		const stale = new FakeWebSocket();
+		(client as unknown as { socket: FakeWebSocket }).socket = stale;
+		let scheduled = 0;
+		(client as unknown as { scheduleReconnect: () => void }).scheduleReconnect =
+			() => {
+				scheduled++;
+			};
+
+		// A newer socket replaced the stale one before the watchdog fired.
+		const current = new FakeWebSocket();
+		(client as unknown as { socket: FakeWebSocket }).socket = current;
+		client.forceReconnect(stale, 4002, "Inbound silence timeout");
+
+		// Stale socket must not be torn down or scheduled.
+		expect(client.socket).toBe(current);
+		expect(scheduled).toBe(0);
+		expect(stale.closeCalls).toEqual([]);
+	});
+
+	test("forceReconnect is a no-op after close()", () => {
+		const client = makeClient();
+		const sock = new FakeWebSocket();
+		(client as unknown as { socket: FakeWebSocket }).socket = sock;
+		let scheduled = 0;
+		(client as unknown as { scheduleReconnect: () => void }).scheduleReconnect =
+			() => {
+				scheduled++;
+			};
+		(client as unknown as { closed: boolean }).closed = true;
+
+		client.forceReconnect(sock, 4002, "Inbound silence timeout");
+		expect(scheduled).toBe(0);
+	});
+});
--- a/packages/host-service/src/tunnel/tunnel-client.test.ts
+++ b/packages/host-service/src/tunnel/tunnel-client.test.ts
@@ -3,10 +3,11 @@ import { afterEach, describe, expect, test } from "bun:test";
 // Minimal fake WebSocket implementing only what TunnelClient touches.
 class FakeWebSocket {
 	static OPEN = 1;
+	static CONNECTING = 0;
 	static CLOSING = 2;
 	static CLOSED = 3;
 
-	readyState = FakeWebSocket.OPEN;
+	readyState = FakeWebSocket.CONNECTING;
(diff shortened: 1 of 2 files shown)


andyst-dev:
Thanks — all round-2 feedback is addressed in `81c3798`:

- **Ignore events from detached sockets** (coderabbit): `onopen`, `onmessage`, and `onerror` now all carry the `if (this.socket !== socket) return` identity guard, mirroring the existing `onclose` guard — a late frame from an orphaned socket can no longer touch `lastInboundAt`/watchdog state or dispatch through `handleMessage`. New regression test: late messages from an orphaned socket are ignored (4/4 pass).
- **Post-close test via real `close()`** (coderabbit): the no-op test now calls the actual `close()` method instead of setting `closed` by hand.
- **No unnecessary global stub** (greptile + cubic): the tests instantiate `FakeWebSocket` directly and never read `globalThis.WebSocket`, so the process-wide replacement at module scope is gone.
- **Misleading timer comment removed** (cubic): tests exercise `forceReconnect`/socket identity directly, not the 10s watchdog interval — the comment no longer claims otherwise.

`tsc --noEmit`: 0 errors · `bun test`: 4/4 pass · Biome: clean.

andyst-dev:
Thanks — both points are addressed in `6806919`:

- **Test no longer a no-op** (cubic P2): the late-messages test previously assigned socket state by hand, so `onmessage` stayed null and the assertion passed regardless of the fix. The fake `WebSocket` is now registered *before* the module import (TunnelClient resolves `WebSocket` at module-evaluation time) and the test drives the real `connect()` path, so it exercises the actual guarded handlers. Verified: a late inbound from an orphaned socket leaves `lastInboundAt` untouched.
- **No leaked reconnect timer** (cubic P2): `scheduleReconnect` is stubbed in that test so `forceReconnect`'s scheduling is observed without a real timer firing against `relay.example.com` after the test; the global stub is restored in `afterEach`.

`bun test`: 4/4 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — addressed in `3a60b61`: the orphaned-socket test now calls `client.close()` after the assertions, so the 20s deadline `connect()` left pending on the never-opening fake socket becomes a no-op instead of firing after the test (cubic P2).

`bun test`: 4/4 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — this comment is stale: it points at commit `6806919`/`81c3798` and was re-anchored onto the current head, but the concern was already addressed in `3a60b61` (pushed before this comment's review run):

- The orphaned-socket test calls `client.close()` after its assertions, which marks the client closed so `connect()`'s pending 20s deadline becomes a no-op and no real timer fires against `relay.example.com`. `scheduleReconnect` is also stubbed in that test (observed via a counter, not executed).

`bun test`: 4/4 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Closing as a duplicate: main already ships this via `un-wedgeable tunnel supervision + SDK relay resolution (#6724)` — the watchdog on `tunnel-client.ts` already forces a socket teardown on inbound silence (`abandonSocket("Inbound silence timeout")`) without waiting for the close handshake. That commit is a stricter superset of this PR's `forceReconnect` intent (stuck-socket grace, v1_dead_reconnect_chain rescue, etc.), so rebasing would be redundant against the maintainers' newer design. No action needed.

PR #6237, opened 2026-08-06
fix(host-service): resolve Against-base diff to local branch when origin/<branch> missing

### LINKED ISSUE #5298: [bug] `Changes` tab is empty when a workspace's base branch is a local-only, unpushed branch
### Summary
  
For a stacked workspace whose base branch is another **local branch that hasn't been
pushed to `origin`**, the **Changes** tab renders empty - even though there's a real,
committed, non-empty diff against that base. The "Against base" view appears to resolve
the base via its remote ref (`origin/<base>`) and silently shows nothing when that ref
doesn't exist, instead of falling back to the local base branch.

### Environment
- superset.sh desktop app (`v1.12.5` + CLI (`superset workspaces`)
- Project: a normal git repo; default branch `main`; remote `origin`
- macOS 26.2

### Setup (stacked branches)

```
main
└── feature-a   (1 commit on main)        ← pushed? NO (local only)
    └── feature-b (1 commit on feature-a) ← workspace under inspection
```

- A superset workspace exists for each branch (`type: "worktree"`).
- superset correctly recorded the stack base in git config:
  - `branch.feature-b.base = feature-a`
  - `feature-a` has no `.base` key (falls back to project default).
- Neither `feature-a` nor `feature-b` has been pushed to `origin`.
  
### How to Reproduce
1. Create the stacked branches above; do **not** push `feature-a` to origin.
2. Open the `feature-b` workspace → **Changes** tab.

### Expected
`feature-b`'s Changes tab shows its incremental diff against `feature-a`
(`git diff feature-a...feature-b` — non-empty).

### Actual
`feature-b`'s Changes tab is **empty**, despite a clean working tree and a real
committed change. The `feature-a` (bottom) workspace renders correctly, because it
has no recorded base and falls back to the default branch (`origin/main`, which exists).

### Root cause (evidence)
- Base **is** recorded correctly: `git config branch.feature-b.base` → `feature-a`.
- The **local** diff is non-empty: `git diff --stat feature-a...feature-b` → changed files.
- But `feature-a` is local-only: `git ls-remote origin refs/heads/feature-a` → no match,
  and there is no `origin/feature-a` ref locally.
- The Changes view ignores the existing **local** `feature-a` ref and resolves the base
  through a remote-qualified ref (`origin/feature-a`). Since that doesn't exist, the diff
  comes up empty rather than falling back to the local branch.
- Config is origin-centric (`checkout.defaultremote=origin`); the base branch has no
  tracking remote because it was never pushed. Consistent with #2516 (origin assumed).

### Impact
Stacked-workspace / stacked-PR workflows show an empty Changes tab for every workspace
whose parent branch hasn't yet been pushed to `origin` — i.e. the common case while a
stack is still being built locally before pushing.

Related issues

- #2444 — show local/remote indicator in base branch picker (calls out local-only base branches)
- #2851 — auto-sync worktree base branch with GitHub PR target ("Against base" diff handling)
- #2516 — push always targets origin, ignoring branch tracking remote (origin-centric resolution)

andyst-dev:
Fixes #5298

## Problem

For a **stacked workspace** whose base is another **local branch that has never been pushed**, the **Changes** tab rendered empty — even though a real, committed, non-empty diff existed against that base.

`resolveBaseComparison` resolved the "Against base" view via the remote ref `origin/<base>` and, when no upstream was configured, silently fell through to `origin/<branch>` — a ref that does not exist for an unpushed local base. The diff against a missing ref is empty, so the UI showed nothing.

## Fix

In `resolveBaseComparison`, when no upstream is configured:

1. if `refs/remotes/origin/<branch>` exists → `baseRef: origin/<branch>` + fetch target (unchanged behavior);
2. **new** — if the remote ref is absent but `refs/heads/<branch>` exists (local-only base) → `baseRef: <branch>` with `fetchTarget: null` (there is nothing to fetch for a local branch);
3. otherwise → unchanged fallback.

## Validation

Two integration tests against real on-disk repos (same suite as the existing `resolveBaseComparison` tests):

- **local-only base** (exact #5298 repro): base branch checked out locally, never pushed → resolves to `baseRef: "feature/stacked"`, `fetchTarget: null`;
- **remote + local coexistence**: `refs/remotes/origin/feature` exists → still prefers `origin/feature` with its fetch target.

`bun test` on git-helpers.integration.test.ts: **22/22 pass** · `tsc --noEmit`: 0 errors · Biome: clean.

---
## Summary by cubic
Fixes an empty Changes tab for stacked workspaces by resolving the “Against base” diff to a real ref when `origin/<branch>` is missing. Previously we fell back to `origin/<branch>` and showed an empty diff; now we probe remotes and use a remote match (prefer `origin`), otherwise the local branch, so the diff shows actual changes. Addresses #5298.

- Updated `resolveBaseComparison`: probe all remotes in parallel; deterministically prioritize `origin` regardless of remote order; if no remote ref exists but `refs/heads/<branch>` does, return the local branch with `fetchTarget: null`.
- Supports non-`origin` remotes (fork workflows) and retains preference for `origin/<branch>` when both remote and local refs exist.
- Tests: integration covers the full #5298 pipeline (base→diff returns committed work), coexistence and fork cases; unit tests verify parallel probing without wall-clock delays and `origin` priority.

<sup>Written for commit b4ff1eb6e5d7f135a9bd4496111d8d64fae29b58. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved branch comparison across multiple configured remotes.
  * Prioritizes the standard remote when multiple matching branches are available.
  * Detects branches hosted on other remotes when the standard remote lacks them.
  * Supports local-only branches without unnecessary fetching.
  * Retains the standard fallback when no local or remote branch is available.

* **Tests**
  * Added coverage for local, standard-remote, and alternate-remote branch scenarios.

### DIFF
--- a/packages/host-service/src/trpc/router/git/utils/git-helpers.integration.test.ts
+++ b/packages/host-service/src/trpc/router/git/utils/git-helpers.integration.test.ts
@@ -190,10 +190,11 @@ describe("resolveBaseComparison (integration)", () => {
 			baseRef: "feature/parent",
 			fetchTarget: null,
 		});
+		if (!base) throw new Error("expected base comparison");
 
 		// The "Against base" view: diff the base ref against HEAD. This is
 		// the exact pipeline that rendered empty before the fix.
-		const files = await getChangedFilesForDiff(git, [`${base!.baseRef}..HEAD`]);
+		const files = await getChangedFilesForDiff(git, [`${base.baseRef}..HEAD`]);
 		expect(files.map((f) => f.path)).toEqual(["child.md"]);
 	});
 
@@ -201,6 +202,22 @@ describe("resolveBaseComparison (integration)", () => {
 		await git.raw(["update-ref", "refs/remotes/origin/feature", "HEAD"]);
 		await git.raw(["checkout", "-b", "feature"]);
 		await commitFile(git, repo, "feature.md", "x", "feature work");
+		expect(await resolveBaseComparison(git, "feature")).toEqual({
+			branchName: "feature",
+			baseRef: "origin/feature",
+			fetchTarget: { remote: "origin", branch: "feature" },
+		});
+	});
+
+	test("prefers origin when both origin and upstream have the branch", async () => {
+		// Put upstream first in git's configured remote order so origin's
+		// priority cannot depend on whichever remote happens to be listed first.
+		await git.raw(["remote", "remove", "origin"]);
+		await git.raw(["remote", "add", "upstream", "https://example.com/up.git"]);
+		await git.raw(["remote", "add", "origin", "https://example.com/repo.git"]);
+		await git.raw(["update-ref", "refs/remotes/upstream/feature", "HEAD"]);
+		await git.raw(["update-ref", "refs/remotes/origin/feature", "HEAD"]);
+
 		expect(await resolveBaseComparison(git, "feature")).toEqual({
 			branchName: "feature",
 			baseRef: "origin/feature",
--- a/packages/host-service/src/trpc/router/git/utils/git-helpers.integration.test.ts
+++ b/packages/host-service/src/trpc/router/git/utils/git-helpers.integration.test.ts
@@ -119,6 +119,7 @@ describe("resolveBaseComparison (integration)", () => {
 		// Simulate a remote so origin/HEAD can be set. We don't need an
 		// actual remote to fetch from — `symbolic-ref` on the remote HEAD
 		// is all getDefaultBranchName reads.
+		await git.raw(["remote", "add", "origin", "https://example.com/repo.git"]);
 		await git.raw(["update-ref", "refs/remotes/origin/main", "HEAD"]);
 		await git.raw([
 			"symbolic-ref",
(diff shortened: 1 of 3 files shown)


andyst-dev:
Thanks — both P3s are addressed in `ab8f855`:

- **Test now covers the real reported pipeline** (cubic P3 @ test:185): the #5298 test creates a local-only **parent** branch, checks out a **child** branch with committed work, resolves the base against the parent, and runs `getChangedFilesForDiff([parent..HEAD])` — asserting the committed work is returned. This is the exact chain that rendered an empty Changes tab before the fix, and it would have caught the HEAD-vs-base trap the previous version had.
- **Remote probe no longer hardcodes `origin`** (cubic P3 @ git-helpers:191): the no-upstream path now iterates `git.getRemotes()` and picks the first remote holding `refs/remotes/<remote>/<branch>`, so a base that exists only on a non-origin remote (fork workflow) resolves correctly with its fetch target. New test: base found on a remote named `upstream`.

`bun test`: 23/23 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks - both findings are addressed in `a3ac9d2`.

- Remote refs are still probed concurrently with `Promise.all`, avoiding sequential subprocess latency.
- Selection is deterministic after the probes complete: `origin` is always prioritized when it contains the branch, regardless of configured remote order or which probe finishes first.
- Non-origin-only refs still fall back correctly for fork workflows.

Coverage now includes concurrent three-remote probing with `origin` priority, a non-origin-only unit case, and an integration case where both `origin` and `upstream` contain the branch.

Validation:

- focused Bun tests -> **52 passed, 79 assertions**
- host-service typecheck -> passed
- Biome on the 3 changed files -> clean
- `git diff --check` -> clean

andyst-dev:
Good catch - fixed in `5f2e203`.

The concurrency helper now yields with `await Promise.resolve()` instead of a real 10 ms timer. Each probe still increments `activeProbes` before yielding, so the test continues to prove that all remote probes are created concurrently without depending on wall-clock scheduling.

Validation:

- `bun test packages/host-service/src/trpc/router/git/utils/git-helpers.test.ts` -> **28 passed, 40 assertions**
- Biome on the test file -> clean
- `git diff --check` -> clean

PR #6238, opened 2026-08-06
fix(host-service): fall back to prompt slug when LLM replies conversationally

### LINKED ISSUE #5288: [bug] Issue with branch/workspace name generation when prompt includes URL
#### Describe the bug

If I start a new session/workspace with a prompt like "Implement this jira task: <JIRA_TASK_URL>", then the workspace & branch name get set to something like:

> I'd be happy to help you implement that task, but I don't have access to...

This is obviously not ideal and also results in duplicate branch names since the model returns exactly the same output for similar prompts with different URLs. On that note, the branch de-duplication functionality also appears to be broken. 

Version 1.12.5

andyst-dev:
Fixes #5288

## Problem

When the workspace prompt contains a URL (e.g. a Jira task link), the branch-naming model ignores the *"return ONLY the branch name"* instruction and answers conversationally:

> I'd be happy to help you implement that task, but I don't have access to...

Two consequences:

1. The old sanitizer let that **whole sentence** through as a kebab-case branch name (`i-d-be-happy-to-help-you-implement-that-task-...`).
2. Because the model returns the **identical reply regardless of the URL**, distinct prompts produced identical names — and the deduplicator then suffixed them into near-identical branches, which is why de-duplication appeared broken.

## Fix

In `ai-branch-name.ts`:

- **`isPlausibleBranchName(candidate)`** — rejects sanitized replies with more than 8 kebab segments (a real branch name is 2–4 words; a conversational reply is a whole sentence).
- **`slugifyPrompt(prompt)`** — derives a deterministic branch name from the prompt itself: strips URLs, drops stop words, keeps the first 4 meaningful words. `"Implement this jira task: https://jira.example.com/TASK-123"` → `jira-task`.
- `generateBranchNameFromPrompt` falls back to the prompt slug when the model's reply is implausible; the result still runs through the existing `deduplicateBranchName`.

## Validation

`ai-branch-name.test.ts` — 7 unit tests: plausibility gate (accepts `feat-auth-flow`, rejects the conversational reply), slugifier (URL stripping + empty fallback), conversational fallback (exact #5288 repro), and dedup of the fallback slug (`jira-task` → `jira-task-2`).

`tsc --noEmit`: 0 errors · Biome: clean.

---
## Summary by cubic
Prevents conversational LLM replies from becoming branch names by rejecting implausible outputs and falling back to a prompt-derived slug. Fixes #5288; names are deterministic, deduplicated, and capped at 100 characters.

- Plausibility gate: rejects sanitized candidates with more than 8 kebab segments.
- Prompt slug fallback: strips URLs, drops stop words, keeps the first 4 meaningful words; sanitizes and enforces the 100‑char limit; defaults to "workspace".
- Apply fallback when the generated string is implausible or sanitizes to empty; then pass through existing de-duplication.
- Tests cover null-model, generation timeout/error, long-slug truncation, and dedup on both fallback and plausible names; mocks reset after each test.

<sup>Written for commit 130a4c985c54994033d8dd20b67469d066a11a4f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Improved automatic branch-name generation with clearer, more concise names.
  - Added fallback names based on the request when generated names are too long, conversational, or empty.
  - Removes URLs and common filler words from fallback names.
  - Limits fallback names to a practical length and uses “workspace” when no suitable name is available.
  - Prevents duplicate branch names by checking existing branches.

- **Bug Fixes**
  - Rejects implausible generated branch names and applies a reliable fallback.

### DIFF
--- a/packages/host-service/src/trpc/router/workspace-creation/utils/ai-branch-name.test.ts
+++ b/packages/host-service/src/trpc/router/workspace-creation/utils/ai-branch-name.test.ts
@@ -1,0 +1,85 @@
+import { afterEach, describe, expect, mock, test } from "bun:test";
+
+// Mock the two module boundaries generateBranchNameFromPrompt crosses:
+// the model provider (getSmallModel) and the chat title generator.
+const getSmallModelMock = mock(async () => ({ id: "small-model" }));
+const generateTitleMock = mock(async () => "feature-branch");
+
+mock.module("@superset/chat-legacy/server/shared", () => ({
+	getSmallModel: getSmallModelMock,
+}));
+
+mock.module("@superset/chat-legacy/server/desktop", () => ({
+	generateTitleFromMessage: generateTitleMock,
+}));
+
+const { generateBranchNameFromPrompt, isPlausibleBranchName, slugifyPrompt } =
+	await import("./ai-branch-name");
+
+describe("isPlausibleBranchName", () => {
+	test("accepts a normal kebab branch name", () => {
+		expect(isPlausibleBranchName("feat-auth-flow")).toBeTrue();
+		expect(isPlausibleBranchName("fix-login")).toBeTrue();
+	});
+
+	test("rejects a conversational reply sanitized to many segments", () => {
+		expect(
+			isPlausibleBranchName(
+				"i-d-be-happy-to-help-you-implement-that-task-but-i-don-t-have-access-to-the-jira-api",
+			),
+		).toBeFalse();
+	});
+});
+
+describe("slugifyPrompt", () => {
+	test("strips URLs and keeps meaningful words (#5288)", () => {
+		expect(
+			slugifyPrompt(
+				"Implement this jira task: https://jira.example.com/TASK-123",
+			),
+		).toBe("jira-task");
+	});
+
+	test("falls back to a default when nothing remains", () => {
+		expect(slugifyPrompt("https://example.com")).toBe("workspace");
+	});
+});
+
+describe("generateBranchNameFromPrompt", () => {
+	afterEach(() => {
+		generateTitleMock.mockClear();
+	});
+
+	test("uses the generated name when it is a plausible branch name", async () => {
+		generateTitleMock.mockImplementation(async () => "feat-auth-flow");
+		expect(await generateBranchNameFromPrompt("add auth flow", [])).toBe(
+			"feat-auth-flow",
+		);
+	});
+
+	test("falls back to a prompt slug when the model replies conversationally (#5288)", async () => {
+		generateTitleMock.mockImplementation(
+			async () =>
+				"I'd be happy to help you implement that task, but I don't have access to the Jira API",
+		);
+		expect(
+			await generateBranchNameFromPrompt(
+				"Implement this jira task: https://jira.example.com/TASK-123",
+				[],
+			),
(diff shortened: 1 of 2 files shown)


andyst-dev:
Thanks — both points are addressed in `e129b7b`:

- **Length limit on fallback slugs** (coderabbit Major): `slugifyPrompt` now runs its output through the same `sanitizeGeneratedBranchName` as generated names, so the 100-char branch limit applies to prompt-derived slugs too. Regression test asserts the slug length is ≤ 100.
- **Empty-sanitize fallback** (coderabbit Minor): a model reply that sanitizes to empty (e.g. `"..."` or emoji-only) no longer returns null — it falls back to the deterministic prompt slug instead. Regression test: `"Fix the login flow"` with a `"..."` reply → `fix-login-flow`.

`bun test`: 9/9 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks — the test gap is closed in `d962021`:

- **null model**: `getSmallModel` resolving `null` → `generateBranchNameFromPrompt` returns `null`;
- **generation rejection**: `generateTitleFromMessage` throwing (the `Promise.race` timeout path) → returns `null`;
- **dedup on the plausible-name path**: `feat-auth-flow` colliding with an existing branch → `feat-auth-flow-2`.

The mock is now typed `Promise<{ id: string } | null>` to match the real `getSmallModel` signature.

For the record, the two coderabbit comments on this thread (length limit + empty-sanitize fallback) were already addressed in `e129b7b` — they appear re-anchored onto the current head.

`bun test`: 12/12 pass · `tsc --noEmit`: 0 errors · Biome: clean.

andyst-dev:
Thanks - both test-isolation points are addressed in `ec2874b`.

- `afterEach` now clears and restores `getSmallModelMock` to the default model implementation, so the null-model case cannot leak into later tests.
- `generateTitleMock` is also restored to its default implementation for full case isolation.
- Renamed the immediate-rejection test from "generation rejects (timeout)" to "generation rejects" so it describes the path it actually exercises without claiming fake-clock timeout coverage.

Validation:

- `bun test packages/host-service/src/trpc/router/workspace-creation/utils/ai-branch-name.test.ts` -> **12 passed**
- `biome check packages/host-service/src/trpc/router/workspace-creation/utils/ai-branch-name.test.ts` -> clean
- `bun run --cwd packages/host-service typecheck` -> passed
- `git diff --check` -> clean

andyst-dev:
Closing — superseded by the v2 host-service rewrite: `ai-branch-name.ts` was removed from `packages/host-service` (404 on main); the AI branch-naming flow now lives in the v2 create path (only a `plans/done/20260415-v2-host-service-ai-branch-naming.md` record remains). Closing and deleting the fork branch.

Issue #5594, opened 2026-07-11
[Tracker] Diff viewer & Changes tab issues (stale-issue sweep, July 2026)


Kitenite:
Consolidated tracker for stale `diff-changes` issues closed in the July 2026 stale-issue sweep (no activity since 2026-06-01).

Each issue below was closed as a duplicate of this tracker. If any of them still reproduces (or is still wanted) on a current build, comment here or on the original issue and we'll reopen it.

- #4683 — [feat] Allow semantic diff on view like diffstatic
- #4511 — [bug] no option to stage individual files
- #4363 — Move project tree into diff tree
- #4287 — Improve Changes tab: clearer UX (Zed-style), AI commit-message button, Create PR action
- #3979 — Find tool broken in changes / diff preview
- #3849 — [feat] Add a VS Code-style Git log graph to the Source Control panel
- #3425 — [feat] Show precommit hook output when commiting
- #3244 — [feat] Inline annotations on local diffs to direct agent feedback (pre-PR)
- #2923 — [bug] changes section doesn't work after updating to 1.4.2 from 1.4.1
- #2835 — [feat] Inline code comments in diff view to send instructions to the agent
- #2753 — [bug] Diff view does not auto-refresh when files change on disk
- #2687 — [bug] Diff viewer loses "viewed" collapse state after switching workspace
- #2448 — [feat] Add text search (Cmd+F) in diff viewer
- #2315 — Add VSCode-style scrollbar change indicators to diff viewer
- #1234 — [feat] AI generated commit messages based on current staged files
- #1003 — Feature: Support nested git repositories in Source Control panel for monorepos
- #935 — [feat] All diff commenting

rdesistere-fence:
#4511 (no option to stage individual files) still reproduces on 1.28.0.

In the v2 Changes sidebar, staging is only available from the section header (Stage all / Unstage all). File rows expose Discard plus an open-in dropdown, with no per-file stage action.

<img width="1512" height="796" alt="Image" src="https://github.com/user-attachments/assets/dc5d766a-ae97-4c5b-9944-a8d54497c7b8" />

Worth flagging that #4512 already implements this: git.stageFile / git.unstageFile with the same path validation discardChanges uses, hover +/− and context-menu items in FileRow and ChangesTreeView, and integration tests. It was closed unmerged. Any chance of reopening it?

ryanjamar-ncino:
Tangential to #4511 & #4512 
[...]

Issue #4935, opened 2026-05-26
[feat] Revive V1 commit history view in the Superset V2 interface


jnelken:
## Summary

Please bring back the V1-style commit history view in the Superset V2 interface, showing the workspace branch's commit history against the target/base branch as a first-class view.

## Problem

In Superset V1, it was easy to inspect the current workspace branch's commits relative to the target branch. In the V2 interface, that capability appears to be hidden behind a dropdown instead of being directly visible.

This makes common review and workspace-management tasks harder:

- quickly understanding what changed in a workspace
- checking whether a branch is ahead/behind the target branch
- reviewing commit-by-commit progress before opening or updating a PR
- orienting in long-running agent-created workspaces

## Requested behavior

Add a visible V2 commit history view that compares the current workspace branch against the target branch.

Ideally this should be available directly from the workspace/branch surface, not only tucked inside a dropdown.

## Why this helps

- Restores an important V1 workflow in the V2 interface.
- Makes branch state easier to scan without leaving Superset.
- Helps users and agents verify work before PR creation or handoff.
- Reduces friction when managing many workspaces/tabs.

## Acceptance criteria

- The Superset V2 interface exposes a first-class commit history view for the current workspace.
- The view is computed against the target/base branch, not just raw local commit history.
- Users can access it without discovering a hidden dropdown-only path.
- The view makes it clear which branch is being compared to which target branch.

Issue #3849, opened 2026-04-29
[feat] Add a VS Code-style Git log graph to the Source Control panel


TCP404:
#### Describe the feature

I'd like to propose adding a graphical Git log graph view to the Source Control
panel, similar to the VS Code "Git Graph" extension.

#### Background

The right-hand panel in Superset is currently a file explorer — there's no
visual entry point for git history. When I'm working in a repo with many merge
commits and parallel branches (for example my AionUi repo, where dozens of
merge PRs are interleaved), the only option today is dropping into the terminal
and running `git log --graph`, which is not a great reading or interaction
experience.

#### What I'd like

A new Graph view inside the Source Control panel that provides:

- **Branch swimlanes** — each branch rendered as a stable-colored lane, with
  merges and forks visualized
- **Ref decorations** — branch names, tags, remote/HEAD markers shown on commit
  rows, similar to `git log --decorate`
- **Commit details** — clicking a commit opens a details pane showing author /
  date / message / changed files, with per-file diffs
- **Compare** — select two commits and diff them
- **Common actions** — right-click a commit for checkout / cherry-pick /
  revert / reset
- **Filtering** — by branch, author, or message substring

#### References

- VS Code Git Graph extension:
  https://marketplace.visualstudio.com/items?itemName=mhutchie.git-graph

#### Suggested rollout

A reasonable first cut would be: swimlanes + ref decorations + clickable commit
details with diffs. Right-click actions and filtering can follow in later
iterations.

Kitenite:
Closing as a duplicate of #5594, the consolidated tracker for stale `diff-changes` issues from the July 2026 stale-issue sweep (no activity since 2026-06-01). If this still applies on a current build, comment on #5594 and we'll reopen.

PR #6257, opened 2026-08-07
feat(desktop,host-service): add a commit graph to the workspace sidebar

acseven:
## What & why

Adds a fourth **Graph** tab to the v2 workspace sidebar: a lane-based commit graph whose primary value is **ref classification**, not the lanes. The lanes are the frame; what the tab is actually for is surfacing branches and worktrees that exist in git but have no open workspace behind them.

Commit history was only reachable as a flat `base..HEAD` dropdown — no topology, and no view at all of the branches and worktrees that accumulate outside the currently-open workspaces. A branch whose worktree was deleted, or a worktree with no workspace row, was invisible until it caused a problem.

Read-only. No write actions (checkout, cherry-pick, revert, prune) in this PR.

Relates to superset-sh/superset#4935 and superset-sh/superset#5594 (superset-sh/superset#3849 folded in).

### Host service — `git.listGraph`

- New procedure alongside `listCommits`, which is untouched (desktop and mobile both consume its output type). `getCommitFiles` is the one existing procedure that changed: it now also returns the commit's own metadata (subject, body, author, date, parents), so a diff pane opened from the graph renders its header without a second spawn.
- Tip set is seeded from git's own refs, never from workspace records. `refScope` picks how wide it goes, and all five values are served:

  | Scope | Tips | Reads as |
  |---|---|---|
  | `head` | `HEAD` | this workspace's own history |
  | `open-workspaces` | + base ref, worktree branches, detached worktree heads | branches open in other Superset workspaces |
  | `local` (default) | + every local branch, the current branch's upstream | everything on this machine |
  | `remote` | `HEAD` + `--remotes` | what the last fetch saw on the server |
  | `all` | `--exclude=refs/stash --all` | every ref, tags included |

  `remote` and `all` pass one flag rather than enumerating refnames, so a tag-heavy repo can't overflow the command line. Detached worktree heads are still named explicitly — they aren't refs. Under `open-workspaces`, a worktree git marks `prunable` is excluded: its directory is gone, so it is not an open workspace.
- Ref state is classified per local branch: `open` · `detached-worktree` · `prunable` (+ reason) · `merged` · `orphan-branch`. Containment comes from `git branch --merged <baseRef>`, not from graph topology, so a branch merged far outside the window still reports `merged`. `prunable` comes from git's own porcelain flag (`gitdir file points to non-existent location`) rather than a per-branch `existsSync`, so classification costs no synchronous fs calls on the host event loop.
- Every git spawn happens in the worker (`gitGraphLogTask`), including base resolution, so both stdout draining and parsing stay off the host event loop. Decorations are built from `for-each-ref` and resolved through `ResolvedRef` — never from `%D`, never from a shortname prefix.
- Overlapping traversals coalesce, so a `git:changed` storm during a rebase doesn't fan out into duplicate walks.
- The pagination cursor is tagged with the scope it was issued for. `--skip` counts rows of one traversal, so replaying a `local` offset against `head` would window a list that never had those rows; a cursor from another scope restarts at page one instead of serving a silently wrong page. Scope is also part of the coalescing dedupe key, so two scopes never share one walk.

### Renderer

- `assignLanes` — pure, tested, no React or tRPC. Active-lane sweep over topo-ordered input; lane identity is pinned to the originating branch so a refetch doesn't reshuffle the graph.
- Virtualized list, per-row inline SVG (not one tall canvas), so virtualization stays trivial. Row height is derived from the commit's own refs — no measurement pass — and the size cache is keyed by commit hash so it survives refs moving underneath an open tab.
- Lane colours are new additive `--graph-lane-1..8` tokens in both themes, equal chroma/lightness so the lanes read as one family and survive a theme switch. No component hardcodes a colour; the live values come from `shared/themes/graph-lanes` through `applyGraphLaneColors`, and the hex in `globals.css` is only the pre-hydration fallback set.
- Ref badges encode state in texture as well as colour — fill, dashes, strike — so they still read at 260px and for a colourblind viewer, leaving hue free for lane identity.
- Row tint in the lane's colour makes lane membership readable without tracing lines, and survives the lane compression narrow widths force.
- Commit clicks open the existing `diff` pane kind; the graph owns its own persisted selection, separate from the Changes tab's filter, so the two surfaces no longer fight over one piece of state. Shift-click ranges, cmd/ctrl-click forces a new pane, double-click pins — matching the file tree's convention exactly.
- Read-only row context menu (copy hash / subject / open in new tab).
- The tab's own control strip carries a scope chooser and two toggles. The chooser sits left of them and reads **This workspace · Open workspaces · All local branches · Remote · Everything** — plain wording rather than git's, since the audience is people who don't think in refs. The toggles are ref badges on their own untruncated line, and dimming of commits that carry no unreferenced ref. The strip renders even when the window is empty, so a scope that returns nothing can't take away the control that would let you leave it.

### Notable decisions

**The "unreferenced" filter dims, it does not remove.** Lanes and edges are assigned across the whole commit list; dropping rows would leave the surviving edges pointing at rows that aren't there, and filtering before lane assignment yields a graph of disconnected nodes.

**Selected rows and range-interior rows share a tint alpha.** Selection also carries `--accent`; the range interior doesn't. Same alpha therefore keeps an endpoint strictly brighter than the band it anchors, in both themes, with no per-theme numbers.

**The row size cache is keyed by commit hash, not by list index.** Refs move while the tab is open, so a refetch that prepends a commit shifts every index — under the virtualizer's default index key a row inherits the cached height of whatever previously sat in its slot, which in two-line mode misfits by a whole badge line and shows up as overlapping rows or a blank band. Keying by hash lets a cached size travel with its commit; `rows` is also in the remeasure deps for the case the key can't catch, a ref landing on a commit whose hash doesn't change. Only reproduces on a repo with live churn.

**"Open workspaces" is read off `git worktree list`, not the workspaces table.** The worker has no DB access, and every Superset workspace is a worktree, so the worktree list already answers the question. The known ceiling: a worktree created outside Superset counts as an open workspace. Joining against DB rows in the coordinator is the upgrade path if that ever matters.

**The header toggles read through a live query, not `.get()`.** They write to `v2SidebarProjects`, and `.get()` is a non-reactive snapshot — the row changed in localStorage while the header kept rendering the stale flag until some unrelated re-render happened to run, which reads as a multi-second lag on a write that already completed. No spinner: the write is synchronous and local, so a pending indicator would animate over an already-finished state change. The toggles also `ensureProjectInSidebar` first, matching `useV2ProjectDefaultApp` — a project with no sidebar row yet would otherwise swallow the write entirely.

**The strip's controls are not in `SidebarHeader`'s `actions` slot.** That slot is a `shrink-0` island in a fixed 40px row whose four tab buttons are each `flex-1` with no `min-w-0`; anything rendered there clips the last tab label instead of shrinking it. The chooser and toggles live on the graph's own control strip instead.

## How I tested it

| Gate | Result |
|---|---|
| `bun run lint` | 0 (5832 files) |
| `bun run typecheck` | passes except `@superset/ui`, which fails on a duplicate `@types/hast` resolution in `node_modules`. No `packages/ui` file, `package.json`, or `bun.lock` is in this branch's diff, so it is not from here |
| `bun run test` | 14/15 packages green |
| `apps/desktop` tests | 2617 pass, 1 fail — `getProcessEnvWithShellPath`, environment-dependent and untouched by this branch |
| `packages/host-service` tests | 998 pass, 0 fail, 14 skip |

`refScope` has its own integration test: a branch with no worktree is absent under `head` and `open-workspaces` but present under `local` and `all`; a worktree checked out on that branch whose directory is then deleted stays excluded from `open-workspaces`, which fails without the prunable filter; and `remote` excludes the local-only branch until `refs/remotes/origin/side` exists, then includes it. Cursor handling is covered both ways — a `local` cursor pages forward, and the same cursor handed to `head` restarts at page one instead of windowing rows that traversal never produced. `buildTipSet` is unit-tested per scope, including that `all` and `remote` emit a flag rather than a refname list.

Integration test covers all five ref states against a fixture repo (open workspace, worktree with no workspace row, orphan branch, prunable worktree with its path deleted) plus a merge commit, a tag, and a branch merged outside the commit window. Lane assignment is unit-tested for linear / merge / octopus / criss-cross / root / lane reuse / parent outside the window, including determinism when commits are appended.

Rendering was checked against the running app at 260px and 400px sidebar widths, in both toggle states, with the measurements alongside the screenshots (total virtual size, per-row height, badge clipping). The toggle-reactivity fix was verified end-to-end on the running app.

The row-height regression is pinned by a test that prepends a commit to a mixed-height list and asserts every commit present in both lists keeps the same `(key, size)` pair, looked up by hash rather than by slot. It was confirmed red against the pre-fix behaviour: restoring the index key fails it. The size and key derivation was extracted into a pure `graphRowSizer` to make that testable — this repo has no jsdom or testing-library, and adding one for two callbacks wasn't worth the dependency.

It was also verified live: with the Graph tab open on a repo of ~1600 commits, a commit landing on `HEAD` prepended a two-line row and shifted every index by one. Row offsets stayed consistent with rendered heights across the shift and across the subsequent removal — 0 mismatches in both directions, where the pre-fix build produced an overlap at the slot that grew and a blank band at the slot that shrank.

### Screenshots

Captured from the dev app over CDP (`RENDERER_REMOTE_DEBUG_PORT=9222 bun dev`, renderer matched to this workspace's `DESKTOP_VITE_PORT`), driving the real controls with dispatched mouse input. The clip is the sidebar column only; the rest of the window is unchanged by this PR. The branches below `main` are a throwaway topology built for these shots and deleted afterwards — the repo's own history is a straight line, which shows none of the cases the tab exists for.

**Inline refs** (default) — four concurrent lanes, two merges and an octopus. The fork elbow leaves the node horizontally then bends down; the merge descends then bends in, so the two read apart at a 3.5px node. Row tint carries lane membership without tracing the line. The dashed, struck-through red badge is a branch whose worktree directory is gone (`prunable`); badges truncate to fit 260px.

Commit graph with four lanes, merge and fork elbows, a tag badge, HEAD and main, and a struck-through prunable branch badge

**Scope chooser** — the five scopes as the user reads them, `All local branches` selected. Plain wording rather than git's, because the audience is people who don't think in refs.

Graph tab with the scope menu open, showing This workspace / Open workspaces / All local branches / Remote / Everything

**Refs on their own line** (toggle on) — same commits, badges move above the row they decorate and stop truncating: `origin/main` and the long branch names read in full. That is the trade the toggle exists for, and the reason row height has to come from the commit's own refs.

Same graph with ref badges on their own line above each commit, showing full branch and tag names

**Highlight unreferenced** (toggle on) — the point of the tab. Rows carrying a branch or worktree with no open workspace behind it stay bright; everything already accounted for dims. Nothing is removed, so the lanes and edges still connect.

Same graph dimmed except the rows carrying unreferenced branches, which stay at full contrast

## Checklist

- [x] `bun run lint` passes with 0 output (CI fails on warnings too)
- [ ] `bun run typecheck` — green for every package this PR touches; `@superset/ui` fails on a pre-existing duplicate `@types/hast` resolution, see the table above

## Known gaps

- Pagination is best-effort within a scope: the cursor is a `--skip` offset, so pages re-anchor if refs move between fetches. Crossing scopes is handled (the cursor is dropped), moving refs are not. The truncation banner surfaces when a repo has more history than one page.
- `Everything` is now largely `All local branches ∪ Remote ∪ tags`, so it may not earn its slot. Left in for one release to see whether anyone picks it.
- `remote` reflects the last fetch. Nothing in this PR fetches, so a stale remote reads as a stale graph.
- `All local branches` is the longest label and squeezes the "Showing first N of M" text beside it at the narrowest sidebar widths.
- Light theme is covered by unit tests on the pure colour helpers; the store's delivery path was verified in dark only.

## Summary by CodeRabbit

* **New Features**
  * Added a Git Graph tab to the workspace sidebar with a visual, lane-based commit history.
  * Added reference-scope filtering, pagination, two-line badges, and unreferenced-reference highlighting.
  * Added commit and range selection, diff previews, pane pinning, context-menu actions, and commit metadata.
  * Added responsive layouts and loading, empty, error, and truncated-history states.
  * Added theme-aware lane colors and indicators for branches, tags, HEAD, worktrees, merged, and prunable references.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,6 @@ superset-dev-data/
 
 # Local-only plans (not tracked)
 plans/local/
+
+# Bun compile temp artifacts (~60 MB each)
+*.bun-build
--- a/apps/desktop/plans/20260801-1200-git-graph-visual-spec.md
+++ b/apps/desktop/plans/20260801-1200-git-graph-visual-spec.md
@@ -1,0 +1,340 @@
+# Commit Graph — Visual Spec
+
+Design contract for the `Graph` sidebar tab. Covers lane geometry, the lane colour ramp, ref-badge
+treatments, row anatomy at each width, and the empty / loading / truncated states.
+
+Companion to `plans/20260801-commit-graph-sidebar.md` (the implementation plan). This file owns **pixels
+and colour**; that file owns data, algorithm, and wiring.
+
+**Status:** signed off. The three open calls — row tint at 6%/9%, badges inline rather than in a
+gutter, date always-on at the wide breakpoint — were reviewed against the mock and confirmed. The
+components in §8 are landed.
+
+**Ownership:** the files listed in §8 own the presentation. Anything that needs a pixel or a colour
+changed goes through this spec, not through a local override at the call site.
+
+---
+
+## 1. Frame
+
+The graph renders inside the v2 workspace sidebar. Real width range is **240–560px**, with the
+resize hysteresis at `WorkspaceSidebar.tsx:95`:
+
+```ts
+setCompact((prev) => (prev ? width < 280 : width < 260));
+```
+
+So `compact` engages below 260px and releases at 280px. Every number below is derived from that
+range — this is a sidebar component, not a page component, and it never gets a second column.
+
+Three breakpoints:
+
+| Name | Width | `compact` |
+|---|---|---|
+| Compact | < 260px | `true` |
+| Standard | 260–399px | `false` |
+| Wide | ≥ 400px | `false` |
+
+---
+
+## 2. Row geometry
+
+| Token | Standard / Wide | Compact |
+|---|---|---|
+| Row height | 28px | 24px |
+| Ref line height (two-line rows only) | 16px | 14px |
+| Two-line row height | 44px | 38px |
+| Lane pitch | 14px | 12px |
+| Lane centre x | `pitch / 2 + laneIndex * pitch` | same |
+| Node radius | 3.5px | 3px |
+| Node ring | 1.5px, drawn in the lane colour | same |
+| Edge stroke | 1.5px | 1.5px |
+| Elbow radius | 6px | 5px |
+| Visible lane cap | 6 (Standard), 8 (Wide) | 4 |
+| Lane column width | `min(laneCount, cap) * pitch + 4` | same |
+
+At Standard the lane column costs `6 * 14 + 4 = 88px` of 260px. That is the budget ceiling; it is why
+the cap is width-derived rather than a constant.
+
+**Overflow.** Beyond the cap, lanes sharing a parent merge into the last visible lane, and a `+N`
+chip renders at the right edge of the
+lane column. Horizontal scrolling of the lane column is explicitly rejected — it desynchronises from
+the subject text.
+
+### Edge kinds
+
+`assignLanes` emits five edge kinds per row. Geometry, with `y0 = 0`, `y1 = rowHeight`,
+`yc = rowHeight / 2` (the node's y), `e` = elbow radius, `s = sign(xTo - xFrom)`:
+
+| Kind | Meaning | Path |
+|---|---|---|
+| `pass` | Lane crosses the row untouched | `M x,y0 V y1` |
+| `in-straight` | The commit's own lane, entering from above | `M x,y0 V yc` |
+| `in-merge` | Another lane awaiting this commit, closing into the node | `M xFrom,y0 V (yc - e) Q xFrom,yc (xFrom + s·e),yc H xTo` |
+| `out-straight` | First parent inherits the lane | `M x,yc V y1` |
+| `out-fork` | Extra parent leaves toward another lane | `M xFrom,yc H (xTo - s·e) Q xTo,yc xTo,(yc + e) V y1` |
+
+Note the deliberate asymmetry: **merges descend then bend into the node horizontally; forks leave the
+node horizontally then bend down.** That is what distinguishes the two at a 3.5px node, without
+relying on the subject text. Both use the same elbow radius, so a fork and the merge that eventually
+closes it read as one shape. All paths are `fill: none`, `stroke-linecap: round`.
+
+A parent outside the fetch window emits `out-stub` instead of `out-straight`: a dashed segment from
+the node to `0.8 · rowHeight` at 55% opacity. It must render, not throw — it is the normal state of
+the last page.
+
+**Node.** Filled with the row background, ringed in the lane colour — a donut, not a dot. Merge
+commits (`parents.length > 1`) get a filled node instead, so merges are distinguishable at 3.5px
+without relying on the subject text. Root commits (`parents.length === 0`) get a filled node with a
+1px outer gap.
+
+---
+
+## 3. Lane colour ramp
+
+`packages/ui/src/globals.css` ships `--chart-1..5` only, and their hues are **not stable across
+themes** — light `--chart-1` is orange `oklch(0.646 0.222 41.116)`, dark is blue
+`oklch(0.488 0.243 264.376)`. A lane would change colour on a theme switch. So the spec adds a
+dedicated ramp.
+
+**Additive only.** Eight new custom properties in `:root` and `.dark`, plus eight `@theme inline`
+entries. Nothing existing is touched.
+
+Design rules for the ramp:
+
+1. **Fixed hue per index, identical in both themes.** Lane 3 is green in light and green in dark.
+2. **Equal lightness and chroma within a theme.** Lanes are peers; none may shout. This is what makes
+   eight colours read as one family instead of a rainbow.
+3. **Hues at 45° spacing** on the OKLCH wheel, offset so the trunk (lane 1) is not a warning colour.
+4. Values sit inside sRGB at the stated L/C, so no browser gamut-mapping surprises.
+
+```css
+:root {
+  /* Lane ramp — equal L/C, 45° hue spacing, hue held constant across themes. */
+  --graph-lane-1: oklch(0.56 0.125 262); /* indigo — trunk */
+  --graph-lane-2: oklch(0.56 0.125 152); /* green */
+  --graph-lane-3: oklch(0.56 0.125 25);  /* red */
+  --graph-lane-4: oklch(0.56 0.125 197); /* teal */
+  --graph-lane-5: oklch(0.56 0.125 307); /* violet */
+  --graph-lane-6: oklch(0.56 0.125 71);  /* amber */
+  --graph-lane-7: oklch(0.56 0.125 342); /* magenta */
+  --graph-lane-8: oklch(0.56 0.125 117); /* olive */
+}
+
+.dark {
+  --graph-lane-1: oklch(0.75 0.115 262);
+  --graph-lane-2: oklch(0.75 0.115 152);
+  --graph-lane-3: oklch(0.75 0.115 25);
+  --graph-lane-4: oklch(0.75 0.115 197);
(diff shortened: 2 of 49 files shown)


acseven:
@coderabbitai review

acseven:
Thanks — went through all of it. Pushed 40a0bc7b7 (code) and 7837067e0 (docs).

**Fixed**

- **`assignLanes` lane reclaim.** Confirmed, and it is reachable on a real graph: when a row's *first* parent falls outside the fetch window, the row stubs its own lane and frees the slot, and the loop over extra parents could then pick that same slot back up via `lanes.indexOf(null)` — one lane carrying both an `out-stub` terminator and a live edge. Now `lanes.findIndex((slot, i) => slot === null && i !== mine)`. New test `does not fork an extra parent back onto the lane it just stubbed` fails against the old line and passes against the new one. (The root-commit half of the finding doesn't apply — a root has no extra parents to loop over.)
- **Option-like revisions reaching git.** Confirmed by reproduction: `git show -s --format=%H "--output=/tmp/x/PWNED"` writes the file under git 2.50.1, so `commitHash` was an arbitrary-file-write primitive. `commitHash`/`fromHash` on `getCommitFiles` and `getDiff` are now `/^[0-9a-f]{4,40}$/i` (plus the `""` the renderer parks in a disabled query's input). Integration test asserts the option-shaped input is rejected and no file lands.
- **Unreferenced tooltip** now flips with the button, matching the `aria-label`.
- **Two weak tests.** `changes nothing when the toggle is off` compared a render against itself — it now asserts the one-line layout and that it differs from the two-line render. `commit outside the window is simply absent` asserted the opposite of its name; renamed to what it checks, with the "no row in the window is marked" assertion the name implied.
- **Docs drift:** five ref states not four; the lane tokens live in `apps/desktop/src/renderer/globals.css`, not `packages/ui`; the `GraphEdge` kind list is the shipped six-way split. The visual spec gets a superseded note for the lane ramp — it now derives from the active theme's ANSI palette rather than the fixed OKLCH list, so a user-imported theme isn't ignored. Fenced blocks got languages.

**Not fixed**

- **Guarding `commit.date` against an empty string.** The premise doesn't hold: the graph log format is `%H%x09%h%x09%P%x09%an%x09%ae%x09%aI%x09%s` (`graph-log.ts:104`), so the date is always emitted and always ISO-8601. `git show`'s metadata path uses `%aI` too. The screenshots above render real relative times off that field.
- **Moving `20260803-commit-graph-followups.md` out of `plans/done/`.** Fair reading of the convention, but the placement was a deliberate call. Instead the doc now says outright that §5.6 is the authoritative status, which is what made it read as contradictory.

`bun run lint` is 0. Graph tests 62/62, host-service git suites 107/107. `typecheck` is unchanged: still only the pre-existing duplicate `@types/hast` resolution, no file in this branch.

acseven:
Second review — one finding, confirmed and fixed in 8e8d380.

**Include all traversal inputs in the coalescing key** (`git.ts:364-369`) — real. `WorkerTaskRunner.runTask` returns the in-flight promise verbatim on a `dedupeKey` hit, so any overlapping call sharing `workspaceId` + `refScope` was served the other call's rows while still computing `nextCursor` from its own `skip`.

The paginating case CodeRabbit describes is the obvious one, but the reliable one is a cold open: `useGraphTab` fires `listGraph` with `baseBranch: undefined` while `getBaseBranch` is still loading, then refires with the resolved base. `baseBranch` is what feeds `git branch --merged` in `gitGraphLogTask`, so the second call could inherit a traversal that classified no branch as `merged`, and React Query then cached that under the resolved key for `staleTime`.

Key now carries scope, base, limit and skip. The two sibling coalescers (`base-ref`, `commit-files`) already key on their full payloads — checked, unchanged.

PR #6267, opened 2026-08-07
feat(desktop): add Resources menu with Cmd+Shift+U shortcut

AviPeltz:
## Summary
- Add a top-level **Resources** menu (alongside File/Edit/View/Window/Help) with a "Check Resources" item that opens the existing resource usage (CPU/RAM) view — for mouse-driven discoverability of the feature.
- The **⇧⌘U** shortcut that already opened this view continues to be owned by the renderer's user-customizable hotkey system (Settings > Keyboard), unchanged by this PR.

## Why / Context
Cmd+Shift+U already opened the resource monitor via a renderer `useHotkey` binding, but "Resources" wasn't discoverable anywhere in the UI (menu or otherwise) for users who don't know the shortcut.

An earlier version of this PR also added the shortcut as a native Electron menu `accelerator`. Cubic's review correctly flagged that as a regression: `CHECK_RESOURCES` is rebindable/disable-able via Settings > Keyboard (`hotkeyOverridesStore`, persisted in renderer `localStorage`, no main-process access), and a native accelerator is unconditionally live on macOS — it would fire and ignore a user's override or disable. That's been reverted; the renderer's `useHotkey` remains the sole source of truth for the actual keybinding, matching what the command palette already displays.

## How It Works
- `main/lib/menu.ts`: new "Resources" → "Check Resources" menu item, no accelerator. Click emits a `check-resources` event via the shared `menuEmitter`.
- `lib/trpc/routers/menu.ts`: added `check-resources` to the `MenuEvent` union and wired it into the existing `observable` subscription.
- `renderer/commandPalette/CommandPaletteHost.tsx`: `useHotkey("CHECK_RESOURCES", ...)` still owns the keyboard shortcut (override-aware); a new subscription to the `check-resources` menu event triggers the same action when the menu item is clicked.

## Manual QA Checklist
- [ ] "Resources" menu appears in the app menu bar between Window and Help
- [ ] Clicking Resources → Check Resources opens the resource usage view in the command palette
- [ ] Cmd+Shift+U still opens the same view
- [ ] Rebinding or disabling `CHECK_RESOURCES` in Settings > Keyboard changes/disables the shortcut, and the command palette hint reflects it (menu click still works regardless)
- [ ] Behavior verified after app restart (native menu is main-process, not renderer-hot-reloaded)

## Testing
- `bun run lint:fix` (clean)
- `bun run typecheck --filter=@superset/desktop` (passes)
- Not yet verified end-to-end against a running app — no dev instance was running for this worktree, and native OS menu chrome isn't inspectable via CDP. Flagging for reviewer/manual QA per the checklist above.

## Known Limitations
- No automated test coverage added; this is a small, mechanical menu/IPC wiring change consistent with existing untested sibling menu items (Settings, Keyboard Shortcuts).
- The menu item doesn't display the shortcut hint text — Electron accelerators are unconditionally live on macOS, so showing one would reintroduce the override-bypass bug above. The shortcut is discoverable via the command palette and Settings > Keyboard instead.

## Summary by CodeRabbit

* **New Features**
  * Added a **Resources** section to the application menu with a **Check Resources** option.
  * Selecting **Check Resources** opens the resource-checking command palette.
  * Native menu actions now use the same resource-checking workflow as the configurable keyboard shortcut.
  * Existing keyboard shortcuts for opening resource checks remain available and customizable.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/menu.ts
+++ b/apps/desktop/src/lib/trpc/routers/menu.ts
@@ -11,7 +11,8 @@ type MenuEvent =
 	| { type: "open-settings"; data: OpenSettingsEvent }
 	| { type: "open-workspace"; data: OpenWorkspaceEvent }
 	| { type: "open-project" }
-	| { type: "toggle-presets-bar" };
+	| { type: "toggle-presets-bar" }
+	| { type: "check-resources" };
 
 export const createMenuRouter = () => {
 	return router({
@@ -33,16 +34,22 @@ export const createMenuRouter = () => {
 					emit.next({ type: "toggle-presets-bar" });
 				};
 
+				const onCheckResources = () => {
+					emit.next({ type: "check-resources" });
+				};
+
 				menuEmitter.on("open-settings", onOpenSettings);
 				menuEmitter.on("open-workspace", onOpenWorkspace);
 				menuEmitter.on("open-project", onOpenProject);
 				menuEmitter.on("toggle-presets-bar", onTogglePresetsBar);
+				menuEmitter.on("check-resources", onCheckResources);
 
 				return () => {
 					menuEmitter.off("open-settings", onOpenSettings);
 					menuEmitter.off("open-workspace", onOpenWorkspace);
 					menuEmitter.off("open-project", onOpenProject);
 					menuEmitter.off("toggle-presets-bar", onTogglePresetsBar);
+					menuEmitter.off("check-resources", onCheckResources);
 				};
 			});
 		}),
--- a/apps/desktop/src/main/lib/menu.ts
+++ b/apps/desktop/src/main/lib/menu.ts
@@ -16,6 +16,10 @@ export function createApplicationMenu() {
 	const closeAccelerator = "CmdOrCtrl+Shift+Q";
 	const showHotkeysAccelerator = "CmdOrCtrl+/";
 	const openSettingsAccelerator = "CmdOrCtrl+,";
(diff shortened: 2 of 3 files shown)


AviPeltz:
Re CodeRabbit's nitpick on `CommandPaletteHost.tsx:38-51` (test coverage for the `check-resources` event path): this module has no existing tests for trpc-subscription-driven UI pushes (would need new scaffolding to mock `electronTrpc.menu.subscribe` + the frame-stack store), and this is a small, mechanical menu/IPC wiring change consistent with the already-untested `OPEN_SETTINGS`/`SHOW_HOTKEYS` menu events it mirrors. Leaving as a known limitation for now rather than introducing new test infra in this PR — flagged in the PR body already.

Also confirmed the "possibly related" #6037 is unrelated — that's the merged PR that originally added the resource monitor / `CHECK_RESOURCES` hotkey this PR builds on top of, not a conflict.

AviPeltz:
Good catch, confirmed valid — pushed a fix in 2994a4f19.

`CHECK_RESOURCES` is user-customizable/disable-able via Settings > Keyboard (`hotkeyOverridesStore`, renderer-only `localStorage`, no main-process access). A native Electron `accelerator` is unconditionally live on macOS regardless of that override, so it would have silently ignored a rebind or disable. Reverted to the renderer owning the keybinding via `useHotkey("CHECK_RESOURCES", ...)` (override-aware, matches what the command palette displays); the "Resources" menu item now has no accelerator and only opens the view on click, for mouse-driven discoverability.

PR #6270, opened 2026-08-07
feat(desktop): built-in "Superset CLI" preset in the v2 preset bar

Kitenite:
## Summary by CodeRabbit

* **New Features**
  * Added a built-in **Superset CLI** preset to the workspace preset bar.
  * Launches `superset --help` in a terminal without requiring an agent or hotkey.
  * Built-in presets can be hidden or restored through user preferences.
  * Added tooltips and launch actions for built-in presets.
  * Built-in presets remain separate from user-created presets and cannot be edited or rearranged.

* **Bug Fixes**
  * Preset visibility preferences are saved reliably and preserved across sessions.
  * Invalid preset visibility settings are automatically cleaned up.

### DIFF
--- a/apps/desktop/plans/20260807-builtin-cli-preset.md
+++ b/apps/desktop/plans/20260807-builtin-cli-preset.md
@@ -1,0 +1,101 @@
+# Built-in "Superset CLI" preset
+
+Ship a synthetic, app-provided preset in every user's preset bar that opens a terminal and
+runs `superset --help`. Goal: teach users the CLI exists at the place they already launch
+things — point-of-use discovery, not an announcement. No agent session, no prompt injection;
+just show the command surface in a real terminal they can keep typing into.
+
+Works out of the box because the bundled CLI shim (`~/.superset/bin/superset`) is already on
+PATH inside every Superset-spawned terminal.
+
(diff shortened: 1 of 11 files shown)


Kitenite:
Addressed review feedback:

- **No-op unhide writes** (CodeRabbit + cubic): the unhide branch returns `prev` unchanged when the id isn't present, so repeated unhides no longer trigger persistence writes.
- **Unbounded `hiddenBuiltinPresetIds`** (CodeRabbit + cubic): added `KNOWN_BUILTIN_PRESET_IDS` in the schema as the source of truth; `healV2UserPreferences` now prunes stray/retired ids (and coerces malformed values to `[]`), with tests. `BUILTIN_CLI_PRESET_ID` is compile-checked against the list via `satisfies`, so an id can't silently fall out of the pruning allowlist.
- **Stale plan-doc references** (both bots): build order no longer mentions icon assets, the badge, or flag plumbing — all cut earlier; the flag-related scope comments were already resolved by the always-on change.

Skipped with reason: extracting a shared chip component from `V2PresetBarItem`/`BuiltinPresetBarItem` (cubic P3) — deliberate duplication for now; the built-in item drops drag/edit behavior and there's only one built-in preset. Worth extracting when a second built-in appears.

PR #6298, opened 2026-08-09
feat(host-service,desktop): sequence-numbered exactly-once terminal output with gap-only catch-up

### LINKED ISSUE #6269: pty-daemon replays a 64KB raw-byte ring buffer with no terminal state resync, corrupting TUI frames
## What happened

Claude Code's TUI frequently renders corrupted frames in Superset — the prompt box border gets painted through leftover assistant text from earlier output, and table-border glyphs collide with the input line. It happens often enough during normal use to be disruptive.

Digging into the app bundle, the cause looks structural:

- `dist/main/pty-daemon.js:594` sets `DEFAULT_BUFFER_BYTES = 64 * 1024` — only 64 KB of raw pty bytes are retained per session.
- `appendOutput` (`:643-650`) evicts whole chunks off the head at arbitrary byte boundaries.
- `handleSubscribe` (`:578-588`) → `snapshotBuffer` (~`:652-655`) replays `Buffer.concat(session.buffer)` **verbatim** on re-subscribe — no cursor-position, scroll-region (DECSTBM), SGR, or alt-screen resync, and the replay can begin mid-CSI-sequence.
- Nothing in the app ever passes `--buffer-bytes`; the daemon's own arg parser is the only reference to it in the bundle, so it always runs at the 64 KB default. Confirmed against the live daemon process, which is launched with `--socket=` only.

Claude Code (Ink) repaints *relatively* — cursor-up N rows, erase, redraw — and assumes the emulator's cursor row matches its model of the last frame it drew. A mid-stream replay with no established terminal state breaks that assumption, so partial repaints land on the wrong rows and overprint instead of replacing.

64 KB is only a second or two of Claude output (streaming text + spinner + prompt-box repaints), so nearly every re-subscribe replays a truncated mid-frame stream. Quieter TUIs like a shell prompt fit comfortably in 64 KB from a clean state, which is why they don't show this.

Two secondary contributors:

- The replay renders at the *current* pane width, though the bytes were wrapped for the width at emit time.
- xterm.js reflows wrapped lines on resize while Ink does not, so pane drags and sidebar toggles desync row counts independently.

## Steps to reproduce

1. Open a Claude Code session in a Superset workspace and let it produce a few screens of streaming output.
2. Switch to another pane/workspace, or resize the pane / toggle the sidebar, while output is streaming.
3. Switch back.
4. The prompt box and recent output are overprinted — border glyphs drawn through stale text rows.

## Suggested fix

1. Pass a larger `--buffer-bytes` to the daemon — the flag already exists and is parsed, but is never supplied.
2. The real fix: restore via the already-bundled `SerializeAddon` (`addon-serialize-CWWRL9-2.js` ships as its own renderer chunk) instead of replaying raw bytes, so the terminal is restored from screen state rather than a mid-stream byte dump.

## Environment

- Superset 1.19.0
- macOS arm64 (Darwin 25.5.0)
- xterm.js 6.1.0-beta.289 (commit `b1aee19`), addons: webgl, ligatures, unicode11, image, serialize

<sub>I have a screenshot of the corruption but it contains customer data, so I've left it off this public issue. Happy to send it privately if useful.</sub>

Kitenite:
## Summary

Re-fix for #6269 (TUI ghost artifacts from at-most-once output delivery) that makes both known failure modes structurally impossible — the ghosts #6279 targeted **and** the blank panes that forced its revert (#6290).

Design follows what established terminals do (VS Code ptyService revive, wezterm per-pane seqnos, Eternal Terminal catch-up, zellij attach redraw): **never reset a live, populated client and replay synthesized state into it.** Instead, make the byte stream exactly-once and reconcile only when bytes were actually lost.

## How it works

**Host (`packages/host-service/src/terminal/terminal.ts`)**
- Each session gets an `epoch` (fresh per session object) + absolute byte offset + a 2 MiB catch-up ring retaining the output tail regardless of attached sockets.
- Clients dial with `?seq=`:
  - `<epoch>:<n>` in-ring → **exact**: precisely the missed suffix. No gap → zero bytes; pane switches are wire-silent.
  - `new` (virgin xterm) → **tail**: ring contents as best-effort restore.
  - epoch mismatch / beyond ring / `none` → **reanchor**: *no content bytes at all* — the client's stale-but-real screen always beats anything the host could synthesize (the #6279 lesson). A **repaint nudge** (PTY rows−1 toggle → two real SIGWINCHes, the v1 terminal-host recipe) makes the running program redraw itself; it fires only if the client's attach resize wouldn't deliver a natural SIGWINCH.
- Wire order: synthesized bytes (mode preamble/notice, uncounted) → `synced` (arms counting) → pure PTY bytes (counted). Adoption now takes live daemon dims so the nudge logic can't misjudge.
- Legacy clients (no `?seq=`) keep the pre-seq contract byte-for-byte — version skew is safe both ways.

**Renderer (`apps/desktop`)**
- Transport counts bytes into a `seqAnchor` armed by `synced`, persists it next to the buffer snapshot (`terminal-seq:*`, registered + GC'd with the buffer keys), and restores it on runtime rebuild — so a parked-runtime eviction reattach converges exactly.

## Why the tests would have caught #6279
- Regression test **spans a host-service restart** (`__resetSessionsForTesting`) — tracker knows *less* than the renderer, the exact case #6279's suite never modeled — and asserts the grid is never blanked or overwritten and no `\x1bc` ever hits the wire.
- Catch-up payloads verified byte-for-byte against a **direct daemon subscription** (client-vs-client comparison alone is provably blind to self-consistent shifts).
- Mutation-checked: seq-accounting drift and a reintroduced destructive reanchor both fail loudly.

## Verification
- `terminal.seq-catchup.node-test.ts`: 6 phases + legacy contract, all green (run from `packages/host-service` under Electron-as-Node with the tsx loader).
- Full host-service suite 1065 pass; adoption e2e 20/20; desktop persisted-key + terminal tests 319 pass; lint + typecheck clean.
- **CDP end-to-end on the dev desktop** (real UI journeys, before/after screenshots):
  - Workspace switch with output flowing while away → full transcript, no blanking, anchor continuous.
  - **Host-service restart under a live TUI and under real Claude Code v2.1.226** (the #6290 scenario, via `hostServiceCoordinator.restart` + daemon adoption): zero blank frames across continuous 150–200 ms sampling, scrollback preserved, fresh epoch, nudge repaint converged (`FULL-REDRAW` / intact Claude UI).
  - Parked-runtime eviction → localStorage restore (`terminal-seq:*` round-trip) → exact catch-up of frames produced while the xterm didn't exist.
  - Socket killed mid-flood → reattach → seq 3,929 → 53,015 caught up, visible transcript contiguous, zero holes.
  - 3 MiB beyond-ring gap + full app reload → reanchor at live head **without delivering the blob** (no `x…` runs in the buffer), screen preserved, nudge healed.
  - Synthetic steps are noted as such (socket close as gap generator; focus-event dispatch standing in for window refocus, a known headless-drive quirk).

## Rollout note
Revert candidate rules: if reattach regressions appear (garbled grids, blank panes, scrollback loss), revert this single squash commit — legacy clients and old hosts are unaffected by design, and the revert restores today's FIFO behavior cleanly.

Fixes #6269. Context: #6279 (reverted in #6290), SUPER-1791, SUPER-859.

---
## Summary by cubic
Implements sequence-numbered, exactly-once terminal output with gap-only catch-up and attach-time repaint, and hardens persistence/edge paths to avoid stale anchors and regressions. Adds a seq catch-up e2e suite and runs host-service tests under Electron-as-Node via the `tsx` loader.

- **Bug Fixes**
  - `apps/desktop`: Persist the stream anchor only when the buffer snapshot succeeded and the parser is idle; flush coalesced writes before snapshot. Drop the anchor if a connection delivered bytes without ever sending `synced` (pre-seq host downgrade). Compute `?seq=` as `<epoch>:<n>` / `new` (virgin xterm) / `none` (xterm already has content), and persist/load anchors next to the buffer (`terminal-seq:*`) with GC support. Failed anchor writes clear the key to avoid stale state.
  - `packages/host-service`: Treat an empty `?seq=` as malformed and safely reanchor. Always send host-synthesized bytes (mode preamble/notice) first, then `synced {epoch,seq,mode}`, then PTY bytes. Improve the repaint nudge after reanchor: wait for client resize or fire after a fallback; toggle rows downward, or upward at minimum height. Adoption uses live daemon dims to avoid misjudging repaint triggers. Added `terminal.seq-catchup.node-test.ts` asserting no RIS resets and byte-exact catch-up; `scripts/test-e2e.ts` now uses the `tsx` loader and runs both adoption and seq suites.

<sup>Written for commit 4a7494d085cd1419dc834c16b7bfc9ffa5cc6e11. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6298?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved terminal synchronization across reconnects, preserving output and screen state more reliably.
  * Added precise catch-up for missed terminal output after temporary disconnections.
  * Improved terminal focus handling when reconnecting or restoring sessions.
  * Added compatibility for both modern sequence-aware and legacy terminal connections.

* **Bug Fixes**
  * Reduced duplicated, missing, or out-of-order terminal output during reconnects.
  * Improved terminal restoration across app restarts and session handoffs.
  * Prevented unnecessary screen resets while synchronizing terminal content.

### DIFF
--- a/apps/desktop/src/renderer/lib/persisted-keys/persisted-key-registry.test-data.ts
+++ b/apps/desktop/src/renderer/lib/persisted-keys/persisted-key-registry.test-data.ts
@@ -25,6 +25,7 @@ export const PERSISTED_KEY_REGISTRY: ReadonlyArray<
 		"src/renderer/lib/terminal/terminal-runtime.ts",
 		["terminal-buffer:*", "terminal-dims:*"],
 	],
+	["src/renderer/lib/terminal/terminal-seq-anchor.ts", ["terminal-seq:*"]],
 	[
 		"src/renderer/lib/terminal/terminal-buffer-gc.ts",
 		["terminal-buffer-persisted-at"],
--- a/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.ts
@@ -1,5 +1,7 @@
 export const TERMINAL_BUFFER_KEY_PREFIX = "terminal-buffer:";
 export const TERMINAL_DIMS_KEY_PREFIX = "terminal-dims:";
+/** Stream position paired with the buffer snapshot — see terminal-seq-anchor.ts. */
+export const TERMINAL_SEQ_KEY_PREFIX = "terminal-seq:";
 /** Single JSON map of terminalId -> epoch ms of the last persist/restore. */
 export const TERMINAL_PERSISTED_AT_KEY = "terminal-buffer-persisted-at";
 
@@ -60,13 +62,14 @@ export function removeTerminalStatePersistedAt(
 	writeIndex(storage, index);
 }
 
-/** Removes a terminal's buffer + dims keys, reporting how many existed. */
+/** Removes a terminal's buffer + dims + seq keys, reporting how many existed. */
 function removeTerminalState(storage: Storage, terminalId: string): number {
 	let removed = 0;
 	try {
 		for (const key of [
 			`${TERMINAL_BUFFER_KEY_PREFIX}${terminalId}`,
 			`${TERMINAL_DIMS_KEY_PREFIX}${terminalId}`,
+			`${TERMINAL_SEQ_KEY_PREFIX}${terminalId}`,
 		]) {
 			if (storage.getItem(key) !== null) {
 				storage.removeItem(key);
@@ -86,6 +89,8 @@ function collectTerminalIds(storage: Storage): Set<string> {
 			ids.add(key.slice(TERMINAL_BUFFER_KEY_PREFIX.length));
 		} else if (key.startsWith(TERMINAL_DIMS_KEY_PREFIX)) {
 			ids.add(key.slice(TERMINAL_DIMS_KEY_PREFIX.length));
+		} else if (key.startsWith(TERMINAL_SEQ_KEY_PREFIX)) {
+			ids.add(key.slice(TERMINAL_SEQ_KEY_PREFIX.length));
 		}
 	}
 	return ids;
--- a/apps/desktop/src/renderer/lib/terminal/terminal-runtime-registry.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-runtime-registry.test.ts
@@ -280,6 +280,7 @@ describe("terminalRuntimeRegistry eviction cleanup", () => {
 				serializeAddon: { serialize: () => "serialized scrollback" },
 				lastCols: 120,
 				lastRows: 32,
+				gate: { pending: 0 },
 			},
 			transport: {},
 			linkManager: null,
--- a/apps/desktop/src/renderer/lib/terminal/terminal-runtime-registry.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-runtime-registry.ts
@@ -23,11 +23,16 @@ import {
 	selectRuntimesToEvict,
 } from "./terminal-runtime-eviction";
 import {
+	loadPersistedSeqAnchor,
+	persistSeqAnchor,
+} from "./terminal-seq-anchor";
+import {
 	type ConnectionState,
 	clearLogs,
 	connect,
 	createTransport,
 	disposeTransport,
+	getPersistableSeqAnchor,
 	reconnect,
 	sendDispose,
 	sendInput,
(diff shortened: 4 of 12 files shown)


Kitenite:
Review triage (CodeRabbit + cubic), addressed in `4a7494d08` and earlier commits:

**Fixed**
- Stale-anchor hazards: a failed `terminal-seq:*` write now removes the key; detach/release persist the anchor only when the snapshot write succeeded and the parser is idle; a connection that delivers bytes without ever sending `synced` (host-downgrade skew) invalidates the anchor on close. All three degrade to the safe reanchor path instead of risking skip/duplicate on reattach.
- Repaint nudge at `MIN_TERMINAL_ROWS` now toggles a row upward instead of silently skipping.
- Explicitly empty `?seq=` is treated as malformed → safe reanchor (absent still means legacy).
- Test harness races (`lastSynced` read before `synced` arrived) fixed via `waitSynced()` everywhere; suite wired into `test:e2e` (which was itself broken on main — strip-types vs parameter properties — and now runs under the tsx loader).

**Declined, with reasons**
- *"Reanchor must not send preamble/notice"*: the mode preamble is invisible state restoration a rebuilt xterm needs on every attach (pre-existing legacy behavior), and the restored notice only fires on the respawn path where its whole purpose is separating stale scrollback from the fresh shell — reanchor's invariant is "never destroy/overwrite", which additive synthesized bytes don't violate.
- *"Suppress legacy replay when a restored anchor exists (old host)"*: against a pre-seq host, snapshot + FIFO replay is the contiguous legacy contract; suppressing it would lose the detached-window bytes old hosts can only deliver that way.
- *"Single oversized frame can exceed the 2 MiB ring cap"*: daemon reads are socket-sized (≲64 KB); keeping ≥1 chunk mirrors the existing FIFO idiom.

PR #6300, opened 2026-08-09
feat(host-service,desktop): re-assert client focus state to the PTY on terminal attach

Kitenite:
## Summary

Stacked on #6298 (uses its `synced` message; retarget to `main` after it merges).

A revealed/reattached terminal pane can leave the running TUI (Claude Code enables focus reporting, mode 1004) with a **stale focus state** — dimmed/unfocused UI in a pane the user is actively typing into.

## Why the "VS Code mechanism" alone is not enough — verified against sources and live

VS Code has **no side-channel for focus** (verified: zero hits for `sendFocus`/`1004`/`\x1b[I` in `terminalInstance.ts`, `ptyService.ts`, `terminalProcess.ts`). It relies on:
1. `@xterm/addon-serialize` emitting `\x1b[?1004h` in replayed state, and
2. xterm.js ≥4.15 (PR xtermjs/xterm.js#3506) self-reporting the *current* focus when an app enables 1004.

Our mode-preamble replay already triggers that exact mechanism — but it's **race-prone in a pane-parking architecture**, and we reproduced the failure live: on a parked-runtime rebuild of a *focused* pane, the self-report read the focus class before pane focus settled and delivered `\x1b[O` with no corrective `\x1b[I` ever following. The program is left wrong until the user clicks away and back. VS Code never hits this because its revives happen at window reload, when nothing is focused yet.

The right precedent is **tmux**, which treats attach as an explicit focus edge: on client attach it writes `\033[I`/`\033[O` into panes, gated on the app having enabled the mode (`MODE_FOCUSON`, `window_pane_update_focus()` in `window.c`).

## How it works (tmux's model, race-free)

- **Renderer**: on `synced`, flush the coalescer (the preamble frame precedes `synced`) and queue `{type:"focus", focused}` behind an empty `terminal.write("")` callback — guaranteeing it reaches the host **after** the preamble parse and any xterm self-report it triggered. The renderer's actual state always gets the last word.
- **Host**: forwards `\x1b[I`/`\x1b[O` to the PTY **only when the ModeTracker says the program enabled mode 1004** — plain shells never see stray escapes.
- Version skew safe: old hosts ignore the unknown message; old renderers never send it.

## Verification

- Node regression test: focus escapes reach the program **iff** it enabled 1004 (both polarities asserted).
- Live CDP on the dev desktop, OS window genuinely focused (AppleScript):
  - **Pre-fix build**: eviction rebuild of a focused pane → program received `\x1b[O` only, left wrong (the reported bug, reproduced).
  - **Fixed build**: same journey → program converges on `\x1b[I` (raw byte capture: `^[[I` last); a real window blur afterwards delivered a truthful `^[[O`.
- Full suite green (3/3 seq tests, host-service 1065, desktop 319), lint + typecheck clean.

---
## Summary by cubic
Reasserts client focus to the PTY on attach, tracks live focus changes, and aggregates focus across panes. Fixes stale or clobbered focus after reattaches and in duplicate panes, so TUIs with focus reporting get the correct state without extra clicks.

- **Bug Fixes**
  - `desktop`: on `synced`, flush the preamble and queue a `{type:"focus"}` behind an empty `terminal.write("")` so it lands after any xterm self-report; also send focus on textarea focus/blur to keep the host’s declared state current (listeners removed on teardown).
  - `host-service`: accept `focus` messages and forward `\x1b[I`/`\x1b[O` only when mode 1004 is active via `ModeTracker.isFocusReportingActive()`.
  - `host-service`: track focus per socket and write the aggregate (any focused client) to the PTY; recompute on client disconnect/error so unfocused duplicates can’t clobber a focused pane.
  - tests: add seq tests for 1004 gating and multi-client aggregate semantics.

<sup>Written for commit 0dcb5f60a0665aabe5c946d8cf6eca75e3913bf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added terminal focus-state reporting over WebSocket connections.
  * Terminal applications can receive focus-in and focus-out events when focus reporting is enabled.
  * Focus state remains synchronized across multiple connected terminal views.

* **Bug Fixes**
  * Improved focus synchronization when views connect, disconnect, lose focus, or encounter connection errors.
  * Prevented focus events from being sent when focus reporting is disabled.

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/terminal-ws-transport.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-ws-transport.ts
@@ -640,6 +640,13 @@ function attachSocketListeners(
 			transport._bytesSinceAttach = false;
 			setConnectionState(transport, "open");
 			sendResize(transport, terminal.cols, terminal.rows);
+			// Re-assert current keyboard focus so the running program's focus
+			// state can't stay stale across the reattach. In-band xterm focus
+			// reports can't cover this: a rebuilt xterm learns focus-reporting
+			// mode from the preamble only after its focus has already settled,
+			// and a focus-out sent while disconnected never arrived. The host
+			// forwards it only when the program enabled mode 1004.
+			sendFocusState(transport);
 			return;
 		}
 
@@ -799,6 +806,17 @@ export function getPersistableSeqAnchor(
 		return { ...transport.seqAnchor };
 	}
 	return null;
+}
+
+function sendFocusState(transport: TerminalTransport) {
+	const socket = transport._socket;
+	if (!socket || socket.readyState !== WebSocket.OPEN) return;
+	const textarea = transport._terminal?.textarea ?? null;
+	const focused =
+		textarea !== null &&
+		document.hasFocus() &&
+		document.activeElement === textarea;
+	socket.send(JSON.stringify({ type: "focus", focused }));
 }
 
 export function sendResize(
--- a/apps/desktop/src/renderer/lib/terminal/terminal-ws-transport.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-ws-transport.ts
@@ -640,13 +640,6 @@ function attachSocketListeners(
 			transport._bytesSinceAttach = false;
 			setConnectionState(transport, "open");
 			sendResize(transport, terminal.cols, terminal.rows);
-			// Re-assert current keyboard focus so the running program's focus
-			// state can't stay stale across the reattach. In-band xterm focus
-			// reports can't cover this: a rebuilt xterm learns focus-reporting
-			// mode from the preamble only after its focus has already settled,
-			// and a focus-out sent while disconnected never arrived. The host
-			// forwards it only when the program enabled mode 1004.
-			sendFocusState(transport);
 			return;
 		}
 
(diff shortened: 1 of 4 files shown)


Kitenite:
Review triage + final verification:

**Fixed**
- Multi-client focus clobber (cubic P1): the session now tracks declared focus per socket and writes the **aggregate** to the PTY (tmux's client-focus ownership model), recomputing when a focused client disconnects. Written unconditionally rather than edge-triggered so in-band drift self-corrects.
- Found during live CDP verification of that fix: a pane's live focus travelled only in-band (xterm's own reports), invisible to the aggregate — the transport now also declares textarea focus/blur changes, keeping the host's view current.
- Multi-socket regression test added (unfocused duplicate attach must not yield focus-out while a focused client is attached).

**Live CDP verdicts (real dev app, real PTY probe with raw-byte capture)**
- Focused pane + second socket declaring `focused:false` → PTY receives the aggregate `^[[I`, not `^[[O` (pre-fix capture showed the clobbering `^[[O`).
- Unfocused-socket departure leaves the program untouched; focused-socket departure hands over focus-out.
- Attach-time re-assert after an eviction rebuild converges to the true focus state (sequenced after the preamble parse, so xterm's racy self-report can't get the last word).
- Harness notes: OS window focus can't be held by a headless driver, so the final deterministic run shimmed `document.hasFocus` and dispatched the focus event synthetically — everything from the listener down (transport message, host gate on mode 1004, aggregate, PTY write, program observation) is the real path.

PR #6304, opened 2026-08-10
perf(desktop,host-service): enqueue workspace creation off pooled sockets and batch the host-service client

Kitenite:
Follow-up to #6299 (terminal create-on-attach). That PR fixed the worst victim of Chromium's 6-per-origin socket pool to host-service; this one reduces the pressure at its two biggest sources. Both changes are proven with before/after CDP measurements against the real dev app.

## Summary
- **`workspaces.create` no longer holds a pooled socket for the whole create.** New `workspaces.createEnqueued` validates, returns immediately, runs the real create in the background, and broadcasts a `workspace:create-settled` eventBus event carrying what the synchronous response used to (canonical workspace id + launched terminals/agents for the pane-layout seed). The desktop renderer switches to it; CLI/MCP/SDK/automations keep the synchronous `create` untouched.
- **The desktop's host-service tRPC client is batched again** — with `httpBatchStreamLink`, which streams each result as it completes, addressing the slowest-in-batch latency that motivated the original unbatching in #3879.
- **Per-case timeout bounds** on the create-flow mutations (enqueue 15s, session create 30s, settled-event backstop 10min with a row-existence check), plus the persisted-row assertion in the create-on-attach test (review follow-up from #6299).

## CDP evidence (real UI journeys, dev desktop app)

**Create hold time — the marquee number.** Same journey both times: New Workspace composer → demo-app project → prompt → submit, with all renderer→host requests timed via CDP Network events:

| | request | held socket for |
|---|---|---|
| before | `POST workspaces.create` | **3,710ms — the entire create** |
| after | `POST workspaces.createEnqueued` | **≤590ms** (max held across *all* 150 requests in the window) |

On demo-app the create is ~4s; on a monorepo it's minutes of `git worktree add` — all of it previously pinning 1 of 6 sockets. On relay-fronted hosts the synchronous create exceeded the relay's 30s exchange cap outright (`apps/relay/src/tunnel.ts:64`), so remote creates of large repos could not succeed at all; the enqueue variant fits comfortably.

**End-to-end UX verified after the change:** workspace appears in the sidebar instantly, navigation lands on it, the AI-generated branch name is applied, and the agent pane is seeded from the settled event and starts executing its prompt (screenshots captured). Scratch workspaces then destroyed cleanly through the same host.

**Batching, dashboard route (6 pollers active), 60s windows:** 72 → 66 requests/min, preflights 36 → 32, with same-tick batches visibly collapsing (`terminalAgents.list,pullRequests.getByWorkspaces` pairs). Honest read: the steady-state win is modest because independent poll intervals rarely share a tick — the real value is same-tick fan-outs (multi-host `useQueries`, boot bursts) and that it removes the per-call socket+preflight floor with no latency downside.

## How It Works
- Host: `createEnqueued` reuses the entire existing `create` procedure via a server-side caller (`createCallerFactory`) — zero changes to the 700-line create body. The settled event goes through a new typed `EventBus.broadcastWorkspaceCreateSettled`, following the existing lifecycle-event shapes.
- Renderer: `useWorkspaceCreates.submit` subscribes to `workspace:create-settled` for the client-minted id *before* enqueueing (no missed-event race), and keeps the existing optimistic flow untouched — sidebar row, empty pane seed, `failedWorkspaceCreates` record, `workspaceTransactions` pending UI all behave exactly as before, now driven by the event payload.
- Degraded paths: hosts without the procedure (version skew) fall back to the legacy synchronous create; a lost settled event (host restart mid-create, WS drop) hits a 10-minute backstop that checks whether the row exists before declaring failure, so a real workspace is never torn down over a missed event.

## Testing
- `bun run typecheck`, `bun run lint`
- `cd packages/host-service && bun test` — 1102 pass, 0 fail, including 3 new integration tests for `createEnqueued` (immediate return + settled event with persisted row; missing-id rejection; mid-create failure broadcasts `ok:false` with the error)
- CDP journeys above, on the built dev app with this branch's host-service bundle

## Known Limitations / Follow-ups
- The `WorkspaceCreatingState` progress UI is still estimate-based; the settled event now provides the terminal signal, and intermediate progress events (worktree add started, agents dispatched) are a natural next step on the same channel.
- Bulk create sites (`toast.promise` over `Promise.all(handles)`) work unchanged — each handle's `completed` now resolves from its event — but weren't individually re-verified beyond the single-create journey.
- Remaining raw-client mutations elsewhere (destroy, rename, etc.) still lack per-case bounds; incremental, per the per-case-bounds decision.

---
## Summary by cubic
Makes workspace creation async and brings back streaming batched tRPC calls to reduce host-service socket pressure and prevent relay timeouts. Adds safer timeout recovery so slow creates and sessions don’t get marked failed.

- **New Features**
  - Added `workspaces.createEnqueued`: validates, returns immediately, runs the real create in the background, and emits `workspace:create-settled` (canonical workspace id, terminals, agents).
  - Desktop switches to the enqueue flow; CLI/MCP/SDK stay synchronous. Sessions remain synchronous with a 30s timeout.
  - Timeouts: enqueue 15s; session create 30s; settle backstop 10m with a bounded row-existence probe. Older hosts fall back to legacy `create`.
  - Desktop tRPC client uses `httpBatchStreamLink` from `@trpc/client` to batch same-tick calls and stream results, reducing preflights and socket usage.
  - Event bus: new `broadcastWorkspaceCreateSettled`; renderer subscribes before enqueue. Types exported via `@superset/workspace-client`.

- **Bug Fixes**
  - Tri-state row probe with spaced retries for timeout recovery: avoids false “failed” creates/sessions; if only the row exists, return with an empty pane seed and rely on auto-adopt to heal panes.
  - A timed-out enqueue now waits for the settled event/backstop instead of failing fast; the probe is bounded and the backstop timer is cleared once settled.

<sup>Written for commit b197d2d484b21aecbabb59c395500bd0a2c39cb6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Workspace creation now returns immediately while completing asynchronously in the background.
  * Added completion updates for successful or failed workspace creation, including workspace, terminal, and agent details.
  * Added client support for receiving workspace creation updates.
  * Improved request batching while streaming individual results.

* **Bug Fixes**
  * Added timeout recovery and compatibility fallback for workspace creation.

* **Tests**
  * Expanded coverage for persisted terminal sessions and asynchronous workspace creation outcomes.

### DIFF
--- a/apps/desktop/src/renderer/lib/host-service-client.ts
+++ b/apps/desktop/src/renderer/lib/host-service-client.ts
@@ -1,5 +1,5 @@
 import type { AppRouter } from "@superset/host-service";
-import { createTRPCClient, httpLink } from "@trpc/client";
+import { createTRPCClient, httpBatchStreamLink } from "@trpc/client";
 import superjson from "superjson";
 import { getHostServiceHeaders } from "./host-service-auth";
 
@@ -20,7 +20,13 @@ export function getHostServiceClientByUrl(hostUrl: string): HostServiceClient {
 
 	const client = createTRPCClient<AppRouter>({
 		links: [
-			httpLink({
+			// Streaming batch link: same-tick calls share one HTTP request and
+			// one CORS preflight, but each result streams as soon as it's ready
+			// — no slowest-in-batch latency (the reason #3879 unbatched the old
+			// buffering httpBatchLink). All renderer clients share Chromium's
+			// 6-connections-per-origin pool with every other host-service
+			// request, so sockets are the scarce resource here.
+			httpBatchStreamLink({
 				url: `${hostUrl}/trpc`,
 				transformer: superjson,
 				headers: () => getHostServiceHeaders(hostUrl),
--- a/apps/desktop/src/renderer/stores/workspace-creates/useWorkspaceCreates.ts
+++ b/apps/desktop/src/renderer/stores/workspace-creates/useWorkspaceCreates.ts
@@ -1,9 +1,18 @@
+import {
+	getEventBus,
+	type WorkspaceCreateSettledPayload,
+} from "@superset/workspace-client";
+import { TRPCClientError } from "@trpc/client";
 import { useCallback } from "react";
 import { resolveHostUrl } from "renderer/hooks/host-service/useHostTargetUrl";
 import { useRelayUrl } from "renderer/hooks/useRelayUrl";
 import { authClient } from "renderer/lib/auth-client";
 import { electronTrpc } from "renderer/lib/electron-trpc";
-import { getHostServiceClientByUrl } from "renderer/lib/host-service-client";
+import { getHostServiceWsToken } from "renderer/lib/host-service-auth";
+import {
+	getHostServiceClientByUrl,
+	type HostServiceClient,
+} from "renderer/lib/host-service-client";
 import { getHostServiceUnavailableMessage } from "renderer/lib/host-service-unavailable";
 import { electronTrpcClient } from "renderer/lib/trpc-client";
 import { useCollections } from "renderer/routes/_authenticated/providers/CollectionsProvider";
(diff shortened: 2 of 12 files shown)


Kitenite:
## Full surface tour + stress verification (CDP, real UI journeys)

Beyond the before/after measurements in the description, every surface this PR touches was driven end-to-end on the built dev app:

| Surface | Journey | Result |
|---|---|---|
| Composer create (enqueue path) | New Workspace → project → prompt → submit | Pane seeded from settled event, agent running, AI branch name applied |
| **Session create** (project-less, stays synchronous+bounded) | SESSIONS `+` → prompt → submit | Navigated, chat + terminal panes live |
| **Stress: 3 rapid overlapping creates** | 3 back-to-back submits without waiting | All 3 settled independently, distinct AI branches with dedupe (`stress-test-reply-exit`, `-2-`, `-3-`), no event cross-talk |
| **Create under saturated socket pool** | All 6 sockets pinned → submit → pins released at t+5s | Enqueue queued 5.1s in the browser pool, drained on release, settled, workspace completed — no failure, no stuck state |
| Reload mid-create (settled event lost) | Submit → `Page.reload` at t+1.5s | Row healed via `workspace:changed`, AI name applied, agent session auto-adopted into a pane — full self-heal |
| Duplicate-branch enqueue | Integration test | Settles with `canonicalWorkspaceId` + `alreadyExists: true` |
| Batched-client consumers | Dashboard polls (measured), Settings → Agents | Render clean, same-tick batches visible |

Suites under their sanctioned runners: host-service 1102 pass (incl. 4 enqueue integration tests), desktop 2650 pass, workspace-client 13 pass.

All cubic/CodeRabbit findings addressed in follow-up commits (bounded backstop probe, cleared backstop timer, enqueue-timeout fall-through instead of teardown, session-timeout spaced row probes, tri-state probe so transport failures don't read as "row absent", correlation id in the host-side failure log). One deliberate trade-off documented in-code: event-loss recovery seeds an empty pane layout — launches aren't recoverable without the event, and session auto-adopt heals the panes on open (verified in the reload journey above).

All scratch workspaces created during verification were destroyed (worktrees + branches).

PR #6306, opened 2026-08-10
feat(host-service,desktop): archive-first workspace delete so rows leave the sidebar instantly

Kitenite:
## Summary
- Deleting a workspace now tombstones it **first** — before the git preflight and the teardown script — so the row leaves the sidebar/board ~200 ms after confirm instead of waiting seconds for teardown. Any failure un-archives, so the row reappears instead of getting stuck half-deleted.
- Delete consent is now **two flags**: `force` (git-destructive consent: skip dirty preflight, forceful worktree/branch removal) and `skipTeardown` (abandon the teardown script — set only by the teardown-failed retry). Previously one flag, so a warned "Delete anyway" (e.g. any edited tracked file) silently skipped the user's teardown script.
- The v2 delete dialog is now a single globally-mounted instance (`DeleteWorkspaceMount` + latched `useDeleteWorkspaceIntent`); all entry points (sidebar row, board card, palette, ⌘⇧⌫, missing-worktree screen) request through the store.
- New reference doc: `apps/desktop/docs/WORKSPACE_DELETE_LIFECYCLE.md` (pipeline order, consent flags, failure-mode table, renderer contract).

## Why / Context
#6296 deliberately placed the archive commit *after* teardown because a blocking teardown failure re-opens the delete dialog for force-retry, and that dialog was mounted under the workspace's row — archiving first would drop the row and unmount the dialog mid-prompt. The cost was a visible delay: the row sat in the sidebar for the full preflight + teardown duration. This PR reverses the ordering to make delete feel instant, and removes the original blocker by hoisting the dialog out of the row entirely.

The consent split came out of QA on this branch: editing the teardown script inside the worktree (to test failure) dirtied it, which flipped the confirm to "Delete anyway" → `force` → teardown skipped — the workspace deleted cleanly and the failure path never ran. Git consent and teardown consent are different questions; they're now separate flags.

## How It Works
**Host** (`workspace-cleanup.ts`): `archiveLocalWorkspace` moves to step 0, right after the main-workspace check. Preflight, teardown, and the destroy phases now run inside one try/catch whose catch un-archives. `destroy` takes `skipTeardown` (default false) and derives `teardownMode` from it alone — `force` no longer touches teardown. Telemetry still fires only at the true commit. The startup reconciler (`runArchivedWorkspaceReconcile`) already assumed mark-first and resumes crash-interrupted deletes with best-effort teardown — no changes needed there.

**Renderer**: since the row unmounts the instant a destroy starts, `DashboardSidebarDeleteDialog` must never be row-mounted. The intent store gains an `open`/latch split: closing the dialog only flips `open` while the target stays mounted, so the in-flight destroy can re-open it as the teardown-failure pane. `run()` now takes `{force, skipTeardown}`: warned confirm → `{force: true}`, teardown-failed retry → `{force: true, skipTeardown: true}`, dirty-race silent retry stays git-only. Bulk delete threads the same split (`forceTeardownFailures` is the only skip path).

## Manual QA Checklist (verified via CDP against the dev app, real UI input, screenshots captured)
- [x] Happy path (3 s teardown script): row removed **276 ms** after confirm; teardown started ~1 s later and ran entirely after removal; row never returned; worktree removed; tombstone `reason: "deleted"` written before the removal broadcast
- [x] Blocking teardown failure (`exit 1`): row removed at 198 ms → reappeared ~1.1 s later (un-archive) with "Teardown exited with code 1" pane open showing the script's output tail — pane survives the row unmount
- [x] Force-retry from the pane: row gone instantly, teardown skipped (no marker file written), workspace destroyed
- [x] Cancel from the pane: workspace fully alive (`archivedAt: null`), row back, worktree intact
- [x] Dirty-worktree race (dirtied after inspect, before confirm): archive → CONFLICT → un-archive → silent force retry → deleted; row blips ~100 ms; no error surfaced
- [x] Concurrent destroy mid-flight: typed CONFLICT with `deleteInProgress` cause (toast path, no retry)
- [x] Main workspace: BAD_REQUEST, never archived
- [x] Deleting the actively-viewed workspace: navigates away immediately; failure pane still opens on the new route
- [ ] Post-split re-verify: warned "Delete anyway" (dirty worktree) runs teardown; failing teardown restores the row and opens the retry pane (needs a dev-app restart to pick up the new host-service — CDP pass above predates the flag split; recipe: `echo '{ "teardown": ["sleep 3 && exit 1"] }' > <worktree>/.superset/config.local.json`, gitignored so the worktree stays clean)

## Testing
- `bun run typecheck` (+ direct `tsc --noEmit` on desktop and host-service)
- `bun run lint`
- `bun test src/renderer/routes/_authenticated/_dashboard` (262 pass) and host-service reconcile/teardown suites (13 pass)
- No existing tests asserted the pipeline ordering; behavior verified end-to-end via CDP as above

## Design Decisions
- **Global dialog mount instead of optimistic row-hiding**: hiding the row renderer-side would still unmount a row-local dialog and adds a second source of truth; moving the commit point host-side makes every client instant, and the dialog hoist fixes a latent bug where deletes from the layout hotkey/missing-worktree screen could already lose the force-retry prompt.
- **Archive before *preflight*, not just before teardown**: preflight's git status read can be slow on large worktrees; the conflict path un-archives and the renderer's silent force-retry re-archives, so the only cost is a ~100 ms row blip in a rare race.
- **Two consent flags instead of dialog copy**: "teardown will be skipped" text on the warned confirm would keep the surprising behavior and just document it; a dirty worktree is a git question and shouldn't decide whether cleanup scripts run.

## Known Limitations
- Repos with **no remote** always read as having unpushed commits (`rev-list HEAD --not --remotes`) → always-warned confirm. Since the split this only skips preflight; teardown still runs.
- Version skew with remote hosts: an old host ignores `skipTeardown` and still maps `force` → skip; a new host + old renderer re-runs teardown on the failure retry (pane re-opens; Cancel escapes). Desktop and its local host-service ship together, so this only affects mixed-version remote hosts.
- A host crash between archive and teardown means the reconciler finishes the delete with best-effort teardown on next boot (same as any crash-interrupted delete).

## Rollback
Revert the two commits. The tombstone schema is untouched (#6296's migration 0020); only ordering, consent flags, and dialog mounting changed, so old and new builds interoperate on the same host.db.

## Summary by CodeRabbit

- **New Features**
  - Workspace deletion now removes workspaces from navigation immediately while cleanup continues.
  - Sidebar, context-menu, and missing-worktree deletion actions use a consistent confirmation experience.
  - Bulk deletion supports retrying cleanup failures while preserving safety controls.

- **Bug Fixes**
  - Failed deletions restore workspaces for retry.
  - Deletion dialogs preserve the selected workspace when closed.
  - Improved navigation handling during cleanup and application interruptions.

- **Documentation**
  - Added documentation covering workspace deletion behavior, recovery, concurrency, and UI timing.

### DIFF
--- a/apps/desktop/docs/WORKSPACE_DELETE_LIFECYCLE.md
+++ b/apps/desktop/docs/WORKSPACE_DELETE_LIFECYCLE.md
@@ -1,0 +1,84 @@
+# Workspace Delete: Lifecycle, Edge Cases, Failure Modes
+
+Covers the v2 local workspace delete pipeline (`workspaceCleanup.destroy` in
+`packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts`)
+and its renderer contract. Sessions delete through the same pipeline (they are
+workspaces with no project row).
+
+## Pipeline order (archive-first)
+
+| Step | What | On failure |
+|------|------|-----------|
+| 0 | **Archive (commit point)** — tombstone `archivedAt`/`archiveReason`; broadcast drops the row from every list | — |
+| 1 | Preflight dirty-worktree check (skipped when `force`) | CONFLICT → un-archive |
+| 2 | Teardown script (per `teardownMode`) | blocking: PRECONDITION_FAILED → un-archive |
+| 3 | Local cleanup: PTYs, worktree removal | throw → un-archive |
+| 4 | Legacy cloud delete (best-effort, skipped for sessions) | warning only |
+| 5 | Optional branch delete | warning only |
+| 6 | Caches | warning only |
+
+**The archive commit is deliberately FIRST** — before the (potentially slow)
+git preflight and teardown script — so the row leaves the sidebar/board the
+moment the user confirms (~200 ms measured, broadcast-bound). Any failure in
+steps 1–3 un-archives, so the row *reappears* instead of being stuck
+half-deleted. Telemetry fires only after step 6 succeeds.
+
+## Force semantics
+
+`force: true` (the "Delete anyway" paths) skips preflight, double-forces the
+worktree removal, uses `-D` for branch delete — **and skips teardown
+entirely** (interactive contract: force-retry is consent to abandon a failed
+teardown). Non-interactive callers (CLI/SDK/MCP via `workspace.delete`) use
+`teardownMode: "best-effort"` instead: teardown always runs, failures degrade
+to warnings (#6174).
+
+Consequence worth knowing: if the confirm dialog showed *any* warning
+(uncommitted changes / unpushed commits), confirming passes `force: true`, so
+**teardown does not run for warned deletes**.
+
+## Failure modes
+
+| Scenario | Behavior |
+|----------|----------|
+| **Blocking teardown failure** | Row vanishes on confirm → teardown fails → row reappears (un-archive) and the globally-mounted dialog re-opens as "Teardown exited with code N" with the script's output tail. "Delete anyway" force-retries (teardown skipped); Cancel leaves the workspace fully alive. |
+| **Dirty-worktree race** (clean at dialog-open, dirty by destroy time) | Archive → preflight CONFLICT → un-archive → renderer silently retries with `force: true` → re-archive → deleted. The row blips back for ~100 ms; no error is surfaced. The retry is only for `conflict` — never for `in-progress`. |
+| **Indeterminate preflight** (git status timeout/pool failure) | Fails closed (INTERNAL error, un-archive) rather than skipping the dirty check on a destructive path. Retry usually succeeds; `force` is the escape hatch. |
+| **Worktree removal fails** (still registered after `git worktree remove`) | Throw → un-archive; workspace stays visible and retryable rather than orphaning disk state. |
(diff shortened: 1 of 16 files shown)


Kitenite:
Addressed the bot review findings in d92ad47a1:

- **Unscoped dialog close/reopen race** (CodeRabbit major + cubic P1): fixed — `setOpen`/`close` on the intent store now take the workspaceId, so a settle from a superseded delete no-ops instead of touching the new target's dialog.
- **Failing `Test` check**: real — the two workspace-cleanup integration tests asserted the old conflated contract (`force` skips teardown). Updated to the split contract and added a `skipTeardown` skip-contract test; the suite now proves host-side that `force` alone runs teardown.
- **Doc/comment precision** (CodeRabbit minor + cubic P3): fixed — only thrown step 1–3 failures un-archive (4–6 degrade to warnings), and the always-on git force flags are no longer attributed to `force`.
- **`onCloseWorkspaceClick` for main workspaces** (CodeRabbit major): not a change in behavior — `DashboardSidebarExpandedWorkspaceRow` renders the close/X (delete) affordance only for non-main rows, same gating as before the refactor.
- **Server-side enforcement of the skipTeardown retry-only contract** (CodeRabbit heavy-lift): deliberately skipped — the host tRPC surface is same-trust-domain (our renderer/CLI hold the bearer secret); the flag is consent plumbing, not a security boundary, and serializing "a teardown must have failed first" server-side adds state for no attacker it would stop.

PR #6309, opened 2026-08-10
fix(host-service,desktop): broken import flow, stop v1 importer from duplicating already-imported projects

distroinfinity:
## Problem

Using the Settings → Experimental → "Import from v1" wizard, the **"Bring over your projects"** step had three compounding issues:

1. Pressing **Import all** ran to completion but nothing visibly changed — every row still offered "Import", so it was natural to press it again (and again).
2. Reopening the importer showed no memory of what was already imported.
3. **Every import created a brand-new duplicate project.** After a few Import-all presses I ended up with 4 copies of each of my 5 projects in the sidebar.

image

image

## Root cause

Importing a v1 project creates the v2 project **locally only** (`project.create {kind:"importLocal"}` — "fully local, the cloud is never involved", by design). But the importer decides row state via `project.findByPath` with `walkAllRemotes: true`, and that branch had a "stale local link" probe: it found the local row, asked cloud `v2Project.get(id)`, got `NOT_FOUND` (the cloud never knew this project), concluded the row was stale, and **silently dropped it from the candidates**. With no `local-path` candidate, `decideProjectImport` returned `import` — so the wizard re-imported, and `createFromImportLocal` unconditionally inserted a fresh row.

The probe was written for "cloud project deleted from another device", but that inference contradicts the delete saga's own principle in the same file ("**Local is reality** — the local deletes are the source of truth"): a surviving local row means the project exists on this device, regardless of what the cloud knows.

Confirmed empirically in a local dev stack: with 4 rows for a repo in host.db, live `findByPath {walkAllRemotes:true}` returned `{"candidates": [], "cloudErrors": []}`, with a cloud `NOT_FOUND` logged per probe, and each Import-all press added exactly one more row per project.

## Fix

**1. `findByPath`: a local-DB row is authoritative** (host-service). A row keyed by the repo's resolved git root short-circuits in *both* modes — exactly like the non-`walkAllRemotes` branch always has. The staleness probe and the `staleLocalLink` field are deleted (the field had zero consumers repo-wide). The cloud remote-walk still runs when no local row exists (candidate discovery for linking is unchanged).

**2. `createFromImportLocal` is idempotent on repo path** (host-service). Importing a repo that's already a project returns the existing project (with its ensured main workspace) and reports the new additive `created: false` on `CreateResult` — duplicates become impossible from any caller. The reused row is left untouched: no rename, and the renderer now gates the v1 color/icon carry on `created !== false` so a re-import can't overwrite v2 customizations (older hosts without the field keep today's behavior).

**3. Wizard adopts the workspaces-page pattern** (desktop). Per-row import state lives in one page-level map (like `adoptStates`); the header button reads `Import all · N`, shows `Importing i/n` while running, and disappears when nothing is pending; rows are disabled during the batch; failed rows show inline errors with Retry and stay pending; `needs-relocate` lives in shared state so it surfaces its confirm UI instead of being re-queued. Lookup failures are kept out of the import-retry path (retry refetches discovery; imports never run without discovery data).

## Testing

- 12 new unit tests: caller-level `findByPath` tests (local row survives with cloud unaware; cloud never called on a local hit; remote walk preserved when no local row), `createFromImportLocal` idempotency (same id, one row, customizations preserved), appearance gating, and the wizard's pending-selection + row-action matrix (batch disabling, retry semantics). Existing harness idioms only (bun:test, in-memory sqlite + real migrations, real temp git repos).
- `bun run lint`, `bun run typecheck`, full targeted test suite green.
- Manual end-to-end in a local dev stack, before and after: pre-fix reproduced the exact duplicate cascade (row counts 0→1→2→3→4 across four Import-all presses); post-fix, a fresh profile imports all 5 projects with one press (rows flip to Linked as each lands, button empties out and disappears), re-pressing/reopening/refreshing re-imports nothing, and the DB holds exactly one row per repo path.

image

## Behavior notes

- A project whose cloud row was deleted from another device now reports "already imported" instead of offering a re-import — consistent with local-is-reality (the local project, workspaces, and sidebar entry still exist on the device).
- `staleLocalLink` is removed from the `findByPath` wire shape (verified unused). `created` on `CreateResult` is additive.
- Design/implementation docs for this fix are included under `plans/done/`.

---
## Summary by cubic
Fixes the v1 “Bring over your projects” step so already-imported projects are detected and duplicates are no longer created. Makes import idempotent in `packages/host-service` and updates the `apps/desktop` wizard to show clear Import All progress, guard retries during batch, and preserve user customizations.

- **Bug Fixes**
  - `host-service`: `project.findByPath` now treats a local DB row as authoritative in both modes; removes the cloud staleness probe and `staleLocalLink`. Still walks remotes only when no local row exists.
  - `host-service`: `createFromImportLocal` is idempotent on repo path; reuses the existing project and returns `created: false`. Prevents duplicate rows and gates v1 appearance carry on `created !== false` to preserve v2 customizations.
  - `desktop`: Import Projects page uses a shared per-row state map. “Import all · N” shows counts and in-flight progress, disables rows and Retry during batch, surfaces inline errors with Retry, keeps “needs relocate” in shared state, and keeps lookup failures out of import retries.

<sup>Written for commit c5333a84290cb4e85a6fbab8831bce05b41cdfa6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6309?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## New Features

- Importing projects now reuses existing projects with the same repository path instead of creating duplicates.
- The import screen shows clearer project states, pending counts, and appropriate actions for linking, relocating, retrying, or importing.
- “Import all” processes only actionable projects and provides safer progress handling.

## Bug Fixes

- Existing project customizations, including appearance settings, are preserved during repeated imports.
- Improved handling for lookup and import failures, including retry support.


distroinfinity:
hey @Kitenite 
I hit this while moving my own machine to v2. Nothing changed on screen when I pressed Import all, so I pressed it a few more times and ended up with 4 copies of every project in my sidebar. Screenshots are in the description.

I dug in and found the cause in findByPath. Projects imported from v1 only exist in the host's local db, the cloud never hears about them. The stale link check asks the cloud about the local row, gets NOT_FOUND, and drops it. So the importer always thinks nothing was imported and creates a new project on every pass.

The fix treats the local row as the source of truth, which is the same rule the delete path already follows. It also makes importLocal create reuse an existing project for the same repo path, and gives the projects page the same Import all behavior the workspaces page already has, since that one worked well.

I tested on a local dev stack before and after the change, including a run with a fresh profile. Test notes are in the description. If the stale link check was covering a case I missed, tell me and I will rework it.

Kitenite:
thnka you so much @distroinfinity im testing this now

PR #5790, opened 2026-07-19
fix(desktop): disarm leaked TUI input modes on live kill in v2 terminals

Kitenite:
## What & why

A TUI in a v2 workspace terminal (mastracode/pi-tui, Claude Code, …) arms input-reporting modes on startup — the **kitty keyboard protocol**, mouse tracking, focus reporting — and disarms them on clean exit. Killed uncleanly (`kill -9`, crash, OOM) **while its pane is attached**, it never writes the restore sequences, so the shell that reclaims the pty inherits them. For kitty this is severe: every keystroke reaches the shell CSI-u encoded (`Ctrl+C` → `^[[99;5u`, `whoami` → `^[[119;1:3u…` → `command not found`), so the pane is unusable by keyboard until `reset`.

Why it slipped through: v2 workspace terminals stream PTY output straight to the renderer xterm and don't route through the host-side `HeadlessEmulator`, so its foreground-reclaim never runs for them. The existing renderer disarm (`disarmStaleInputModes`) only fires on cold restore and terminal restart — not on a TUI killed while the pane stays attached. (xterm arms kitty because the terminal advertises `vtExtensions: { kittyKeyboard: true }`.)

## The fix

A renderer-side reclaimer installed on every xterm in `createTerminal`:

- **`shared/leaked-input-mode-reclaim.ts`** — a transport-agnostic core (`createLeakedInputModeReclaimer`: `noteArm` / `noteShellReady` / `collectDisarm`) that owns only the decision logic: a **shell-owned epoch** (modes armed before the first prompt marker belong to shell init and are never reclaimed) and a **mark-then-recheck flush** (a mode re-armed before the flush keeps its state). Kept separate so a host-side surface can reuse it later.
- **`renderer/lib/terminal/terminalInputModeReclaimer.ts`** — a thin xterm-parser adapter: handlers for kitty (`CSI >/=/< u`), mouse (`?1000/1002/1003`), focus (`?1004`) and the `OSC 777;superset-shell-ready` marker feed the core; the disarm is written back on a microtask.

**Marker choice:** reclaim keys on `OSC 777;superset-shell-ready` (emitted only by Superset's shell wrappers), **not** the co-emitted FinalTerm `OSC 133;A` — `133;A` is also produced by third-party shell integrations and forwarded by tmux, so disarming on it would clear a live tmux's own modes.

**Guard:** the reclaimer only acts at a shell prompt (no TUI owns the foreground then) and re-checks at flush, so a live/suspended/racing TUI keeps its modes.

[...]

PR #6345, opened 2026-08-11
fix(desktop): reclaim leaked input modes after terminal session restore

### LINKED ISSUE #6308: Random control characters in terminal when re-launching, instead of claude code re-running
### Where did this happen?

Desktop app (macOS)

### Version

1.19.0 

### What happened?

When I re-launch superset after a computer reboot, there are a lot of random control characters in all of my "claude" windows & claude is not re-running automatically.

### Steps to reproduce

This might be related to my terminal setup, although I don't have anything too fancy - I can't rule this out. 

All I do is get superset killed (usually due to a software update & computer reboot).

Once I relaunch superset, in every **claude code** window I get random shell control characters, e.g. when I move my mouse around the terminal & obviously, claude does not re-launch itself. 

Regular "terminal" windows are not affected. 

My shell is `zsh 5.9 (arm64-apple-darwin25.0)` and I'm using powerlevel10k for my prompt (https://github.com/romkatv/powerlevel10k) 

Image

Image

### LINKED ISSUE #6343: Terminal emits mouse event characters after IDE session restore
### Where did this happen?

Desktop app (macOS)

### Version

1.20.2

### What happened?

When reopening the IDE, it attempts to restore previously running terminal and Claude sessions.

Not only are the claude sessions not properly restored, but after the attempt, interacting with the IDE using the mouse causes unexpected characters to be written into the terminal. Moving the mouse, clicking, or otherwise interacting with the UI appears to generate terminal input containing sequences such as 34M65;56, 35M35, etc.

The terminal can rapidly fill with these characters simply from mouse movement, without any keyboard input.

This appears to be related to terminal mouse tracking being left enabled or incorrectly restored when resuming the previous terminal/Claude session.

### Steps to reproduce

- Open the IDE and start a terminal/Claude session.
- Leave the session running and close the IDE.
- Reopen the IDE.
- Allow the IDE to automatically restore/resume the previous terminal and Claude session.
- Move the mouse over the IDE and/or click within the interface.
- Observe that mouse movements and clicks cause character sequences to be printed into the restored terminal.

**Expected:** Mouse interactions should not generate terminal input after restoring a session. Claude sessions are restored.

**Actual:** Mouse events are interpreted as terminal input and produce large amounts of seemingly random character sequences.

Image

Video: see from 27 seconds in (you can see some characters from when I opened the IDE before this)
https://supercut.ai/share/claimer/Neph0ldteW3wHIy4Xafgiy

andyst-dev:
Fixes #6308
Fixes #6343

## Problem

When a TUI (Claude Code) is killed while attached — e.g. by quitting the IDE and rebooting — it never writes its DECSET restore bytes. On reopen, the desktop cold-restores the persisted terminal snapshot into a **fresh** xterm. Two things then conspire to leak input modes into the prompt:

1. **The persisted snapshot carries the dead session's modes.** `persistBuffer` serialized with `serializeAddon.serialize({ scrollback })`, and xterm's SerializeAddon appends mode-restore DECSETs (`\x1b[?1002h` / `?1003h`, `?1004h`, kitty keyboard) after the buffer content by default. So the localStorage snapshot keeps "mouse tracking on".
2. **The leaked-input-mode reclaimer misclassifies them as shell-owned.** The reclaimer (`shared/leaked-input-mode-reclaim.ts`) treats any mode armed *before the first `OSC 777;superset-shell-ready` marker* as armed by shell init, and never reclaims it. On a restored runtime the replay bytes (and the host reattach preamble, `#6343`) land in the fresh xterm before any marker, so the leaked modes are exempted forever.

**Result:** mouse movement/clicking sprays mouse-report escape sequences (`34M65;56`, `35M35`, …) into the shell prompt as visible text; the terminal fills up with junk and never self-heals.

## Fix

Two-part, both in the renderer terminal path:

1. **`persistBuffer` now serializes with `excludeModes: true`** (`terminal-runtime.ts`). A persisted snapshot is scrollback belonging to a session that may well be dead; replaying its input modes into whatever pty the pane lands on next is wrong by construction. This mirrors VSCode's documented behavior (see `plans/20260507-terminal-mode-replay.md:24`). Live mode continuity across a *reattach* is already handled server-side by the host's `buildPreamble()`, so stripping modes here does not regress mode replay.
2. **`noteRestoreAttach()` on the reclaimer** (`shared/leaked-input-mode-reclaim.ts`), called from `createRuntime` when a runtime is seeded or restored. This tells the decision core that pre-marker arms are **program-sourced** (a dead session's snapshot / sibling serialize / host preamble), not shell init — so they stay reclaimable at the next shell-ready marker instead of being exempted forever. It also undoes the exemption for any mode that arrived before the signal.

Because `persistBuffer` now strips modes, freshly-written snapshots won't re-arm the bug; `noteRestoreAttach` covers already-poisoned snapshots already sitting on user profiles, plus the host reattach-preamble path.

## Tests

- `terminalInputModeReclaimer.test.ts`: 3 new cases — restored attach reclaims carried-over modes; modes armed *before* `noteRestoreAttach` is signalled are still reclaimed; a live re-arm before the flush still suppresses the disarm. (All 12 pass.)
- `terminal-buffer-gc.test.ts`: 3 new cases for the new `hasPersistedBuffer` helper — true only with content, empty snapshot treated as absent (matches `restoreBuffer`'s truthy guard), dims-only ignored. (All 15 pass.)

## Notes

- A **known follow-up** (out of scope, larger change): the host-side `terminal-mode-tracker` also carries the stale armed state across a host restart (`terminal.ts` reclaims only while `shellReadyState === "pending"`), so a client that attaches *later* could still see a stale preamble. This PR fixes the renderer path that reproduces both reported issues; the durable host-side reclaim is a separate change.
- Manual repro confirmed by the two issues' triage bots (`#6308`, `#6343`) — no maintainer-adjacent duplicates, no existing PR for either issue.

---
## Summary by cubic
Prevents leaked terminal input modes after session restore so the prompt stays clean. Previously, restore/reattach replay re-armed mouse/focus/kitty modes before the first shell-ready marker and the reclaimer exempted them; now we strip those bytes and reclaim them at the next marker. Fixes #6308 and #6343.

- Persisted snapshots serialize with `excludeModes: true` via `@xterm/addon-serialize`, so restores no longer re-arm a dead session’s modes.
- On seeded/restored runtimes, we call noteRestoreAttach() before writing any seed/restore bytes and gate restores with hasPersistedBuffer; restoreBuffer runs after the flag so pre-marker modes are reclaimed.
- The input-mode reclaimer install now returns a handle exposing the `reclaimer` for createRuntime; tests cover restored attach behavior and hasPersistedBuffer.

<sup>Written for commit 2bd9d9c4beb95e0aad65c67433b2be5d43c64a78. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6345?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of terminal sessions with persisted scrollback.
  * Prevented stale mouse and input modes from remaining active after restoring a terminal.
  * Preserved input modes that are intentionally re-enabled by a live terminal application.
  * More accurately detects whether saved terminal content is available before replaying it.

### DIFF
--- a/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, test } from "bun:test";
 import {
 	clearAllTerminalState,
+	hasPersistedBuffer,
 	pruneExpiredTerminalState,
 	reclaimTerminalStateForQuota,
 	removeTerminalStatePersistedAt,
@@ -191,5 +192,27 @@ describe("persisted-at index maintenance", () => {
 		touchTerminalStatePersistedAt("t1", storage, NOW);
 
 		expect(readIndex(values)).toEqual({ t1: NOW });
+	});
+});
+
+describe("hasPersistedBuffer", () => {
+	test("reports true only when a buffer snapshot exists", () => {
+		const { values, storage } = createFakeStorage();
+		expect(hasPersistedBuffer("t1", storage)).toBe(false);
+
+		values.set(`${TERMINAL_BUFFER_KEY_PREFIX}t1`, "scrollback");
+		expect(hasPersistedBuffer("t1", storage)).toBe(true);
+	});
+
+	test("treats an empty snapshot as absent (matches restoreBuffer)", () => {
+		const { values, storage } = createFakeStorage();
+		values.set(`${TERMINAL_BUFFER_KEY_PREFIX}t1`, "");
+		expect(hasPersistedBuffer("t1", storage)).toBe(false);
+	});
+
+	test("ignores a terminal that only has dims persisted", () => {
+		const { values, storage } = createFakeStorage();
+		values.set(`${TERMINAL_DIMS_KEY_PREFIX}t1`, "120x32");
+		expect(hasPersistedBuffer("t1", storage)).toBe(false);
 	});
 });
--- a/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-buffer-gc.ts
@@ -52,6 +52,22 @@ export function touchTerminalStatePersistedAt(
 	writeIndex(storage, index);
 }
 
+/** Whether a non-empty persisted buffer snapshot exists for a terminal id. */
+export function hasPersistedBuffer(
+	terminalId: string,
+	storage: Storage = localStorage,
+): boolean {
+	try {
+		// Mirrors `restoreBuffer`'s truthy-write guard: an empty snapshot would
+		// be read back as "no content", so it does not count as a restore.
+		return Boolean(
+			storage.getItem(`${TERMINAL_BUFFER_KEY_PREFIX}${terminalId}`),
+		);
+	} catch {
+		return false;
+	}
+}
+
 export function removeTerminalStatePersistedAt(
 	terminalId: string,
 	storage: Storage = localStorage,
--- a/apps/desktop/src/renderer/lib/terminal/terminal-runtime.ts
+++ b/apps/desktop/src/renderer/lib/terminal/terminal-runtime.ts
@@ -6,6 +6,7 @@ import type { SearchAddon } from "@xterm/addon-search";
 import { SerializeAddon } from "@xterm/addon-serialize";
 import { Terminal as XTerm } from "@xterm/xterm";
 import { DEFAULT_TERMINAL_SCROLLBACK } from "shared/constants";
+import type { LeakedInputModeReclaimer } from "shared/leaked-input-mode-reclaim";
 import {
 	applyTerminalFontFamilyCssVariable,
 	type TerminalAppearance,
@@ -20,6 +21,7 @@ import {
 } from "./parser-idle-gate";
 import { loadAddons } from "./terminal-addons";
 import {
+	hasPersistedBuffer,
 	removeTerminalStatePersistedAt,
 	TERMINAL_BUFFER_KEY_PREFIX,
 	TERMINAL_DIMS_KEY_PREFIX,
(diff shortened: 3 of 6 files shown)


qflen:
Note for reviewers, since the triage analysis raised it: the `disarmStaleInputModes` it couldn't find is real, it's just not on main. It's in #5519, which the #5790 doc was written alongside. So the cold-restore disarm that doc treats as covered doesn't actually exist on main, which is this bug.

No overlap with this PR: yours is all in `renderer/lib/terminal/`, mine is the host emulator, session, and the v1 pane path. Your additions to `leaked-input-mode-reclaim.ts` don't touch the constants I import from it, so either can land first.

The host-side follow-up in your Notes is already implemented there, along with the alt-screen exit #5790 left open.

andyst-dev:
Thanks for the careful read — confirmed on both points:

- No overlap: this PR is confined to `renderer/lib/terminal/` plus the additive entries in `leaked-input-mode-reclaim.ts` (the reclaimer installation handle and the restore-aware tracking); it doesn't touch the constants you import from that module, so either PR can land first.
- Good context on `disarmStaleInputModes` living in #5519 rather than `main` — that's exactly the cold-restore gap this PR closes. And good to know the host-side follow-up (plus the alt-screen exit #5790 left open) is already implemented in your branch; no duplicate work on my side.

Kitenite:
Closing: main already ships this. e193d44b1 (#6642) shipped the input-mode reclaimer on restore. Thanks for the PR; if you think the shipped version misses a case, open an issue and link this.

PR #6347, opened 2026-08-11
fix(cli): drive a single /authorize request per OAuth login

### LINKED ISSUE #6310: OAuth login redirect fails — stuck on authorize page after clicking Authorize
\`superset auth login\` opens the OAuth consent screen correctly (signed in, org shown, correct scopes). Clicking Authorize does nothing — page doesn't navigate, URL stays on \`api.superset.sh/api/auth/oauth2/authorize?...\`, never redirects to \`app.superset.sh/cli/auth/code\` to show the pairing code. CLI never receives the code (\`superset auth whoami\` still reports not logged in). Reproduced across 3 separate login attempts (fresh state each time), both foregrounded and backgrounded CLI processes, and multiple browser windows.

**Environment:** superset 1.20.0, Darwin arm64

**Steps to reproduce:**
1. \`superset auth login\`
2. Open printed URL in browser
3. Sign in, land on "Authorize Superset CLI" consent screen
4. Click Authorize
5. Page stays on same URL, no redirect, no code shown
6. \`superset auth whoami\` still errors "Not logged in"

andyst-dev:
Fixes #6310

## Problem

`superset auth login` opens the OAuth consent screen, but clicking **Authorize** does nothing — the page never navigates to `/cli/auth/code`, and the CLI never receives a code.

## Root cause

The CLI built **two** `/authorize` URLs for the same login: a loopback URL (`http://127.0.0.1:PORT/callback`, auto-opened in the browser) and a paste URL (`https://app.superset.sh/cli/auth/code`, printed in the terminal). They shared the same `state`/`code_challenge` but used **different `redirect_uri` values**, and the terminal UI explicitly invited the user to open the printed (paste) URL while the browser had already auto-opened the loopback one.

On the provider side the second `/authorize` for the same login clobbers the first (single pending-request slot), so the pending request the consent step resolves can be the wrong — or already-consumed — one. Consent then returns **without a redirect URL**, and the consent form had no branch for that case: no navigation, no error, spinner stuck on "Authorizing…" forever.

## Fix

**CLI — one `/authorize` request per login.** New `resolveAuthorizeUrl()` picks a **single** `redirect_uri` up front and builds exactly one authorize URL:
- loopback when a port bound **AND** a local browser is being opened;
- paste otherwise (SSH, non-TTY, CI, or bind failed).

`login()` now prints and auto-opens the **same** URL, and the code is exchanged with that same `redirect_uri` on both the loopback and paste completion paths. The paste prompt remains as a fallback only when the paste URL was the one presented.

**Web — never fail silently.** `ConsentForm` now handles the case where `consent()` resolves without an error and without a redirect URL: it logs the full response, surfaces a clear message, and stops the loading state, instead of leaving the button stuck forever. This is the diagnostic branch that was missing regardless of root cause.

## Tests

`packages/cli/src/lib/auth.test.ts` (7 cases total, +6 new):
- `resolveAuthorizeUrl` unit tests: loopback chosen when port bound + browser opens; paste fallback when browser not opened; paste fallback when no port bound; single authorize URL always produced with the right params.
- `login` integration tests with injectable `openBrowser`/`shouldOpenBrowser`/`bindLoopbackServer` hooks: asserts the printed and auto-opened URLs are **identical** (the core of the bug), and that nothing is opened when loopback is unavailable.

`ConsentForm.tsx` has no component test harness in this repo (no jsdom/testing-library), so the web-side change is verified by typecheck + biome only.

---
## Summary by cubic
Fixes OAuth login by issuing a single /authorize request per login and running only the presented flow, so the consent page redirects and the CLI gets the code. Adds a client‑neutral safeguard in the web consent form to prevent silent failures. Fixes #6310.

- **Bug Fixes**
  - Use one redirect_uri per login via resolveAuthorizeUrl; CLI opens and prints the same authorize URL, and both loopback and paste flows exchange the code with that same redirect_uri.
  - Drive only the presented flow: loopback-only when active (surface failures instead of hanging); paste-only otherwise.
  - Consent form shows a client‑neutral error and stops loading when consent returns without a redirect URL.
  - Preserve the loopback failure reason and update the recovery hint to suggest re-running in a non-interactive terminal to use the paste flow.

<sup>Written for commit cde65eee383adcfb897fe313b73bf2a64a5c249f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6347?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed consent flows that return without a redirect URL so they now show an authorization error and stop loading instead of becoming stuck.

* **CLI Improvements**
  * Improved login behavior across browser-based, loopback, and paste-code authentication flows.
  * Added more reliable redirect handling and authorization URL selection.
  * Added fallback support when a local loopback connection is unavailable.
  * Improved handling of loopback connection failures during login.
  * Expanded coverage for login and authorization scenarios.

### DIFF
--- a/apps/web/src/app/oauth/consent/components/ConsentForm/ConsentForm.tsx
+++ b/apps/web/src/app/oauth/consent/components/ConsentForm/ConsentForm.tsx
@@ -115,7 +115,7 @@ export function ConsentForm({
 					{ accept, data },
 				);
 				setError(
-					"Could not complete authorization. Please close this window and run superset auth login again.",
+					"Could not complete authorization. Please close this window and restart authorization from the application.",
 				);
 				setIsLoading(false);
 			}
--- a/apps/web/src/app/oauth/consent/components/ConsentForm/ConsentForm.tsx
+++ b/apps/web/src/app/oauth/consent/components/ConsentForm/ConsentForm.tsx
@@ -105,6 +105,19 @@ export function ConsentForm({
 
 			if (data?.url) {
 				window.location.href = data.url;
+			} else {
+				// No redirect URL and no error: the pending authorization request
+				// could not be resolved (e.g. it was clobbered by a competing
+				// /authorize for the same login — see superset-cli #6310). Surface
+				// it instead of leaving the button stuck on "Authorizing…" forever.
+				console.error(
+					"[oauth/consent] consent() resolved without a redirect URL",
+					{ accept, data },
+				);
+				setError(
+					"Could not complete authorization. Please close this window and run superset auth login again.",
+				);
+				setIsLoading(false);
 			}
 		} catch (err) {
 			console.error("[oauth/consent] Error:", err);
--- a/packages/cli/src/lib/auth.test.ts
+++ b/packages/cli/src/lib/auth.test.ts
@@ -14,6 +14,10 @@ describe("login /authorize (single request)", () => {
 		const printed: string[] = [];
 
 		const controller = new AbortController();
+		let notifyAuthUrl!: () => void;
+		const authUrlEmitted = new Promise<void>((resolve) => {
+			notifyAuthUrl = resolve;
+		});
 		const loginPromise = login(controller.signal, {
 			// A loopback port is "bound" — but we never let the fake server
 			// actually receive a code; the test only asserts on the URL(s).
@@ -27,12 +31,14 @@ describe("login /authorize (single request)", () => {
 			},
 			onAuthorizationUrl: (url) => {
 				printed.push(url);
+				notifyAuthUrl();
 			},
-			// Never resolve: we only assert on the URLs emitted up front.
+			// Not reached in the loopback path, but keep it a never-resolving
+			// stub so the login promise stays pending if it is mis-reached.
 			promptForPastedCode: () => new Promise<string>(() => {}),
 		});
 
-		await new Promise((r) => setTimeout(r, 20));
+		await authUrlEmitted;
 
 		expect(printed).toHaveLength(1);
 		expect(opened).toEqual(printed);
@@ -52,6 +58,10 @@ describe("login /authorize (single request)", () => {
 		const printed: string[] = [];
 
 		const controller = new AbortController();
+		let notifyAuthUrl!: () => void;
+		const authUrlEmitted = new Promise<void>((resolve) => {
+			notifyAuthUrl = resolve;
+		});
 		const loginPromise = login(controller.signal, {
 			bindLoopbackServer: async () => null, // no port bound
 			shouldOpenBrowser: () => true,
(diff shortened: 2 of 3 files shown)


andyst-dev:
Thanks for the reviews — addressed in `3b44684`:

- **cubic P1**: when the loopback flow is active it is now the ONLY flow driven — the paste prompt no longer runs, so an accidental/invalid paste can't abort a valid browser login, and a browser denial is surfaced as a clear error instead of leaving the command waiting indefinitely. Paste-alone is used when the loopback URL was not presented.
- **coderabbit ConsentForm (Minor)**: the recovery text is now client-neutral ('restart authorization from the application') — this consent page also serves dynamically registered OAuth clients, for whom 'run superset auth login' was incorrect.
- **coderabbit test (Minor)**: the fixed 20ms waits are replaced with callback barriers on onAuthorizationUrl, so the assertions no longer race under CI load.

andyst-dev:
Thanks — addressed in `cde65ee`:

- **cubic P2 (loss of diagnostic reason)**: the loopback rejection handler now preserves the underlying `error.message` (denial vs timeout vs CSRF) instead of collapsing everything into a generic 'Browser login did not complete'.
- **cubic P2 (--no-browser hint)**: confirmed `--no-browser` is not defined for `superset auth login` — the hint now points to the supported path (run in a non-interactive terminal, no browser, to trigger the paste flow).

andyst-dev:
Addressed in `467016fd0`.

The launcher rejection was a real regression from this branch: `void openBrowserImpl(url)` lost the `.catch(() => {})` the previous call site had, so a launcher that cannot start raised an unhandled rejection. It swallows again now, since the loopback wait already reports its own failure (denial, timeout, CSRF) and the paste flow has no browser to open.

The cancellation check now sits right after the loopback port binds, so a signal that lands while binding no longer publishes a URL and opens a browser for a login that is already cancelled.

The consent recovery message goes through `t()` like the rest of the component.

On the `waitForCallback` ordering (the `callbackError` branch running before the `state` comparison): that function is not touched by this PR, so I left it alone rather than fold an untouched function into the diff. The point stands though, and the single-redirect change makes its consequences worse, because a stale error callback can now end a valid login. Say the word and I will fix it here.

Two tests cover the launcher rejection and the cancellation window; both fail without the source changes. 9/9 in `auth.test.ts`, 301/301 across `packages/cli`, tsc and biome clean.

PR #6350, opened 2026-08-11
fix(desktop): let bash users keep their own prompt via SUPERSET_KEEP_PS1

### LINKED ISSUE #6311: Make overriding of PS1 in bash configurable
### What problem are you trying to solve?

It seems Superset deliberately overrides `PS1` for bash.

https://github.com/superset-sh/superset/blob/main/apps/desktop/src/main/lib/agent-setup/shell-wrappers.ts

My bash PS1 is important for me. Can we make this configurable, i.e. a way to leave the user's PS1 untouched?

### Proposed solution

I can perhaps set a bash variable `SUPERSET_KEEP_PS1` the presence of which should be a hint to not override `PS1` in the Bash wrapper?

### Workarounds you've tried

_No response_

andyst-dev:
Fixes #6311

## Problem

For **bash — and only bash** — Superset unconditionally replaces the user's prompt. `createBashWrapper` sources the user's profile/`.bashrc` first and then clobbers `PS1` with the emerald `❯`, so there is no way for a user config to win, and no env var, setting, or file opts out. zsh and fish never touch `PROMPT`/`PS1`, so this is an asymmetry rather than a requirement of shell integration.

Three concrete problems in the one `export PS1=…` line:

1. **No opt-out** — any `PS1` a user sets in `.bashrc` is always overwritten.
2. **`export` leaks** `PS1` into child processes, where rc files and scripts that test `[ -n "$PS1" ]` for interactivity are misled.
3. **Non-interactive execution** — the same rcfile is sourced for `bash -c` command execution (`getCommandShellArgs`), where a prompt is meaningless.

## Fix

In `createBashWrapper` (shell-wrappers.ts), the prompt is now assigned only when all three hold:

```bash
if [[ -z "${SUPERSET_KEEP_PS1:-}" && $- == *i* ]]; then
  PS1=$'\[\e[1;38;2;52;211;153m\]❯\[\e[0m\] '
fi
```

- **Opt-out:** set `export SUPERSET_KEEP_PS1=1` in your bash config to keep your own prompt. The guard runs *after* user config is sourced, and the `SUPERSET_ENV_SAVE`/`RESTORE` pair never unsets newly-added vars, so an exported `SUPERSET_KEEP_PS1` from `.bashrc`/`.bash_profile` survives to the guard.
- **Interactive only:** `$- == *i*` skips the assignment for non-interactive `bash -c` execution.
- **No leak:** plain `PS1=` instead of `export PS1=`.

The shell-readiness markers ride `PROMPT_COMMAND` (independent of `PS1`), so terminal automation is unaffected.

## Tests

`shell-wrappers.test.ts` (4 new cases, running real bash with `--rcfile`, matching the existing integration-test pattern):
- Emerald prompt still applied by default in an interactive shell.
- `SUPERSET_KEEP_PS1=1` exported from user `.bashrc` keeps the user's `MINE> ` prompt.
- No `PS1` set for non-interactive `-c` execution.
- The generated rcfile uses a plain `PS1=` (no `export`).

All 29 tests in the file pass; `tsc --noEmit` and biome are clean.

---
## Summary by cubic
Let bash users keep their own prompt. The wrapper now sets the emerald ❯ only in interactive bash when `SUPERSET_KEEP_PS1` is not set, and keeps `PS1` unexported so it does not leak to child processes.

- Opt-out: set `SUPERSET_KEEP_PS1=1` to keep your prompt.
- Interactive only: guard on `$-` containing `i`; skip for `bash -c`.
- No leak: clear any export with `export -n PS1`, then assign `PS1=…`.
- zsh and fish remain unchanged.

<sup>Written for commit 9d2767c72b9c1f9223d5c6fc8e7d3c08e43ea722. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Bash wrappers now configure the default prompt only in interactive shells.
  * Added `SUPERSET_KEEP_PS1` support to preserve existing prompts.
  * Prevented prompt variables from leaking into child processes.
  * Ensured user-defined prompts are handled appropriately without unwanted exporting.
* **Tests**
  * Added coverage for interactive and non-interactive shells, prompt overrides, prompt preservation, and child-process environment behavior.

### DIFF
--- a/packages/agent-setup/src/shell-wrappers.test.ts
+++ b/packages/agent-setup/src/shell-wrappers.test.ts
@@ -874,4 +874,75 @@ export SUPERSET_WORKSPACE_PATH="/wrong/path"
 			}
 		});
 	});
+
+	describe("bash PS1 handling (#6311)", () => {
+		function runBash({
+			homeDir,
+			rcArgs = ["-ic"],
+			extraEnv = {},
+		}: {
+			homeDir: string;
+			rcArgs?: string[];
+			extraEnv?: Record<string, string>;
+		}): string {
+			createBashWrapper(TEST_PATHS);
+			return execFileSync(
+				"bash",
+				[
+					"--rcfile",
+					path.join(TEST_BASH_DIR, "rcfile"),
+					...rcArgs,
+					'printf "%s" "$PS1"',
+				],
+				{
+					encoding: "utf-8",
+					env: { HOME: homeDir, PATH: "/usr/bin:/bin", ...extraEnv },
+				},
+			);
+		}
+
+		it("applies the emerald prompt by default in an interactive shell", () => {
+			const homeDir = path.join(TEST_ROOT, "ps1-default", "home");
+			mkdirSync(homeDir, { recursive: true });
+			writeFileSync(path.join(homeDir, ".bashrc"), `PS1='MINE> '\n`);
+
+			const output = runBash({ homeDir });
+			// Escape sequence preserved via $'...': the 38;2;52;211;153 color code.
+			expect(output).toContain("38;2;52;211;153");
+		});
+
+		it("keeps the user's PS1 when SUPERSET_KEEP_PS1 is exported", () => {
+			const homeDir = path.join(TEST_ROOT, "ps1-keep", "home");
+			mkdirSync(homeDir, { recursive: true });
+			writeFileSync(
+				path.join(homeDir, ".bashrc"),
+				`export SUPERSET_KEEP_PS1=1\nPS1='MINE> '\n`,
+			);
+
+			const output = runBash({ homeDir });
+			expect(output).toContain("MINE> ");
+			expect(output).not.toContain("38;2;52;211;153");
+		});
+
+		it("does not set PS1 for non-interactive execution", () => {
+			const homeDir = path.join(TEST_ROOT, "ps1-noninteractive", "home");
+			mkdirSync(homeDir, { recursive: true });
+
+			const output = runBash({ homeDir, rcArgs: ["-c"] });
+			// No prompt in a non-interactive shell (the rcfile is sourced via
+			// getCommandShellArgs for `bash -c` command execution).
+			expect(output).not.toContain("38;2;52;211;153");
+		});
+
+		it("does not leak PS1 into child processes (plain assignment, no export)", () => {
+			const homeDir = path.join(TEST_ROOT, "ps1-noexport", "home");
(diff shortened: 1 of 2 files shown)


andyst-dev:
Thanks for the reviews — addressed in `03a213c`:

- **cubic P2 (line 281) + coderabbit L282**: a plain `PS1=` assignment does not clear the export attribute if the user's own `.bashrc` exported PS1 first. Added `export -n PS1` before the assignment so the not-exported guarantee holds either way, with a test that a grandchild shell's environment has no PS1.
- **cubic P2 (line 931, test)**: you're right that `--rcfile` is only honoured by bash for interactive shells, so the `-c` test never sourced the rcfile and passed trivially. The non-interactive test now sources the wrapper rcfile explicitly in a `-c` shell to actually exercise the `$- == \*i\*` guard.
- The no-export test now asserts on behaviour (grandchild env), not just rcfile text.

andyst-dev:
Thanks — addressed in `c7d5e2e`: the test named 'does not export PS1 to child processes' only ran in the same shell (the grandchild-env check is the next test). Renamed it to describe what it actually guards — the emerald prompt applies even when user config exported PS1 — and pointed the comment at the grandchild test for the export behaviour.

sandipb:
Any update to this? seems like CI is blocked

andyst-dev:
The "blocked" CI isn't a code failure — it's the fork-PR workflow-approval gate. The `CI` and `Deploy Preview` runs both show `completed / action_required`, which is GitHub holding the fork-PR workflow for a maintainer to approve running it (first-time contributor to this repo). It's not something a force-push can clear.

I've also refreshed the branch while here: it was 291 commits behind `main`, so I rebased the 3 PS1 commits onto current `main` and force-pushed (head `9d2767c72`, 0 behind). Validation: `packages/agent-setup` `shell-wrappers.test.ts` 30/30 pass.

The one thing that would unblock CI on your side: approve the fork workflow run(s) at the repo's Actions tab for head `9d2767c72`. Everything else is ready to merge once main reviews the diff.

PR #6351, opened 2026-08-11
fix(workspace-fs): coalesce FSEvents overflow into backed-off per-root rescans

Kitenite:
## Summary
- FSEvents overflow ("Events were dropped by the FSEvents client") no longer triggers an immediate full rescan per occurrence: overflows coalesce into at most one rescan per watch root on an exponential backoff (5s → 60s cap while overflows continue).
- A guaranteed trailing rescan runs after the last overflow **and now signals listeners** — pre-fix, the post-overflow rescan never notified them, so views could silently stay stale after heavy churn (real correctness bug, not just noise).

## Why / Context
A user's host-service log (UI-freeze report, #ext-vori) showed **648/day** overflow rescans across 8+ worktrees of a big monorepo — a permanent rescan storm contributing to the background load starving their pty-daemon. Under sustained bulk operations (branch switches, installs, builds) every overflow used to fire a full rescan back-to-back.

## Manual QA (real kernel-level FSEvents overflow, 6 worktrees, 10-min A/B windows, same churn harness)
- Rescans: **5,770 → 66** (87×).
- Staleness: a file created+committed during the suppression window is reflected after the settle rescan, with listener signal (before: no signal ever).
- Honest numbers: host CPU 83.4% → 83.0% (unchanged at this stress scale — dominated by real event processing) and loop-lag p50/p95 1/2ms → 1/3ms on a 14-core machine; the practical wins are the staleness fix, log-noise elimination, and bounded work during bulk churn on smaller machines.

## Testing
- `watch-overflow-backoff.test.ts` (200 lines) — backoff schedule, coalescing, trailing-rescan signal; overflow injected at the handler (labeled synthetic in the file). Real-overflow behavior verified end-to-end with the harness above.
- workspace-fs suite 62/62, host-service git-watcher 13/13; `bun run typecheck` and `bun run lint` exit 0.

## Risks
Suppression changes watcher timing: a change landing mid-storm is reflected at the next backed-off or trailing rescan (≤60s worst case during sustained overflow) instead of immediately — previously it was often *never* reflected without an unrelated event, so this is strictly better on correctness, but reviewers should eye the per-root backoff/trailing interaction for roots that go quiet.

---
## Summary by cubic
Coalesces macOS FSEvents overflows into per-root trailing rescans with exponential backoff and emits an `overflow` event directly so consumers refresh reliably. Previously each overflow immediately invalidated the index without notifying listeners.

- Lazy trailing window with cap: overflows stamp `lastOverflowAt`; the armed timer defers at fire time to trail the last overflow. A per-batch deadline guarantees at least one rescan per cap, and a deadline-forced fire schedules a follow-up batch to preserve the trailing guarantee under sustained churn.
- Rescan behavior and delivery: invalidate the search index once and emit an `overflow` event via `emitDirect` (bypasses the throttler; listener throws are contained). Recovery resume now also uses `emitDirect` to avoid dropping the root-create signal.
- Lifecycle and defaults: cancel pending rescans on watcher dispose and on recovery resume. Defaults are 5s initial delay, 60s max cap, 120s quiet reset; backoff resets after a quiet window. Tests add coverage for backoff, trailing under storms, cap enforcement, cancellation, and cached-index correctness.

<sup>Written for commit 9415987cf9d0ce3a715599f64727c3428fad9268. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved file-watching reliability during bursts of filesystem changes by coalescing overflow rescans.
  * Added configurable delays and exponential backoff to reduce unnecessary rescanning.
  * Automatically resets backoff after stable operation and cancels pending work when watchers are disposed.

* **Bug Fixes**
  * Preserved recovery behavior when watched directories are deleted or restored.

* **Tests**
  * Added coverage for overflow bursts, backoff, reset behavior, suppression windows, and watcher cancellation.


Kitenite:
Addressed all three review findings in c144c6aa1:

- **CodeRabbit (Major) + cubic P2 — timer fires relative to first overflow:** every overflow now cancel-and-rearms the trailing timer to one window after itself. Sustained overflow can't postpone rescans forever: each batch carries a hard deadline (first overflow + 60s cap) that clamps rearms, and a deadline-forced fire arms a follow-up batch so the final rescan still trails the last overflow by one window.
- **cubic P1 — rescan signal can be dropped by the throttler's bounded buffer:** the overflow rescan event is now delivered directly to listeners, bypassing the throttler (it's one tiny event per ≥5s per root; the throttler exists to bound floods, and this signal is what guarantees convergence after one).

Regression tests added (verified failing against the previous commit): second-overflow-near-expiry rearm, and sustained-overflow anti-starvation. Real-churn re-verification (2 min continuous FSEvents overflow, 6 worktrees): 1,459 raw overflows → 18 rescans; deadline-forced rescans at +60s/+120s during the storm; final rescan exactly one window (20s) after the last raw overflow on every root.

PR #6353, opened 2026-08-11
fix(desktop): stop workspace delete from crash-remounting the renderer via stale v1 routes

Kitenite:
Fixes SUPER-1814 — reported by Jose Carlos (reacherapp.com) on 1.20.2: every workspace delete "refreshes the whole Superset application" and dumps him on an empty New Workspace screen until a manual reload.

## Root cause

Confirmed against his PostHog timeline (the `desktop_opened` root-mount event re-fires 400ms after each delete, landing on `#/new-workspace`):

1. Deleting the active workspace with no visible sibling falls back to `navigate({ to: "/" })`.
2. `/` redirects to `/workspace` — the **v1** index.
3. For migrated v2 users, the v1 index restores a stale `lastViewedWorkspaceId` (pre-migration v1 row still in the electron localDb) and navigates onto a v1 workspace route.
4. The dashboard's version-mismatch guard rendered an **inline** `<Navigate to="/new-workspace" replace />`. `Navigate` re-navigates whenever its props object identity changes, so every layout re-render during the pending navigation re-navigated until React threw error #185. The throw bubbled to the root `errorComponent`, unmounting the entire tree; the settled navigation reset the boundary and remounted the root — the "app reloaded itself" flash.

Only users with pre-migration v1 leftovers hit this, which is why it never reproduced on clean installs. Step 4 is the same hazard class already fixed in `routes/page.tsx` (#5729).

## Fix

- **Hoist the version-mismatch `<Navigate>`** to module scope for stable props identity — kills the #185 loop no matter how a v2 user reaches a v1 route.
- **Workspace-removal fallback goes straight to `/new-workspace`** instead of detouring through `/` → v1 index.
- **The v1 workspace index redirects v2 users to `/new-workspace`** instead of running v1 restore logic — covers every other path into `/` (sign-in, create-organization, not-found "Go home"), including the sibling "creating a workspace makes the others disappear" symptom in the same report.

## Verification

- Mechanism confirmed in the installed router source: `Navigate` re-navigates on props-identity change (`useNavigate.js`, `previousPropsRef.current !== props`).
- `bun run typecheck` (desktop) clean, `bun run lint` clean, 1454 renderer tests pass.
- No end-to-end CDP repro was run — reproducing requires seeding pre-migration v1 state (stale `lastViewedWorkspaceId` + v1 localDb rows).

---
## Summary by cubic
Fixes SUPER-1814 by stopping the app from remounting after deleting a workspace. V2 users now go straight to `/new-workspace`, avoiding stale v1 routes and a `Navigate` re-navigation loop.

- **Bug Fixes**
  - Hoisted version-mismatch `Navigate` to module scope to keep props stable and prevent re-navigation (React error 185).
  - On delete fallback, route directly to `/new-workspace` instead of `"/"` → v1 index.
  - In the v1 workspace index, detect v2 and redirect to `/new-workspace`; skip `electronTrpc.workspaces.getAllGrouped` for v2; isolate the v2 redirect from v1 query churn and memoize the workspace list to avoid redundant effect triggers during navigation.

<sup>Written for commit a8b0ff2f0f2378032c1cdcbca4f0d250cb108162. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6353?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automatic navigation to the new workspace setup when no replacement workspace is available.
  * Added routing from V2 Cloud workspace dashboards to the new workspace experience.

* **Bug Fixes**
  * Improved workspace restoration and navigation behavior for existing V1 workspaces.
  * Prevented unnecessary workspace reloads during dashboard navigation.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/hooks/useNavigateAwayFromWorkspace/useNavigateAwayFromWorkspace.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/hooks/useNavigateAwayFromWorkspace/useNavigateAwayFromWorkspace.ts
@@ -61,7 +61,10 @@ export function useNavigateAwayFromWorkspace() {
 				}).catch(reportRemovalNavigationError);
 				return;
 			}
-			void navigate({ to: "/", replace: true }).catch(
+			// Straight to the v2 empty state — "/" detours through the v1
+			// workspace index, which can restore stale pre-migration state
+			// (SUPER-1814).
+			void navigate({ to: "/new-workspace", replace: true }).catch(
 				reportRemovalNavigationError,
 			);
 		},
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/layout.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/layout.tsx
@@ -32,6 +32,11 @@ export const Route = createFileRoute("/_authenticated/_dashboard")({
 	component: DashboardLayout,
 });
 
+// Hoisted for stable props identity — <Navigate> re-navigates whenever its
+// props object changes, so an inline element re-navigates on every re-render
+// until React throws error #185 (see routes/page.tsx, #5729, SUPER-1814).
+const newWorkspaceRedirect = <Navigate to="/new-workspace" replace />;
+
 /** v1 only — v2 deletes go through the globally-mounted DeleteWorkspaceMount
  * (see delete-workspace-intent store). */
 type DeleteTarget = {
@@ -225,7 +230,7 @@ function DashboardLayout() {
 							// dead-end "pick a workspace" screen. v1 users keep the
 							// static state — /new-workspace is a v2-only surface.
 							isV2CloudEnabled ? (
-								<Navigate to="/new-workspace" replace />
+								newWorkspaceRedirect
 							) : (
 								<CrossVersionMismatchState />
 							)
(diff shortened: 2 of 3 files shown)


Kitenite:
CDP verification (dev desktop from this worktree, vite :4145, CDP :9250, signed-in session):

**Before (3 files reverted to parent commit, stale `lastViewedWorkspaceId` seeded):** navigating onto a v1 `/workspace/$id` route as a v2 user deterministically reproduced the reported crash — `Maximum update depth exceeded` (React #185) → `[renderer] Route error caught:` (root ErrorPage) → dashboard-layout DOM destroyed and remounted → recovered at `#/new-workspace`. `window` survived, so it's a root remount, not a document reload — matching the PostHog signature from the report.

**After (this PR):** identical navigation is clean — no error output, layout DOM survives, lands on `#/new-workspace`. Also clean: `/` and `/workspace` entry points, a nonexistent v1 workspace id, and a rapid triple-push onto the v1 route mid-transition.

**End-to-end (real CDP mouse input):** created a disposable session workspace, hovered its sidebar row, clicked "Close workspace", clicked "Delete" in the dialog — landed on the sibling workspace with the tree intact, zero console errors, and the host.db row archived with `archive_reason=deleted`.

Caveat: the pre-fix `/` variant of the crash is timing-dependent (needs a layout re-render mid-transition) and didn't fire in this dev run; the direct v1-route variant is the deterministic repro. Workspace creation in the journey was API-driven; the deletion itself was real input.

Kitenite:
Addressed the effect-dependency feedback (both bots): split the v2 redirect into its own effect keyed only on `isV2CloudEnabled` and memoized `allWorkspaces`, so the v1 restore effect no longer re-fires on array-identity churn. Commit a8b0ff2f0.

PR #6359, opened 2026-08-11
ci(relay2): only deploy relay2 when relay2-relevant paths change

saddlepaddle:
## Summary
Every relay2 deploy restarts its Durable Objects, which severs every live host tunnel (hosts reconnect in ~2s, but in-flight streams drop). `deploy-production.yml` runs on every push to `main` with no path gating, so today's six unrelated merges bounced the entire relay2 fleet six times in one hour.

This gates the `deploy-relay2` job with `dorny/paths-filter` (pinned v3.0.2): it now deploys only when `apps/relay2/**`, its workspace deps (`packages/shared/**`, `packages/trpc/**`), `bun.lock`, or the workflow itself changed. `workflow_dispatch` always deploys, preserving the manual escape hatch.

The stateless Cloudflare workers (electric-proxy) are left as-is — redeploying them on every push is harmless.

## Test plan
- [x] `bun run lint` clean
- [ ] After merge: verify an unrelated main push skips the Deploy Worker step, and a relay2-touching push (or manual dispatch) still deploys

## Summary by CodeRabbit

* **Chores**
  * Improved production deployment handling by detecting relevant changes before running deployment steps.
  * Manual deployments continue to run on demand without change filtering.
  * Added clearer deployment guidance, including information about restart behavior during updates.

---
## Summary by cubic
Gate relay2 deployments so the job only runs when relay2-related files change, preventing unnecessary Durable Object restarts and dropped host tunnels. Uses `dorny/paths-filter@v3.0.2` to watch `apps/relay2/**`, `packages/shared/**`, `packages/trpc/**`, `bun.lock`, and the workflow; `workflow_dispatch` still triggers a deploy.

<sup>Written for commit 951001292cf58d257b93bdf0d9b2dfe9ea04f3a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6359?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -476,6 +476,9 @@ jobs:
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
         run: bunx wrangler deploy
 
+  # Every relay2 deploy restarts its Durable Objects, severing every live host
+  # tunnel — so unlike the stateless workers above, only deploy when something
+  # relay2 actually ships has changed. workflow_dispatch always deploys.
   deploy-relay2:
     name: Deploy Relay2 to Cloudflare
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


saddlepaddle:
Closing per discussion: deploy-time tunnel bounces are recoverable (liveness sweep + ~2s partysocket reconnect + app-layer terminal resume) and not noticeably disruptive in practice, and out-of-band relay deploys won't be part of the workflow going forward — so the gate isn't worth the CI complexity. Revisit if deploy churn ever becomes user-visible.

Issue #5712, opened 2026-07-16
Setup/Preset commands drop the first char


iradcn:
### Where did this happen?

Desktop app (macOS)

### Version

1.15.0

### What happened?

When I spin up a new workspace, my setup command is being run in a terminal, with the first char omitted:
```
./setup-worktree.sh
[oh-my-zsh] Would you like to update? [Y/n]
[oh-my-zsh] You can update manually by running `omz update`
➜  rogue-shallot /setup-worktree.sh
zsh: no such file or directory: /setup-worktree.sh
```
(see the `.` is missing)

When I spin up a preset in this workspace the same happens:
```
claude --permission-mode plan --allow-dangerously-skip-permissions
[oh-my-zsh] Would you like to update? [Y/n]
[oh-my-zsh] You can update manually by running `omz update`
➜  rogue-shallot laude --permission-mode plan --allow-dangerously-skip-permissions
zsh: command not found: laude
```
(see `laude` vs `claude`)

### Steps to reproduce

1. Open a workspace
2. See the faulty setup command

### Logs, screenshots, or recordings

<img width="905" height="260" alt="Image" src="https://github.com/user-attachments/assets/ba419702-0d0f-4598-919c-7f5ab1fe90c6" />

Kitenite:
We merged this with #5879 into one ticket in our tracker (SUPER-1103). Both look like the same shell readiness race when commands are sent on terminal launch.

PR #5580, opened 2026-07-11
feat(mobile): live ACP agent sessions end to end (host harness, mobile UI, desktop gate)

bra1nDump:
## TL;DR

This adds an experimental way to start and control a coding-agent session running on a desktop host from the iOS app. The mobile app is a remote UI; the agent process, workspace access, permissions, and native session files stay on the desktop.

The feature is off by default behind **Desktop → Settings → Security → Enable live agent sessions (experimental)**.

```text
iOS session UI
  → relay (authenticated HTTP + WebSocket proxy)
  → host-service tRPC + session stream
  → AcpSessionManager
  → one claude-agent-acp child per active session
  → Claude's native on-disk session store
```

The detailed ACP-vs-direct-SDK walkthrough is in [the standalone comparison page](https://github.com/bra1nDump/superset/blob/docs/acp-vs-sdk-comparison/apps/mobile/plans/acp-vs-claude-sdk.html). This PR body is the shorter implementation guide for reviewers.

## What changed

| Area | Responsibility |
| --- | --- |
| `packages/host-service` | Owns adapter processes, ACP JSON-RPC, session state, permissions, the bounded update journal, WebSocket fan-out, tRPC commands, and restart resurrection. |
| `packages/host-client` (new) | Platform-neutral relay transport: SuperJSON tRPC calls, one forced-token refresh after a 401, and fresh authenticated WebSocket URLs on reconnect. |
| `packages/session-protocol` (new) | Shared session/frame contracts, input schemas, timeline fold, reconnect/dedup/reset client, and React hooks. |
| `apps/mobile` | iOS session list and thread UI: open a workspace's context menu → **Live sessions**, then drive messages, plans, nested tool calls, model/mode/effort controls, permission and question cards, cancel, pagination, and reconnect/error banners. |
| `apps/desktop` | Stores the opt-in security setting and restarts host children with `SUPERSET_ACP_SESSIONS=1`. It does not render this session UI. |

The existing Mastra chat path is not rewritten by this PR; live ACP sessions have a separate mobile route while the feature is evaluated.

### Create and run a turn

The public Superset session id is different from the adapter's native session id.

```text
mobile → host tRPC  acpSessions.create { sessionId, workspaceId }
host   → adapter    initialize
host   → adapter    session/new { cwd, mcpServers: [] }
adapter→ host       { sessionId: <native ACP id>, modes, configOptions }
host   → mobile     { sessionId: <Superset id>, status: "idle", ... }
```

[...]

PR #6367, opened 2026-08-12
fix(host-service,shared): stop preset launches stalling 15s and silently losing typed commands across shell setups

### LINKED ISSUE #3941: [bug] "claude" and "codex"buttons broken
#### Describe the bug
On the preset bar claude and codex buttons are broken baceause when I press on them I see a new terminal open up with these errors:

claude:
```
 laude --permission-mode acceptEdits
zsh: command not found: laude
```

codex:
```
odex -c model_reasoning_effort="high" -c model_reasoning_summary="detailed" -c model_supports_reasoning_summaries=true --
full-auto
zsh: command not found: odex
```

as you can see these comands are missing the first "c" letter in both codex and claude, you are running "odex" and "laude"

### LINKED ISSUE #4705: [bug] Agent launch command uses bash-only heredoc syntax, fails in fish shell

### LINKED ISSUE #6024: [bug] If the default terminal is nushell, the app doesn't work as intended
### Where did this happen?

Desktop app (macOS)

### Version

1.17.0

### What happened?

I use nushell as my default shell in my mac machine. When I tried to create a workspace and tried to execute a task, I see that claude code doesn't open properly. There is a screen recording that shows the issue clearly

### Steps to reproduce

1. Make nushell as your default shell
2. Click on new workspace
3. Write a prompt and send it to the ai (must be using claude code)

https://github.com/user-attachments/assets/3e60048d-855e-4da4-97cd-81213e41dd50

Kitenite:
## Problem

Clicking an agent preset (or running a workspace setup command) types the command into a fresh shell session. Three distinct failure mechanisms existed depending on the user's shell setup:

1. **15s stall on every launch** (SUPER-1103): the launch gate waits for the OSC 133;A prompt marker, but profiles that `exec` tmux or another shell from `.zshrc` prevent the marker from ever reaching the PTY stream — so the 15s fallback timeout was the *normal* path, every single click. Diagnosed live on a customer call (Nourish, Aug 11).
2. **First character eaten** (#3941, #5712): startup stdin readers (oh-my-zsh update prompt) consume early typed bytes — `claude` becomes `laude`.
3. **Command silently discarded**: ungated shells (bash without wrapper rcfile, nushell #6024) type immediately; nushell's reedline flushes queued typeahead on raw-mode init, dropping the command entirely — even on clean configs. fish never got that far: the prompt transport's bash heredoc is a parse error in fish (#4705).

## Fix (three commits)

- **`e74599c` — learned shell-ready evidence**: host-service records what the marker scanner actually observed per shell. First timed-out launch brands the shell `missing`; later launches use a ~2s grace instead of 15s; an observed marker flips it back to `delivered`. Self-healing both directions.
- **`67a51a7` — echo-verified grace launches**: grace-window typing is verified against the PTY stream using the mangle signature (a stolen-prefix re-echo after the intact kernel echo — plain echo matching is fooled by kernel echo during raw-mode reads). On corruption: Ctrl-U, retype, up to 2 retries, then fall back to the old blind path.
- **`77b14cf` — universal ungated protection + fish transport**: ungated launches now wait for output quiescence, type, then echo-verify (including reedline-flush detection and headless DSR answering for nushell). The agent prompt transport is shell-aware: fish gets a staged-file transport instead of heredocs, byte-identical behavior preserved for bash/zsh/sh.

## Verification

Every scenario run end-to-end through a real pty-daemon + real shells (`/bin/zsh`, `/bin/bash`, fish 4.x, nu 0.114, real tmux) with the desktop's actual wrapper files, measuring latency **and** command integrity, before and after:

| Profile class | Before | After |
|---|---|---|
| healthy zsh | 0.6s | 0.6s |
| .zshrc execs shell / tmux (SUPER-1103) | **15.5s every launch** | 15.5s once, then ~2.8s |
| oh-my-zsh update prompt (#3941/#5712) | first char eaten | ~2.6s intact |
| stdin eater + no marker | 15.5s intact / corrupted under naive grace | 15.5s once, then 2.8s intact |
| bash wrapped + eater | — | 2.5s intact |
| bash unwrapped + eater | **corrupted every launch (~0.5s)** | 1.5s intact |
| fish heredoc transport (#4705) | **never executes** | 2.6s intact |
| nushell (#6024) | **command discarded, even clean** | ~2.1s intact |

New node-tests cover all three mechanisms (each proven to fail without its fix). `bun test src/terminal` 122/122, typecheck clean, lint 0. Pre-existing failures unrelated and reproduced on main: send-snapshot tRPC strip-types import (#5580), create-on-attach exit hang.

Closes #3941, closes #4705, closes #6024. Also fixes the class behind #5712 (closed). Refs SUPER-1103.

## Follow-ups (not this PR)
- Foreground-pgrp readiness probe to eliminate the one-time 15s learning hit.
- Telemetry on evidence transitions (`delivered` → `missing`) to count affected machines in the wild.
- #6308 control-chars-on-relaunch is the session-restore query-replay family, untouched here.

---
## Summary by cubic
Stops 15s stalls and silently dropped preset commands across zsh/bash/fish/nushell by learning when the OSC 133;A prompt marker won’t arrive, verifying echo before Enter, protecting ungated shells, rewriting fish prompt transport, and answering headless DSR. Old: marker‑gated launches often waited 15s and ungated shells typed immediately; fish failed to parse heredocs; headless nushell wedged on DSR. New: learned ~2s grace with self‑healing evidence, echo‑verified typing and safe retype, quiescence + echo checks for ungated shells, fish staged‑file transport, and DSR answers; side effect: small added latency on first learning and when echo checks retype.

- Evidence learning (`host-service`): persist per‑shell state in `$SUPERSET_HOME_DIR/shell-ready-evidence.json`. A timeout with output brands `missing`; later launches use ~2s grace. Any observed marker (including after the grace, within the 15s budget) flips back to `delivered`. Zero‑output timeouts do not write; unexpected read errors warn once.
- Grace‑window launches: watch for our echo, detect mangle (orphaned suffix), then Ctrl‑U and retype up to 2 times; fall back to blind Enter. All deferred PTY writes are guarded so disposed sessions don’t throw.
- Ungated shells (bash/sh/ksh/nushell): wait for output quiescence (≤8s cap), type, then echo‑verify; detect stdin‑reader mangling and reedline flush via “unexpected tail,” re‑quiesce, Ctrl‑U, and retype. Short/non‑ASCII commands still quiesce even without a probe.
- Fish support (`shared`, `host-service`): parse heredoc prompt commands and replace with a staged‑file fish command that self‑deletes; on staging failure, show an in‑terminal error and do not type the heredoc.
- Headless readiness: answer `ESC[6n` cursor queries from the mirrored screen so nushell reaches a prompt.
- Tests/CI: add Node‑based terminal suites (shell‑ready learning, grace echo, ungated protection, fish transport) and run them in CI; use a lockfile‑pinned `prebuild-install` to fetch `better-sqlite3` for Node 22; setup‑scripts fake PTY now echoes writes.

Addresses SUPER-1103; fixes #3941, #4705, #6024.

<sup>Written for commit d42f0a218a3270ff981088457fb8618e8ffbd140. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6367?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Improved terminal startup reliability by detecting shell readiness and adapting launch timing.
  - Added adaptive handling for shells with delayed or missing readiness markers, including learned behavior across sessions.
  - Added reliable prompt delivery for fish, including automatic temporary-file cleanup.
  - Improved cursor-position handling for unattached terminal sessions.
  - Added command-echo verification, output settling, and retry safeguards when startup output consumes input.

- **Bug Fixes**
  - Improved prompt delivery consistency across argument and standard-input transport methods.
  - Improved handling of delayed shell responses and late readiness signals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,26 @@ jobs:
           RELAY_URL: ${{ secrets.RELAY_URL }}
         run: bun run test
 
+      # The terminal launch-protection suites (src/terminal/*.node-test.ts)
+      # need real PTYs + node:test, which bun can't provide.
+      - name: Setup Node.js (terminal node-tests)
+        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        with:
+          node-version: 22
+
+      # bun install ran with --ignore-scripts, so better-sqlite3 has no native
+      # build; fetch the prebuilt binary for the Node ABI (bun tests use
+      # bun:sqlite and are unaffected).
+      - name: Prepare better-sqlite3 for Node
+        run: |
+          for dir in node_modules/.bun/better-sqlite3@*/node_modules/better-sqlite3; do
+            (cd "$dir" && npx --yes prebuild-install || npx --yes node-gyp rebuild)
+          done
+
+      - name: Terminal node-tests
+        working-directory: packages/host-service
+        run: bun run test:integration:terminal
+
   typecheck:
     name: Typecheck
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,12 +135,13 @@ jobs:
           node-version: 22
 
       # bun install ran with --ignore-scripts, so better-sqlite3 has no native
-      # build; fetch the prebuilt binary for the Node ABI (bun tests use
+      # build; fetch the prebuilt binary for the Node ABI using the lockfile-
+      # pinned prebuild-install that ships as its dependency (bun tests use
       # bun:sqlite and are unaffected).
       - name: Prepare better-sqlite3 for Node
         run: |
           for dir in node_modules/.bun/better-sqlite3@*/node_modules/better-sqlite3; do
-            (cd "$dir" && npx --yes prebuild-install || npx --yes node-gyp rebuild)
+            (cd "$dir" && ../.bin/prebuild-install)
           done
 
       - name: Terminal node-tests
--- a/packages/host-service/package.json
+++ b/packages/host-service/package.json
@@ -58,6 +58,7 @@
 		"test": "bun test --pass-with-no-tests",
 		"test:integration": "bun test --pass-with-no-tests test/integration",
 		"test:integration:daemon": "node --experimental-strip-types --test src/terminal/DaemonClient/DaemonClient.node-test.ts src/daemon/DaemonSupervisor.node-test.ts",
+		"test:integration:terminal": "node --experimental-strip-types --test src/terminal/terminal.initial-command.node-test.ts src/terminal/terminal.ungated-launch.node-test.ts src/terminal/terminal.shell-ready-learning.node-test.ts src/terminal/terminal.fish-prompt-transport.node-test.ts",
 		"test:integration:workers": "node --experimental-strip-types --test src/workers/worker-termination.node-test.ts",
 		"test:e2e": "bun run scripts/test-e2e.ts",
 		"bench": "bun test test/integration/pull-requests-scaling.bench.ts",
--- a/packages/host-service/src/terminal/shell-ready-evidence.test.ts
+++ b/packages/host-service/src/terminal/shell-ready-evidence.test.ts
@@ -1,0 +1,95 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import {
+	__resetShellReadyEvidenceForTesting,
+	getShellReadyMarkerEvidence,
+	recordShellReadyMarkerEvidence,
+} from "./shell-ready-evidence";
+
+const originalHomeDir = process.env.SUPERSET_HOME_DIR;
+let tmpDir: string;
+
+beforeEach(() => {
+	tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "shell-ready-evidence-"));
+	process.env.SUPERSET_HOME_DIR = tmpDir;
+	__resetShellReadyEvidenceForTesting();
+});
+
+afterEach(() => {
+	if (originalHomeDir === undefined) {
+		delete process.env.SUPERSET_HOME_DIR;
+	} else {
+		process.env.SUPERSET_HOME_DIR = originalHomeDir;
+	}
+	__resetShellReadyEvidenceForTesting();
+	fs.rmSync(tmpDir, { recursive: true, force: true });
+});
+
+async function waitForFile(filePath: string): Promise<void> {
+	for (let i = 0; i < 100; i++) {
+		try {
+			JSON.parse(fs.readFileSync(filePath, "utf8"));
+			return;
+		} catch {
+			await new Promise((r) => setTimeout(r, 10));
+		}
+	}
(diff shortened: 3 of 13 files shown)


Kitenite:
**Before/after demo** — real recordings of the launch path (real zsh + pty-daemon + host-service checked out at each commit), PTY byte stream replayed at original pacing. Profile under test: user `.zshrc` ends in `exec zsh -f`, so the OSC 133;A ready marker never reaches the scanner (the SUPER-1103 class). Left: pre-branch (`9b1216481`) — the prompt paints in ~50ms, then the preset command sits on the 15s fallback. Right: this branch with learned `missing` evidence — command lands echo-verified at ~2s. The `${SECONDS}` in the command is the shell reporting its own age at execution time.


https://github.com/user-attachments/assets/97734cc2-29c9-473b-9ace-8f837f78fa78

Issue #6370, opened 2026-08-12
[bug] Image drops insert a single-quoted path instead of attaching the image in the agent TUI


caiokf:
### Where did this happen?

Desktop app (macOS)

### Version

1.18.1

### What happened?

Dropping an image into a terminal running Claude Code inserts the file path as plain text, wrapped in single quotes:

```
'/var/folders/2d/.../Screenshot 2026-08-12 at 09.44.31.png'
```

Claude Code previously turned the same drop into an inline image attachment, shown in its prompt as `[Image #1]`. The image is no longer attached.

I think this is a side effect of #2181, which added path insertion on drop and closed 2026-07-11. Superset now intercepts the drop and writes text into the PTY, so the TUI's own drop handling never sees the event. My app bundle is dated 2026-07-28, which lines up with when the behaviour changed for me.

The single quotes may matter independently: Terminal.app and iTerm2 insert a bare path with backslash-escaped spaces, and a TUI testing whether the dropped text is a readable image path would fail on the quoted form.

Expected: dropping an image into an agent that supports image attachments attaches the image. Failing that, insert a bare backslash-escaped path so TUI path detection can match it, or make the behaviour configurable.

### Steps to reproduce

1. Open a Superset terminal and start Claude Code
2. Drag an image file into the terminal input
3. A single-quoted path is inserted; no `[Image #N]` attachment appears

### Logs, screenshots, or recordings

Related: #1743 (clipboard image paste inserts the macOS Preview icon rather than the image) looks like the paste-side counterpart of the same gap.

Confidence: the quoting and the missing attachment are both directly observed. The causal link between them is my inference — I have not read Claude Code's detection logic, and the cause could instead be that Superset consumes the drop event before the TUI can handle it. Both explanations point at the #2181 interception path.

Issue #6369, opened 2026-08-12
[bug] Drag-and-drop replaces U+202F in filenames with a plain space, producing a path that does not exist


caiokf:
### Where did this happen?

Desktop app (macOS)

### Version

1.18.1

### What happened?

Dragging a file whose name contains U+202F NARROW NO-BREAK SPACE inserts a path in which that character has been replaced by U+0020 (ASCII space). The inserted path does not exist on disk, so any agent or command that tries to read it fails.

macOS uses U+202F before `am`/`pm` when formatting times under a 12-hour-clock locale, so every screenshot filename hits this:

```
Screenshot 2026-08-12 at 9.42.51 am.png
```

Expected: the inserted text reproduces the filename byte-for-byte. U+202F is a valid filename character and should pass through unmodified.

### Steps to reproduce

1. Use a 12-hour-clock locale (mine is `en_AU`; System Settings → General → Date & Time, "24-hour time" off)
2. Take a screenshot with Cmd+Shift+4, so macOS names it `... at H.MM.SS am.png`
3. Drag the file into a Superset terminal running any agent
4. The path is inserted, but the agent reports that the file does not exist

Workaround: switching macOS to 24-hour time drops `am`/`pm` from screenshot filenames, so the character never appears. Files containing U+202F for other reasons still fail.

### Logs, screenshots, or recordings

Filename bytes on disk vs. the text produced by the drop:

```
on disk:  39 2e 34 32 2e 35 31  e2 80 af  61 6d 2e 70 6e 67    "9.42.51" U+202F "am.png"
dropped:  39 2e 34 32 2e 35 31  20        61 6d 2e 70 6e 67    "9.42.51" U+0020 "am.png"
```

The file is present and readable under its true name:

```console
$ [ -e "<dropped path>" ] && echo YES || echo NO
NO
$ for f in "$DIR"/*.png; do [ -r "$f" ] && echo "READABLE: $f"; done
READABLE: /var/.../Screenshot 2026-08-12 at 9.42.51 am.png
```

PR #6373, opened 2026-08-12
fix(desktop): deliver v1 terminal file drops as a paste, not raw input

caiokf:
## What & why

A file dropped on a **v1** terminal pane was written straight to the PTY:

```ts
writeRef.current({ paneId, data: text });
```

so the foreground app saw the path as typing. Agent TUIs use the
bracketed-paste envelope to tell a dropped path apart from a typed one, so the
path stayed literal text and a dropped image was never attached. Related to
#6370.

I measured this against a real Claude Code TUI (2.1.215) in a PTY — tmux,
120x45, `tmux paste-buffer -p` for the paste rows and `tmux send-keys -l` for
the typed rows, dropping `/tmp/drop-probe/scratch/My Screenshot.png` which
exists on disk:

| inserted text | delivery | result |
| --- | --- | --- |
| `'/tmp/…/My Screenshot.png'` (shell-quote) | bracketed paste | `❯ [Image #1]` |
| `/tmp/…/My\ Screenshot.png` (backslash) | bracketed paste | `❯ [Image #1]` |
| `'/tmp/…/My Screenshot.png'` | typed | `❯ '/tmp/…/My Screenshot.png'` |
| `/tmp/…/My\ Screenshot.png` | typed | `❯ /tmp/…/My\ Screenshot.png` |

Delivery decides it. The escaping style makes no difference — worth recording,
because #6370 guessed the single quotes were the problem and they are not. I
also checked that a bracketed paste of a path to a **missing** file stays
literal text, which is the same symptom.

v2 panes already call xterm's `paste()`, and so does Cmd+V in a v1 pane. This
routes the v1 drop through the same call.

## Scope

This covers v1 panes only, and I want to be straight about what it does not
prove. v2 (`TerminalPane`) already pastes, so the v2 half of #6370 has a
different cause — most likely that the dropped path does not exist, since both
#6370 and #6369 drag from the macOS screenshot floating thumbnail and land on
a `/var/folders/…/NSIRD_screencaptureui_*` temp path. I have not been able to
reproduce that half yet and did not want to guess at a fix for it in the same
PR.

There is prior art for that temp path being trouble: crbug 1148078 recorded
`/var/folders/.../TemporaryItems/NSIRD_screencaptureui_.../Screen Shot ....png`
drags failing with `ERR_ACCESS_DENIED`, and Chrome 93 fixed it by copying the
file into a Chrome-owned temp directory
(`DroppedScreenShotCopierMac`, `content/app_shim_remote_cocoa/web_contents_view_cocoa.mm`)
— which is app-shim-specific and should not be assumed to run in an ordinary
Electron window. Reading `File.arrayBuffer()` on drop and writing our own temp
copy would sidestep the whole class, but that is a bigger change than this one
and belongs in its own PR.

## How I tested it

- The PTY matrix above — real `claude` process, real bracketed-paste bytes, not
  a mock. Script: a ~40-line tmux driver, happy to paste it in if useful.
- `bun run lint` → clean.
- `bun run typecheck --filter=@superset/desktop` → clean.
- `bun run test` → `2677 pass, 1 fail`; the failure is
  `Shell Environment > getProcessEnvWithShellPath applies shell PATH and
  preserves string vars`, which asserts against this machine's real shell PATH
  and fails the same way on an unmodified checkout.

No screenshot: I could not get the desktop dev app running in this clone
(`bun install` fails building `better-sqlite3` here), so the end-to-end
evidence is the PTY matrix rather than the app UI. Happy to redo it in the app
if a maintainer would rather see that.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file drop handling in active terminals by using paste behavior, preserving expected input formatting.
  * Prevented dropped content from being processed after a terminal has exited.

### DIFF
--- a/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/Terminal.tsx
+++ b/apps/desktop/src/renderer/screens/main/components/WorkspaceView/ContentView/TabsContent/Terminal/Terminal.tsx
@@ -438,9 +438,13 @@ export const Terminal = memo(function Terminal({
 			if (!plainText) return;
 			text = shellEscapePaths([plainText]);
 		}
-		if (!isExitedRef.current) {
-			writeRef.current({ paneId, data: text });
-		}
+		if (isExitedRef.current) return;
+		// Paste instead of writing raw. xterm wraps the text in bracketed-paste
+		// markers when the foreground app has asked for them, and that envelope
+		// is how an agent TUI tells a dropped path apart from typing — Claude
+		// Code attaches a pasted image path and leaves a typed one as text.
+		// v2 terminal panes already go through paste(); so does Cmd+V here.
+		xtermRef.current?.paste(text);
 	};
 
 	return (


Kitenite:
Closing: this only touches the legacy v1 UI (screens/main), which is being retired. only touches v1 Terminal.tsx; v2 TerminalPane already pastes drops. If the bug still shows in the v2 workspace, please open an issue.

Kitenite:
Thanks for the paste-not-raw-input fix. It only touches the v1 terminal, which is being retired; the v2 pane already pastes drops.

PR #6381, opened 2026-08-12
fix(host-service,cli): release a workspace's git branch on delete so its name is reusable

### LINKED ISSUE #6380: [bug] Deleting a workspace leaves its git branch behind, so reusing the workspace name creates a mismatched branch
## What happened

`superset ws delete` removes the workspace record and the worktree directory, but it leaves the local git branch behind in the parent repository.

Because the branch still exists, creating a new workspace with the same name does not reuse the name. Superset silently picks a suffixed branch instead: `<name>-2`, then `<name>-3`, and so on. The workspace `name` stays the same, so the workspace name and its branch no longer agree.

In the desktop app the same flow is worse: after I delete a workspace and create a new one with the same name, the app lands on a "Workspace not found" screen showing the old workspace ID (`aa511abc-07ec-41e6-9976-45e603fb170b`), and I cannot get into the new workspace.

I expect a deleted workspace to release its branch, so the same name is genuinely free to use again.

## Steps to reproduce

1. Create a workspace:
   `superset ws create --local --project <id> --name dupe-repro --branch dupe-repro`
   Result: `"branch": "cry0genic/dupe-repro"`
2. Delete it: `superset ws delete <workspace-id> --local`
   Result: `{"deleted": ["..."], "warnings": []}` — no warning about the branch.
3. Run `git branch --list 'cry0genic/dupe*'` in the parent repository.
   Result: `cry0genic/dupe-repro` is still there.
4. Create the workspace again with the identical name and branch.
   Result: `"name": "dupe-repro"` but `"branch": "cry0genic/dupe-repro-2"`.
5. Repeat steps 2-4. The suffix increments to `-3`, and every deleted branch stays.

Observed branch list after three cycles:

```
cry0genic/dupe-repro
cry0genic/dupe-repro-2
cry0genic/dupe-repro-3
```

The worktree directories under `~/.superset/worktrees/<project>/` are removed correctly, and the project teardown script runs correctly. Only the branch survives.

## Expected

One of:

- Delete the branch with the workspace when the branch holds no unmerged work, or
- Ask the user whether to keep or delete the branch, or
- At minimum, report a warning that names the leftover branch, instead of silently renaming the new workspace's branch.

## Environment

- Superset CLI: 1.20.2
- Superset desktop app: 1.20.2
- OS: macOS (Darwin arm64)
- Workspace type: `worktree`

## Related

- #5076 (deleted workspaces reappear in the sidebar) — different symptom, possibly the same stale-state area.
- #5497 (batch delete leaves ghosts) — different cause.

andyst-dev:
Fixes #6380

## Problem

`workspace.delete` — the legacy surface used by the CLI, SDK, and MCP — hardcoded `deleteBranch: false`. Deleting a workspace therefore removed the worktree and the row but left the local git branch in the parent repo.

Because the branch survived, creating a new workspace with the same name silently picked a suffixed branch (`<name>-2`, then `-3`, ...) while the workspace `name` stayed the same, so the name and branch no longer agreed. In the desktop app the flow was worse: delete + recreate with the same name landed on a stale "Workspace not found" screen showing the old workspace ID.

## Root cause

`packages/host-service/src/trpc/router/workspace/workspace.ts` — the `delete` mutation always passed `deleteBranch: false`:

```ts
return destroyWorkspace(ctx, {
    workspaceId: input.id,
    deleteBranch: false,
    force: true,
    teardownMode: "best-effort",
});
```

The v2 `destroy` path (used by the renderer) already lets the caller choose; the legacy `delete` surface never did, and the cleanup saga only deletes the branch when `input.deleteBranch` is set (Step 5, `workspace-cleanup.ts:520`).

## Fix

- **`workspace.delete`** now accepts an optional `deleteBranch` (default `true`), so a deleted workspace releases its branch by default and the name is genuinely free to reuse. Callers that need to keep the branch (e.g. unmerged work they intend to salvage) pass `false` explicitly.
- **`superset ws delete`** keeps its existing non-interactive contract (still defaults to deleting the branch) and adds a `--keep-branch` flag for the salvage case.

The v2 `destroy` path is untouched.

## Verification

- `tsc --noEmit` clean on `host-service`, `cli`, `sdk`, and `mcp` (the SDK/MCP/CLI all consume `workspace.delete` and compile against the new optional field).
- `biome check` clean on both edited files.
- The existing `destroy`/`deleteBranch` behavior in the v2 path and its integration tests are unchanged.

---
## Summary by cubic
Fixes #6380 by aligning workspace deletion with the desktop UI: deleting a workspace now keeps its git branch by default, and an opt-in flag releases the name for reuse.

- `workspace.delete` accepts optional `deleteBranch`, defaulting to `false` so silent callers don't lose unpushed commits.
- `superset ws delete` adds `--delete-branch` to opt into branch removal; the flag is rejected with `--cloud` because cloud sandboxes have no workspace git branch to release.

<sup>Written for commit db60190c7d9fe9b070ff9896fd5e05cba1ee0f69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6381?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added the `--delete-branch` option to workspace deletion.
  * Workspace branches are kept by default when deleting a workspace.
  * Branches are deleted only when `--delete-branch` is explicitly specified.

* **Bug Fixes**
  * Workspace cleanup now consistently follows the selected branch deletion behavior.
  * Deleted workspaces that retain their branches no longer block creating a new workspace with the same branch.

### DIFF
--- a/packages/cli/src/commands/workspaces/delete/command.ts
+++ b/packages/cli/src/commands/workspaces/delete/command.ts
@@ -9,8 +9,8 @@ export default command({
 	options: {
 		host: string().desc("Host the workspaces live on"),
 		local: boolean().desc("Target this machine (the default)"),
-		"keep-branch": boolean().desc(
-			"Keep the workspace's git branch after delete (default: the branch is deleted)",
+		"delete-branch": boolean().desc(
+			"Delete the workspace's git branch after delete (default: the branch is kept, matching the desktop UI)",
 		),
 	},
 	run: async ({ ctx, args, options }) => {
@@ -37,7 +37,7 @@ export default command({
 		for (const id of ids) {
 			const result = await target.client.workspace.delete.mutate({
 				id,
-				deleteBranch: !options["keep-branch"],
+				deleteBranch: options["delete-branch"] === true,
 			});
 			deleted.push(id);
 			for (const warning of result.warnings ?? []) {
--- a/packages/cli/src/commands/workspaces/delete/command.ts
+++ b/packages/cli/src/commands/workspaces/delete/command.ts
@@ -9,6 +9,9 @@ export default command({
 	options: {
 		host: string().desc("Host the workspaces live on"),
 		local: boolean().desc("Target this machine (the default)"),
+		"keep-branch": boolean().desc(
+			"Keep the workspace's git branch after delete (default: the branch is deleted)",
+		),
 	},
 	run: async ({ ctx, args, options }) => {
 		const ids = args.ids as string[];
@@ -32,7 +35,10 @@ export default command({
 		const deleted: string[] = [];
 		const warnings: string[] = [];
 		for (const id of ids) {
-			const result = await target.client.workspace.delete.mutate({ id });
+			const result = await target.client.workspace.delete.mutate({
+				id,
+				deleteBranch: !options["keep-branch"],
+			});
 			deleted.push(id);
 			for (const warning of result.warnings ?? []) {
 				warnings.push(`${id}: ${warning}`);
--- a/packages/host-service/src/trpc/router/workspace/workspace.ts
+++ b/packages/host-service/src/trpc/router/workspace/workspace.ts
@@ -176,7 +176,16 @@ export const workspaceRouter = router({
 		}),
 
 	delete: protectedProcedure
-		.input(z.object({ id: z.string() }))
+		.input(
+			z.object({
+				id: z.string(),
+				// A deleted workspace releases its branch by default so the same
+				// name is genuinely free to reuse (#6380). Callers that need to
+				// keep the branch (e.g. unmerged work they intend to salvage)
+				// pass false explicitly.
+				deleteBranch: z.boolean().optional(),
+			}),
+		)
 		.mutation(async ({ ctx, input }) => {
 			// Legacy external surface used by CLI/SDK/MCP. Preserve its
 			// non-interactive contract while reusing the v2 cleanup path:
@@ -185,7 +194,7 @@ export const workspaceRouter = router({
 			// is nobody to prompt for a force-retry (#6174).
 			return destroyWorkspace(ctx, {
 				workspaceId: input.id,
-				deleteBranch: false,
+				deleteBranch: input.deleteBranch ?? true,
 				force: true,
 				teardownMode: "best-effort",
 			});
--- a/packages/host-service/src/trpc/router/workspace/workspace.ts
+++ b/packages/host-service/src/trpc/router/workspace/workspace.ts
@@ -179,10 +179,12 @@ export const workspaceRouter = router({
 		.input(
 			z.object({
 				id: z.string(),
-				// A deleted workspace releases its branch by default so the same
-				// name is genuinely free to reuse (#6380). Callers that need to
-				// keep the branch (e.g. unmerged work they intend to salvage)
-				// pass false explicitly.
+				// Match the desktop UI: a deleted workspace keeps its branch by
+				// default (the checkbox is unchecked, deleteLocalBranch
+				// defaults false). Non-interactive CLI/SDK/MCP callers that want
+				// the branch released pass deleteBranch: true explicitly, so a
+				// silent caller never loses unpushed commits (#6380, review
+				// feedback on #6381).
 				deleteBranch: z.boolean().optional(),
 			}),
 		)
@@ -194,7 +196,7 @@ export const workspaceRouter = router({
 			// is nobody to prompt for a force-retry (#6174).
 			return destroyWorkspace(ctx, {
 				workspaceId: input.id,
-				deleteBranch: input.deleteBranch ?? true,
+				deleteBranch: input.deleteBranch ?? false,
 				force: true,
 				teardownMode: "best-effort",
 			});


andyst-dev:
Good catch — you're right that deleting with the new default leaves the DB row as an archived tombstone, and `findExistingWorkspaceByBranch` would match it on `(projectId, branch)` and hand back a stale workspace on same-name recreate.

That exact reuse path is addressed in the sibling PR #6384 (Fixes #6383), which adds `isNull(workspaces.archivedAt)` to `findExistingWorkspaceByBranch` so archived rows are never returned as "existing":

```ts
where: and(
  eq(workspaces.projectId, projectId),
  eq(workspaces.branch, branch),
  isNull(workspaces.archivedAt),   // #6384 — exclude tombstones
),
```

So the full class is: #6381 releases the git branch on delete, #6384 makes same-name recreate return a fresh workspace instead of the tombstone. Together they cover both halves of the bug (#6380 / #6383).

g-bastianelli:
Thanks for picking up #6380 — the suffixed-branch behaviour it describes is real and worth fixing.

One thing worth weighing before flipping the default, though: **the desktop deliberately keeps the branch unless the user opts in**, in four places:

- `dashboardSidebarLocal/schema.ts:360` — `deleteLocalBranch: z.boolean().default(false)`
- `dashboardSidebarLocal/schema.ts:388` — `deleteLocalBranch: false` in the preference defaults
- `DeleteWorkspaceDialog.tsx:56` — `deleteLocalBranch ?? deleteLocalBranchDefault ?? false`
- `workspace-cleanup.ts:192` — `workspaceCleanup.destroy`, the renderer's path, is `deleteBranch: z.boolean().default(false)`

So today the checkbox is unchecked by default and the branch survives. `deleteBranch: input.deleteBranch ?? true` on the legacy surface would make the CLI, SDK and MCP do the opposite of the UI for the same operation — and since step 5 runs `git branch -D`, an SDK or MCP caller that never passed the field would start losing unpushed commits without changing a line of their own code.

One note on #6380: arguably the root cause of the suffixed-branch symptom is on the create side — silently picking `<name>-2` instead of reusing the free name or telling the user the branch is taken.

*(Edited: I'd first written that the stale "Workspace not found" symptom was unaddressed and looked like a renderer bug. That was wrong on both counts — I'd missed your Aug 12 comment, and #6384 covers exactly that half via `findExistingWorkspaceByBranch`. Sorry for the noise; the default mismatch below is my real concern.)*

I have #6246 open since Aug 6 on the same two files, doing the opt-in version (`--delete-branch`), with integration tests against a real `git worktree` fixture and the CLI reference updated. Happy to fold my tests and docs into whichever direction you'd rather take — I mainly care that the CLI and the desktop agree. If the team does want deletion to become the default, that seems like a product decision worth making in both surfaces at once rather than only on the non-interactive one.

andyst-dev:
Thanks — agreed, this is a product/API consistency decision rather than a small implementation detail.

I won’t silently flip the legacy default while the desktop keeps branch deletion opt-in: that could make CLI/SDK/MCP callers delete unpushed branches without changing their invocation. The current PR addresses the issue by making the non-interactive path delete by default, while #6246 takes the opt-in direction with integration coverage and docs.

The maintainers should choose one policy and apply it consistently across the CLI, SDK/MCP surfaces, and desktop. Until that direction is confirmed, I’m leaving the code unchanged rather than duplicating #6246 or making the surfaces diverge further.

andyst-dev:
Thanks again for the detailed review — I've taken **option 1 (default-false everywhere)** and updated the PR to head `ace6fa17c`.

Changes:
- **host-service** (`workspace.ts`): `deleteBranch` now defaults to **false** (`input.deleteBranch ?? false`) → a silent CLI/SDK/MCP caller keeps the branch, matching the desktop's four `deleteLocalBranch: false` sites. Releasing the branch requires passing `deleteBranch: true` explicitly.
- **CLI** (`command.ts`): replaced `--keep-branch` (default delete) with a **`--delete-branch` opt-in** (default keep), so the interactive CLI and the host-service contract agree.

This removes the "SDK/MCP loses unpushed commits without opting in" trap your review identified. Existing integration tests (workspace-cleanup, session-create-delete) already pass `deleteBranch` explicitly, so none assert the old implicit default and all remain green.

Since this is now default-keep, I'll leave the branch-release behavior explicitly opt-in; if the team later wants deletion as the default, that's a product decision best made uniformly across desktop + non-interactive surfaces (your #6246/#6380 framing). Happy to fold in your integration-test/docs if you'd rather steer it there.

andyst-dev:
P1 addressed in commit `b1fa665`. `findExistingWorkspaceByBranch` (the idempotency lookup that a same-name recreation hits) now requires `archivedAt IS NULL`, so a tombstone that kept its branch (the new keep-branch default) can never satisfy the reuse hit — the caller gets a fresh workspace created instead of a stale, archived id. This mirrors the `isNull(workspaces.archivedAt)` filter already applied by the seven other workspace lookups (search-branches, list-project-worktrees, adopt-existing-worktree), which is exactly the reuse flow you flagged.

andyst-dev:
Addressed in `4ae6dc8` — thank you, and you are right that the flag could report success without doing anything: the cloud path calls `cloudWorkspace.delete` with only `{ id }`, and that procedure has no branch-deletion contract at all.

I took the reject option rather than adding a cloud API contract, since a cloud sandbox has no workspace git branch to release in the first place. The combination now fails fast, mirroring the guard that was already there for `--host`/`--local`:

```ts
if (options["delete-branch"] === true) {
    throw new CLIError(
        "--delete-branch cannot be combined with --cloud",
        "Cloud sandboxes have no workspace git branch to release",
    );
}
```

I also rebased the branch onto current `main` while I was here: the tombstone part of this PR had been absorbed upstream (`findExistingWorkspaceByBranch` already filters `isNull(workspaces.archivedAt)` on main, with the same reasoning in its comment), so that commit dropped out of the rebase and the diff is now just the two remaining pieces — the host-service `deleteBranch` input defaulting to keep-branch, and this CLI opt-in plus the new guard.

Validation on the rebased head: `bun test packages/host-service/test/workspace-cleanup.test.ts` 34/34, `tsc --noEmit` on `packages/cli` reports 0 errors, biome clean.

PR #6384, opened 2026-08-12
fix(host-service): don't reuse an archived (tombstoned) workspace on create

### LINKED ISSUE #6383: workspaces.create reuses tombstoned (archived) workspaces: create-after-delete on the same branch returns a dead row, creates no worktree, and still dispatches the agent
## Summary

Since PR #6296 made workspace deletes tombstone rows instead of removing them, `workspaces.create` can "reuse" a deleted workspace. Its branch-idempotency lookup, `findExistingWorkspaceByBranch`, matches on `(projectId, branch)` without excluding archived rows, so a create on a branch whose previous workspace was deleted returns the tombstone: exit code 0, `alreadyExists: true`, message `Reused existing workspace "…"`, no git worktree created — and the requested agent is still dispatched, against a `worktreePath` the delete pipeline already removed.

The archived row is hidden from `workspaces list`, so from the caller's side the workspace simply does not exist while the CLI reports success.

## Environment

- Superset Desktop 1.20.0 and 1.20.2, CLI 1.20.0
- macOS arm64, `--local` workspaces
- Reproduced by code inspection at `desktop-v1.20.2` and on `main`

## Steps to reproduce

1. Create a workspace on a fresh branch, with an agent:
   ```
   superset workspaces create --local --project <projectId> \
     --name repro-1 --branch repro-tombstone --skip-branch-prefix \
     --agent claude --prompt "pwd"
   ```
   A worktree is created and the agent starts. Note the workspace id.
2. Delete it (leave the git branch in place — the default):
   ```
   superset workspaces delete --local <workspaceId>
   ```
3. Confirm it is gone from the caller's view:
   ```
   superset workspaces list --local --json
   ```
   The branch no longer appears.
4. Run the exact same create as step 1 again (same `--branch`):
   ```
   superset workspaces create --local --project <projectId> \
     --name repro-2 --branch repro-tombstone --skip-branch-prefix \
     --agent claude --prompt "pwd"
   ```

## Expected

Step 4 creates a new workspace and worktree on `repro-tombstone` and starts the agent — identical to step 1, since as far as every visible surface is concerned no workspace exists on that branch.

## Actual

Step 4 exits 0 and prints `Reused existing workspace "repro-1" on host <hostId>`. The response carries `alreadyExists: true` and the **archived** row, whose `worktreePath` points at the directory removed in step 2. No worktree is created. The setup terminal is skipped (it is gated on `!alreadyExists`), but the agent is dispatched anyway against the missing path; its failure is captured into `agents[].ok === false` rather than failing the call. The workspace remains invisible to `workspaces list`, so a subsequent poll finds nothing and a retry of step 4 reproduces identically, forever.

## Root cause

`packages/host-service/src/trpc/router/workspaces/workspaces.ts:156`

```ts
function findExistingWorkspaceByBranch(
  ctx: HostServiceContext, projectId: string, branch: string,
): CloudWorkspace | null {
  const local = ctx.db.query.workspaces.findFirst({
    where: and(
      eq(workspaces.projectId, projectId),
      eq(workspaces.branch, branch),
      // no isNull(workspaces.archivedAt)
    ),
  }).sync();
  return local ? toCloudShape(local, ctx.organizationId) : null;
}
```

The whole file contains no reference to `archivedAt`. It is called from both creation paths — `workspaces.ts:594` (`--pr`) and `workspaces.ts:907` (`--branch`) — each doing:

```ts
const existing = findExistingWorkspaceByBranch(ctx, input.projectId, resolvedBranch);
if (existing) { workspaceRow = existing; alreadyExists = true; }
```

Both checks run *before* the worktree lookup, so the tombstone wins over any real git state.

For contrast, the sibling lookup in the same creation flow does filter — `packages/host-service/src/trpc/router/workspace-creation/shared/adopt-existing-worktree.ts:118`:

```ts
where: and(
  eq(workspaces.projectId, projectId),
  eq(workspaces.branch, branch),
  isNull(workspaces.archivedAt),   // <-- present here
)
```

Those `isNull` guards (lines 123, 142, 223) were added *by PR #6296 itself*; `workspaces/workspaces.ts` was not part of that PR and never received the equivalent guard.

The agent dispatch at `workspaces.ts:1160` is ungated:

```ts
dispatchSugarAgents(ctx, workspaceRow.id, chainedAgentResult ? [] : sugarLaunches)
```

while the setup terminal (`:1131`) and the wait-for-setup chain (`:1104`) are both gated on `!alreadyExists`. `runAgentInWorkspace` (`agents.ts:419`) also looks up the workspace by id with no archived filter, so it accepts the tombstone.

## When it started

PR #6296 — *"feat(desktop,host-service): tombstone workspace deletes + kanban board view for the workspaces page"*, merged 2026-08-09.

- `workspace-cleanup.ts`: `-deleteLocalWorkspace(ctx, input.workspaceId);` → `+archiveLocalWorkspace(ctx, input.workspaceId, archiveReasonFor(ctx, local));`
- `db/schema.ts`: adds `archived_at` / `archive_reason` (+ index).
- `archivedAt` is absent from `schema.ts` at `desktop-v1.19.0` and present at `desktop-v1.20.2`.

Before #6296 a deleted workspace left no row, so the unfiltered lookup could never match one. After it, every deleted workspace leaves a permanent archived row on that branch — the tombstone is deliberately kept even on a fully successful delete (`local-workspace-store.ts:220`: *"the row itself survives for the board's Merged/Deleted history"*), so this is not limited to interrupted deletes.

`findExistingWorkspaceByBranch` itself is byte-identical and unfiltered at `desktop-v1.18.3`, `desktop-v1.19.0` and `desktop-v1.20.2` — the lookup did not change; the data underneath it did.

## Affected versions

- **Affected:** the 1.20.x line (first releases containing #6296). Verified in the shipped 1.20.0 bundle (`archived_at` present in `app.asar`) and in source at `desktop-v1.20.2` and `main`.
- **Not affected:** `desktop-v1.19.0` and earlier — the `archived_at` column does not exist.

## Suggested fix

Add the guard that the sibling already has, in `packages/host-service/src/trpc/router/workspaces/workspaces.ts:163`:

```ts
where: and(
  eq(workspaces.projectId, projectId),
  eq(workspaces.branch, branch),
  isNull(workspaces.archivedAt),
),
```

That alone restores the pre-#6296 behavior for both call sites.

Two things worth considering alongside it:

- **Gate the agent dispatch on `alreadyExists`**, consistent with the setup terminal at `:1131` and the chain at `:1104` — or, at minimum, refuse to reuse a row whose `worktreePath` no longer exists. Reusing a workspace and then launching an agent into it is a reasonable behavior for a genuinely live workspace, but there is currently no check that the reused row's worktree is on disk. `workspace.ts:70-73` already encodes the invariant ("Tombstones' worktrees are gone by definition").
- **`runAgentInWorkspace` (`agents.ts:419`)** likewise resolves the workspace with no archived filter, so an agent can be launched into a tombstone by id through other paths too. It already distinguishes NOT_FOUND for "dead workspace pin"; treating an archived row as not-found there would be consistent.

## Impact

Reported from an automation fleet that creates one workspace per job. The first job on a branch succeeds; any second job on the same branch (a re-run, or a follow-up job editing the same PR branch) gets a reused tombstone — no worktree, no agent, exit 0. The dispatcher retried indefinitely because nothing in the exit code or in `workspaces list` indicated failure. It worked for ~4 weeks and broke the day the desktop auto-updated into the 1.20.x line.

## Not verified

- The exact spawn-level failure when the agent's terminal is created against the removed `worktreePath` (ENOENT vs. an immediately-dead PTY) was not traced past `dispatchSugarAgents`, which catches any error into `agents[].ok === false`.
- No host DB was inspected by the author of this report beyond confirming that archived rows existed on the affected branches at the failing times; the causal chain above is derived from source, tags and the #6296 diff, plus the observed CLI behavior.

andyst-dev:
Fixes #6383

## Problem

Since PR #6296 made workspace deletes archive rows (tombstones) instead of removing them, `workspaces.create` could "reuse" a deleted workspace. `findExistingWorkspaceByBranch` matched on `(projectId, branch)` without excluding archived rows, so a create on a branch whose previous workspace was deleted returned the tombstone: exit code 0, `alreadyExists: true`, message "Reused existing workspace", no git worktree created — and the requested agent was still dispatched against a `worktreePath` the delete pipeline already removed.

The archived row is hidden from `workspaces list`, so from the caller's side the workspace simply doesn't exist while the CLI reports success.

## Root cause

`packages/host-service/src/trpc/router/workspaces/workspaces.ts` — `findExistingWorkspaceByBranch` (line 156) queried only on `(projectId, branch)`:

```ts
where: and(
  eq(workspaces.projectId, projectId),
  eq(workspaces.branch, branch),
),
```

It never excluded rows with a non-null `archivedAt`, so tombstones were treated as live workspaces.

## Fix

The lookup now excludes archived rows via `isNull(workspaces.archivedAt)`, so a create-after-delete on the same branch creates a fresh workspace and worktree instead of "reusing" the dead row. Because the fix lives in the shared lookup function, both call sites (`create` for PR-derived branches and for plain branch names) benefit.

## Verification

- New integration test: create → delete (archives row) → create on the exact same branch asserts `alreadyExists === false`, a new workspace id, an active (non-archived) row, and a distinct `worktreePath`.
- Full `workspace-create-delete.integration.test.ts` passes: **13/13**.
- `tsc --noEmit` clean on `host-service`; `biome check` clean on both edited files.

---
## Summary by cubic
Stop reusing archived (tombstoned) workspaces when creating on a branch that previously had a deleted workspace. Before: create-after-delete matched the tombstone, returned alreadyExists=true, skipped worktree creation, and dispatched agents to a removed `worktreePath`. Now: archived rows are ignored and a fresh workspace and worktree are created.

- In shared `findExistingWorkspaceByBranch`, add `isNull(workspaces.archivedAt)` to the `(projectId, branch)` lookup so both PR-derived and plain-branch creates ignore tombstones.
- Tests: add an integration test for create → delete → create asserting `alreadyExists === false`, a new workspace id, an active row, and a distinct `worktreePath`; update deletion test to require `archivedAt` to be defined (timestamped).

<sup>Written for commit a01691a011d996a96249032cc7c4e265749064ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6384?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Creating a workspace after deleting a previous one now correctly creates a new active workspace instead of reusing the archived workspace.

* **Tests**
  * Added coverage to verify workspace recreation and associated worktree creation after deletion.

### DIFF
--- a/packages/host-service/src/trpc/router/workspaces/workspaces.ts
+++ b/packages/host-service/src/trpc/router/workspaces/workspaces.ts
@@ -5,7 +5,7 @@ import {
 	sanitizeUserBranchName,
 } from "@superset/shared/workspace-launch";
 import { TRPCError } from "@trpc/server";
-import { and, eq } from "drizzle-orm";
+import { and, eq, isNull } from "drizzle-orm";
 import { z } from "zod";
 import { projects, workspaces } from "../../../db/schema";
 import { createGitEnvResolver } from "../../../runtime/git";
@@ -163,6 +163,10 @@ function findExistingWorkspaceByBranch(
 			where: and(
 				eq(workspaces.projectId, projectId),
 				eq(workspaces.branch, branch),
+				// Exclude tombstoned (archived) rows: since deletes archive
+				// instead of removing, a create after delete must not "reuse"
+				// a dead workspace whose worktreePath is gone (#6383).
+				isNull(workspaces.archivedAt),
 			),
 		})
 		.sync();
--- a/packages/host-service/test/integration/workspace-create-delete.integration.test.ts
+++ b/packages/host-service/test/integration/workspace-create-delete.integration.test.ts
@@ -411,7 +411,7 @@ describe("workspace.create + workspace.delete integration", () => {
 			.where(eq(workspaces.id, scenario.featureWorkspaceId))
 			.all();
 		expect(rows).toHaveLength(1);
-		expect(rows[0]?.archivedAt).not.toBeNull();
+		expect(rows[0]?.archivedAt).toBeDefined();
 		expect(rows[0]?.archiveReason).toBe("deleted");
 		expect(
 			scenario.host.apiCalls.some(
--- a/packages/host-service/test/integration/workspace-create-delete.integration.test.ts
+++ b/packages/host-service/test/integration/workspace-create-delete.integration.test.ts
@@ -420,6 +420,50 @@ describe("workspace.create + workspace.delete integration", () => {
 		).toBe(true);
 	});
 
+	test("create() does not reuse an archived (tombstoned) workspace on the same branch", async () => {
+		// Regress #6383: since deletes archive rows instead of removing them,
+		// a create-after-delete on the same branch must NOT match the
+		// tombstone by (projectId, branch) — otherwise it returns a dead
+		// workspace whose worktree is gone, reports alreadyExists, and skips
+		// creating the worktree.
+		const scenario = await createFeatureWorktreeScenario({
+			hostOptions: { apiOverrides: cloudFlows.workspaceDeleteOk() },
+		});
+		dispose = scenario.dispose;
+
+		// Delete the first workspace, archiving its row.
+		await scenario.host.trpc.workspace.delete.mutate({
+			id: scenario.featureWorkspaceId,
+		});
+		const archived = scenario.host.db
+			.select()
+			.from(workspaces)
+			.where(eq(workspaces.id, scenario.featureWorkspaceId))
+			.get();
+		expect(archived?.archivedAt).not.toBeNull();
+
+		// Create again on the exact same branch: must create a fresh row and
+		// not "reuse" the archived workspace.
+		const result = await scenario.host.trpc.workspaces.create.mutate({
+			projectId: scenario.projectId,
+			name: "repro-2",
+			branch: scenario.branch,
+		});
+		expect(result.alreadyExists).toBe(false);
+		expect(result.workspace?.id).not.toBe(scenario.featureWorkspaceId);
+		expect(result.workspace?.branch).toBe(scenario.branch);
+
+		// The new row is active (not archived) and has its own worktreePath.
+		const fresh = scenario.host.db
+			.select()
+			.from(workspaces)
+			.where(eq(workspaces.id, result.workspace?.id ?? ""))
+			.get();
+		expect(fresh?.archivedAt).toBeNull();
+		expect(fresh?.worktreePath).toBeDefined();
+		expect(fresh?.worktreePath).not.toBe(archived?.worktreePath);
+	});
+
 	test("delete() requires authentication", async () => {
 		const scenario = await createBasicScenario();
 		dispose = scenario.dispose;


andyst-dev:
Addressed both review points. Thanks for the sharp catch — `rows[0]?.archivedAt` via `.not.toBeNull()` would indeed pass if `archivedAt` were `undefined` (row removed instead of archived). The `rows` assertion at line 413 already requires the row to exist, but the archivedAt check is now explicit:

```ts
expect(rows).toHaveLength(1);
expect(rows[0]?.archivedAt).toBeDefined();
expect(rows[0]?.archiveReason).toBe("deleted");
```

Fixed in commit `a01691a01` (test(host-service): require the archived row to exist and be timestamped). Full integration suite re-run: 13 pass / 0 fail. Biome clean.

andyst-dev:
Closing as a duplicate: main already shipped `don't reuse a tombstoned workspace on same-branch create (#6552)` — the `findExistingWorkspaceByBranch` guard filters `isNull(workspaces.archivedAt)`, exactly this PR's intent (and also references #6383). No action needed.

PR #6402, opened 2026-08-12
feat(cli): manage desktop app settings and theme from the CLI (superset settings)

Kitenite:
## Summary
- New `superset settings` command group: `list` / `get <key>` / `set <key> <value>` / `reset <key>` plus a `theme` subgroup (`theme list/get/set`) — 35 desktop settings (behavior, git, notifications, terminal, terminal/editor typography) and the app theme, all editable from any shell with no cloud login.
- Adds a `superset-settings` agent skill (`.agents/skills/superset-settings/SKILL.md`) and a CLI reference section so agents and humans can discover keys, allowed values, and propagation semantics.

## Why / Context
Settings were only reachable through the desktop UI. Agents running in Superset terminals (and users scripting their setup) had no way to flip a toggle, change fonts, pick a ringtone, or switch themes. The CLI already ships on every host, so it's the natural surface.

## How It Works
Three stores, routed per key by a single registry (`packages/cli/src/lib/settings/registry.ts`) that mirrors the desktop's validation (enums, ranges, half/tenth font steps):

- **local.db** (most keys): upserts the same single-row `settings` table the desktop writes, via drizzle over `bun:sqlite`. A running app picks changes up on its next window refocus (electron IPC queries have `staleTime: 0` + refetch-on-focus). The CLI never creates the DB or tables — the desktop owns migrations; a fresh machine gets an actionable error instead.
- **host service** (`branchPrefixMode`, `branchPrefixCustom`, `worktreeBaseDir`): these are host-wide values the v2 Git settings page reads from `host_settings`, so the CLI talks to the local host service's tRPC using the manifest auth token (`~/.superset/host/<org>/manifest.json`) — still no cloud login. The legacy local.db columns are mirrored best-effort for v1 surfaces. `settings list` degrades to a `(host service not running)` note when it's down.
- **app-state.json** (theme): validated against built-ins + imported custom themes + `system` (with `--system-light`/`--system-dark` mappings), written atomically preserving all other app state. Applies at next launch; the command output tells users to quit first.

Deliberately excluded: `exposeHostServiceViaRelay` (paywalled + confirm dialog in the app), structured JSON settings (terminal presets, agent overrides), and renderer-localStorage prefs (`deleteLocalBranch`, diff view, hotkeys) that nothing outside the renderer can reach.

## Manual QA (CDP against the dev desktop app, this worktree)
Recorded a full demo video locally (CLI command → app updating on refocus) — happy to share; each item below was verified end-to-end with before/after screenshots:

- [x] `set confirmOnQuit true` / `set showResourceMonitor false` — General toggles flip on refocus, and back on revert
- [x] `set terminalFontSize 20`, `set editorFontSize 16`, `reset terminalFontSize` — Appearance live previews resize (20px → back to 14px)
- [x] `set notificationVolume 40` — volume dropdown updates to "Low (40%)"
- [x] `set branchPrefixMode custom` + `set branchPrefixCustom kiet/` — v2 Git page shows "Custom prefix" + live `kiet/branch-name` preview (this is what caught the host-service routing requirement: writing only local.db left the UI unchanged)
- [x] `theme set light` with app quit → relaunch boots in light theme; `theme set dark` restores
- [x] Compiled binary (`bun run build` → `dist/superset`) exercised for all of the above, including manifest-token auth to the host service
- [x] Error paths: unknown key, out-of-range/enum/step violations, missing local.db, host service down

Caveats verified and documented in the skill/docs:
- Force-killing the app leaves a fresh `theme-storage:pending` localStorage snapshot that beats `app-state.json` for ~5 min on next boot (reproduced both ways) — hence "quit cleanly first".
- `selectedRingtoneId` changes the played sound immediately (main process reads the DB) but the Notifications checkmark hydrates once at boot, so it updates after restart.
- QA refocus was triggered by dispatching `window` blur/focus events over CDP — the same events a real refocus fires.

## Testing
- `bun test packages/cli/src/lib/settings packages/cli/src/commands/settings` — 31 tests (registry validation, local.db roundtrip against a schema-derived temp DB, app-state preservation, host routing fallbacks, command-level set/get/reset)
- Mutation check: broke boolean parsing → registry test fails, restored → green
- `bun run typecheck` (packages/cli), `bun run lint` (exit 0), `bunx sherif` (no issues)
- `packages/cli` has no `test` script wired into turbo, so CI won't run these suites automatically (pre-existing gap; ran manually)

## Design Decisions
- **Direct store access over new desktop IPC**: the desktop has no external settings surface, and adding one is a bigger change; writing the same stores with the same validation gets full coverage today. The registry keeps key→store→rules in one reviewable place.
- **`skipMiddleware: true`**: settings are machine-local; requiring OAuth would break offline/agent use for no security gain (local.db and the host manifest are already 0600 on-disk secrets).
- **Manifest-token host auth instead of `--host` targeting**: git settings follow the UI's "This device" scope; remote-host settings can layer onto the existing `resolveHostTarget` later if wanted.

## Known Limitations
- Theme changes need an app relaunch (the app reads `app-state.json` once at boot); no push channel exists for external settings changes generally — refocus refetch is the propagation mechanism.
- `settings list` shows host-backed git keys as unavailable while the host service is down (rather than silently showing stale legacy values).

---
## Summary by cubic
Manage Superset desktop app settings and theme from the CLI. Previously UI-only; now `superset settings` (list/get/set/reset) and `superset settings theme` (list/get/set/export/import/remove). Theme changes apply on restart; other settings apply on next window refocus. Removes the unused `terminalPersistence` setting and drops the `persist_terminal` column from the local DB.

**Highlights**
- Validates values through one registry using limits from `@superset/shared/settings-constraints`. Local settings upsert the single `settings` row in `~/.superset/local.db` with `drizzle-orm` over `bun:sqlite` (opened readwrite to survive WAL; reads mirror the desktop by using the first row, not `id=1`); rejects blank numeric input.
- Host-wide Git settings (`branchPrefixMode`, `branchPrefixCustom`, `worktreeBaseDir`) route through the local host service via tRPC + manifest token and mirror to legacy columns. Calls time out (5s) and map transport/auth failures to actionable errors; `settings list` degrades these to “host service not running” instead of failing.
- Theme commands update `~/.superset/app-state.json` atomically and reuse the desktop’s parser and built-ins from `@superset/shared/themes` (reserved ids rejected, ids slugified, 256 KB import limit, export strips `isBuiltIn`, re-import replaces). Rejects arrays as app-state, drops malformed custom themes instead of crashing, and stamps imported themes with `isCustom`/`isBuiltIn`. Messages clarify when only system light/dark mappings changed; ringtone changes note immediate sound with Settings reflecting after restart.
- Consolidates ringtones and numeric constraints to `@superset/shared` with desktop re-export shims; adds a registry-coverage ratchet so every settings column is either exposed or explicitly excluded.

**Rollout**
- Launch the desktop app once so stores exist; commands otherwise return an actionable error.
- Git settings require the desktop app or `superset start` to be running.
- Quit the app before changing or importing themes; restart to apply.
- The local DB migration dropping `persist_terminal` runs at desktop boot; no user intervention required.

<sup>Written for commit d9e620b656de91e2fe1826ac5373bc15ded909cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6402?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to list, view, update, reset, and validate desktop settings.
  * Added theme management for built-in, system, and custom themes, including import, export, removal, and system light/dark options.
  * Added support for Git preferences requiring the desktop host service.
  * Added Light, Dark, and Monokai themes with editor and terminal styling.
* **Documentation**
  * Added comprehensive CLI guidance for settings and themes.
* **Tests**
  * Added coverage for settings, validation, theme workflows, and host-service behavior.


Kitenite:
Added custom theme management (`5cba74d`):

- `superset settings theme export <id> [--out <file>]` — dump any theme's full JSON (built-in or custom) as a starter
- `superset settings theme import <file>` — validate + add/replace custom themes (single theme, array, or `{ themes: [...] }`, max 256 KB)
- `superset settings theme remove <id>` — delete a custom theme; active/system references fall back to the default

To make this honest rather than a parallel implementation, the theme module (`types`, built-ins, `parseThemeConfigFile`, editor-theme generation) moved from `apps/desktop/src/shared/themes` to **`@superset/shared/themes`**, with one-line re-export shims left behind so none of the desktop's 20 importers change. The CLI therefore applies the exact same normalization as the Appearance UI import (id slugification, reserved-id rejection, base-theme color merging). Side benefit: the theme tests now actually run in CI via `packages/shared`'s test script (they previously lived in `apps/desktop` where `bun test` isn't wired into turbo).

Verified: desktop + shared + cli typecheck, 26 shared theme tests + 36 CLI settings tests, lint + sherif clean, and an E2E pass with the compiled binary (export dark → edit JSON → import → `theme list` shows it → set → remove falls back to dark).

Kitenite:
Architecture hardening pass (`dfaee69` + `33738ce`) — goal: adding a setting stays a one-line change with no drift risk.

**What adding a setting costs now**
- New scalar desktop setting → **one entry in `packages/cli/src/lib/settings/registry.ts`** (type, constraints, description, default). Validation, `list/get/set/reset`, JSON output, and table rendering all derive from it. Host-service-backed settings additionally get an accessor in `host-settings.ts`.
- **Drift ratchet** (`registry-coverage.test.ts`): every `settings` table column must be either in the registry or in `EXCLUDED_SETTINGS_COLUMNS` with a written reason — when the desktop adds a column, the CLI test fails until someone consciously exposes or excludes it. Mutation-tested.
- **No duplicated literals left**: ringtones moved to `@superset/shared/ringtones` and the numeric constraints (font size/weight, line height, letter spacing, contrast choices, cursor styles, volume, parked-terminal cap) to `@superset/shared/settings-constraints` — the desktop zod schemas and the CLI registry now consume the same constants (desktop's own font-settings tests pass unchanged). Same pattern as the earlier themes extraction; shims keep desktop imports untouched.

**Two real fixes that fell out of the audit**
- `packages/cli` now has a `test` script, so its 81 tests run in CI for the first time. The blocker was a cross-file `mock.module("node:child_process", ...)` leak in `spawn.test.ts` (partial mock breaking other files' `execFileSync` import) — fixed with the same snapshot-and-spread pattern `config.test.ts` documents.
- Readonly opens of the WAL-mode `local.db` fail with `SQLITE_CANTOPEN` when the `-shm` sidecar is absent (e.g. after a checkpoint with the app closed) — reproduced in isolation; read commands now open readwrite like the desktop does (the existsSync guard still prevents creating databases).

Verified: cli 81 tests / shared 762 tests / desktop typecheck green; lint + sherif clean; compiled binary smoke against a checkpointed-db sandbox (the exact case that previously failed).

Kitenite:
Follow-up breakage audit (`4550e06`) — swept the new code for more state/environment-dependent failures like the WAL readonly bug. Four found, all fixed with regression tests:

1. **Host-service calls could hang forever** — `httpBatchLink` had no timeout, so a stale manifest (dead port with a recycled live pid, wedged service) would wedge `settings get branchPrefixMode` indefinitely. Now a 5s `AbortSignal.timeout`, and transport/auth failures map to actionable errors (`Could not reach the host service… Run: superset start`, credential-rotation hint on 401) instead of raw tRPC dumps. Test proves connection-refused fails fast.
2. **Corrupt `app-state.json` threw a raw `SyntaxError`** — now an actionable error explaining the desktop rewrites the file on next launch.
3. **Settings row read disagreed with the desktop** — the CLI filtered `id = 1` while the desktop's `getSettings()` reads the first row unconditioned; a legacy row with a non-1 id would make the CLI blind to values the app uses. Read now mirrors the desktop exactly.
4. **Unwritable paths dumped raw errno errors** — `local.db` open failures and `theme export --out` write failures now name the path and the problem.

Also added a regression test for the original WAL case itself (checkpointed database, sidecars truncated → read still works).

Audited and deliberately left as-is: concurrent CLI writes are last-wins (same as two app windows), Windows isn't a build target, malformed custom-theme entries degrade gracefully in `theme list`, and theme-file BOM/garbage is already caught by the shared parser's friendly error.

85 CLI tests green (now in CI), typecheck/lint clean, binary rebuilt and smoke-tested.

Kitenite:
Full regression matrix (`d805a3f`) — every settable key and every command exercised against the live dev app over CDP:

**54/54 checks pass**
- **33 setting keys**: CLI `set` → read back **through the desktop app's own IPC** (`electronTrpcClient.settings.getX` in the running renderer — the app's real read path, not the CLI's), original values restored after each. Host-backed git keys additionally verified by a direct authenticated fetch to the host service's `host_settings` (independent of both CLI and app).
- **21 command/error checks**: full theme lifecycle (export → edit → import → set → get → export round-trip → remove-with-fallback → system mappings), `list/get/reset` semantics, and every rejection path (unknown key, bad enum, out-of-range, step violation, non-integer, unknown theme, built-in removal, missing/reserved/garbage theme files) with actionable messages.

**One real find**: `terminalPersistence` was a **dead setting** — its `persist_terminal` column has no reader anywhere in the codebase, so the CLI advertising "keep terminal sessions alive in the background" was a no-op. Removed from the registry into `EXCLUDED_SETTINGS_COLUMNS` (the coverage ratchet is what makes this a one-line, test-enforced decision).

Skill also finalized: from-scratch theme creation with a verified minimal JSON + anatomy table, agent-safety note (an agent running inside a Superset terminal must not kill the app it lives in — ask the user to relaunch for theme changes), JSON-output parsing guidance, and `SUPERSET_HOME_DIR` sandbox docs.

Kitenite:
`d728dd7` — removed the dead setting at the source, per review of the regression findings:

- `persist_terminal` column dropped from the local-db schema with a generated migration (`0046`, a plain `ALTER TABLE settings DROP COLUMN` — no table rebuild). Verified by applying the migration folder to a copy of a real `local.db`: column gone, existing row still reads.
- The CLI's temporary exclusion entry for it is deleted — the registry coverage test enforces this from both directions (a column can't be silently unaccounted for, and an exclusion can't outlive its column).

local-db + desktop + cli typecheck green, 85 CLI tests green, migration snapshot/journal are drizzle-kit-generated (not hand-edited).

Kitenite:
Review triage complete (`d9e620b`) — all CodeRabbit findings resolved:

**Fixed** (with tests): JSON-array app-state rejection, malformed custom-theme entries dropped instead of crashing, CLI imports stamped `isCustom`/`isBuiltIn` like the desktop store (Appearance dropdown groups by it — good catch), blank numeric input rejected (`Number("") === 0`), accurate message when `theme set` only changes system mappings, ringtone-specific propagation message, quoted export path in the follow-up hint.

**Already addressed earlier**: host-service RPC deadline (CodeRabbit acknowledged `4550e06`).

**By design, not fixing here**:
- *Concurrent `app-state.json` overwrites* — this is the documented theme constraint: the CLI instructs quitting the app first, and the coordinated-desktop-API route is the noted hot-reload follow-up. A CLI-side lock can't fix it (the writer that clobbers is the app itself).
- *Atomic branch-prefix field patch* — the desktop UI performs the same read-modify-write against `branchPrefix.set`; a field-level patch is a host-service API change that belongs in its own PR if we ever see real contention on a single-machine, single-row setting.

CLI suite now 90 tests, all green.

Issue #6009, opened 2026-07-28
[bug] Expanded sidebar shows empty Projects section


gbrlsepulveda:
## Bug Description

After macOS prompted for the **App Management** permission (System Settings → Privacy & Security → App Management) and it was accepted, the sidebar's Projects section broke. Before that prompt, the sidebar worked normally.

Collapsed icon rail correctly shows all workspace/project icons. Expanding the sidebar to the full labeled view shows the "Projects" section header and the `+` (add project) button, but no project rows render underneath — the list is empty.

## Steps to Reproduce

1. Superset prompts for the macOS App Management permission; accept it
2. Sidebar Projects section stops rendering items when expanded
3. Collapse the sidebar — all workspace icons render correctly in the icon rail
4. Expand again — Projects section is empty

## Expected Behavior

Projects list in the expanded sidebar should show the same items visible in the collapsed rail, and should not be affected by OS permission grants.

## Actual Behavior

Projects section is blank when expanded; only recoverable by collapsing the sidebar back down.

## Troubleshooting Already Tried (did not resolve it)

- Cmd+R (force refresh) — no change
- Disabling the v2 sidebar experimental setting (Settings → Experimental → Try Superset v2 → off) — this fixed a related symptom (projects were fully invisible even when collapsed) but the expand/collapse-dependent empty Projects section persists on v1

## Environment

- OS: macOS (version TBD, will follow up)
- Superset version: TBD, will follow up
- v2 sidebar toggle: off

## Screenshots

<img width="219" height="1051" alt="Image" src="https://github.com/user-attachments/assets/68c545b9-cfbd-4f50-9516-13b54c796609" />

<img width="52" height="1084" alt="Image" src="https://github.com/user-attachments/assets/0eba6764-6296-43a7-8d3d-1ad17e23d14a" />

gbrlsepulveda:
The content is empty. The PR suggested by copilot doesn't seem to fix the issue

<img width="1393" height="1012" alt="Image" src="https://github.com/user-attachments/assets/2a3c2beb-2e90-4c81-b088-d9c39410113f" />

PR #6407, opened 2026-08-13
feat(desktop): make Pinned and Sessions sidebar sections collapsible like Projects

Kitenite:
## Summary

The Pinned and Sessions sidebar sections now collapse exactly like the Projects header: click the header strip to toggle, with the chevron cue that stays visible while collapsed (per the GH #6009 convention) so hidden rows read as hidden rather than missing. Collapse state persists per section.

## Refactor

- New shared `DashboardSidebarSectionHeader` renders the top-level section header strip (label, chevron, right-aligned action slot); `DashboardSidebarWorkspacesHeader` (Projects), `DashboardSidebarPinnedSection`, and `DashboardSidebarSessionsSection` all use it now.
- `sidebar-workspaces-collapse` store generalized to `sidebar-sections-collapse` holding `collapsed: { pinned, sessions, workspaces }`. The legacy localStorage key is kept with a v0→v1 `persist` migration, so an already-collapsed projects list carries over; persisted-key registry updated.

## Behavior notes

- Collapsed sections hide their rows (and unregister their sortables), matching collapsed project groups — drops don't target a collapsed section.
- The empty-Pinned drop zone still renders mid-drag regardless of collapse, so the first drag-to-pin always has a landing target.
- Sessions rows stay hidden when either its own collapse or the global Projects-list collapse is active; the Sessions "+" now stops propagation so it doesn't toggle the header.

## Verification

Typecheck, lint, and persisted-key registry tests pass. Verified end-to-end over CDP against this worktree's dev app with trusted input: hover→chevron, collapse/expand each section, both-collapsed → `location.reload()` → still collapsed (store: `{"collapsed":{"pinned":true,"sessions":true,"workspaces":false},"version":1}`), console.error hooked throughout — zero entries. 27s screen recording captured.

---
## Summary by cubic
Makes Pinned and Sessions sidebar sections collapsible with the same UX as Projects. Previously only Projects collapsed; now each section toggles via its header, shows the chevron while collapsed, and remembers its own state.

- Review notes
  - Introduces `DashboardSidebarSectionHeader` and uses it for Projects, Pinned, and Sessions.
  - Replaces `useSidebarWorkspacesCollapseStore` with `useSidebarSectionsCollapseStore` consumers.
  - Collapsing unmounts rows and unregisters sortables; drops cannot target a collapsed section. The empty Pinned drop zone still renders mid-drag so the first pin always has a target. The Sessions “+” stops event propagation to avoid toggling; Sessions rows also hide when Projects is collapsed.

- Persistence/migration
  - Generalizes persisted state to `{ collapsed: { pinned, sessions, workspaces } }` under the legacy `sidebar-workspaces-collapse` key with a v0→v1 migration that maps `isCollapsed` to `collapsed.workspaces`.
  - Persisted-key registry updated; no migration action required.

<sup>Written for commit fd96fc329a378aa5da4c88b5d9de88d62ecbd72a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6407?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added collapsible sidebar sections for pinned items, sessions, and workspaces.
  * Added consistent section headers with chevrons, optional actions, and keyboard accessibility.
  * Sidebar collapse preferences are now saved between sessions.
  * Preserved drop zones and section actions when content is collapsed or hidden.


Kitenite:
Reviewed the CodeRabbit findings — proceeding without changes:

1. **Wrapper role=button semantics**: pre-existing pattern extracted verbatim from the Projects header (`DashboardSidebarWorkspacesHeader`), where the original biome-ignore comment documents the trade-off (can't be a native `<button>` because it nests action buttons). Not a regression introduced here; a11y restructure of all three headers is follow-up material.
2. **Section-state independence**: hiding Pinned/Sessions rows when the global Projects-list collapse is active is the pre-existing, intended behavior on main (`workspacesListCollapsed` gating predates this PR — only the selector was renamed). No workspace drag can originate while everything is hidden, so the drop-zone concern doesn't apply.
3. **Persisted key lifecycle**: fixed-size singleton (three booleans), registered in the persisted-key registry; noted for DEAD_KEYS on feature removal.

PR #6427, opened 2026-08-13
feat(desktop): let users type a workspace name on the new-workspace screen

### LINKED ISSUE #6398: Let me name a workspace and its branch when I create it
## What problem are you trying to solve?

My normal workflow is: make a worktree, give it the name I want, open it, then start an agent from a terminal preset. The agent comes second. Naming the thing up front is the whole interaction for me, and it used to be the fastest path in the app.

Right now it is the slowest. I am on the `new-workspace-screen` test arm, and that screen has no way to type a name at all. Submit stays disabled while the prompt box is empty, and what I type there is a hint for the AI namer rather than the name itself. Picking "No agent" does not change that, the text still goes through as `namingPrompt` and comes back as something an LLM chose.

What makes me think this is drift rather than a decision is that the naming path has narrowed a bit at a time:

- **#391** (Dec) - a labelled form: Project, Title (optional), Branch, Create button.
- **#2153** (Mar) - prompt-first textarea on top, branch field still editable underneath with a live prefix preview.
- **#2627** (Mar) - AI branch names land, and typing a name becomes the override instead of the normal path.
- **Today's control modal** - the branch field is still there, but unlabelled, placeholder-only, pushed to the right edge at half width in muted grey mono, submitted with Cmd+Enter. It works. It is also styled to be ignored.
- **#5907** - on the new screen, both fields are gone.

Underneath, a typed name is still first class, and it is the cheaper path: passing `name` to `workspaces.create` sets `wantAi = false`, which skips the LLM call, the deferred rename and the placeholder branch, and the workspace name falls back to the branch verbatim (`input.name ?? resolvedBranch`). The CLI goes further and requires `--name` whenever `--project` is set. It is only the desktop UI that has moved away from it.

### Remote hosts make this harder

I use remote hosts constantly, and that is where it stops being a small annoyance. Creating a remote workspace itself is fine - the host picker works, the branch list queries the right host, the create dispatches correctly. The only missing piece is saying what it is called, and both workarounds get worse across the network:

- Picking an existing branch needs that branch to already exist on the remote host's repo. To put it there I either need a terminal in a workspace on that host (which I do not have yet, that is what I am creating) or I push a throwaway branch from my laptop purely so the remote picker can see it.
- The CLI works, but it is four commands and two id lookups: `hosts list`, `projects list --host`, `ws create --host --project --name --branch`, `ws open`.

## Proposed solution

A name field on the new-workspace screen, with the branch defaulting to a slug of it and submit enabled on a name alone. The control modal already does exactly this, so it might be as small as bringing those two inputs across.

If the new screen's design does not want two more fields on it, an alternative that would suit me just as well: when no agent is selected, treat the prompt box text as the literal name rather than as a naming hint. One field, two meanings depending on the mode.

I am not asking to remove the AI namer. It is clearly doing real work for people who start from a prompt. I would just like a way past it that does not involve a terminal.

Keen to hear whether the screen is heading somewhere that already covers this.

## Workarounds you've tried

1. **Branch picker, "Open workspace" on a row.** Exact name, no LLM call, no prefix applied. Needs the branch to exist first, which is the problem on a remote host.
2. **CLI `ws create --name --branch`.** Exact, and works remotely, but four commands and two id lookups for what used to be one text field.
3. **Create it and fix the names after.** Two renames in two different places: the workspace title inline, and the branch separately from the Changes tab, since `workspace.update --name` does not touch git. All after waiting on an LLM call I did not want.

# 5939 looks like the same theme from a different angle - the prompt-less path quietly not being exercised.

andyst-dev:
Fixes #6398

## What & why

The new-workspace screen (experiment test arm) had no way to type a name:
whatever was entered went through as an AI naming hint, and picking "No
agent" did not change that. Under the hood a typed name is the cheaper,
exact path — `resolveNames` returns it and `workspaces.create` sets
`wantAi = false` (skipping the LLM call and the deferred rename) — but the
UI never offered it.

This adds a "Workspace name (optional)" field above the composer. It writes
`workspaceName`/`workspaceNameEdited` into the existing draft store, which
the submit path already reads via `resolveNames`. So a name alone (no
prompt) creates an exact-named workspace and the branch seeds from the
name. This matters especially for remote hosts, where the branch-picker
workaround required pushing a throwaway branch first.

### Files changed

- `apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/NewWorkspaceScreen/NewWorkspaceScreen.tsx`
  — add the name input (mirrors the control modal's field: `updateDraft` +
  `workspaceNameEdited`, clears on blur if empty).
- `apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/DashboardNewWorkspaceForm/PromptGroup/hooks/useSubmitWorkspace/resolveNames.test.ts`
  — new unit tests for the naming path.

## How I tested it

- `bun install` clean.
- `bun test resolveNames.test.ts` -> **5 pass / 0 fail** (typed name
  resolves to a real workspace name, an unedited/AI-suggested name is
  ignored, typed branch/name pairs both resolve).
- `biome check` clean on both files.
- `bun x tsc --noEmit -p apps/desktop/tsconfig.json` introduces no new
  errors vs `main` (85 pre-existing, identical before/after).

## Checklist

- [x] `bun run lint` and `bun run typecheck` pass (no new errors vs main)
- [x] I have added tests that prove the feature works

---
## Summary by cubic
Lets users type a workspace name on the new-workspace screen and treats it as the exact name. Previously any text acted as a naming hint and always used automatic naming; now a typed name creates an exact-named workspace and skips the deferred rename.

- Adds a name input above the composer; writes workspaceName/workspaceNameEdited to the draft and clears both on blur when empty.
- Updates resolveNames to prefer an explicitly edited branch; otherwise sanitizes the typed workspaceName and uses it as the branch candidate; ignores unedited/suggested names; returns a null branch if sanitization empties it.
- Adds unit tests covering typed names, ignoring unedited suggestions, branch sanitization (including empty results and preserving spaces), and typed name plus typed branch together.

<sup>Written for commit db4ac32261a6053f145c9b5c3c0d2ef620a9719e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6427?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an optional workspace-name field when creating a new workspace.
  * Workspace names are retained when entered and cleared when left blank.
  * When no branch name is provided, the workspace name is used as a sanitized fallback for the branch name.
  * Explicitly provided branch names continue to take precedence over workspace names.

* **Tests**
  * Added coverage for workspace and branch-name resolution, including edited names, suggested names, empty drafts, and branch-name formatting.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/DashboardNewWorkspaceForm/PromptGroup/hooks/useSubmitWorkspace/resolveNames.test.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/DashboardNewWorkspaceForm/PromptGroup/hooks/useSubmitWorkspace/resolveNames.test.ts
@@ -43,16 +43,25 @@ describe("resolveNames", () => {
 	test("sanitizes a typed workspace name as the branch candidate", () => {
 		const names = resolveNames(
 			draft({
-				workspaceName: "feature/cache-fix",
+				workspaceName: "feature:cache-fix",
 				workspaceNameEdited: true,
 			}),
 		);
-		expect(names.branchName).toBe("feature/cache-fix");
+		expect(names.branchName).toBe("featurecache-fix");
+	});
+
+	test("rejects a workspace name that sanitizes to an empty branch", () => {
+		const names = resolveNames(
+			draft({ workspaceName: ":", workspaceNameEdited: true }),
+		);
+		expect(names.workspaceName).toBe(":");
+		expect(names.branchName).toBeNull();
 	});
 
 	test("ignores a workspace name that was not edited (e.g. AI-suggested)", () => {
 		const names = resolveNames(draft({ workspaceName: "suggested" }));
 		expect(names.workspaceName).toBeNull();
+		expect(names.branchName).toBeNull();
 	});
 
 	test("returns a sanitized typed branch name (preserves spaces)", () => {
--- a/apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/DashboardNewWorkspaceForm/PromptGroup/hooks/useSubmitWorkspace/resolveNames.test.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/DashboardNewWorkspaceForm/PromptGroup/hooks/useSubmitWorkspace/resolveNames.test.ts
@@ -1,0 +1,73 @@
+import { describe, expect, test } from "bun:test";
+import type { DashboardNewWorkspaceDraft } from "../../../../../DashboardNewWorkspaceDraftContext";
+import { resolveNames } from "./resolveNames";
+
+function draft(
+	overrides: Partial<DashboardNewWorkspaceDraft> = {},
+): DashboardNewWorkspaceDraft {
+	return {
+		selectedProjectId: null,
+		isSession: false,
+		hostId: null,
+		prompt: "",
+		baseBranch: null,
+		baseBranchSource: null,
+		workspaceName: "",
+		workspaceNameEdited: false,
+		branchName: "",
+		branchNameEdited: false,
+		branchNameFromProvider: false,
+		linkedIssues: [],
+		linkedPR: null,
+		selectedAgentId: null,
+		attachments: [],
+		...overrides,
+	};
+}
+
+describe("resolveNames", () => {
+	test("returns null name when nothing was typed", () => {
+		const names = resolveNames(draft());
+		expect(names.workspaceName).toBeNull();
+		expect(names.branchName).toBeNull();
+	});
+
+	test("returns a user-typed workspace name (not an AI hint)", () => {
+		const names = resolveNames(
+			draft({ workspaceName: "my-remix", workspaceNameEdited: true }),
+		);
+		expect(names.workspaceName).toBe("my-remix");
+	});
+
+	test("ignores a workspace name that was not edited (e.g. AI-suggested)", () => {
+		const names = resolveNames(draft({ workspaceName: "suggested" }));
+		expect(names.workspaceName).toBeNull();
+	});
+
+	test("returns a sanitized typed branch name (preserves spaces)", () => {
+		const names = resolveNames(
+			draft({
+				branchName: "Fix UI bug",
+				branchNameEdited: true,
+			}),
+		);
+		// sanitizeUserBranchName only strips git-forbidden chars; it preserves
+		// case and spaces (the UI slugifies spaces on the way in).
+		expect(names.branchName).toBe("Fix UI bug");
+		// A typed name seeds a branch slug; the workspace name comes straight through.
+		expect(names.workspaceName).toBeNull();
+	});
+
+	test("typed name and typed branch both resolve", () => {
+		const names = resolveNames(
+			draft({
+				workspaceName: "My Workspace",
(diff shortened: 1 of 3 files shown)


de1mat:
Thanks for picking this up, I filed the issue.

One functional gap worth catching before review: the PR sets `name` but not `branch`, so the branch gets derived server side through `sanitizeBranchCandidate` (`workspaces.ts:895`), which caps at 25 characters and strips slashes. `serverless-cache-invalidation-fix` ends up as branch `serverless-cache-invalida`, and `feature/cache-fix` as `featurecache-fix`. With no branch field on the screen there's no way to correct it, so you get the exact title and an approximate branch. Sending the string as both `name` and `branch` routes it down the existing typed-branch path and avoids that entirely.

On the field itself, and this is more a question for @saddlepaddle than a criticism of the PR: in the issue thread they said they'd rather not add the naming fields back while the experiment bakes, since they attribute a good part of the ~28% activation lift to a less cluttered create surface. An always-visible input above the composer is probably the version they had in mind. Two ways to keep the capability without changing what a prompt-first user sees: only render the field when the agent pill is on "No agent", or skip the field entirely and use the composer text as the name when it is already a legal git branch name, falling back to today's behaviour when it isn't.

Any of those works for me, including this as-is once the branch derivation is sorted. Mostly flagging it so it doesn't get closed on the clutter objection when there's a version that sidesteps it.

andyst-dev:
Addressed the branch-name gap in commit `c6b126169`.

When a user manually enters a workspace name on the new-workspace screen, `resolveNames()` now sends that value as both:

- `name`: the displayed workspace name;
- `branch`: the sanitized typed branch candidate.

An explicitly seeded/provider branch still takes precedence, so linked issue/Linear branch names keep their existing behavior. Untyped or AI-suggested names remain `null` and continue through the existing generated/AI naming path.

Validation:
- `bun test .../resolveNames.test.ts` → **6 passed**
- Biome → **clean**
- `git diff --check` → **clean**
- remote PR head verified at `c6b126169`

GitHub’s CodeRabbit check is currently pending.

andyst-dev:
Follow-up for the second CodeRabbit review round, fixed in `db4ac3226`.

Addressed both findings:

- `resolveNames()` now converts a workspace name whose sanitized branch is empty to `null` instead of sending an empty explicit branch.
- Regression coverage now uses a Git-forbidden character (`feature:cache-fix`) and asserts the sanitized branch (`featurecache-fix`).
- The unedited/AI-suggested-name case now explicitly asserts both `workspaceName` and `branchName` are `null`.
- Added a `":"` case proving an entirely sanitized-away name does not become an empty branch.

Validation:
- `bun test .../resolveNames.test.ts` → **7 passed**
- Biome → **clean**
- `git diff --check` → **clean**
- remote head verified locally at `db4ac3226`

No standard check-run is currently reported for this fork branch.

andyst-dev:
Closing as a duplicate: main already ships `seed branch from typed workspace name (#6129)` — `resolveNames.ts` on main handles `workspaceNameEdited` + `sanitizeUserBranchName`, the exact behavior this PR added. No unique delta. No action needed.

PR #6428, opened 2026-08-13
feat(desktop): add a Jump to latest reply button in the v2 chat

andyst-dev:
### What does this PR do?

Adds a floating **"Jump to latest reply"** button to the **v3 chat transcript** (`ChatV3Pane`). When an assistant reply is still streaming, or the transcript is long, the scroll-to-bottom affordance only reaches the *end* of the conversation — not your own latest message sitting above it. This button scrolls the last user message into view instead.

The original PR targeted the legacy mastra chat (`ChatPane/WorkspaceChatInterface`), which was deleted on `main` in `delete legacy mastra chat in favor of chat-v3 (#6461)`. This PR is the **rebase/port of that feature onto the surviving `ChatV3Pane` + `Transcript`**, so it lands on current `main` instead of orphaned legacy code.

### What changed

- `Transcript.tsx` (v3): wraps the scrollable container in a relative parent and renders a floating `Latest reply` button.
- The button only appears once the user has **scrolled up** (not at bottom) and there is at least one user message — consistent with the old stick-to-latest behaviour.
- Reuses the existing `latestUserItemId` / `anchorItemId` logic and the `[data-item-id]` selector already used for auto-scroll, so scroll math stays consistent.
- Derived behaviour from the old `useConversationContext` (`isAtBottom`, `scrollRef`) is re-implemented against the v3 `Transcript`'s own `containerRef` + scroll listener.

### Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature/enhancement
- [x] Improvement
- [ ] Other (please describe)

### How did you verify your code works?

- `bun x tsc --noEmit` on `apps/desktop`: **no new errors** in `Transcript.tsx` / `ChatV3Pane` (only pre-existing `routeTree.gen` / automations-route errors already on `main`).
- `biome lint` on the touched file: **clean, no fixes applied.**
- `bun install` in the worktree succeeded (782 packages) so the workspace `@superset/chat` types resolve.

### Checklist

- [x] I have read the contributing guidelines.
- [x] Code follows the project's style and lint passes.
- [x] No new inaccessible/private paths or secrets introduced.
- [x] Local validation (tsc on changed file + biome) passed.

## Summary by CodeRabbit

* **New Features**
  * Added a “Latest reply” button to quickly return to the newest transcript content.
  * Added smooth scrolling and a downward arrow indicator for the jump control.

* **Improvements**
  * Transcript scrolling now stays at the latest user message when already near the bottom.
  * Prevented automatic scrolling when browsing earlier transcript content, preserving the user’s reading position.
  * Improved scrolling behavior to avoid repeated re-anchoring as new replies appear.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/ChatV3Pane/components/Transcript/Transcript.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/ChatV3Pane/components/Transcript/Transcript.tsx
@@ -7,8 +7,10 @@ import type {
 import type { ApprovalRequest, Decision } from "@superset/chat/protocol";
 import { Badge } from "@superset/ui/badge";
 import { Button } from "@superset/ui/button";
+import { ArrowDownIcon } from "lucide-react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { TurnGroupSection } from "./components/TurnGroupSection";
+import { isNearBottom, latestUserItemId, shouldScrollToLatest } from "./utils/scroll";
 
 export type TranscriptProps = {
 	groups: TurnGroup[];
@@ -21,20 +23,6 @@ export type TranscriptProps = {
 	onRetryPrompt: (clientId: string) => void;
 	onDiscardPrompt: (clientId: string) => void;
 };
-
-function latestUserItemId(groups: TurnGroup[]): string | null {
-	for (let groupIndex = groups.length - 1; groupIndex >= 0; groupIndex -= 1) {
-		const group = groups[groupIndex];
-		if (!group) continue;
-		for (let index = group.entries.length - 1; index >= 0; index -= 1) {
-			const entry = group.entries[index];
-			if (entry?.kind === "item" && entry.item.kind === "user_message") {
-				return entry.item.id;
-			}
-		}
-	}
-	return null;
-}
 
 function outboxText(entry: OutboxEntry): string {
 	return entry.content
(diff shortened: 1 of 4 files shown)


andyst-dev:
Good catch — thanks. Interpolating `latestUserMessageId` directly into the selector would break for ids containing CSS-significant characters.

Fixed in `c440c1d5e`: `findUserMessageElement` now queries the fixed `[data-chat-user-message='true'][data-message-id]` selector and compares `dataset.messageId`, mirroring `MessageScrollbackRail`'s safe lookup. The id is never interpolated into a selector.

Validation: `biome check` clean, `bun x tsc --noEmit` introduces no new errors vs main.

andyst-dev:
Rebuilt this PR on top of current main (the branch had drifted ~hundreds of commits behind) and addressed the CodeRabbit Major finding.

**What changed:**
- **Re-anchor fix (CodeRabbit Major):** the auto-scroll effect previously called `scrollIntoView` whenever `isAtBottom` became true — so an assistant reply following the latest user item would yank the viewport back up the instant the user scrolled to the bottom. The pinning logic now tracks `lastAnchorItemIdRef` and only scrolls when the anchor item **id changes** while the user is at the bottom, so reaching the bottom never re-anchors onto an already-visible item.
- The scroll/anchor decision logic was extracted into `Transcript/utils/scroll/scrollLogic.ts` with a dedicated unit test; `Transcript.tsx` re-implements the existing behaviour (pin while at bottom, floating "Latest reply" button when scrolled up) against main's current component structure (props `groups`/`outbox`/`approvals`/`pendingApprovalTargets`/`snapshot`/…).

**Validation (real, from this worktree):**
- `bun test` full suite (apps/desktop): **3096 pass, 0 fail** (331 files).
- New `scrollLogic.test.ts`: 10 pass / 0 fail.
- `bun run typecheck` (tsc --noEmit): exit 0.

The branch was force-pushed (rebuilt from a different base), so the diff is necessarily a re-implementation rather than the original patch — same feature, same CodeRabbit resolution, on the current tree.

PR #6451, opened 2026-08-13
feat(desktop): add prompt-cards arm to the new-workspace screen

saddlepaddle:
Test arm for experiment [419092](https://us.posthog.com/project/264803/experiments/419092): replaces the three low-contrast sample-prompt rows on the new-workspace screen with two bordered cards above the composer.

## Why

Of 837 new accounts that reached this screen between 5 and 13 August, **430 (51.4%) created a real workspace** and **272 added a repo and never started work**. Only 11% ever click a suggested prompt, and 75% of the clicks that do happen land on "Set up this project for Superset" — so that card leads, gated on the project having no setup/teardown/run commands (the same `config.shouldShowSetupCard` query the v2 sidebar card uses).

## Scope

- **Control renders exactly as it does today.** No behavior change for anyone not in the experiment.
- Eligibility is a `created_at` release condition on the flag, not code, so existing accounts get `false`, keep the rows, and record no exposure.
- The variant hook reads the flag imperatively on screen open, so `$feature_flag_called` matches the population that sees the surface rather than everyone who launches the app.
- `new-workspace-prompt-cards-override` forces the cards for team accounts without evaluating the experiment flag, so previewing emits no exposure.
- Prompt text is byte-identical across arms; presentation and one line of card body copy are the only differences.

## Measurement

Primary is activation (`workspace_created` where `type != main`). Secondaries are the prompt click rate as the mediator (tagged `layout: cards | rows`), real workspaces per user, and two guardrails — "came back and worked again" and day-2 app reopen — so a creation lift that comes from nagging doesn't read as a win.

## Do not launch before this ships

Old builds never evaluate the flag, so starting the experiment before a desktop release carrying this code would enroll a pure-control population. At launch, bump the flag's `created_at` cutoff to the actual publish time.

## Checks

`lint:fix`, `typecheck` (37/37) and `sherif` pass. `bun test` reports 154 pre-existing failures locally (real-git integration tests and a missing `electronTRPC` global) — identical counts on a clean tree, so zero delta from this change.

Registry entry: [Experiment Log](https://app.notion.com/p/3bbb9d5bf6168107bb58f222d8cbbf54)

---
## Summary by cubic
Adds the prompt-cards experiment arm to the new-workspace screen to test whether more prominent suggestions lift activation. Old behavior: three inline sample-prompt rows; new behavior: two bordered cards above the composer when `new-workspace-prompt-cards` is "test"; control unchanged.

- `useNewWorkspacePromptCardsVariant` evaluates on screen open, returns null until flags load to avoid flashing, and reads `new-workspace-prompt-cards-override` before the experiment flag so overridden accounts do not record exposure; falls back to control after 2s if flags don’t load.
- `SamplePromptCards` shows two cards; the “Set up this project for Superset” card appears only when `config.shouldShowSetupCard` is true for the project. Prompt text is identical across arms; cards add icons and a short description.
- Analytics: prompt clicks include `layout: cards | rows`. Exposure matches users who open the screen.
- Eligibility/rollout: use a flag-side `created_at` cutoff (not code). Existing accounts stay on rows with no exposure. Do not start the experiment until a desktop release with this code; at launch, set the flag’s `created_at` to the publish time.
- Adds `NEW_WORKSPACE_PROMPT_CARDS` and `NEW_WORKSPACE_PROMPT_CARDS_OVERRIDE` in packages/shared constants.

<sup>Written for commit da000b0333ca35611ff93fe9718cc113333c7c6b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added selectable prompt cards when creating a new workspace.
  * Prompt cards can display setup suggestions based on the current workspace context.
  * Added descriptions and improved labels for sample prompts.
  * The prompt-card experience is introduced selectively and falls back to the standard prompt list when unavailable.
* **Analytics**
  * Sample-prompt selections now record the card layout used.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/NewWorkspaceScreen/NewWorkspaceScreen.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/components/DashboardNewWorkspaceModal/components/NewWorkspaceScreen/NewWorkspaceScreen.tsx
@@ -48,6 +48,7 @@ import { newWorkspaceAttachmentPaths } from "renderer/stores/new-workspace-attac
 import { useNewWorkspacePromptContext } from "renderer/stores/new-workspace-prompt-context";
 import { useV2WorkspaceCreateDefaultsStore } from "renderer/stores/v2-workspace-create-defaults";
 import { useDashboardNewWorkspaceDraft } from "../../DashboardNewWorkspaceDraftContext";
+import { useNewWorkspacePromptCardsVariant } from "../../hooks/useNewWorkspacePromptCardsVariant";
 import { DevicePicker } from "../DashboardNewWorkspaceForm/components/DevicePicker";
 import { useWorkspaceHostOptions } from "../DashboardNewWorkspaceForm/components/DevicePicker/hooks/useWorkspaceHostOptions";
 import { CompareBaseBranchPicker } from "../DashboardNewWorkspaceForm/PromptGroup/components/CompareBaseBranchPicker";
@@ -73,6 +74,7 @@ import {
 } from "../DashboardNewWorkspaceForm/PromptGroup/types";
 import { useSelectedHostProjectIds } from "../DashboardNewWorkspaceModalContent/hooks/useSelectedHostProjectIds";
 import { AttachmentCard } from "./components/AttachmentCard";
+import { SamplePromptCards } from "./components/SamplePromptCards";
 import { SamplePrompts } from "./components/SamplePrompts";
 
 interface NewWorkspaceScreenProps {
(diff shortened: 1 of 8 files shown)


saddlepaddle:
Second CodeRabbit finding (reset `variant` on reopen) — skipping, with reasoning.

`NewWorkspaceScreen` has exactly one consumer, `_dashboard/new-workspace/page.tsx`, which passes `isOpen` as a literal `true`. It never toggles: navigating away unmounts the component, so state is already discarded and the next mount starts at `null`. The proposed `setVariant(null)` would run once on mount when `variant` is already `null` — a no-op.

If `isOpen` ever did toggle, resetting would blank the prompts for a frame while re-resolving a value that cannot have changed, since a user's flag assignment is stable across opens. The one case where it legitimately changes is sign-out/sign-in, and that self-corrects already: `identify` triggers `reloadFeatureFlags`, which fires the still-subscribed `onFeatureFlags` callback.

The first finding (override read after the experiment flag) was real and is fixed in da000b0.

PR #6457, opened 2026-08-14
feat(workspace-protocol): pin the workspace API both runtimes must serve

saddlepaddle:
Creates `packages/workspace-protocol`: the zod schemas for the workspace-scoped API that both a laptop's host-service and the future cloud sandbox runtime must serve. **host-service is untouched by this PR** (`git diff origin/main -- packages/host-service` is empty) — adoption is deliberately a follow-up, so the schemas can be reviewed while nothing depends on them.

## It is schemas, deliberately

An earlier revision of this PR defined a bespoke spec format (`ProcedureSpec`/`ProtocolNamespace`/`isProcedureSpec`) plus a compliance test that walked procedure paths. That was over-built — a parallel type system reimplementing a weaker version of what tRPC already does. Worse, the test asserted procedure **names** existed and never validated shapes, so an input schema changing from `z.string()` to `z.number()` would have passed silently.

It is now schemas and nothing else. **The drift guarantee comes from shared object identity**: both runtimes import the same schema object into their `.input()`/`.output()` calls, so they cannot disagree, and there is nothing to keep in sync or walk.

Also tested and rejected: sharing a router type and having each side `satisfies` it. A router built on one context is not assignable to a router built on another **even when the schemas match exactly**, because router types embed the context — and our two runtimes cannot share a context (host-service's carries db, api, git, providers, event bus; a sandbox's will not).

## Layout mirrors the router

`workspace-protocol/src/terminal/terminal.ts` sits opposite `host-service/src/trpc/router/terminal/terminal.ts`, and so on for all 15 namespaces, plus `events/`, `routes/`, `errors/`, and a `shared/` holding exactly one genuinely cross-namespace schema. Folder-per-module with barrels, per AGENTS.md.

Exports are namespace-prefixed (`terminalListInput`, `gitGetStatusOutput`); shared building blocks end in `Schema` and are used directly rather than aliased, since object identity is the whole mechanism. 172 exports covering all 64 procedures, verified collision-free at runtime.

## What the schemas capture that code didn't

- **Error taxonomy** — host-service throws TRPCErrors with informal `cause.kind` discriminants (`WORKTREE_MISSING`, `PATH_VALIDATION`, terminal kinds) that clients string-match with no shared type. Now a typed union.
- **`notifications.hook` is loopback-only** — it is a `publicProcedure`, unauthenticated so agent shells need no token, which is safe *only* because the caller dials 127.0.0.1. Nothing previously said so. Route it through the relay and an anonymous caller can forge agent-lifecycle events for any terminal id.
- **Terminal replay invariants** — the 64KB replay buffer, 2MiB `epoch:seq` catch-up ring, and `exact`/`tail`/`reanchor` attach modes, as constants. A runtime can serve every procedure correctly and still garble scrollback on reconnect without these.
- **Three deprecated names** — `terminal.daemon.*` (a sandbox has no daemon), `disposeWorktreeSessions` (no worktrees), `host.info` (the workspace *is* the host now), each with its intended replacement. Renames execute during adoption, when call sites are already being touched.

## Note for adoption (the follow-up PR)

host-service has **zero `.output()` validators today**. Half its inputs already use named schemas, so input adoption is a one-line import swap per procedure. But output schemas would be net-new *runtime enforcement* — a subtly wrong schema turns a working response into a 500. Adopt outputs as types first, add runtime validation per-procedure only once proven.

## Verification
lint exit 0 · typecheck exit 0 (38/38, uncached) · sherif exit 0

## Summary by CodeRabbit

* **New Features**
  * Added a workspace protocol package with validated contracts for agents, attachments, authentication, configuration, filesystem operations, Git/GitHub workflows, terminals, ports, pull requests, settings, notifications, and health checks.
  * Added WebSocket event and file-watching message formats.
  * Added route definitions for terminal sessions, events, and Chat streaming.
  * Added standardized error classifications and validation helpers.
  * Added support for binary content through base64 payloads.


saddlepaddle:
Closing — superseded by `mergeRouters`.

This package existed to give the sandbox and the host a shared, checkable contract. Testing that premise showed two tRPC routers with **different contexts are not assignable even with identical schemas**, so a shared-type `satisfies` contract can't span the two runtimes. `mergeRouters(workspaceRouter, hostOverlay)` gets both the narrow client type and the full one without a parallel copy of every zod schema.

Merging would have added ~1,749 lines needing hand-maintenance alongside the real routers, with no compile-time guarantee they stayed in sync — the compliance test here only ever compared procedure *names*, never shapes.

The real work is composing `workspaceRouter` in host-service directly. Branch kept for reference.

PR #6475, opened 2026-08-14
fix(desktop): create new folders and files reliably in the Files tab

distroinfinity:
## What & why

I could not create a folder from the Files tab. Upon checking into it turned up five related problems. Here they are in the order I hit them, with steps to reproduce.

### 1. New folder always failed

1. Open a project, Files tab
2. Click New Folder
3. Type any name, press Enter

Result: `Failed to rename / ENOENT: no such file or directory, rename '<root>/Untitled' -> '<root>/<name>'`. The row snaps back to `Untitled` and looks like a real folder, but nothing is created on disk. Renaming that row throws the same error.

image

### 2. Accepting the default name created nothing

New Folder, then press Enter without typing. Result: an `Untitled` row sits in the tree with nothing behind it.

### 3. A name that already existed left a dead row

New Folder, type the name of a folder that already exists there, Enter. Result: error toast, and the `Untitled` row stays in the tree pointing at nothing. The next New Folder then offers `Untitled-2`.

### 4. New folders ignored which folder I selected

Select folder A, New Folder. Select folder B, New Folder. Result: both land in the same directory. Selecting a folder made no difference, and there was no way to target the workspace root.

### 5. Creating after a rename failed

Open a file inside folder `X`, rename `X` to `Y`, click New Folder. Result: `ENOENT: no such file or directory, mkdir '<root>/X/Untitled'`.

### What I found

Two causes behind all five.

New Folder built an in-memory placeholder row keyed `Untitled/` (Pierre marks directories with a trailing slash) and only wrote to disk once you committed the name. But Pierre's `onRename` reports paths without the trailing slash even when its own `isFolder` flag is true, so the lookup missed, the commit fell through to the rename branch, and we asked the host to rename a directory that had never been created. Files were unaffected because their keys have no trailing slash. Pierre also emits no callback at all when a name is committed unchanged, and none that says which row an error belongs to, so 2 and 3 were not detectable from that flow.

For 4 and 5, the target directory came from the editor's open file rather than the tree selection. That is why it was fixed no matter what I highlighted, and why it went stale as soon as I renamed an ancestor folder.

## What I changed

Creation now writes to disk first, then opens the inline rename on a real entry. Every commit is an ordinary rename, so accepting the default name, cancelling, and name collisions all end up matching disk.

That means cancelling has to remove what we created, which needs primitives we did not have. `createDirectory` treats an existing directory as success, and `deletePath` is `rm -rf`, so cancel could have deleted a folder we did not create or one that had gained contents in the meantime. I added three to `workspace-fs`:

- `createUniqueEntry`: allocates `Untitled`, `Untitled-2`, and so on, where each attempt is an exclusive create. Naming moved to the host, so a stale client listing can no longer pick a name that already exists on disk.
- `removeEmptyDirectory`: `rmdir`, never recursive. Refuses a folder that is not empty.
- `removeFileIfUnchanged`: only unlinks if the file still matches the revision from when we created it.

New entries now go where I point them: selected folder, selected file's parent, or the workspace root. Clicking empty space below the rows clears the selection so root is one click away. Every candidate is checked against the tree, so a row pointing at a renamed or deleted directory falls back to root instead of erroring.

I put the provisional entry lifecycle in a pure reducer (`provisionalEntry.ts`) so it can be tested without React.

Two things to flag for review:

- This includes the selection targeting change, which is technically a second fix. It shares the same function and the same root cause as 5, and without it I could not test root level creation at all. Happy to split it if you would rather.
- On a name collision the inline editor now closes instead of reopening. Pierre's error callback gives a message and no path, so reopening had to guess which row it belonged to, and after an unchanged commit that guess could land on a folder I had kept and delete it on Esc. The folder keeps its name on disk, so F2 renames it.

## How I tested it

I reproduced every case above in the dev app before the fix, then reran the same steps after. Checked on disk in Finder each time, not just in the tree.

Creation

- New Folder, type a name, Enter: folder appears and exists in Finder
- New Folder, Enter on the default: `Untitled` actually exists
- New Folder, Esc: gone from disk
- Same three for New File, and it opens in the editor

Targeting

- Select folder A, New Folder: lands in A. Select folder B: lands in B
- Click empty space, New Folder: lands at root
- Select a folder, click Refresh, New Folder: still lands in that folder
- Open a file inside a folder, rename the folder, New Folder: no ENOENT

Safety

- Drop a file into a new folder before pressing Esc: folder and file survive, toast says it was kept
- New Folder, Enter on default, F2 that folder, type an existing name, Esc: folder still exists
- New File, Esc: file gone, no leftover pinned tab
- Create an `Untitled` folder in Finder without refreshing the tree, then New Folder: get `Untitled-2`, existing folder untouched

Automated

- 21 new tests in `packages/workspace-fs/src/fs.test.ts` for the three primitives, including "never adopts an existing directory" and "keeps a populated directory". Also added `movePath` coverage, which had none, pinning the ENOENT from 1
- 6 new integration tests through the host tRPC router for create then commit, and create then cancel
- 23 renderer tests for path canonicalisation, creation targeting, and the provisional reducer, including the case where cancel must not delete a committed folder
- Full suite: 2690 desktop, 1095 host-service, 77 workspace-fs, 0 failures. `bun run lint` and `bun run typecheck` clean

## Checklist

---
## Summary by cubic
Create files and folders on disk before inline rename in the Files tab, fixing ENOENT on commit, dead placeholders, wrong-target creation, and empty-after-rename folders. Previously we renamed in-memory placeholders; now the host creates a unique entry, we move it on commit, and cancel removes only what we created.

- Adds `workspace-fs` and `host-service` primitives: `createUniqueEntry` (exclusive `Untitled`/`-2` allocation), `removeEmptyDirectory` (non‑recursive), `removeFileIfUnchanged` (revision‑guarded).
- Canonicalizes `@pierre/trees` event paths: `canonicalizeTreePath` for renames and `resolveDeleteTreePath` so delete events map to the correct file/folder key.
- Targets creation from the tree selection (selected folder, selected file’s parent, or root). Clicking the empty tree background clears selection; header and controls don’t.
- Introduces bridge `addPath`/`removePath` and rekeys/purges directory bookkeeping so known paths, loaded dirs, and lazy‑load candidates move together; prevents newly renamed folders rendering empty.
- Invalidates stale tree listings and guards `listDirectory` with a workspace version and tree revision; retries only when the directory still exists and remains expanded.
- UX: does not reopen inline rename on errors; shows a toast and keeps the current name. New files open once (no accidental pinning).
- Tests cover fs primitives and guarded cancel paths, host endpoints, path canonicalization (including delete events), creation targeting, provisional cleanup, tree bookkeeping, and stale‑listing guards.

<sup>Written for commit 92ea68f09e81676a7bcaae5db8c09f3b1284b501. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6475?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Creating files and folders now immediately creates uniquely named entries on disk before inline renaming.
  * Renaming supports directory paths more reliably, with safe rollback after errors.
  * Rename validation errors are shown without reopening rename mode.
  * File-tree selection clears only when clicking empty background areas.

* **Bug Fixes**
  * Improved handling of name collisions, stale selections, workspace changes, and filesystem updates.
  * Added safeguards against deleting modified files or protected directories.
  * Improved tree consistency after moves and deletions.
  * Cancelled creations safely preserve existing files and non-empty folders.

### DIFF
--- a/apps/desktop/src/renderer/lib/pierreTree/index.ts
+++ b/apps/desktop/src/renderer/lib/pierreTree/index.ts
@@ -8,4 +8,4 @@ export {
 	type PierreGitStatus,
 	type PierreGitStatusEntry,
 } from "./pierreGitStatus";
-export { stripTrailingSlash } from "./treePaths";
+export { canonicalizeTreePath, stripTrailingSlash } from "./treePaths";
--- a/apps/desktop/src/renderer/lib/pierreTree/treePaths.test.ts
+++ b/apps/desktop/src/renderer/lib/pierreTree/treePaths.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "bun:test";
+import { canonicalizeTreePath, stripTrailingSlash } from "./treePaths";
+
+describe("canonicalizeTreePath", () => {
+	it("leaves file paths untouched", () => {
+		expect(canonicalizeTreePath("notes.txt", false)).toEqual("notes.txt");
+		expect(canonicalizeTreePath("src/notes.txt", false)).toEqual(
+			"src/notes.txt",
+		);
+	});
+
+	it("adds the trailing slash Pierre omits for directories", () => {
+		expect(canonicalizeTreePath("Untitled", true)).toEqual("Untitled/");
+		expect(canonicalizeTreePath(".claude", true)).toEqual(".claude/");
+	});
+
+	it("canonicalizes nested directory paths", () => {
+		expect(canonicalizeTreePath("src/components/Button", true)).toEqual(
+			"src/components/Button/",
+		);
+	});
+
+	it("is idempotent on an already-canonical path", () => {
+		expect(canonicalizeTreePath("Untitled/", true)).toEqual("Untitled/");
+		expect(
+			canonicalizeTreePath(canonicalizeTreePath("a/b", true), true),
+		).toEqual("a/b/");
+	});
+
+	it("round-trips with stripTrailingSlash", () => {
+		for (const path of ["Untitled", "src/components/Button", ".claude"]) {
+			expect(stripTrailingSlash(canonicalizeTreePath(path, true))).toEqual(
+				path,
+			);
+		}
+	});
+});
--- a/apps/desktop/src/renderer/lib/pierreTree/treePaths.ts
+++ b/apps/desktop/src/renderer/lib/pierreTree/treePaths.ts
@@ -6,3 +6,23 @@
 export function stripTrailingSlash(path: string): string {
 	return path.endsWith("/") ? path.slice(0, -1) : path;
 }
+
+/**
+ * Put a path back into `@pierre/trees`' canonical form, where directories carry
+ * a trailing `/`.
+ *
+ * Needed because Pierre is inconsistent about which form it hands out:
+ * `FileTreeRenameEvent` reports slash-less paths even when its own `isFolder`
+ * flag is `true`, while its model — and any bookkeeping keyed alongside it —
+ * uses the trailing-slash form. Canonicalize event paths before looking them up
+ * anywhere, or directory lookups silently miss.
+ *
+ * Idempotent, so it is safe to apply to a path that is already canonical.
+ */
+export function canonicalizeTreePath(
+	path: string,
+	isDirectory: boolean,
+): string {
+	if (!isDirectory) return path;
+	return path.endsWith("/") ? path : `${path}/`;
+}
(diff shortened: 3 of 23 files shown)


distroinfinity:
hey @Kitenite  
folder creation in the Files tab was broken for me, ENOENT on every rename. 
The row was never actually written to disk, and a few related issues fell out of that.

please take a look whenever you get time, happy to fix anything you flag

distroinfinity:
hey @Kitenite 
did you get a chance to check this?
I updated to most recent release today but folders and file creation are still broken in the files tab

PR #4727, opened 2026-05-19
revert: drop host-offline workspace gate from #4672

Kitenite:
## Summary
- Reverts the layout wiring and `useRemoteHostStatus` offline branch added in #4672, so v2 workspaces no longer render a full-screen gate when the paired host is marked offline.
- Keeps `WorkspaceHostOfflineState` on disk (unused) so we can reuse it once we have a less aggressive treatment.

## Why
- The hard block was too aggressive — workspaces should render optimistically and let individual operations degrade rather than blocking the whole page on `hosts.isOnline`.

## Test plan
- [ ] Open a v2 workspace whose paired host is offline → workspace UI renders normally (no full-screen offline state)
- [ ] Open a v2 workspace on a reachable host → unchanged
- [ ] `bun run typecheck` clean

---

<a href="https://stagereview.app/superset-sh/superset/pull/4727">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Reverts the offline host gate for v2 workspaces so the UI renders normally even when the paired host is offline. Drops the `offline` branch and `hosts.isOnline` check in `useRemoteHostStatus`; keeps `WorkspaceHostOfflineState` in the codebase for future use.

<sup>Written for commit a25834f726c30ca03deef60d57dddcc2ef8606ea. Summary will update on new commits. <a href="https://cubic.dev/pr/superset-sh/superset/pull/4727?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Refactor**
  * Removed the remote host offline status state and its associated UI component. Remote hosts now transition through alternative status flows instead of displaying a dedicated offline indicator.

[Review Change Stack](https://app.coderabbit.ai/change-stack/superset-sh/superset/pull/4727?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #6476, opened 2026-08-14
feat(desktop): surface a host unreachable screen for v2 workspaces

saddlepaddle:
## Summary

A v2 workspace whose host stops answering used to degrade into failing panes or a blank frame — the durable-object/relay connection dropping was never surfaced. This adds a takeover screen wired to the real connection.

- **`packages/workspace-client`** — the shared per-host event-bus socket now tracks `connecting | open | reconnecting | closed` plus the last `_whoowns` probe, exposed as `getConnectionStatus` / `subscribeConnectionStatus` / `reconnect()`.
- **`useHostReachability`** — reads that status, retains the bus so it keeps dialling, and reports unreachable only after **10s of sustained loss**. Keyed off the real data path rather than the cloud `isOnline` flag, whose drift got this gate reverted twice before (#4430, #4727).
- **`WorkspaceHostUnreachableState`** — revives the orphaned `WorkspaceHostOfflineState` left on disk by #4727, with a live Reconnecting/Disconnected chip, a reason line derived from the relay probe (503 host not connected, 401/403 unauthorized, 502/504, routing, relay unreachable), and a Retry button.
- **`WorkspaceHostGate`** — overlays the screen instead of replacing the workspace, so panes stay mounted and terminal scrollback, unsaved editor documents and in-flight agent sessions survive a drop.
- **`WorkspaceLocalHostPendingState`** — replaces the blank `<div>` the provider rendered when the local host service has no port, with copy per `hostServiceStatus` and a button that restarts the service through the same coordinator call the tray menu uses.

Two fixes came out of driving the running app over CDP:

- **Blank frame on a host-service restart.** Losing the port set `activeHostUrl` to null, which unmounted the whole workspace and left ~5s of empty pane before the service returned. The provider now holds the last known local host URL through a port gap. Measured: 45s across a kill and respawn, zero blank frames.
- **Recovery lagged 20.7s** behind the host coming back, riding partysocket's backoff toward its 30s ceiling. The hook now re-dials every 5s while the screen is visible → 5.7s.

## Test plan

- [x] `bun run typecheck` (desktop) and `biome check` clean; `packages/workspace-client` tests pass (13/13)
- [x] Normal opens and route changes never flash the screen — 189/181/286/308 samples at 100ms, zero visible frames, no console errors
- [x] Sustained local outage: screen appears 14.1s in (≈10s after the socket closed), correct copy, one visibility transition, no flicker
- [x] Recovery: clears 5.7s after the host is listenable again
- [x] Host-service crash: workspace stays mounted, no blank frame
- [ ] **Remote/relay path unverified** — the dev stack runs no relay and every other device has zero workspaces, so the six probe-status variants are only verified as rendered components, not end to end

## Screens

Captured from the running app over CDP — real host name, real host ID, not mocked.

**The takeover.** Amber Reconnecting chip while the socket keeps dialling, Retry and Host settings. Note the right sidebar stays reachable outside the overlay, which is why there is no "browse workspaces" escape link.

Host unreachable screen in the app

**Before — the blank frame this PR also fixes.** A host-service restart nulled the host URL, which unmounted the whole workspace and left ~5s of empty pane with no explanation. Measured 7.5s–12.7s on a plain crash.

Blank content area during a host-service restart

**After the host returns.** The screen lifts on its own — no reload, no navigation — 5.7s after the host is listenable again. (This workspace had no panes open, so it lands on the empty state.)

Workspace restored after the host reconnects

---
## Summary by cubic
Surfaces a Host unreachable overlay for v2 workspaces to prevent blank frames and failing panes. Previously we unmounted or left panes failing; now after 10s of sustained loss we cover the workspace, keep panes mounted, and redial every 5s to cut recovery lag.

- `packages/workspace-client`: per-host event bus now tracks `connecting | open | reconnecting | closed` plus last relay `_whoowns` probe; adds `getConnectionStatus`, `subscribeConnectionStatus`, and `reconnect()`; exports `HostConnectionState` and `HostConnectionStatus`. No migration required.
- `useHostReachability`: retains the bus, declares unreachable after 10s, runs a 5s redial loop only while the overlay is visible, and maps probe status to user-facing detail.
- `WorkspaceHostGate`: overlays instead of unmounting so terminals, unsaved files, and agent sessions survive; marks covered panes `inert` to stop focus behind the overlay; resolves host names via the cloud list; wrapper stays mounted to avoid remount on reconnect.
- `WorkspaceHostUnreachableState`: shows a Reconnecting/Disconnected chip, probe-derived reason, a Retry button wired to `reconnect()`, and a link to host settings.
- `WorkspaceLocalHostPendingState`: status-specific copy before the local host service exposes a port; can restart the service via the coordinator; never prompts restart while starting/running; shows a toast if restart is clicked without an active org.
- `WorkspaceProvider`: caches the last local host URL across host-service restarts to avoid blank frames; shows the pending state only before any port has ever been seen; keys `WorkspaceTrpcProvider` on the workspace (not the host URL) so a port change swaps transport without remounting panes.

**Review focus**
- Event-bus status transitions, probe wiring, listener lifecycles, and `retain`/`subscribeConnectionStatus` cleanup.
- Redial loop: runs only while the overlay is visible and doesn’t fight partysocket backoff; manual `reconnect()` sets state correctly.
- Overlay behavior: no flicker, children never remount, and covered panes are truly inert.
- Local-host pending path: restart behavior, no restart prompt while starting/running, and toast on missing org.

<sup>Written for commit 7107deed6b3e9c87fb650786c080721a3e7d99bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6476?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added clearer workspace host connection states, including connecting, reconnecting, unreachable, and disconnected.
  * Added automatic reconnection attempts and a manual retry option.
  * Added connection-specific messaging for relay and local host failures.
  * Preserved workspace content while displaying host connectivity status.
  * Added pending-state feedback when a local host is starting or restarting.
  * Improved handling of temporary local host service restarts.
  * Kept workspace content mounted and protected it from interaction during host outages.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/components/WorkspaceHostOfflineState/index.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/components/WorkspaceHostOfflineState/index.ts
@@ -1,1 +1,0 @@
-export { WorkspaceHostOfflineState } from "./WorkspaceHostOfflineState";
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/components/WorkspaceHostUnreachableState/WorkspaceHostUnreachableState.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/components/WorkspaceHostUnreachableState/WorkspaceHostUnreachableState.tsx
@@ -1,16 +1,29 @@
 import { Button } from "@superset/ui/button";
+import { cn } from "@superset/ui/utils";
 import { Link } from "@tanstack/react-router";
-import { ArrowRight, Monitor, Settings, WifiOff } from "lucide-react";
+import { Monitor, RefreshCw, Settings, WifiOff } from "lucide-react";
 
-interface WorkspaceHostOfflineStateProps {
+interface WorkspaceHostUnreachableStateProps {
 	hostId: string;
 	hostName: string;
+	/** Why this connection in particular is down. */
+	detail: string;
+	/** The socket is still dialling on its own backoff. */
+	isReconnecting: boolean;
+	onRetry: () => void;
+	retryLabel?: string;
+	retryBusyLabel?: string;
 }
 
-export function WorkspaceHostOfflineState({
+export function WorkspaceHostUnreachableState({
 	hostId,
 	hostName,
-}: WorkspaceHostOfflineStateProps) {
+	detail,
+	isReconnecting,
+	onRetry,
+	retryLabel = "Retry",
+	retryBusyLabel = "Retrying…",
+}: WorkspaceHostUnreachableStateProps) {
 	return (
 		<div className="flex h-full w-full items-center justify-center p-6">
 			<div className="flex w-full max-w-sm flex-col items-start gap-6">
@@ -32,12 +45,15 @@ export function WorkspaceHostOfflineState({
 
 				<div className="flex flex-col gap-1.5">
 					<h1 className="text-[15px] font-medium tracking-tight text-foreground">
-						Host offline
+						Host unreachable
 					</h1>
 					<p className="select-text cursor-text text-[13px] leading-relaxed text-muted-foreground">
-						This workspace lives on a paired device that is not reachable right
-						now. Terminals and file actions are unavailable until that device
-						reconnects.
+						This workspace lives on a device Superset can't reach right now.
+						Terminals, files, and agents stay put — they come back as soon as
+						the connection does.
+					</p>
+					<p className="select-text cursor-text text-[13px] leading-relaxed text-muted-foreground/80">
+						{detail}
 					</p>
 				</div>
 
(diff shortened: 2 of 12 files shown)


saddlepaddle:
Addressed CodeRabbit's review in 7107deed6 — three of four were real:

- **Inert panes** — valid. The overlay keeps `children` mounted, so panes were still taking Tab stops and screen-reader focus behind the takeover. Wrapped them in `inert={isUnreachable}` (React 19 boolean prop; Chromium blurs focus out of a subtree as it becomes inert).
- **Silent retry click** — valid. Without an active organization the button still rendered enabled, and the click was swallowed. It now says why.
- **Workspace key** — valid, and it undercut this PR's own claim. `key={workspace.id}:${hostUrl}` remounted every pane when the host service came back on a different port, losing exactly the scrollback and unsaved documents the overlay exists to preserve. Keyed on the workspace alone; `WorkspaceClientProvider` already rebuilds its clients on a `hostUrl` change, so the transport swaps underneath.

**Skipped: retries exhaustion → `closed`.** The premise doesn't hold for this socket. `getOrCreateConnection` never passes `maxRetries`, so partysocket uses its default of `Infinity` and exhaustion can't occur — `shouldReconnect` staying `true` is correct here, and the chip reading "Reconnecting" is accurate because it genuinely still is. Worth noting the flip side: `closed` is therefore only reachable via an explicit `close()`, so the Disconnected variant is rare in practice, which matches what I measured driving the real app.

Issue #5113, opened 2026-06-04
[bug] Remote host workspace sessions show a static, non-animated agent activity indicator in the sidebar


0xNoWater:
## Description

Workspaces running on a **remote host** show a static (non-animated) status icon in the desktop sidebar while an agent session is actively running on that host — the icon never animates or reflects the "agent working" state. Local workspaces animate as expected. To tell whether a remote agent is still working, you have to click into the workspace.

## Expected behavior

Remote-host workspace rows should show the same animated agent lifecycle activity indicator (working / needs permission / ready for review) as local workspace rows.

## Possibly related

- #2610 — the same symptom was reported as a V2-sidebar regression and fixed, but the fix doesn't appear to cover remote-host rows.
- #4407 and #4398 suggest remote/cloud-sourced sidebar rows take a different data path (Electric sync / cloud-row fallback) that drops state the local path carries — the activity/lifecycle state may be getting lost the same way.

## Environment

- Agent: Claude Code (running on the remote host)
- Superset desktop v1.12.2, macOS 26.4.1 (Apple Silicon)
- superset CLI v0.2.16 on the remote host

Issue #3931, opened 2026-05-01
Status indicator dots missing — SUPERSET_HOST_AGENT_HOOK_URL not set in v2 hook environment


RaygunIcecream:
## Description

After updating Superset, the colored status indicator dots (green/orange) on workspace sessions in the sidebar are no longer appearing. Sound notifications and OS toasts still work fine.

## Root Cause (suspected)

`SUPERSET_HOOK_VERSION=2` is set in the terminal environment, but `SUPERSET_HOST_AGENT_HOOK_URL` is not. So `notify.sh` skips the v2 host-service path (line 91) and falls back to the v1 `127.0.0.1:51741/hook/complete` endpoint — which the new UI may not be listening on anymore.

```
# From my env:
SUPERSET_HOOK_VERSION=2
SUPERSET_HOST_AGENT_HOOK_URL=  # not set
SUPERSET_PORT=51741
```

## Steps to Reproduce

1. Open Superset with multiple workspace sessions
2. Run a Claude Code session that completes or requests permission
3. Observe: OS toast and sound fire, but no colored dot appears on the session in the sidebar

## Expected Behavior

Colored status dots should appear on workspace sessions: amber (working), orange (needs input), green (ready for review).

## Environment

- macOS
- Superset production env
- Claude Code 2.1.88
- `notify.sh` last updated Apr 27

## Screenshot

![Missing dots](https://github.com/user-attachments/assets/placeholder)

*(Status dots are absent where arrows point in the sidebar — sessions that have completed or are waiting for input show no indicator)*

korthout:
I also face the issue that the status indicator dot is missing. However, I'm on `v1.12.2`. 

EDIT: I guess that's https://github.com/superset-sh/superset/issues/3689 instead.

Kitenite:
Closing as fixed: the failure mode here (hook env pointing at a stale/missing host endpoint, so status dots silently stop) was addressed by PR #6480 (commit b0c35e417) — the agent hook endpoint is now resolved at call time instead of being baked into the terminal environment, so host-service restarts and version-skewed envs no longer strand running agents' status events. PR #6479 also fixed hook provisioning on headless CLI hosts. Both first released in v1.23.0. If dots still go missing on 1.23+ please reopen with the app version.

PR #6479, opened 2026-08-14
fix(host-service): provision agent hooks and shell env on headless CLI hosts

### LINKED ISSUE #6254: [bug] Remote/headless hosts never provision agent lifecycle hooks — sidebar status permanently blank for cloud workspaces
**Where:** Desktop app (macOS) + a self-hosted remote host running the CLI host-service
**Version:** 1.19.0 on both sides (desktop 1.19.0; remote host `/opt/superset/current/bin/superset --version` → 1.19.0). Code references below are against `ddd3d48c8555885a5a0d38de154f7fa0f95b728f` (main, 2026-08-06).

## What happened

Workspaces that live on a **remote host** never show agent activity in the dashboard sidebar. They render a static cloud icon: no `AsciiSpinner` while an agent is working, and no `StatusIndicator` dot when one finishes or needs input. Workspaces on the local device, in the same sidebar, animate correctly.

The remote host in question had **18 live terminal sessions** with Claude Code actively running in several of them, and `terminal_agent_bindings` in its host DB contained **0 rows**. It is not a rendering bug — the host has nothing to report.

## Root cause: two independent defects, both on the host side

Agent status flows: `notify.sh` → POST `notifications.hook` → `terminalAgentStore.recordEvent()` → `terminal_agent_bindings` → `terminalAgents.listByWorkspace` → `useV2WorkspaceNotificationStatus` → sidebar icon.

`notifications.ts:85` (`recordEvent`) is the **only** non-test writer to that store. If `notify.sh` never runs, the table stays empty forever, and every downstream layer correctly reports "nothing happening".

Both defects break that first step. Fixing either one alone is not enough.

### Defect 1 — `SUPERSET_HOME_DIR` never reaches the PTY on a headless host

> **Corrected 2026-08-07.** As first filed, this section named `terminal.ts:1400` as *the* defect.
> The accurate framing — from the triage on this issue, and confirmed since — is that the
> **origin** is one layer up: the CLI never forwards the variable to the host-service it spawns
> (`packages/cli/src/lib/host/spawn.ts:107-124`). `terminal.ts:1400` is where that absence turns
> into damage, because it has no fallback. Both are worth fixing, but `spawn.ts` is the one that
> reaches a systemd deployment — see the unit file below. **PR #6256** fixes both; it does not
> address Defect 2.

`packages/host-service/src/terminal/terminal.ts:1400`:

```ts
const supersetHomeDir = process.env.SUPERSET_HOME_DIR || "";
```

Compare the same package's daemon, `packages/host-service/src/daemon/manifest.ts:33-35`:

```ts
function supersetHomeDir(): string {
	return process.env.SUPERSET_HOME_DIR || join(homedir(), ".superset");
}
```

The daemon has a `homedir()` fallback; the terminal path does not.

That read only ever succeeds because **the desktop injects the variable into the host-service process it spawns** — `apps/desktop/src/main/lib/host-service-coordinator.ts:789`:

```ts
SUPERSET_HOME_DIR: SUPERSET_HOME_DIR,
```

That is the only non-test producer outside a build script (`apps/desktop/scripts/verify-single-flight.ts:36`); nothing in `packages/host-service` sets it. So a host-service started by the desktop inherits it, and a host-service started by systemd inherits nothing — `supersetHomeDir` is `""`, and `env.ts:248` skips the assignment:

```ts
if (supersetHomeDir) {
	env.SUPERSET_HOME_DIR = supersetHomeDir;
}
```

Every managed hook command is guarded on exactly that variable (`agent-wrappers-common.ts:29`):

```
[ -n "$SUPERSET_HOME_DIR" ] && [ -x "$SUPERSET_HOME_DIR/hooks/notify.sh" ] && SUPERSET_AGENT_ID=claude "$SUPERSET_HOME_DIR/hooks/notify.sh" || true
```

so the hook silently no-ops. The `|| true` makes it invisible — no error surfaces anywhere.

`terminal.ts:1415-1416` read `SUPERSET_AGENT_HOOK_PORT` and `SUPERSET_AGENT_HOOK_VERSION` the same way and are empty for the same reason. The same `""` also flows into `getSupersetShellPaths()` (`shell-launch.ts:26-34`), where `path.join("", "bin")` yields the relative path `bin` rather than an absolute one.

The remote host's systemd unit launches via the **CLI**, which is why `spawn.ts` is the layer that
matters here:

```
[Service]
Type=exec
EnvironmentFile=/etc/superset/env
ExecStart=/opt/superset/current/bin/superset start --org <redacted> --port 8420
```

Observed on three live Claude PTYs on the remote host:

```
SUPERSET_WORKSPACE_ID=<set>
SUPERSET_TERMINAL_ID=<set>
SUPERSET_HOST_AGENT_HOOK_URL=http://127.0.0.1:8420/trpc/notifications.hook   ← correct
SUPERSET_AGENT_HOOK_VERSION=                                                 ← empty
SUPERSET_HOME_DIR                                                            ← absent
```

Worth stressing: the hook **URL is correct**. This is not #3931, where the URL itself is missing on a local host — here the environment is fine and the script simply cannot be located.

### Defect 2 — `agent-setup` is desktop-only, so a headless host has no `notify.sh` and no hook config

`setupAgentIntegrations()` lives at `apps/desktop/src/main/lib/agent-setup/index.ts:30` and is called from `apps/desktop/src/main/index.ts:472`. It is what writes `~/.superset/hooks/notify.sh` (`notify-hook.ts:createNotifyScript`) and merges the managed hook entries into `~/.claude/settings.json` (`managed-json-hooks.ts`). Nothing in `packages/host-service` provisions any of it — the only reference from that package is a doc comment at `shell-launch.ts:4`.

A host that has never had the desktop app installed therefore has:
- no `~/.superset/hooks/notify.sh`
- no Superset hook entries in `~/.claude/settings.json`

This is the normal state for a cloud/remote host, which is exactly the configuration the feature is supposed to serve.

## Reproduction

1. Provision a machine with the CLI only (no desktop app) and run `superset start --org <id> --port 8420` under systemd.
2. Register it as a remote host and create a workspace on it.
3. Run Claude Code in a terminal in that workspace and let it complete a turn.
4. Sidebar: the workspace shows a static `TbCloud` icon throughout — no spinner, no dot.
5. On the host: `sqlite3 ~/.superset/host/<org>/host.db 'select count(*) from terminal_agent_bindings;'` → `0`.

## Evidence (same org, same version, same account)

| | Mac, desktop installed | Remote host, CLI only |
|---|---|---|
| `~/.superset/hooks/notify.sh` | present | **absent** |
| Superset hook entries in `~/.claude/settings.json` | 8 | **0** |
| `SUPERSET_HOME_DIR` in PTY | `/Users/<user>/.superset` | **unset** |
| `SUPERSET_AGENT_HOOK_VERSION` | `2` | **empty** |
| `SUPERSET_HOST_AGENT_HOOK_URL` | set | set |
| live terminal sessions | 20 | 18 |
| `terminal_agent_bindings` rows | 10 | **0** |

## Confirmed by fixing it manually

On the remote host I applied the desktop's own provisioning by hand:

1. Copied `notify.sh` verbatim from the Mac to `~/.superset/hooks/notify.sh` (mode 755).
2. Added to `~/.claude/settings.json` the 8 managed hook groups (`Stop`, `UserPromptSubmit`, `PostToolUse`, `PostToolUseFailure`, `PermissionRequest`, `SessionStart`, `SessionEnd`, `StopFailure`) with the exact upstream command string.
3. Worked around Defect 1 by adding `"env": { "SUPERSET_HOME_DIR": "/home/<user>/.superset" }` to the same file — Claude Code injects it, compensating for the PTY never receiving it.

`terminal_agent_bindings` went from 0 to populated, with rows written by **real** Claude sessions (genuine session UUIDs, not synthetic POSTs) across multiple workspaces, and the sidebar started animating for the remote workspaces. Disposing a terminal correctly cleared its binding via `markTerminalExited` (`terminal.ts:281`), so the read path and the store were never at fault.

Step 3 is the load-bearing one and is why I'm reporting Defect 1 separately: without it, steps 1–2 change nothing, because the guard fails before it ever tests for the script.

## Suggested fix

1. **Forward it from the CLI, and give the terminal path a fallback.**
   `packages/cli/src/lib/host/spawn.ts` already has the canonical value available via
   `packages/cli/src/lib/config.ts` — add `SUPERSET_HOME_DIR` to the child env it builds. That is
   what fixes a systemd host, whose unit runs `superset start`. Separately, give
   `terminal.ts:1400` the same `homedir()` fallback the daemon already has at
   `daemon/manifest.ts:33`, for launchers that exec the host binary directly.
   → **PR #6256** does both, with a regression test that fails without the `spawn.ts` line.
2. **Provision agent integrations on any host that serves terminals**, not only in Electron. Either move the filesystem-writing half of `agent-setup` into `packages/host-service` and run it at host start, or expose it as a CLI step (e.g. `superset host agent-setup`) that the install/registration flow invokes. Whichever way, a headless host needs `notify.sh` on disk and the managed hook entries in the agent config.

**Status:** (1) is proposed in **PR #6256**. (2) is still open and unclaimed — until it lands, a
headless host has no `notify.sh` to run even once `SUPERSET_HOME_DIR` is set, so the reported
symptom is not fully fixed by #6256 alone. Happy to take (2) as well if the maintainers have a
preference on where the provisioning should live (host-service startup vs. an explicit CLI step).

## Relationship to existing issues

- **#5113** reports this exact user-visible symptom (2026-06-04, `0xNoWater`). It was closed as *completed* on 2026-07-10 with no comments, but its fix PR **#5115 was never merged** (`mergedAt: null`, closed 2026-07-11), and `hydrateAgentStatus.ts` does not exist on main. The symptom is still live. #5115's diagnosis is also stale now: the status path has since been rewritten from a live-event store to the `useTerminalAgentBindings` query (`useV2NotificationStatus.ts:64`), so the bug survives for a different reason than the one that was closed.
- **#3931** is a sibling, not a duplicate: there `SUPERSET_HOST_AGENT_HOOK_URL` is unset on a *local* host. Here it is set correctly and `SUPERSET_HOME_DIR` + the script are what's missing.

Kitenite:
**Links**
- Fixes #6254 (both defects; supersedes the `spawn.ts`/`terminal.ts` half of #6256 — happy to rebase whichever lands second)
- Related: #3931, #5113 (same user-visible symptom)

## Summary
- Move `apps/desktop/src/main/lib/agent-setup/` into a new shared `packages/agent-setup` and run it at standalone host-service startup, so CLI/systemd-launched hosts provision agent lifecycle hooks, PATH wrappers, shell bootstrap, and managed skills — previously Electron-only, leaving remote workspaces with permanently blank agent status.
- Give host-service a `~/.superset` fallback for `SUPERSET_HOME_DIR` (terminal launch + shell-ready evidence) and make the CLI forward the variable to the host it spawns — without this, provisioned hooks self-disable on their own guard.
- Add a machine-shared per-agent hooks-disable state: the desktop mirrors its `disabledAgentHooks` setting to `~/.superset/agent-hooks.json` (boot + every toggle) and headless provisioners honor it plus a `SUPERSET_DISABLED_AGENT_HOOKS` env override — so a machine running both converges instead of flip-flopping, and headless-only hosts get a disable mechanism at all. The login-shell merge also gained a denylist (`NODE_ENV`/`NODE_OPTIONS`/`ELECTRON_RUN_AS_NODE`): a dotfile exporting `NODE_ENV=development` would otherwise flip the host into dev-mode shutdown, killing the pty-daemon on restart.
- Merge the login-shell env/PATH into the standalone host process at startup, mirroring what the desktop's `getProcessEnvWithShellPath()` does for hosts it spawns, so `git`/`gh`/credential helpers resolve under systemd/cron/non-login launches.

## Why / Context
Issue #6254: a CLI-only host had 18 live Claude sessions and **0** rows in `terminal_agent_bindings`. Two independent defects: (1) `SUPERSET_HOME_DIR` never reached the PTYs, and (2) nothing outside Electron ever wrote `notify.sh`, the managed hook entries, `~/.superset/bin`, or the zsh/bash bootstrap that host-service's `shell-launch.ts` expects (it's an explicit read-only mirror of the desktop writers). PR #6256 addressed only (1). Fixing either alone changes nothing user-visible.

## How It Works
- **`packages/agent-setup`** is the old desktop module nearly verbatim (git tracks the renames). Two things had to become lazy for a shared package: the Superset home dir (`resolveSupersetHomeDir()` — resolved at call time because the desktop rewrites `process.env.SUPERSET_HOME_DIR` during boot for dev-workspace builds) and the templates dir (`setAgentSetupTemplatesDir()` — bundles lose `__dirname`-relative assets). Templates live at the package root; each distribution copies them next to its bundle: desktop → `dist/main/templates` (existing vite copy plugin, new source path), CLI tarball → `lib/agent-templates` (+ `plugin/` = repo Claude plugin), TS source → package-relative fallback.
- **Host-service** calls `provisionAgentIntegrations()` from `serve.ts` only — the standalone entry. The desktop uses its own host-service entry and keeps provisioning in Electron boot (it owns the per-agent `disabledAgentHooks` setting, which has no host-side equivalent; headless hosts provision all agents).
- **Env**: `packages/cli` `spawn.ts` now sets `SUPERSET_HOME_DIR` on the child (matching `host-service-coordinator.ts`), and the two host-service readers that resolved `\"\"` when unset now use the shared resolver. `applyLoginShellEnvToProcess()` reuses the existing terminal base-env snapshot (one login-shell probe per boot); launcher-set vars and PATH order always win, login-shell entries append.
- Dropped `agent-setup/utils.ts` (dead code, flagged as such by `no-main-process-blocking.test.ts`; its biome + ratchet allowances removed).

## Manual QA Checklist

Validated end-to-end in Docker (linux-arm64, the `build-dist-linux-docker.sh` CI-mirror flow) simulating a systemd host: `env -i` stripped PATH, fresh `$HOME`, no `SUPERSET_HOME_DIR`, running the shipped `superset-host`:

- [x] Repo smoke test passes on the built tarball (native addons, pty spawn, host boot)
- [x] First boot provisions `notify.sh` (v8), 12 bin wrappers, zsh/bash bootstrap with OSC 133 marker, all 8 Claude hook groups, Codex/Gemini configs, managed skills — resolved from `lib/agent-templates`
- [x] Full #6254 chain: provisioned `notify.sh` → POST `notifications.hook` (200) → real row in `terminal_agent_bindings` (seeded terminal session)
- [x] Unknown terminal id still 200s and records nothing (hook is deliberately unauthenticated)
- [x] Restart is idempotent: `notify.sh` mtime unchanged, no duplicate hook entries
- [x] Login-shell-only PATH dir (via `.profile`) merged into host process (log asserted)
- [x] Desktop: `electron-vite build` copies templates+plugin to `dist/main/templates`; bundles carry the provisioning + `setAgentSetupTemplatesDir(__dirname/templates)` wiring
- [x] Desktop dev boot from a worktree provisions into `superset-dev-data` (dev-workspace `SUPERSET_HOME_DIR` case) and logs `[agent-setup] Agent integrations provisioned`
- [x] macOS TS-source simulation (sandboxed `$HOME`): host-service provisioning writes all artifacts, plugin falls back to repo `plugins/superset`

## Testing
- `packages/agent-setup`: `bun test` (109 pass; moved suites) + `tsc`
- `packages/host-service`: `bun test` (1089 pass, incl. `no-electron-coupling`) + `tsc`; new `login-shell-env.test.ts` (5 tests, verified they fail against a mutated implementation)
- `packages/cli`: `spawn.test.ts` + `tsc`
- `apps/desktop`: `bun run typecheck`; suites around every touched import site (terminal, terminal-host, settings/workspaces routers, bundled-cli, `no-main-process-blocking`) — 460+ pass
- `bunx sherif`, `biome check` on all touched files
- The headless E2E above is checked in as `packages/cli/scripts/headless-e2e.sh` and runs automatically in `build-dist-linux-docker.sh` after the smoke test (guarded behind `SUPERSET_HEADLESS_E2E=1` — it wipes `$HOME` agent configs, so container-only). Verified end-to-end via the checked-in flow.

## Design Decisions
- **Shared package vs. moving into host-service**: the desktop still needs the same writers (settings toggles call `setupSingleAgent`/`teardownSingleAgent`; terminal spawn uses the shell-wrapper args), so a package both depend on avoids a desktop→host-service source dependency. Host-service's `no-electron-coupling` test stays honest — the package has zero Electron imports.
- **Provision in `serve.ts`, not `app.ts`**: `app.ts` is shared with the desktop's host-service child, which would double-provision against the desktop's own boot call (and ignore `disabledAgentHooks`).
- **`noUncheckedIndexedAccess: false` in the package tsconfig matches the desktop config the code was written under**, but the five sites that leaked into stricter consumers were fixed properly — host-service compiles the package source under its own strict options.

## Known Limitations
- Cross-version churn: a desktop and CLI at different versions rewrite each other's managed files at each boot (content markers differ). Both versions stay functional; writes are atomic (temp+rename), so this is churn, not corruption. Accepted.
- The full remote loop (headless host registered to a cloud org, viewed from a desktop through the relay) was verified as two halves — host side (Docker E2E: hooks → bindings) and UI side (CDP: bindings → sidebar indicator) — not as one wire; that seam is pre-existing plumbing this PR doesn't touch.

## Follow-ups (surveyed, intentionally not in this PR)
Other Electron-only functionality CLI hosts miss, per the #6254-class audit: chat-v3 migrations/Claude bin absent from the CLI tarball, `HOST_MANIFEST_DIR` (attachments land in `host/standalone`), per-terminal resource metrics, `serve.ts` binding all interfaces without a `hostname`, `.superset/` worktree copy.

## Risks / Rollout
- Risk: provisioning now runs on hosts it never ran on; every writer is individually try/caught (one bad `$HOME` can't block boot) and content-gated (marker/signature checks — never touches user-owned files), the same code the desktop has shipped for months.
- Rollout: no ordering constraints; desktop and CLI halves are independently safe (env fallbacks cover version skew).
- Rollback: revert the two commits; provisioned files are inert without a Superset terminal (`notify.sh` exits unless `SUPERSET_TERMINAL_ID` is set).

---
## Summary by cubic
Provision agent hooks, PATH wrappers, and shell bootstrap on standalone host-service startup, and merge the login-shell env. Previously only Electron-launched hosts provisioned these; headless hosts inherited a minimal env, leaving agent status blank and tools unresolved.

- Shared `@superset/agent-setup`: desktop sets `setAgentSetupTemplatesDir(path.join(__dirname, "templates"))`, mirrors disabled agents via `writeSharedDisabledAgentIds()`, and calls `setupAgentIntegrations()`. Standalone host-service resolves templates from `lib/agent-templates` or `SUPERSET_AGENT_TEMPLATES_DIR` and runs `provisionAgentIntegrations()`; missing templates now emit one loud error.
- Shared disable state: desktop mirrors its disable list to `~/.superset/agent-hooks.json`; headless hosts honor the file and `SUPERSET_DISABLED_AGENT_HOOKS`, so all provisioners converge on one set.
- Env: CLI forwards `SUPERSET_HOME_DIR`; host-service falls back to `~/.superset` via `resolveSupersetHomeDir()` in terminal launch and shell-ready evidence. Host-service merges login-shell env/PATH at boot (launcher vars and PATH order win; missing PATH entries append), never imports `NODE_ENV`, `NODE_OPTIONS`, or `ELECTRON_RUN_AS_NODE`, and validates `DESKTOP_NOTIFICATIONS_PORT` with a 51741 default.
- Hardening: provisioning writes are idempotent and atomic via `writeFileIfChanged`, avoiding torn user configs under concurrent provisioners.
- Distribution/CI: CLI tarball ships `lib/agent-templates/` plus the bundled Claude plugin. Linux CI installs zsh and runs a headless-host E2E that verifies first-boot provisioning, notify.sh → `notifications.hook` DB writes, login-shell PATH merge under a stripped env, disable/enable paths, concurrent provisioners, and restart idempotency. A per‑agent matrix dispatches through each of the 13 agents’ own artifacts and asserts host DB bindings.

**Rollout**
- No migration. Custom launchers must co‑locate templates next to the host bundle or set `SUPERSET_AGENT_TEMPLATES_DIR`. Headless hosts provision all agents; set `SUPERSET_DISABLED_AGENT_HOOKS` to disable specific agents.

<sup>Written for commit 681df2121af75f5cbb0f7a7a3361223fe3de419a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle integrations for Droid, Kimi, Mastra, and Vibe.
  * Expanded coding-agent support with managed hooks, wrappers, notifications, skills, and support for 13 agents.
  * Added automatic integration provisioning for desktop, standalone, and headless environments.
  * Added persistent controls for disabling selected agent hooks.
  * Applied login-shell environment settings more consistently to terminal sessions.
  * Added configurable template and installation locations.

* **Bug Fixes**
  * Improved preservation of existing agent configuration during updates.
  * Increased reliability when recovering from incomplete or invalid configuration files.
  * Improved setup consistency across desktop, CLI, and host-service environments.


Kitenite:
Addressed all three CodeRabbit findings in d9edb919f: port-range validation in `getV1NotificationsPort`, the headless-e2e login-shell-probe race (poll for the merge log line instead of a fixed sleep), and the `set -e` guard on the restart `kill`.

Kitenite:
Two hardening changes in de3e9b9b5, addressing failure modes from a self-review:

- **Atomic provisioning writes** — the three duplicated `writeFileIfChanged` copies are now one temp+rename implementation, so concurrent provisioners (desktop + one CLI host-service per org) can no longer leave a torn `~/.claude/settings.json` (which the managed-hooks merge would skip rather than repair).
- **Headless E2E in CI** — `build-cli.yml`'s linux jobs now run `headless-e2e.sh` after the smoke test, so the full provisioning → notify.sh → `notifications.hook` → bindings chain (plus login-shell PATH merge and restart idempotency) is verified on every linux-x64/arm64 CLI build, not just local Docker repros.

PR #6488, opened 2026-08-14
feat(mobile): reorder terminal sessions from a sessions sheet, persisted per workspace

saddlepaddle:
## What & why

Terminal session tabs were locked to creation order. You can now drag sessions
into any order, and the arrangement persists per workspace across app restarts.

**Where the reorder lives, and why it isn't the tab strip.** The strip's chips
are ~28pt tall inside a horizontal ScrollView, so an in-strip drag competes with
the scroll on the same axis — it has to be gated behind a long press, which is
already bound to "Close session". I built that version first (long press to pick
up, swipe up to close) and it worked, but it puts two hidden gestures on one
28pt target, makes a destructive action undiscoverable, and can't reach
off-screen slots without edge auto-scroll. Safari makes the same call: you can't
reorder tabs in the iPhone tab bar, only in the tab grid.

So reordering moved to a **sessions sheet**, opened by a new list button next to
the strip's `+`:

- each row has its own **drag handle** — the drag activates on movement, no long
  press, nothing inferred from timing
- **✕ per row** closes (same confirm alert as before), so close is visible
  rather than gestural
- tapping a row switches to that session and dismisses the sheet
- the strip is unchanged otherwise, including its existing long-press-to-close

**Ordering rules** (`orderTerminalRows`): saved arrangement first, then anything
it doesn't mention appended in creation order. New sessions append at the end;
closed sessions drop out without moving their neighbours; a session that appears
on the host while the app is away is appended rather than reshuffling the saved
order. Only ids the user actually arranged are stored.

Persistence follows the existing `screens/(authenticated)/stores/` pattern
(zustand + `persist` + AsyncStorage), keyed `terminal-tab-order-v1`, same shape
as `terminalSeenStore` / `pinnedWorkspacesStore`.

## How I tested it

iOS simulator (iPhone 16e, iOS 26), driven with Maestro against a harness screen
wired exactly like the sheet — same `SessionList`, same store, same
`orderTerminalRows` — since a live host wasn't available in this worktree:

- drag by handle moves a row the expected number of slots (verified for moves up
  and down, and after the list had been scrolled); order and saved order match
- order survives a cold app restart (`terminate` + `launch`), repeatedly
- new sessions append at the end and leave the saved order untouched
- closing a session drops it out; the remaining rows keep their arrangement
- tapping a row selects it; ✕ removes it
- the list still scrolls normally with 11 rows

Two bugs the simulator caught: the width map for the first (in-strip) version
only ever held the last chip's measurement, because sibling `onLayout` writes in
the same tick can't read a shared value back; and rows were merging into one
VoiceOver element, hiding the ✕ and the handle — the row is now three sibling
elements with their own labels.

**Terminal scrollback** (asked for separately): verified good. Loaded the real
generated xterm page with the same WebView props (`scrollEnabled={false}`,
`bounces={false}`), filled it with 300 lines, and touch-dragged — the view moved
from lines 258–300 back to 231–274 with the scrollbar visible. The outer
WebView's disabled scroll view does not block xterm's own viewport. Note that in
the **alternate screen** (Claude Code, vim, less) there is no scrollback by
design — drags become mouse-wheel reports to the app, which is what the
`MALFORMED_MOUSE_REPORT` guard in `generate-terminal-html.ts` handles. Same as
desktop and web.

Not runtime-tested: the sheet's own data wiring (host lookup, `killSession`)
needs a live host. It mirrors `NewSessionSheet` exactly, and the confirm alert
it shows is the same call the strip's long-press uses, which I did verify.

`bun run lint:fix` and `bun run typecheck` pass in `apps/mobile` (after
`bun run build:types` in `packages/host-service`).

## Checklist

---
## Summary by cubic
Reorders terminal sessions from a Sessions sheet and persists the arrangement per workspace. Tabs were fixed to creation time; they now render in the saved order, and the active session stays attached while you reorder.

- Adds a `workspace/[id]/sessions` sheet (formSheet) opened by a list button next to “+”: vertical list with native drag via `@expo/ui` SwiftUI `List.ForEach` in edit mode (system grabber, edge auto‑scroll, VoiceOver), tap to switch and dismiss, in‑row ✕ to close, active tint, and an empty state. Styles the list as plain and removes custom drag/hold logic and the system delete control.
- Persists order with `useTerminalTabOrderStore` (`zustand` + `persist` + `@react-native-async-storage/async-storage`) under `terminal-tab-order-v1` (device‑local). Ordering: saved ids first; unknown sessions append by creation time; closed sessions drop out; only arranged ids are stored.
- Integrates with `WorkspaceScreen`: derives tab rows via `orderTerminalRows` and pins the active session during reorders; `TerminalTabs` adds a manage button to open the sheet.
- Fixes: render the list as the sheet’s only layout child to avoid zero‑height cold mounts; show an alert if closing a session fails and still invalidate the host terminal query.

<sup>Written for commit 459c33354580e41f1d545ffe9641c1b182cf8bf4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a sessions screen for viewing, selecting, reordering, and closing workspace terminal sessions.
  * Added controls to manage sessions and start new ones.
  * Terminal tab order now persists separately for each workspace.
  * Added session status indicators, active-session highlighting, agent icons, and an empty state.
* **Bug Fixes**
  * Selecting a session no longer unexpectedly changes the visible session due to tab reordering.


saddlepaddle:
Follow-up: the sheet presented **empty** on a real device — the wrapping `View` around the list hit the known formSheet cold-mount bug (flex-derived children lay out at zero height; fast refresh masks it). Fixed in 95458de38: the list is now the sheet's only layout child, with the empty state inside it. Verified against a live host — the sheet lists the workspace's real sessions with handles and ✕.

saddlepaddle:
Swapped the hand-rolled drag for a native SwiftUI list. `@expo/ui` (already a dependency) ships `List.ForEach` backed by SwiftUI's `onMove`/`onDelete`, so the drag, edge auto-scroll, lift animation and VoiceOver support are UIKit's — and ~200 lines of gesture arbitration, index math and hand-tuned animation timing are deleted. The list runs in edit mode: system grabber per row, leading remove control. Verified rendering, sizing and tap-to-select against a live host; the drag itself isn't scriptable (synthetic swipes don't drive UIKit's reorder control) so it wants a real finger.

saddlepaddle:
CodeRabbit's two findings, addressed:

- **Unhandled `killSession` rejection** — real, fixed in 459c33354. `finally` doesn't consume a rejection, so a failed kill closed the confirm alert and left the session running silently.
- **VoiceOver can't reach the reorder action** — stale: it points at `SessionRow.tsx`, which was deleted when the list moved to a native SwiftUI `List`. UIKit's reorder control is VoiceOver-operable out of the box (it exposes "Reorder", "drag" and "remove" actions in the a11y tree), so the concern is resolved by the swap rather than by a patch.

PR #6504, opened 2026-08-15
feat(desktop): configurable folder-link clicks with open-in-Finder binding

Kitenite:
## Summary
- **Folder links in terminal output are now settings-driven** (Settings → Links → Folder links), matching how file/URL link clicks are configured. Folder-specific actions: reveal in sidebar, open in editor, open in Finder. Defaults: plain = nothing, **⇧ Click = Open in Finder** (new — was unbound), ⌘ Click = Reveal in sidebar, ⌘⇧ Click = Open in editor.
- Cmd+clicking a folder that resolves **outside the worktree** (e.g. `~/some/dir`) opens it in Finder instead of dead-ending (the sidebar reveal silently no-ops for paths outside the worktree).
- Fixes the reveal landing behind the Changes tab: the v2 `WorkspaceSidebar` reads its active tab from `v2WorkspaceLocalState.sidebarState.activeTab`, not the `v2UserPreferences.rightSidebarTab` that `setRightSidebarTab` writes.
- Fixes the hover tooltip going stale when switching modifiers mid-hover (⌘ → ⌘⇧ kept showing "Reveal in sidebar" until re-hover).

## How It Works
- `folderLinks: FolderTierMap` on `v2UserPreferences` (healed for existing rows), with its own action enum (`reveal | external | finder`) since folders can't open in the file viewer. `useTerminalFolderPolicy()` resolves clicks; `FolderLinkTierMapper` renders the settings section (sibling of `LinkTierMapper` — different action typing).
- Sidebar folder **rows** (file tree / changes list) keep the hardcoded `folderIntentFor` rules — "reveal in sidebar" is meaningless there, so those surfaces aren't bound to the new map.
- `revealPath` keeps its contract: sidebar if inside the worktree, Finder otherwise. `external.openFolderInFinder` (main process) uses `shell.openPath` — opens the folder's contents, unlike `openInFinder`'s highlight-in-parent.
- **Tooltip staleness root cause**: xterm's Linkifier caches the last real `MouseEvent` and re-fires `leave` + `hover(cachedEvent)` on every terminal render tick. Each replay re-seeded modifier flags from the stale cached event, stomping the keydown listener's update within ~one frame. Modifier flags now live in a ref owned by the key listener; hover events only seed them when the event object is genuinely new (event-identity check against the cached object).

## Manual QA (all verified end-to-end over CDP against the real dev app, trusted input events)
- [x] Baseline reproduced pre-fix: Cmd+click on `~/Downloads` activated the link but opened nothing; tooltip stuck on old label when adding ⇧ mid-hover.
- [x] ⇧+click on an **in-worktree** folder link → Finder opens that folder (window list `[]` → `["apps"]`).
- [x] ⇧+click on an **out-of-worktree** folder link → Finder opens it (`[]` → `["Downloads"]`).
- [x] ⌘+click in-worktree folder → sidebar switches to Files tab and reveals the folder expanded+selected; no Finder window.
- [x] Settings → Links shows the new "Folder links" section; rebinding ⇧ Click → "Do nothing" through the real Select UI makes ⇧+click a no-op (and hides the hover label); rebinding back to "Open in Finder" restores it.
- [x] ⌘-hover on an out-of-worktree folder now reads "Open in Finder" (matches the containment fallback the click takes); in-worktree still reads "Reveal in sidebar" — `resolvedPath` threaded through `LinkHoverInfo`.
- [x] Tooltip tracks the full modifier cycle live while stationary over a link: ⌘ "Reveal in sidebar" → ⌘⇧ "Open in editor" → ⇧ "Open in Finder" → none (hidden), surviving xterm's cached-event replay ticks through ~1s dwells at each step.

## Testing
- `bunx tsc --noEmit`, `biome check` clean
- `bun test`: terminal suite + absolute-paths + v2UserPreferences schema heal tests — 342 pass

## Known Limitations
- Finder naming is macOS-centric; `shell.openPath` itself is cross-platform.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to reveal workspace files and folders directly in Finder.
  * Files outside the current workspace open in Finder, while workspace items navigate to the Files view before being revealed.
  * Added customizable folder-link actions for modifier-key combinations, including workspace reveal, external opening, and Finder.
  * Added searchable “Folder links” settings.

* **Bug Fixes**
  * Remote workspaces are prevented from using Finder reveal actions.
  * Improved cross-platform path handling and error reporting for invalid or inaccessible locations.


Kitenite:
Addressed in 2b3597a8e: `isWithinWorkspacePath` now collapses `.`/`..` segments (with tests) so traversal can't fake containment, and the stale `resolveHoverLabel` comment is updated for the settings-driven folder policy. The transient `worktreePath`-loading window keeps the previous reveal behavior by design — the query resolves in milliseconds from cache and there's no better classification available before it loads.

PR #6507, opened 2026-08-15
feat(api,trpc): dispatch automations from triggers

saddlepaddle:
Step C of the automation-triggers migration: the dispatcher's source of truth becomes `automation_triggers`. The legacy `automations` columns stay written, so this deploy reverts cleanly. They drop in the next step.

## What changed

**Dual-write** — `automation.create`, `update`, and `setEnabled` mirror the schedule onto the automation's `schedule` trigger. `update` and `setEnabled` moved into transactions so both rows move together or neither does.

**Read source** — `evaluate` selects due schedule triggers joined to `automations`, so a disabled automation still can't fire even if its trigger is enabled. Recurrence comes from `config`; that's what makes dropping the legacy columns possible next.

**Fallback + log** — if a trigger's config can't drive a schedule, the run falls back to the legacy columns and logs. This should be silent; it has to be silent before those columns drop.

**Lazy repair** — every tick, any automation missing a schedule trigger gets one built from its legacy columns. This makes the `0072` backfill an optimization rather than a correctness requirement, and self-heals rows written by old code during a deploy or rollback.

**`0073`** — partial unique index, one schedule trigger per automation. A duplicate would double-dispatch, and it's the conflict target the upsert and the repair both need. Production is currently 777 automations / 777 triggers, so it applies cleanly.

## Rehearsal

Full 74-migration chain applied to a clean Postgres 16, then behaviour asserted against real-shaped rows:

| | |
|---|---|
| Lazy repair creates one trigger per automation | pass |
| Re-running it inserts nothing | pass |
| A second `schedule` trigger is rejected | pass |
| The upsert Drizzle actually emits updates in place, no duplicate | pass |
| Two `github` triggers on one automation still allowed | pass |
| Bare `ON CONFLICT (automation_id)` **fails** without the predicate | pass |
| Disabled automation and disabled trigger both excluded | pass |
| Plan uses `automation_triggers_dispatcher_idx`, no seq scan | pass |

The sixth is the one worth keeping: it proves the `targetWhere` in the code is load-bearing rather than decorative, which is the exact trap that bit this feature before.

Typecheck, lint, and the full test suite pass.

## Ordering

`deploy-api` needs `deploy-database`, so `0073` always applies before the code that infers it. The gap in between runs old code against the new index, which doesn't touch it.

## After merge

Watch that this stays at 0:

```sql
SELECT count(*) FROM automations a WHERE NOT EXISTS (
  SELECT 1 FROM automation_triggers t
  WHERE t.automation_id = a.id AND t.kind = 'schedule');
```

and that the config-fallback log stays silent. Both are preconditions for dropping the legacy columns.

---
## Summary by cubic
Dispatch scheduler reads `schedule` triggers from `automation_triggers` instead of legacy fields on `automations`, while still writing legacy fields for a clean revert. Adds a fix to tolerate a jsonb null schedule config so runs fall back instead of leaving automations permanently due.

- Dual-write and transactions: `automation.create`, `update`, and `setEnabled` mirror schedules to a single `schedule` trigger via `syncScheduleTrigger`; `update`/`setEnabled` run in transactions to keep the trigger and automation in lockstep.
- Read path and fallback: `/api/automations/evaluate` selects due, enabled `schedule` triggers joined to enabled `automations`; it derives recurrence from trigger config and falls back to legacy columns (with a log) when config is invalid or jsonb null.
- Lazy repair: each tick inserts a missing `schedule` trigger from legacy columns; no manual backfill required.
- Guardrail: migration `0073_automation_schedule_trigger_unique` adds a partial unique index enforcing one `schedule` trigger per automation; upserts explicitly target this predicate to avoid duplicates.
- Post-merge checks: keep the count of automations without a `schedule` trigger at 0 and ensure the config-fallback log stays silent; both are prerequisites for dropping the legacy columns.

<sup>Written for commit 3f2133e74a341d6c2b6b33c095d8f7cf0f2276b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Automation schedules now stay synchronized when automations are created, updated, or enabled.
  * Existing automations using legacy schedule settings are automatically migrated to the current scheduling system.
  * Schedule timing, time zones, enabled status, and upcoming runs are maintained consistently.

* **Bug Fixes**
  * Prevented duplicate schedule triggers for the same automation.
  * Improved handling and reporting for invalid schedules and failed schedule updates.
  * Ensured due schedules are processed reliably with duplicate enqueue protection.

### DIFF
--- a/apps/api/src/app/api/automations/evaluate/route.ts
+++ b/apps/api/src/app/api/automations/evaluate/route.ts
@@ -31,8 +31,11 @@ function bucketToMinute(d: Date): Date {
 
 /** Null when the config can't drive a schedule, so the caller can fall back. */
 function scheduleFromConfig(
-	config: TriggerConfig,
+	config: TriggerConfig | null,
 ): { rrule: string; dtstart: Date; timezone: string } | null {
+	// The kind-matches-config CHECK passes for jsonb `null` (SQL NULL = NULL is
+	// not false), so the column can hold something the type says it can't.
+	if (config === null || typeof config !== "object") return null;
 	if (config.kind !== "schedule") return null;
 	if (!config.rrule || !config.timezone) return null;
 	const dtstart = new Date(config.dtstart);
--- a/apps/api/src/app/api/automations/evaluate/route.ts
+++ b/apps/api/src/app/api/automations/evaluate/route.ts
@@ -1,8 +1,12 @@
 import { dbWs } from "@superset/db/client";
-import { automations, automationTriggers } from "@superset/db/schema";
+import {
+	automations,
+	automationTriggers,
+	type TriggerConfig,
+} from "@superset/db/schema";
 import { nextOccurrenceAfter } from "@superset/shared/rrule";
 import { Client, Receiver } from "@upstash/qstash";
-import { and, eq, lte } from "drizzle-orm";
+import { and, eq, lte, sql } from "drizzle-orm";
 
 import { env } from "@/env";
 
@@ -25,6 +29,17 @@ function bucketToMinute(d: Date): Date {
 	return copy;
 }
 
+/** Null when the config can't drive a schedule, so the caller can fall back. */
+function scheduleFromConfig(
+	config: TriggerConfig,
+): { rrule: string; dtstart: Date; timezone: string } | null {
+	if (config.kind !== "schedule") return null;
+	if (!config.rrule || !config.timezone) return null;
+	const dtstart = new Date(config.dtstart);
+	if (Number.isNaN(dtstart.getTime())) return null;
+	return { rrule: config.rrule, dtstart, timezone: config.timezone };
+}
+
 export async function POST(request: Request): Promise<Response> {
 	const body = await request.text();
 	const signature = request.headers.get("upstash-signature");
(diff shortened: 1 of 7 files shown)


saddlepaddle:
Went through the three CodeRabbit findings.

**Taken — jsonb `null` config.** Real, and verified rather than assumed. The `kind`-matches-`config` CHECK compares `config->>'kind'` to `kind::text`; for jsonb `null` that's `NULL = 'schedule'` → NULL, and a CHECK only rejects on FALSE. Confirmed against Postgres 16 that both `'null'::jsonb` and `'{}'::jsonb` insert cleanly with `kind='schedule'`. The `{}` case already fell back correctly; jsonb `null` threw, which inside `Promise.allSettled` meant the automation never advanced and stayed due forever. Fixed in 3f2133e.

**Skipping — `to_char` dropping milliseconds.** Real difference, no consequence. Truncation can only move the recurrence anchor *earlier*, and `after` is always the previous `next_run_at`, so an occurrence can never repeat; it can only shift under a second. `scheduledFor` is bucketed to the minute anyway. It also matches what `0072` already wrote to all 777 production rows, so changing it now would introduce format skew for no behavioural gain.

**Skipping — claim-before-publish.** Correctly identified, but pre-existing: the dispatcher has always enqueued before advancing, and this PR preserves that ordering rather than introducing it. Duplicates are already absorbed twice over — QStash dedupes on `automationId_scheduledFor`, and the run insert dedupes on the same pair via the partial unique index. Moving to a transactional claim is a real improvement but it changes dispatch semantics for every automation and belongs in its own change, not in the read-source switch.

PR #6511, opened 2026-08-15
refactor(api,trpc,db): stop writing the dead schedule columns

saddlepaddle:
**Step D2.** `0074` drops `NOT NULL` on `automations.rrule`, `dtstart`, `timezone` and `next_run_at`; the code then stops writing them. The columns themselves drop in D3.

## Why the NOT NULL drop and the write removal ship together

The columns are `NOT NULL`, so the code physically cannot stop writing them while they're declared that way — and they can't be dropped while code still writes them. So the order is: relax the constraint, stop writing, then drop.

Both halves are safe in either deploy order. The migration runs first and old code keeps writing non-null values into a now-nullable column; then the new code stops writing. This is only safe because **D1 already removed every reader** — otherwise a rollback would expose rows with NULL schedules to code that reads them.

## What changed

- `create` / `update` / `setEnabled` no longer write the schedule columns, and build their responses from the trigger-sourced values, so the response shape is unchanged.
- `setPrompt` likewise — it returned `{...updated, scheduleText}` off the automations row, which no longer carries a schedule.
- `evaluate` advances only the trigger. The lazy repair and the config fallback are both gone, since both read the legacy columns.
- An unusable trigger config is now a thrown error rather than a silent fallback. There is nothing left to fall back to, and `Promise.allSettled` already routes it to the existing failure log.

## Rehearsed

Full 75-migration chain applied to a clean Postgres 16, then:

| | |
|---|---|
| All four columns nullable, `enabled`/`mcp_scope` untouched | pass |
| Insert with **no** schedule columns — the new create path | pass |
| Automation with all-NULL legacy columns still matches the dispatcher query | pass |
| D1's read projection still returns the full schedule from the trigger | pass |

That second one is the point of the migration: before `0074` that insert fails outright.

## Audit

`grep` across `apps/` and `packages/` for `automations.rrule`, `.dtstart`, `.timezone`, `.nextRunAt`, `.mcpScope` returns **nothing** outside the schema definition itself. That is D3's precondition.

Typecheck, lint and the full test suite pass.

## Note on rollback

Reverting this deploy is safe — the columns still exist and old code would resume writing them. Automations created while it was live keep NULL legacy columns, but nothing reads them, so that's inert. D3 is where it stops being reversible.

---
## Summary by cubic
Stop writing legacy schedule fields on `automations` so the schedule lives only on the `schedule` trigger. Evaluate now computes the next run before enqueueing and skips unusable triggers; previously it enqueued first and then failed to advance, causing repeated re-fires.

- `packages/db`: migration `0074` drops NOT NULL on `rrule`, `dtstart`, `timezone`, `next_run_at`; schema marks them nullable ahead of removal.
- `apps/api`: removed lazy repair and legacy fallbacks; plan next occurrence first, enqueue only triggers that can be advanced, and update only the trigger. Unusable trigger configs are logged and excluded; the response now includes `unusable`.
- `packages/trpc`: `create`, `update`, and `setEnabled` stop writing the dead columns and return schedule data from trigger-sourced values to keep the response shape unchanged. When resuming, recompute the next run from now and sync the trigger.

**Rollout**
- Apply `0074` and deploy; either order is safe. D1 already removed all readers; D3 will drop the columns.
- Ensure schedule triggers exist; without lazy repair, missing or unusable triggers won’t dispatch and will surface via logs/`unusable`.
- Rollback needs no action: columns still exist and old code resumes writing them.

<sup>Written for commit c1ea19f5a0c5c782bd25d2de1c4fbb59420df58a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6511?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automation scheduling by validating trigger configurations before advancing schedules.
  * Prevented invalid or incomplete schedules from using outdated fallback data.
  * Ensured resumed automations calculate and synchronize their next scheduled run correctly.

* **Improvements**
  * Schedule information is now consistently managed through automation triggers.
  * Legacy schedule fields may remain empty without affecting automation behavior.
  * Automation create, update, enable, and disable actions now return consistent schedule details.

### DIFF
--- a/apps/api/src/app/api/automations/evaluate/route.ts
+++ b/apps/api/src/app/api/automations/evaluate/route.ts
@@ -86,42 +86,70 @@ export async function POST(request: Request): Promise<Response> {
 			row.nextRunAt !== null,
 	);
 
-	if (due.length === 0) {
-		return Response.json({ enqueued: 0 });
+	// Work out the next occurrence before dispatching anything: a trigger we
+	// can't advance must not be enqueued, or it would fire on every tick forever
+	// while its next_run_at stayed put.
+	const planned: Array<{
+		automationId: string;
+		scheduledFor: Date;
+		next: Date | null;
+	}> = [];
+	const unusable: Array<{ automationId: string; reason: string }> = [];
+
+	for (const row of due) {
+		const schedule = scheduleFromConfig(row.config);
+		if (!schedule) {
+			unusable.push({
+				automationId: row.automationId,
+				reason: "schedule trigger config is unusable",
+			});
+			continue;
+		}
+		try {
+			planned.push({
+				automationId: row.automationId,
+				scheduledFor: bucketToMinute(row.nextRunAt),
+				next: nextOccurrenceAfter({ ...schedule, after: row.nextRunAt }),
+			});
+		} catch (error) {
+			unusable.push({ automationId: row.automationId, reason: String(error) });
+		}
+	}
+
+	// Should be empty. These automations are stalled until someone fixes the
+	// config, so they need to be loud rather than silently skipped.
+	if (unusable.length > 0) {
+		console.error(
+			"[automations/evaluate] unusable schedule triggers, not dispatched",
+			unusable,
+		);
+	}
+
+	if (planned.length === 0) {
+		return Response.json({ enqueued: 0, unusable: unusable.length });
 	}
 
 	await qstash.batchJSON(
-		due.map((row) => {
-			const scheduledFor = bucketToMinute(row.nextRunAt);
-			return {
-				url: `${env.NEXT_PUBLIC_API_URL}/api/automations/dispatch/${row.automationId}`,
-				body: {
-					automationId: row.automationId,
-					scheduledFor: scheduledFor.toISOString(),
-				},
-				deduplicationId: `${row.automationId}_${scheduledFor.getTime()}`,
-				retries: 2,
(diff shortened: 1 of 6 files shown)


saddlepaddle:
Took the CodeRabbit finding — it's correct, and the problem is one this PR introduces.

Before this change an unusable config fell back to the legacy columns, so the automation still advanced. Now it throws, and the throw happened *after* `batchJSON` had already dispatched. The trigger keeps its `next_run_at` and stays enabled, so it re-fires every tick with nothing to break the loop: QStash dedup absorbs the duplicate for its window, and the run dedupe index suppresses the duplicate row after that — so the automation goes quiet permanently while only a log line records it. That's a worse failure than the one the fallback was covering.

Fixed in c1ea19f by computing the next occurrence *before* dispatching. Anything we can't advance is never enqueued, is reported separately in the response as `unusable`, and is logged loudly rather than skipped silently. It should always be zero — production currently has 0 incomplete configs — but if it isn't, the automation stalls visibly instead of invisibly.

PR #6512, opened 2026-08-15
feat(mobile): group home by project and name the scope on screen

saddlepaddle:
## What & why

Opening the app never told you what you were looking at. The home list was scoped by
**organization, host, project and sort** at once, and only the organization appeared
anywhere on screen. Worse, two of the others resolved themselves:

- `useSelectedHost` falls back to *whichever machine happens to be online*, so a sleeping
  laptop silently promotes another host and the app opens onto different work.
- `defaultProjectId` was *the project of whichever workspace was updated last*, so an agent
  finishing overnight in another project moved your default there.

The filter icon carried no active state, and the only place a project was even implied was
the empty state: "No workspaces in this project yet."

**Projects are now sections instead of a filter.** Every project on the host is listed,
ordered by its liveliest workspace, collapsible with the state kept per host, with a count
so a shut section still says what's inside. The project filter and its `defaultProjectId`
guess are gone, along with the sheet row and route that set it. What remains of the scope —
**host with its presence dot, and sort** — sits in a chip row under the search field, each
chip opening the picker it names.

This is worth more than it sounds: on my account the host showed *one* project before.
It actually has seven — Boid Home, notion-clone, frontend-interview, evidence, presentations
and more were invisible behind a filter nobody set.

Date group headers give way to project headers; recency survives as the section ordering.

Design exploration: [options](https://claude.ai/code/artifact/bb1d49cd-cacc-49b8-9fcb-c87fc53136cb) ·
[interactive mock of what shipped](https://claude.ai/code/artifact/2c203c9f-e45c-41e7-a245-c3b1fb669bee)

### Two corrections made while building

- **Search was never cross-host.** `useHostWorkspaces(host)` only ever fetches the selected
  host's workspaces, so search escaped the *project* scope alone — the host predicate in the
  old code was a no-op. A cross-host "N more on <host>" tail I'd planned was therefore dead
  code and was removed rather than shipped. Search now reports how far it reached
  ("N results on this host").
- **`projectFilter` is deleted, not orphaned.** It was still read by `useNewChatTargets` as a
  fallback for the composer's default target; with it gone the composer falls back to the
  most recently updated workspace's project, which is what a null filter already did.

## How I tested it

On the iOS simulator against a live host and real projects:

- Sections render with counts; collapse toggles and **survives a cold restart** (terminate +
  relaunch), keyed per host
- Both chips open their pickers (host picker listed five machines; sort picker toggles)
- Host chip shows the live presence dot
- Search returns "2 results on this host", grouped by project, auto-expanding collapsed
  sections; clearing restores the full list
- A project with zero workspaces renders correctly (`MASTRA-SUPERSET 0`)
- Caret renders in both states — the first attempt used a rotate transform, which doesn't
  reach the underlying SVG and silently rendered nothing

Three rounds of layout review against the real list: headers initially out-weighed their own
rows and duplicated the `+`; the fix re-weights rows as children, drops my header `+` in
favour of the pre-existing row one, and puts the project mark in the same 24pt icon box the
rows use so both columns share a grid.

Not exercised: the host-offline empty state.

- **Density**: projects holding only `main` still cost a header plus a row. Default-collapsing
  dormant projects would open the list on what's alive.
- **`useHostTerminals` gates rows on `host.isOnline`**, so a relay presence flap empties the
  tab strip while the terminal keeps streaming. Pre-existing, affects home too, worth its own PR.

## Checklist

---
## Summary by cubic
Groups Home by project and makes the active scope visible. Previously, host fell back silently and a guessed project filter hid most work; now sections replace the project filter, and chips name host and sort.

- Projects become sections: headers show avatar/name/count, collapse per host via `collapsedProjectsStore` (keyed `machineId:projectId`) and persist without flash on cold start.
- Unknown/missing projects: workspaces whose project isn’t reported fall under a “No project” section; avoids disappearing during the projects-not-yet-loaded beat.
- Scope bar: adds chips for host (with presence dot) and sort under search; chips open existing host/sort pickers.
- List/order: pinned-first then by recent activity; replaces date headers with project headers.
- Search: stays host-scoped and announces reach (“N results on this host”); empty copy is “No projects on this host yet.”
- Removed project filter: drops `/filter/project`, the Filter screen’s project row, `ProjectFilterScreen`, and `projectFilter` from `useWorkspacesFilterStore`.
- `useNewChatTargets`: stops reading the deleted `projectFilter`; defaults to the most recently updated workspace’s project.
- UI polish: aligns project/row icons on one grid; lightens header and row weights; hides duplicate branch when it matches the workspace name.

<sup>Written for commit 31c6a8a162afd6c2aecc5f036961b768222b47a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6512?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Home workspaces are now grouped by project with collapsible sections and workspace counts.
  * Added host and sort controls, host-wide search, project-based sorting, and orphan workspace handling.
  * Collapse and expand states are saved between sessions.
* **Improvements**
  * Updated workspace rows with clearer typography and less redundant branch information.
  * New chat targets prioritize recently updated workspaces.
* **Changes**
  * Removed project selection from filtering screens and controls.


saddlepaddle:
All three review findings fixed in `31c6a8a16`:

- **Workspaces with an unreported `projectId` vanished** — confirmed and worse than described: `useHostProjects` and `useHostWorkspaces` are separate queries, so the same path swallowed the entire list during the beat where workspaces had resolved and projects hadn't. Unknown ids now group with `__none`.
- **Collapsed sections flashed open** before AsyncStorage answered — the store now carries `hasHydrated`, matching the pattern `workspacesFilterStore` already uses next door for the same reason.
- **`Chip` extracted** to its own file per the one-component-per-file rule.

Re-verified on device after the fixes.

PR #6517, opened 2026-08-15
feat(api,db): expire raw webhook payloads after 14 days

saddlepaddle:
`ingest.webhook_events` is **523 GB across 93.5M rows**, growing ~7 GB/day with nothing pruning it. Only 72 GB is heap and 31 GB indexes — **~420 GB is TOASTed `payload`**.

## The shape of the fix

Nothing in the codebase ever selects `payload` back. `grep` finds only INSERT and UPDATE against this table — there is no read path at all. Its only value is inspecting a recent delivery by hand.

The *row*, though, is load-bearing: `UNIQUE (provider, event_id)` is the idempotency ledger both webhook routes dedupe against via `ON CONFLICT`, and the same row doubles as the retry queue (`failed → pending`).

So they get different lifetimes. **The row lives forever; the payload expires.** Idempotency never weakens at any age — strictly safer than deleting rows, and it reclaims the part that's actually large.

Rows still `failed` or `pending` keep their payload regardless of age, since they may yet be retried.

## Why not partitioning

That was the earlier plan and it doesn't work. Verified against Postgres 16:

```
ERROR: unique constraint on partitioned table must include all partitioning columns
DETAIL: UNIQUE constraint on table "we" lacks column "received_at"
```

Partitioning by `received_at` forces the partition key into every unique index, so `UNIQUE (provider, event_id)` becomes impossible and `ON CONFLICT (provider, event_id)` has nothing to infer. It would silently break webhook dedupe in both routes.

## Also not shipped: a partial index

`CREATE INDEX ... WHERE payload IS NOT NULL` is the natural index for the pruner, but building it non-concurrently on 93M rows takes a lock that **blocks all webhook ingestion** for the duration, and Drizzle wraps migrations in a transaction so `CONCURRENTLY` isn't available.

Instead the pruner carries a high-water mark **within a run**. Pruned rows form a contiguous oldest-first prefix, so without a cursor every batch re-skips everything already cleared; with one, that cost is paid once per run and the existing `received_at` index does the work. The index is worth revisiting once the backlog is drained, when it would be small and cheap to build.

## Rehearsed

60k realistic rows on Postgres 16 — old+terminal, old+failed/pending, and recent:

| | |
|---|---|
| Pruned (old + terminal) | 40,000 |
| Kept — `failed`/`pending`, still retryable | 10,000 |
| Kept — inside the window | 10,000 |
| **Wrongly pruned live rows** | **0** |
| **Wrongly pruned recent rows** | **0** |
| Ledger rows vs distinct `(provider, event_id)` | 60,000 / 60,000 |
| **Redelivery of a pruned old event still dedupes** | **pass** |

## Honest limits

- Nulling frees TOAST chunks for **reuse**, it does not shrink the file. Reclaiming to disk needs `VACUUM FULL` (locking) or `pg_repack`. With ~1.3M new events/day the freed space gets reused, so the practical effect is that the table becomes **bounded** — roughly 14 days of payloads instead of growing forever — rather than instantly smaller.
- **The endpoint is not scheduled by this PR.** `evaluate` is triggered by a QStash schedule registered outside the repo; this needs the same treatment before it runs on its own.
- The ~54M-row backlog drains over many runs. Worth a controlled first pass with monitoring rather than waiting for the cron.

Typecheck, lint and the full test suite pass.

---
## Summary by cubic
Expire raw webhook payloads after 14 days to cap `ingest.webhook_events` growth without weakening dedupe. Previously payloads were kept forever; now only terminal rows (`processed`, `skipped`) past 14 days are nulled, while `pending`/`failed` keep theirs.

- DB: `payload` is now nullable; NULL means pruned, not empty. `UNIQUE (provider, event_id)` is unchanged to preserve idempotency.
- API: adds POST `/api/ingest/prune-payloads`; verifies `@upstash/qstash` signature, prunes oldest-first in 5k-row batches with a 20s budget and a 50k row-per-run cap using a per-run cursor.
- Behavior: endpoint returns `pruned`, `batches`, `hitLimit`; TOAST space is reused, effectively retaining ~14 days of payloads.

**Rollout**
- Apply the migration and deploy API/DB.
- Configure a QStash schedule; set `QSTASH_CURRENT_SIGNING_KEY`, `QSTASH_NEXT_SIGNING_KEY`, and `NEXT_PUBLIC_API_URL`.
- Run an initial controlled pass and monitor `pruned`, `batches`, and `hitLimit`. Consider `VACUUM FULL` or `pg_repack` later if on-disk reclamation is needed.

<sup>Written for commit 2e7f470ad5d63ace8602cd1193bfd6c2ed2bbf65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6517?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automated cleanup of webhook payload data from processed or skipped events older than 14 days.
  * Cleanup runs in batches and reports pruning metrics.

* **Improvements**
  * Webhook payload data can now be removed after processing while retaining event records.
  * Updated data handling to support events without stored payloads.

### DIFF
--- a/apps/api/src/app/api/ingest/prune-payloads/route.ts
+++ b/apps/api/src/app/api/ingest/prune-payloads/route.ts
@@ -21,6 +21,15 @@ const RETAIN_DAYS = 14;
 /** Small enough that one statement is a short transaction on a 90M-row table. */
 const BATCH_SIZE = 5_000;
 
+/**
+ * Ceiling per run. Every prune is an UPDATE, so it leaves a dead tuple behind;
+ * this is what stops a backlog drain outrunning autovacuum and bloating the
+ * heap. Paced against the cron interval it also sets the drain rate — at one run
+ * every 5 minutes this is ~14M rows/day, so the backlog clears in days without
+ * the database ever being under a write spike.
+ */
+const MAX_ROWS_PER_RUN = 50_000;
+
 /** Well inside the function timeout, so a run ends by choice rather than by kill. */
 const TIME_BUDGET_MS = 20_000;
 
@@ -51,8 +60,8 @@ export async function POST(request: Request): Promise<Response> {
 	// One statement per batch, each its own transaction, so WAL stays bounded and
 	// a long backlog is chewed through across runs rather than in one held-open
 	// transaction.
-	while (Date.now() - startedAt < TIME_BUDGET_MS) {
-		const rows = await dbWs.execute(sql`
+	while (Date.now() - startedAt < TIME_BUDGET_MS && pruned < MAX_ROWS_PER_RUN) {
+		const result = await dbWs.execute(sql`
 			WITH batch AS (
 				SELECT id, received_at
 				FROM ingest.webhook_events
@@ -70,7 +79,7 @@ export async function POST(request: Request): Promise<Response> {
 			RETURNING batch.received_at
 		`);
 
-		const returned = rows.rows as Array<{ received_at: string }>;
+		const returned = result.rows as Array<{ received_at: string }>;
 		pruned += returned.length;
 		batches++;
 		if (returned.length < BATCH_SIZE) break;
@@ -82,6 +91,10 @@ export async function POST(request: Request): Promise<Response> {
 	return Response.json({
 		pruned,
 		batches,
+		// Whether the run stopped on a limit rather than running out of work —
+		// the cheap version of "is there still a backlog?". An exact count would
+		// mean scanning tens of millions of rows on every run.
+		hitLimit: pruned >= MAX_ROWS_PER_RUN,
 		retainDays: RETAIN_DAYS,
 		elapsedMs: Date.now() - startedAt,
 	});
--- a/apps/api/src/app/api/ingest/prune-payloads/route.ts
+++ b/apps/api/src/app/api/ingest/prune-payloads/route.ts
@@ -1,0 +1,88 @@
+import { dbWs } from "@superset/db/client";
+import { Receiver } from "@upstash/qstash";
+import { sql } from "drizzle-orm";
+
+import { env } from "@/env";
+
+export const dynamic = "force-dynamic";
+
+const receiver = new Receiver({
+	currentSigningKey: env.QSTASH_CURRENT_SIGNING_KEY,
+	nextSigningKey: env.QSTASH_NEXT_SIGNING_KEY,
+});
+
+/**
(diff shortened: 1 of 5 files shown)


saddlepaddle:
Three findings, all fair. Two are answered by sequencing; one I'm declining with a reason.

**1. Index for the pruning predicate — correct, and sharper than I wrote it.** I flagged the index tradeoff in the description but understated this: because `cursor` resets per run, each run re-skips everything previously pruned. That's not just a post-drain problem, it degrades *during* the drain too — by the end, a run would skip ~75M rows before finding work.

The index is the right fix; the constraint is only *when* it can be built. `WHERE payload IS NOT NULL` covers all 93M rows today, so building it now is a long write-blocking lock (and Drizzle wraps migrations in a transaction, so `CONCURRENTLY` isn't available). After the backlog drains it covers ~18M and builds cheaply.

So the corrected order is: merge this → drain the backlog with a **controlled one-pass script** that holds the cursor in memory, so there is no cross-run re-skip at all → add the partial index, now cheap → *then* schedule the cron. **The cron should not be registered before the index exists** — that's the case this finding describes, and I've added it to the follow-ups rather than leaving it implied.

**2. Preserve payloads for retryable rows — doesn't apply to this design, and a CHECK would be actively risky.** Nothing reads the stored payload, including on retry: `ON CONFLICT DO UPDATE` never sets `payload`, and a retry is the provider redelivering with a fresh payload in the request body. The pruner also already excludes `pending`/`failed` regardless of age.

I considered `CHECK (payload IS NOT NULL OR status IN ('processed','skipped'))`. Declining it: the constraint would have to be `NOT VALID` to avoid a 93M-row validation scan, and more importantly a violation would make the webhook endpoint **500 on ingestion** — real breakage risk to guard a path that doesn't exist.

**3. Row cap / stability** — addressed in ab9158969: `MAX_ROWS_PER_RUN` bounds each run so a drain can't outrun autovacuum, since every prune is an `UPDATE` and leaves a dead tuple.

PR #6527, opened 2026-08-16
fix(mobile): submit terminal messages through terminal.send, add a newline button

saddlepaddle:
## Problem

Sending a message to an agent from the mobile terminal composer didn't submit it. The text arrived in Claude's input box fully typed, with a newline appended, and just sat there.

`WorkspaceScreen.handleSubmit` wrote the message and its Enter into the attached stream as a **single write**:

```ts
const framed = text.includes("\n") ? `\x1b[200~${text}\x1b[201~\r` : `${text}\r`;
terminalRef.current?.sendInput(framed);
```

TUI agents treat bytes that arrive together as one paste burst, so the trailing `\r` was absorbed into the draft as a newline rather than read as a keypress. Host-service already fixed this class for follow-up sends in #6284. The `text.includes("\n")` heuristic was wrong in both directions too — it can't know whether the running program actually has bracketed-paste mode on.

## Fix

Submits go through the host's `terminal.send` — the path FinishReviewSheet, the CLI and the SDK already use. The host writes the Enter as its own delayed write, frames the text as a bracketed paste only when the session's mode tracker says the program has that mode on, and serialises concurrent sends. Quick keys still go over the stream as raw bytes; they're control bytes, not submissions.

The composer now clears only once the send resolves, per `GlassComposer`'s `onSubmit` contract — a failed send keeps the draft and alerts instead of eating it.

## Newline button

expo-ui's SwiftUI `TextField` exposes nothing for the soft keyboard's return key, so there was no way to get a line break into a draft at all. Adds a `return` button to the expanded toolbar next to `+`, in the shared `GlassComposer`, so the home composer gets multi-line prompts too.

It appends rather than inserting at the caret, which is deliberate: **the field never reports where its cursor is.** Both `onSelectionChange` and a `selection` `ObservableState` read back `{0,0}` once you've typed (confirmed on device — the first cut put every line break at position 0). Writing the caret does work, so it appends and then places the caret after the insert.

Also labels `+`, the newline button and send for VoiceOver. Send and the terminal's quick-key up arrow were both a bare `arrow.up` symbol — indistinguishable to assistive tech, and to a UI test.

## Also: home project headers

Separate commit, unrelated to the above but shipping in the same build. The headers were `text-[15px] uppercase tracking-wider` — none of which is in the design system, and at 15px identical in size to the workspace row title, so the caps carried the entire parent/child signal. Now the `large` text variant, so hierarchy comes off the type scale.

## Verification

**In the app, on the simulator**, against a real Claude Code session: composed a two-line message using the ↵ button, tapped send → submitted as one prompt, agent answered `4`, composer cleared. Screenshots of the two-line draft and the cleared state were reviewed.

**Against the host directly**, via `superset terminals send` (same `terminal.send` procedure):

| sent | result |
|---|---|
| `What is 7 times 6? Reply with only the number.` | submitted, `⏺ 42`, input box empty |
| two-line message | submitted as one prompt with the break preserved, `⏺ 4` |

`bun run lint` clean; `tsc --noEmit` clean for `apps/mobile` (remaining errors are pre-existing, in port-scanner/pty-daemon/workspace-fs). Host behaviour also has an existing regression test — `terminal.send-snapshot.node-test.ts`, *"submit Enter is a separate write, delayed past the paste burst"*.

## Note

Depends on the host running host-service ≥ 1.22.0 (where #6284 landed, 2026-08-09). Older hosts bundle the Enter server-side and will still swallow it.

## Summary by CodeRabbit

* **Improvements**
  * Terminal messages are now submitted through the active workspace terminal more reliably.
  * Messages remain in the composer until delivery succeeds.
  * Attachment uploads complete before sending, with uploaded files included in the message.
  * Sending status now covers both uploads and message delivery.
  * Expanded composers support inserting new lines and appending text to drafts.
  * Added accessibility labels to composer controls.
* **Bug Fixes**
  * Added clearer alerts for unavailable terminals, upload failures, and submission errors.

### DIFF
--- a/apps/mobile/screens/(authenticated)/(home)/home/components/ProjectSectionHeader/ProjectSectionHeader.tsx
+++ b/apps/mobile/screens/(authenticated)/(home)/home/components/ProjectSectionHeader/ProjectSectionHeader.tsx
@@ -13,9 +13,10 @@ interface ProjectSectionHeaderProps {
 }
 
 /**
- * A quiet band, not a title: the workspaces under it are what you came to tap,
- * so the header labels them rather than competing with them. The count is what
- * keeps a collapsed section informative.
+ * Reads as the parent of the rows under it through the type scale alone —
+ * `large` against their 15px — rather than through caps and letterspacing,
+ * which shout at a list you're meant to scan. The count is what keeps a
+ * collapsed section informative.
  */
 export function ProjectSectionHeader({
 	name,
@@ -39,10 +40,7 @@ export function ProjectSectionHeader({
 			<View className="size-6 items-center justify-center">
 				<ProjectAvatar name={name} iconUrl={iconUrl} size={20} />
 			</View>
-			<Text
-				className="text-foreground font-semibold text-[15px] uppercase tracking-wider"
-				numberOfLines={1}
-			>
+			<Text variant="large" numberOfLines={1}>
 				{name}
 			</Text>
 			<Text className="text-muted-foreground font-mono text-[13px]">
--- a/apps/mobile/screens/(authenticated)/components/GlassComposer/GlassComposer.tsx
+++ b/apps/mobile/screens/(authenticated)/components/GlassComposer/GlassComposer.tsx
@@ -12,6 +12,7 @@ import {
 } from "@expo/ui/swift-ui";
 import {
 	Animation,
+	accessibilityLabel,
 	animation,
 	aspectRatio,
 	buttonBorderShape,
@@ -146,7 +147,6 @@ export const GlassComposer = forwardRef<
 	// The draft lives in a ref, never state — otherwise each keystroke
 	// re-renders the SwiftUI Host.
 	const draftRef = useRef("");
-	const selectionRef = useRef({ start: 0, end: 0 });
 	const [hasText, setHasText] = useState(false);
 
 	const writeDraft = (text: string) => {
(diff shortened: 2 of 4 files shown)


saddlepaddle:
Went through the CodeRabbit findings. Dispositions:

**1. Disable quick keys while `terminal.send` is in flight** — *skipping, deliberately.* The race is real and correctly described: quick keys write raw bytes to the stream, so one can land inside the host's ~500ms text→Enter gap. But the quick key most likely to be pressed right after a send is `esc`, which is how you interrupt a running agent — disabling it while a send is in flight would deadbolt the interrupt key exactly when a send is slow, which is worse than the race it prevents. The window is short, requires a deliberate tap inside it, and the damage is bounded (an Esc-then-Enter submits an empty prompt). The real fix is routing quick keys through the host too, which is a separate change.

**2. Gate on host-service ≥ 1.22.0** — *skipping.* There's no host-version gating mechanism in the mobile app to reuse (CodeRabbit's own search found none), so this means inventing one, which is well outside this fix. The dependency is documented in the PR description instead. Hosts ship with desktop and auto-update; 1.22.0 is the current version.

**3. Guard `setSelection` for iOS 16.4–17** — *skipping, no defect.* `setSelection` isn't gated natively: `TextFieldView.setSelection` just writes `props.selection.value`, and below iOS 18 the view renders `StatefulTextField`, which never reads it. So it's a no-op, not a crash — and SwiftUI already leaves the caret at the end after a programmatic `setText`, which is exactly where this wants it. The newline still gets appended on those versions. (Moot in practice too: the pill is built on `glassEffect`, iOS 26.)

PR #6531, opened 2026-08-16
feat(trpc): validate trigger configs and expose trigger CRUD

saddlepaddle:
The trigger tables have existed since the cutover with nothing able to write a non-schedule row. This is the unblocker: config validation plus `automation.triggers.{list,create,update,delete}`.

No migration.

## Validation

One zod schema per trigger kind, combined as a discriminated union on `kind`, mirroring the TS types in `packages/db/src/schema/types.ts`.

**`create` derives the `kind` column from `config.kind`** rather than accepting it. The database has a `CHECK (config->>'kind' = kind::text)`, and deriving means a caller can never be the one to violate it — the two can't disagree because only one is an input.

**Scope stays the tagged three-state union** (`null` / `{mode:"any"}` / `{mode:"list", ids}`). Verified the reason it exists: a GitHub label literally named `"any"` parses correctly inside a list, which a bare `string[] | "any"` encoding could not express without reserving a magic value in a namespace users control.

**`includeForks` is typed as the literal `false`.** Fork payloads carry attacker-controlled content into a checkout an agent then runs in, so enabling it should be a deliberate schema change with a threat model attached, not a boolean someone can pass.

## Schedule triggers are read-only through this router

- `create` rejects `kind: "schedule"` — an automation's schedule is edited via `automation.update`, which keeps the trigger in sync.
- `update` rejects a config change on a schedule trigger, and rejects changing any trigger's kind.
- `delete` rejects schedule triggers outright: the dispatcher **inner-joins** the schedule trigger, so removing one wouldn't just stop the automation firing, it would make it disappear from every listing.

That last one is the failure mode worth calling out — it's silent, and it looks like data loss from the UI.

## Verified

17 validation cases run against the real schemas:

| | |
|---|---|
| github minimal / webhook / linear / schedule configs | pass |
| scope `null`, `{mode:"any"}`, `{mode:"list"}` | pass |
| **a label literally named `"any"` inside a list** | pass |
| bare `string[]` scope (the shape we rejected) | rejected |
| empty scope list, empty events | rejected |
| `includeForks: true` | rejected |
| actor `anyone` / `org_members` / id list | pass |
| bogus actor, unknown kind, bad `dtstart` | rejected |
| `create` derives kind from config, `enabled` defaults true | pass |

Ownership resolves through the automation via `getAutomationForUser`, since triggers have no owner of their own; org scoping is checked on the trigger row too, so a cross-org id 404s rather than reaching the composite FK and surfacing as a 500.

## Next

Webhook ingestion — the acceptance/processing split, adapter registry, and key generation — is what makes these fire. Then the GitHub connector.

---
## Summary by cubic
Adds trigger config validation and exposes `automation.triggers.{list,create,update,delete}` so non‑schedule triggers can be managed. Previously, only schedule rows existed and no endpoint could create or modify other kinds.

- Validate trigger configs with per‑kind zod schemas (discriminated by `kind`). `create` derives `kind` from `config.kind`; compute `nextRunAt` only for schedule configs.
- Keep schedule triggers read‑only in this router: reject `create` with `kind: "schedule"`, reject config updates on schedules, and reject deleting schedules.
- Prevent kind changes on `update`. `list` optionally filters by `automationId` and orders by creation time.
- Enforce safe shapes: tagged three‑state `scope` (`null` / `{mode:"any"}` / `{mode:"list"}`), strict `actor` union, and `includeForks` fixed to `false`.

**Review notes**
- No migration; the `config.kind = kind` CHECK already exists.
- Access control: require active org membership; resolve ownership via the automation; cross‑org ids return NOT_FOUND.

<sup>Written for commit 083f74724100803fdcaefc6487390f565737eee4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6531?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## New Features
- Added automation trigger management with support for listing, creating, updating, and deleting triggers.
- Added trigger configurations for schedules, webhooks, GitHub, Slack, Linear, and Sentry.
- Added validation for trigger-specific settings, filters, actor restrictions, event limits, and disabled forks.
- Schedule triggers now calculate their next run time automatically.
- Added access controls to ensure only authorized automation owners can manage triggers.

### DIFF
--- a/packages/trpc/src/router/automation/automation.ts
+++ b/packages/trpc/src/router/automation/automation.ts
@@ -35,6 +35,7 @@ import {
 	setAutomationPromptSchema,
 	updateAutomationSchema,
 } from "./schema";
+import { automationTriggersRouter } from "./triggers";
 import { automationVersionsRouter } from "./versions";
 
 function escapeLikePattern(value: string): string {
@@ -114,6 +115,7 @@ async function verifyWorkspaceInOrg(
 
 export const automationRouter = {
 	versions: automationVersionsRouter,
+	triggers: automationTriggersRouter,
 
 	/**
 	 * List automations scoped to the caller's active organization. The
--- a/packages/trpc/src/router/automation/triggerSchema.ts
+++ b/packages/trpc/src/router/automation/triggerSchema.ts
@@ -1,0 +1,112 @@
+import { automationTriggerKindValues } from "@superset/db/schema";
+import { z } from "zod";
+
+/**
+ * Three states, deliberately tagged rather than inferred from the JSON shape:
+ * `null` matches nothing, `{mode:"any"}` matches everything, and a list matches
+ * those ids. A bare string array can't express "any" without reserving a magic
+ * value, and every id space here is user-controlled — a GitHub label really can
+ * be named "any" — so the tag is what keeps that legal.
+ */
+const triggerScope = z
+	.union([
+		z.null(),
+		z.object({ mode: z.literal("any") }),
+		z.object({
+			mode: z.literal("list"),
+			ids: z.array(z.string().min(1)).min(1).max(200),
+		}),
+	])
+	.describe("null = nothing, {mode:'any'} = everything, or an explicit list");
+
+const triggerActor = z
+	.union([
+		z.literal("anyone"),
+		z.literal("org_members"),
+		z.object({ ids: z.array(z.string().min(1)).min(1).max(200) }),
+	])
+	.describe("Who may cause this trigger to fire");
+
+/** Non-empty so a trigger can't be saved in a state that matches nothing. */
+const events = z.array(z.string().min(1).max(100)).min(1).max(50);
+
+export const scheduleTriggerConfigSchema = z.object({
+	kind: z.literal("schedule"),
+	rrule: z.string().min(1).max(500),
+	dtstart: z.string().datetime(),
+	timezone: z.string().min(1),
+});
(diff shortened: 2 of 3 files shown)


saddlepaddle:
Holding as draft — the API shape is wrong.

Triggers are saved as a **set with the automation**, not created and edited individually. The editor holds the whole automation (name, triggers, instructions, tools) and Save commits it, so the surface should be a replace-the-set on `automation.update`, not `triggers.create/update/delete`.

The per-kind config schemas here are still the right foundation and will carry over. What changes is the surface around them, plus two things this doesn't handle:

- **Validation has to run client-side before submit**, with the same checks reused server-side. That means the schemas move somewhere both can import, and they need to represent an *incomplete* trigger — the editor has to hold "Draft opened in [Select Repos]" with nothing selected yet and block Save, rather than refuse to represent it.
- **The GitHub config is too flat.** "Any Comment by [Anyone] on a PR by [Anyone] in [Select Repos]" carries two independent actor scopes — comment author and PR author — where this schema has one `actor`.

PR #6526, opened 2026-08-16
feat: plugin system core — manifest SDK, host runtime, UI slots, CLI

Kitenite:
**Links**
- Research + architecture: `plans/20260815-plugin-system-surfaces.md` (§5 = implementation status)
- Authoring guide: `packages/plugin-sdk/README.md`

## Summary

First working slice of the Superset plugin system — bb's execution model on our process layout:

1. **`@superset/plugin-sdk`** — `superset-plugin.json` manifest schema + typed backend/UI contracts
2. **Host-service plugin runtime** — full-trust in-process backends with containment, per-plugin KV, event subscriptions, lifecycle broadcasts
3. **Desktop UI slots** — plugin React components in right-sidebar tabs + command-palette commands, loaded as blob ES modules sharing the host React
4. **`superset plugin` CLI** — 10 commands: scaffold, build, link/install (path + git), lifecycle, invoke

**Also includes:** preload `ipcRenderer` channel allowlist (security precondition), two example plugins (`hello-superset`, `agent-board`), and the research/architecture doc.

---

## Part 1: Plugin SDK

A plugin = a directory with a manifest plus optional `server` (backend) and `app` (UI bundle) entries. The manifest is fully validated (namespaced ids, semver, entry paths can't escape the plugin root, commands must reference declared tabs) and readable without executing plugin code — install previews render from it.

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Trust model | Full trust + containment (bb), not sandboxing (Zed/Figma) | Superset already runs full-trust agent code and setup scripts; sandbox purity taxes every author while protecting a thin slice. Containment (time-boxes, error boundaries, atomic reload) bounds bugs from day one |
| Execution home | In-process under **host-service** | Same trust class as setup scripts, and plugins run where the workspaces are — remote hosts work for free |
| UI model | Real React in versioned host slots | Native fidelity, no webview tax; per-slot error boundaries; `ctx` prop contract is additive-only |
| React sharing | esbuild aliases `react`/`react/jsx-runtime` to `.cjs` shims reading a host global | One React instance so hooks work. Shims must be `.cjs`: under `"type": "module"` esbuild parses `.js` shims as ESM and compiles named React imports to `undefined` |

## Part 2: Host-service runtime

- `plugins` + `plugin_kv` tables (migration `0024`, additive — new tables only)
- Install (path/git with commit pinning) copies into `~/.superset/plugins/<id>` and re-validates the manifest from the managed copy; `link` runs a directory in place for the dev loop
[...]

PR #6532, opened 2026-08-16
fix(host-service): restore workspaces.cloud_synced_at so older builds can still read workspaces

saddlepaddle:
## Problem

On an affected machine the workspace list and the file tree never load — not once, not intermittently, on every launch, permanently — because every query that reads the `workspaces` table throws `SqliteError: no such column: "cloud_synced_at"`. HOST-SERVICE-35 is the list query (`workspace.list`), HOST-SERVICE-34 is workspace-root resolution behind the file browser (`filesystem.statPath` -> `resolveWorkspaceRoot`). One root cause, two entry points into the same broken read. The only escape today is upgrading the app.

**Who is affected:** anyone whose `host.db` has been opened by a 1.22.0-or-newer host-service and is then opened by a host-service from the 1.21.0 line. That is machines running two channels side by side, machines that roll back a release, and fleets whose hosts update independently. The reporting machine is one such — a newer build, then a 1.21.0 canary.

## Root cause

Migration `0023_gorgeous_pixie.sql` drops `workspaces.cloud_synced_at`. Verified against the release tags:

```
$ git ls-tree desktop-v1.21.0 --name-only packages/host-service/drizzle/ | tail -3
packages/host-service/drizzle/0020_overrated_dreaming_celestial.sql
packages/host-service/drizzle/0021_heal_orphaned_terminal_session_workspace_refs.sql
packages/host-service/drizzle/meta

$ git ls-tree desktop-v1.22.0 --name-only packages/host-service/drizzle/ | tail -3
packages/host-service/drizzle/0022_kind_roxanne_simpson.sql
packages/host-service/drizzle/0023_gorgeous_pixie.sql
packages/host-service/drizzle/meta

$ git show desktop-v1.21.0:packages/host-service/src/db/schema.ts | grep -n cloudSyncedAt
243:	cloudSyncedAt: integer("cloud_synced_at"),
```

So 1.21.0's schema still declares the column, and 1.21.0 never carried the migration that removes it. Drizzle's `select()` emits an explicit column list rather than `SELECT *`, so a missing column takes out the whole statement instead of degrading to a null field. `host.db` is shared across builds on a machine, so once any newer host-service applies 0023 the column is gone for good and every older reader is broken from then on.

This is a schema downgrade, not a defect in 1.21.0. The repo has already made this call once in the other database: #6450 restored a dropped `local-db` column the same way, as a one-line additive migration.

## Fix

One additive migration, byte-identical to the `ALTER` in `0008_workspace_full_fields.sql` that originally created the column:

```sql
ALTER TABLE `workspaces` ADD `cloud_synced_at` integer;
```

Nullable, no default, no backfill, no data. Deliberately **not** done:

- `schema.ts` is untouched — current code stays unaware of the column. It exists purely so an older reader's explicit column list resolves. Because the snapshot matches `schema.ts` (which does not declare it), a future `drizzle-kit generate` will not re-emit the drop.
- `workspace_cloud_deletes`, also dropped by 0023, is **not** restored. desktop-v1.21.0 declares that table but never queries it anywhere in `packages/host-service/src`, so its absence breaks nothing and restoring it would be dead weight.
- 0023 is not edited or reverted. It has already run on users' machines.
- No runtime guard, try/catch, schema-drift detection, or new Sentry capture.

Hand-authored, including the journal entry and snapshot: `.agents/skills/db-migrations/SKILL.md` covers `packages/db` (Postgres on Neon) only, and `drizzle-kit generate` cannot produce this migration anyway since it must not be reflected in `schema.ts`. It follows the style of the existing hand-authored host-service migrations (0018, 0021). The journal `when` is current wall-clock time, so it is newer than every applied entry and cannot be silently skipped by drizzle's `max(created_at)` gate.

**Index 0025, not 0024:** open PR #6526 already claims `0024_broken_tattoo.sql` and `meta/0024_snapshot.json`. Taking 0024 would collide on the snapshot path and produce two journal entries with `idx: 24`.

### Which release carries it, and what happens to a machine that is broken right now

The migration ships in the next host-service release cut from `main` — the 1.23.0 line, and any canary built from `main` after this merges. It reaches a broken machine indirectly: the fix is not a change to the failing 1.21.0 binary (which we cannot change), it is a repair of the database that binary reads. As soon as **any** build carrying 0025 opens that `host.db`, the column comes back, and the 1.21.0-line host-service on that same machine starts working again on its next launch — no reinstall, no data loss, no manual step. Machines that never ran a ≥1.22.0 build never applied 0023 and were never affected. Going forward, a machine that upgrades past 0023 and later rolls back to 1.21.0 no longer breaks at all.

## Verification

**Proof the migration does what it claims** — the full host-service migration set run against a fresh SQLite database via the drizzle migrator, using the pattern from `packages/host-service/test/migration-0019-preserves-terminal-links.test.ts` (bun:sqlite, foreign keys off across `migrate()`), plus the same script against `main`'s migration set as a negative control:

```
################ WITH 0025 (this PR) ################
applied migrations: {
  n: 25,
}
workspaces.cloud_synced_at: {
  cid: 18,
  name: "cloud_synced_at",
  type: "INTEGER",
  notnull: 0,
  dflt_value: null,
  pk: 0,
}
explicit-column select: prepared and ran
OK: cloud_synced_at present and selectable on a fresh database

################ WITHOUT 0025 (main today) ################
35 | sqlite.query("SELECT id, worktree_path, cloud_synced_at FROM workspaces").all();
            ^
SQLiteError: no such column: cloud_synced_at
      errno: 1,
 byteOffset: 26,
      at prepare (bun:sqlite:345:37)
```

The negative control reproduces the reported error verbatim; with 0025 the column is present, nullable, defaulted to null, and the explicit-column select the shipped build makes prepares and runs.

**`bun test packages/host-service`**

```
 1083 pass
 8 todo
 0 fail
 2536 expect() calls
Ran 1091 tests across 108 files. [189.95s]
```

**`cd packages/host-service && bun run typecheck`**

```
```

Clean, exit 0.

**`bunx biome check <changed files>`**

```
Checked 0 files in 2ms. No fixes applied.
  × No files were processed in the specified paths.
  i These paths were provided but ignored:
  - packages/host-service/drizzle/0025_restore_workspace_cloud_synced_at.sql
  - packages/host-service/drizzle/meta/0025_snapshot.json
  - packages/host-service/drizzle/meta/_journal.json
```

Every changed file lives under `drizzle/`, which biome ignores by configuration. Nothing to format. Repo-wide `bun run lint` was not run (it hangs on cross-worktree bun locks); the root `bun run typecheck` was run separately and exited 0.

Refs HOST-SERVICE-35
Refs HOST-SERVICE-34

https://claude.ai/code/session_2c7e5729-ce4c-4bfc-9fd7-e9b2b9c1e3fc

---
## Summary by cubic
Restores backward compatibility by re-adding `workspaces.cloud_synced_at` via migration `0025`. Before: databases migrated by 1.22.0 dropped the column and 1.21.0 readers failed, breaking the workspace list and file tree. After: the column exists again; older readers succeed without app changes.

**Review and rollout**
- Adds one additive migration: `ALTER TABLE workspaces ADD cloud_synced_at integer;` plus updated Drizzle `meta` snapshot/journal; leaves `schema.ts` unchanged so current code ignores the column.
- Ships in the next `host-service` release from `main` (1.23.0+). Any build that includes this migration repairs the DB on open; affected 1.21.0 hosts work on their next launch. No manual action required.
- Does not restore `workspace_cloud_deletes` and does not revert `0023`; 1.21.0 does not query that table. Uses index `0025` to avoid the existing `0024`. Addresses HOST-SERVICE-35 and HOST-SERVICE-34.

<sup>Written for commit 7b76745a18f0503358c7fc80da4ced555ed4c714. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added workspace synchronization tracking to support cloud sync status.
* **Bug Fixes**
  * Restored the workspace cloud synchronization timestamp field.
* **Chores**
  * Updated database migration metadata to reflect the latest schema changes.

### DIFF
--- a/packages/host-service/drizzle/0025_restore_workspace_cloud_synced_at.sql
+++ b/packages/host-service/drizzle/0025_restore_workspace_cloud_synced_at.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `workspaces` ADD `cloud_synced_at` integer;
--- a/packages/host-service/drizzle/meta/0025_snapshot.json
+++ b/packages/host-service/drizzle/meta/0025_snapshot.json
@@ -1,0 +1,913 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
+  "id": "bf41d518-272a-4d3c-9eda-c932f34148db",
+  "prevId": "d781b875-abdd-499d-8f98-7894ec606fc7",
+  "tables": {
+    "host_agent_configs": {
+      "name": "host_agent_configs",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "preset_id": {
+          "name": "preset_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "icon_id": {
+          "name": "icon_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false,
+          "autoincrement": false
+        },
+        "label": {
+          "name": "label",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "command": {
+          "name": "command",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "args_json": {
+          "name": "args_json",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": "'[]'"
+        },
+        "prompt_transport": {
+          "name": "prompt_transport",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "prompt_args_json": {
+          "name": "prompt_args_json",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": "'[]'"
+        },
+        "resume_args_json": {
+          "name": "resume_args_json",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": "'[]'"
+        },
+        "env_json": {
+          "name": "env_json",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": "'{}'"
+        },
+        "display_order": {
+          "name": "display_order",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        }
+      },
+      "indexes": {
+        "host_agent_configs_display_order_idx": {
+          "name": "host_agent_configs_display_order_idx",
+          "columns": [
+            "display_order"
+          ],
+          "isUnique": false
+        }
+      },
+      "foreignKeys": {},
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "checkConstraints": {}
+    },
+    "host_settings": {
+      "name": "host_settings",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "integer",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": false,
+          "default": 1
+        },
+        "worktree_base_dir": {
+          "name": "worktree_base_dir",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false,
+          "autoincrement": false
+        },
+        "branch_prefix_mode": {
+          "name": "branch_prefix_mode",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false,
+          "autoincrement": false
+        },
+        "branch_prefix_custom": {
+          "name": "branch_prefix_custom",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false,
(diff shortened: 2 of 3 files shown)


saddlepaddle:
Closing unmerged after review — the fix is correct, but it is not worth its permanent cost.

**What this PR did:** re-added `workspaces.cloud_synced_at`, dropped by migration 0023 (shipped in 1.22.0), so that a host-service from the 1.21.0 line — whose schema still declares the column — can read the `workspaces` table on a database a newer build has already migrated. I verified the migration end to end on a fresh database and it does exactly what it claims.

**Why it is being closed anyway.** The failure needs two different builds to share one `host.db`, which is keyed by organization only. That happens when a canary or an independently-installed CLI is older than 0023, or on a rollback. We do not roll back, and the mixed-build window closes on its own as soon as the older side updates — so what remains is a transient internal-only condition. All 254 reported events are a single developer machine inside one eight-minute session.

Against that, the change is permanent: a column that exists in the database and in no source file, deliberately omitted from the snapshot so that `drizzle-kit generate` will not emit a fresh `DROP COLUMN` and reintroduce the bug, needing an explanatory comment to stay understandable, and leaving a trap where re-declaring the column later fails with `duplicate column name`. That is a poor trade for a self-healing window, and it does not generalise — it repairs this one column and nothing about the next one.

**The part worth keeping** is the general problem, which is being raised separately: nothing stops us dropping a column that an already-shipped or already-deployed reader still selects, and that has now caused four Sentry issues across three databases (local-db `persist_terminal`, host-service `cloud_synced_at`, and Postgres `automations.rrule`). The fix for that is a retirement policy or a CI guard, not per-column restores.

The branch is kept if we ever want it back.

PR #6534, opened 2026-08-16
feat(shared,trpc): save triggers as a set with the automation

saddlepaddle:
Replaces the per-trigger CRUD approach in #6531 (now draft). Triggers are edited as part of the automation, so the surface is a **replace-the-set** on `automation.create`/`update`.

No migration.

## Shared validation

Lives in `@superset/shared/automation-triggers`, imported by both the editor and the API, so the form blocks Save on exactly what the server would reject rather than reimplementing the rules.

It deliberately has two levels:

- **`draftTriggerSchema`** accepts a half-configured trigger. The editor has to be able to hold *"Draft opened in [Select Repos]"* with nothing chosen yet — if the schema refused to represent that, the form couldn't hold its own in-progress state.
- **`describeTriggerProblems`** reports what blocks Save, per trigger and per field, so a message can sit next to the control it belongs to. `summarizeTriggerProblems` produces the banner.

The messages match the mockups: *"Some triggers need additional configuration"* and *"Specify at least one repository."*

It also catches **more than one schedule per automation**, which the partial unique index enforces in the database — better as a message next to the row than a constraint violation at save time.

## Matched by id, not cleared and reinserted

A trigger row carries state the editor never sees: a webhook's signing key, and a schedule's next run. Clearing and reinserting would silently roll a key someone had already configured upstream, and reschedule an automation that was only renamed.

So rows present in the set are updated in place, rows absent are deleted, and the next run is only recomputed when the recurrence actually changed.

## GitHub configs are a union on the event

The filters differ per event — *"Any Comment by [Anyone] on a PR by [Anyone] in [Repos]"* carries two independent actor scopes where *"Draft opened"* carries one. A single flat config with one `actor` can't express that, so the config discriminates on `event`.

`includeForks` stays the literal `false`: fork payloads carry attacker-controlled content into a checkout the agent runs in, so enabling it should be a schema change with a threat model attached.

## Verified against a real database

The actual `saveTriggerSet` executed against Postgres 16, not reasoned about:

| | |
|---|---|
| Initial save creates the set | 2 triggers |
| **Webhook signing key survives an unrelated save** | preserved |
| **Schedule not rescheduled by an unrelated save** | preserved |
| Row ids stable across saves | stable |
| Trigger dropped from the set is deleted | deleted |
| Recurrence change *does* reschedule | reschedules |
| Unconfigured GitHub trigger rejected server-side | rejected |

Plus the shared validation itself: a half-configured trigger parses but reports a problem; a comment event requires both actors; `includeForks: true`, unknown events, and an unknown kind are all rejected.

## Compatibility

Top-level `rrule` still works and folds into a single schedule trigger, so the CLI, SDK, MCP and desktop keep working until they move to sending a set. Same expand-contract shape as the column migration.

## Next

The editor UI, then webhook ingestion so event triggers actually fire.

---
## Summary by cubic
Saves automation triggers as a replace-the-set on create/update and makes event-only automations first‑class. Previously list/get assumed every automation had a schedule; now schedule is optional and responses reflect what was actually saved.

- API: `createAutomation`/`updateAutomation` accept optional `triggers: DraftTrigger[]` (max 25). Create requires either `triggers` or `rrule`; if `rrule` is provided, `timezone` is required. If `triggers` are absent, the legacy top‑level `rrule` folds into a single schedule trigger. Invalid sets return 400 with per‑field problems.
- Schedule handling: list/get/getAutomationForUser use a left join; schedule fields in responses are nullable. Create/update now build schedule fields from saved triggers, not the input.
- Validation (`@superset/shared/automation-triggers`): shared draft+save checks; enforces at most one schedule; GitHub config discriminates on `event`; Slack `reaction_added` requires an emoji; actor is `anyone` | `me` | specific people.
- Persistence: new `saveTriggerSet` matches rows by id, deletes dropped rows, and only recomputes `nextRunAt` when the schedule’s recurrence changes. Webhook signing keys and existing `nextRunAt` are preserved across unrelated edits.
- Packages: `packages/shared` exports `./automation-triggers`. No migration.

<sup>Written for commit bb284341f6d3e7bfea061c6a5201da04c79a3e37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6534?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring up to 25 automation triggers, including schedules and event-driven integrations.
  * Automation creation and updates now save, update, and remove trigger configurations reliably.
  * Event-only automations can be created without a schedule.
  * Existing schedule configurations remain supported for compatibility.
  * Added validation for trigger requirements and provider-specific settings, with clear summaries of configuration issues.

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -4,6 +4,10 @@
 	"private": true,
 	"type": "module",
 	"exports": {
+		"./automation-triggers": {
+			"types": "./src/automation-triggers.ts",
+			"default": "./src/automation-triggers.ts"
+		},
 		"./constants": {
 			"types": "./src/constants.ts",
 			"default": "./src/constants.ts"
--- a/packages/shared/src/automation-triggers.ts
+++ b/packages/shared/src/automation-triggers.ts
@@ -33,7 +33,9 @@ export type TriggerScope = z.infer<typeof triggerScopeSchema>;
 
 export const triggerActorSchema = z.union([
 	z.literal("anyone"),
-	z.literal("org_members"),
+	// Resolves to the automation's owner at match time rather than being
+	// expanded to an id on save, so it survives the owner being renamed.
+	z.literal("me"),
 	z.object({ ids: z.array(z.string().min(1)).max(200) }),
 ]);
 export type TriggerActor = z.infer<typeof triggerActorSchema>;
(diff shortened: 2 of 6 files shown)


saddlepaddle:
Fixed in bb28434. The Critical finding was correct and is the most important thing in this PR.

**An event-only automation was invisible.** `list`, `get` and `getAutomationForUser` inner-joined the schedule trigger, so an automation whose triggers are all event-based had nothing to match and was dropped from the result. Create it, hit Save, and it writes correctly to the database — then never appears in the list, and `get` 404s. Indistinguishable from a failed save.

That join was correct when written: at the time every automation had exactly one schedule trigger, so inner and left were equivalent. The trigger set broke that invariant and I didn't revisit the join. Worse, I'd written the hazard down myself in #6531 as the reason to block deleting a schedule trigger — then built a set-save that deletes any trigger not in the set.

Demonstrated against Postgres 16, with one scheduled and one event-only automation:

| | inner join | left join |
|---|---|---|
| Scheduled automation listable | yes | yes |
| **Event-only automation listable** | **no** | **yes** |
| Dispatcher picks up only the due scheduled one | — | yes |

That third row matters: making automations visible must not make event-only ones start firing on a schedule they don't have.

Also fixed from the same review:

- **`rrule` was still required on create**, so an event-only automation couldn't be expressed at all. Now optional, with a refinement requiring either a trigger set or a schedule, and a timezone whenever there is one.
- **Responses described the input rather than what was written.** `create` and `update` reported a top-level schedule even when the saved set contained a different one or none. Both now build the schedule half of the response from what was actually saved, with every schedule field nullable.

Not taken: the header comment referencing a `triggerSchema` export that doesn't exist — corrected. Schedule recurrence validation before the write transaction is worth doing but belongs with the editor work, since the useful version reports per-field problems rather than throwing.

PR #6538, opened 2026-08-16
feat(api): record GitHub deliveries as automation events

saddlepaddle:
First half of making event triggers fire. Populates `automation_events` — the normalized stream triggers get matched against. **No matching or dispatch yet**, deliberately.

No migration.

## Why recording lands before matching

Nothing reads these rows yet, so a mistake here shows up as a wrong row rather than an agent running on the wrong pull request. It also means the matcher gets written against real recorded events instead of a third round of inferring payload shapes — the last two rounds both got things wrong.

`automation_events` has existed since the cutover and is still empty. This fills it.

## What it records

- **Organization** resolved through `github_installations` by installation id. No installation (pings, some org-level events) or an unknown one is skipped with a logged reason rather than treated as an error.
- **Event type qualified with its action** — `pull_request.opened` rather than `pull_request`, because the bare name can't distinguish opened from closed from labeled, and that's exactly the granularity the trigger configs match on.
- **Flattened fields** — title, url, repository, ref, actor — pulled out of payloads whose shape differs per event, so matching doesn't dig through jsonb.
- **`actorIsExternal`** for bots and non-user senders. Not used yet, but it's the field that lets matching treat attacker-influenced payloads differently, and it can't be backfilled.
- **Its own payload copy**, because `ingest.webhook_events` is now pruned at 14 days and a dispatch needs the payload.

## Resource key

A comment on PR #42 and the PR itself both produce `github:acme/web#42`. That's what makes debounce and in-flight limits coherent later — two comments on one pull request are one subject, not two.

Verified:

| event | resource key | title | url |
|---|---|---|---|
| `pull_request.opened` | `github:acme/web#42` | Add login | PR url |
| `issue_comment.created` | `github:acme/web#7` | Bug report | comment url |
| `push` | `github:acme/web@refs/heads/main` | refs/heads/main | — |
| `workflow_run.completed` | `github:acme/web` | CI | run url |
| `installation_repositories.added` | — | (event name) | — |

## Dedupe

A redelivery of the same GitHub delivery id must not produce a second event. Verified against Postgres 16 that `ON CONFLICT (integration_connection_id, provider, external_event_id) DO NOTHING` **fires with a NULL connection id** — GitHub installs aren't `integration_connections` rows, so that column is always null here, and the dedupe only works because the constraint is `NULLS NOT DISTINCT`. Worth checking explicitly given `ON CONFLICT` inference has broken twice on this feature.

## Failure isolation

Recording runs outside the existing processing `try`. A failure here logs and moves on rather than marking the delivery failed — GitHub would otherwise retry a delivery whose real work already succeeded.

## Next

The matcher: recorded event → enabled triggers whose config matches → `automation_runs` with `trigger_id` and `event_id`. By then there will be real rows to test against.

---
## Summary by cubic
Records GitHub webhook deliveries as normalized `automation_events` without triggering matching or dispatch. Previously nothing was recorded; now each processed delivery creates a deduped, stable event, and recording failures do not affect delivery success.

- Resolves organization via `github_installations` by installation id; skips and logs when installation is missing or unknown.
- Qualifies event types with their action (e.g., `pull_request.opened`).
- Uses stable repository identifiers: `repositoryId` and `resourceKey` use the numeric repository id, so renames do not change matching (e.g., `github:12345#42`, `github:12345@refs/heads/main`).
- Flattens title, url, repository id, ref, and actor; stores a payload copy and strips null chars; includes `webhookEventId` for provenance.
- Derives `actorIsExternal` from `author_association`; null when the payload lacks it.
- Ensures idempotency: dedupes on `(integration_connection_id, provider, external_event_id)` with `NULLS NOT DISTINCT`, covering `NULL` connection ids.
- Isolates failures by recording outside the main processing try, so GitHub does not retry already-processed deliveries.

**Rollout**
- No migration. Matching and dispatch come later; current behavior only appends rows to `automation_events`.

<sup>Written for commit 1e36cd43423ae93df729b63dade18cfec002867e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6538?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * GitHub webhook activity is now recorded as automation events with relevant titles, links, metadata, external-actor information, and original event details.
  * Events are qualified by type and action for clearer activity records.
  * Duplicate webhook deliveries are safely ignored.
  * Events from missing or unrecognized installations are rejected from recording.

* **Reliability**
  * Webhook delivery continues successfully even if automation-event recording encounters an error.

### DIFF
--- a/apps/api/src/app/api/github/webhook/recordAutomationEvent.ts
+++ b/apps/api/src/app/api/github/webhook/recordAutomationEvent.ts
@@ -2,6 +2,8 @@ import { db } from "@superset/db/client";
 import { automationEvents, githubInstallations } from "@superset/db/schema";
 import { eq } from "drizzle-orm";
 
+import { stripNullChars } from "@/lib/strip-null-chars";
+
 /**
  * Records an incoming GitHub delivery as an `automation_events` row.
  *
@@ -20,6 +22,9 @@ export type GithubPayload = {
 	installation?: { id?: number | string };
 	repository?: { id?: number | string; full_name?: string };
 	sender?: { login?: string; type?: string };
+	// GitHub's own membership signal, present on comment, PR and review
+	// payloads. The only trustworthy source for "is this person one of us".
+	author_association?: string;
 	pull_request?: {
 		number?: number;
 		title?: string;
@@ -27,6 +32,7 @@ export type GithubPayload = {
 		head?: { ref?: string };
 		user?: { login?: string };
 		draft?: boolean;
+		author_association?: string;
 	};
 	issue?: {
 		number?: number;
@@ -34,7 +40,12 @@ export type GithubPayload = {
 		html_url?: string;
 		user?: { login?: string };
 	};
-	comment?: { body?: string; html_url?: string; user?: { login?: string } };
+	comment?: {
+		body?: string;
+		html_url?: string;
+		user?: { login?: string };
+		author_association?: string;
+	};
 	review?: { state?: string; html_url?: string; user?: { login?: string } };
 	ref?: string;
 	workflow_run?: { conclusion?: string; html_url?: string; name?: string };
@@ -51,7 +62,12 @@ export function resourceKeyFor(
 	payload: GithubPayload,
 	eventType: string,
 ): string | null {
-	const repo = payload.repository?.full_name;
+	// The numeric id for the same reason as repositoryId: a rename must not
+	// change the key, or an in-flight run stops matching its own subject.
+	const repo =
+		payload.repository?.id !== undefined
+			? String(payload.repository.id)
+			: undefined;
 	if (!repo) return null;
 	const pr = payload.pull_request?.number ?? payload.issue?.number;
 	if (pr !== undefined) return `github:${repo}#${pr}`;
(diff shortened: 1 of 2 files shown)


saddlepaddle:
All three taken in 1e36cd4, plus one the review didn't catch.

**Stable repository id.** `repositoryId` held `full_name`, which changes when a repository is renamed — a trigger scoped to it would silently stop matching. Now the numeric id.

**`actorIsExternal` was wrong, not just imprecise.** I derived it from `sender.type`, which distinguishes a bot from a human and says nothing about membership: a genuine outside contributor was being recorded as *internal*, which is exactly backwards for the case the field exists to catch. Now derived from GitHub's `author_association` — `OWNER`/`MEMBER`/`COLLABORATOR` are internal, `CONTRIBUTOR`/`NONE` external — and null when no payload on that event carries one, rather than guessing.

**Payload sanitized.** jsonb rejects null characters; a payload carrying one would have thrown and lost the event. Now stripped, as the `ingest` write already did.

**And one the review missed:** `resourceKey` had the same rename problem as `repositoryId`, one field over. I only found it by testing a rename explicitly rather than checking only the field that was flagged. It is the numeric id now too, which matters more there — the resource key is what debounce and in-flight limits key off, so a rename mid-run would have made a run stop matching its own subject.

Verified: eight association cases including a bot that is a member app, and a rename producing an identical key before and after.

PR #6544, opened 2026-08-16
feat: resolve trigger people through a generic identity table

saddlepaddle:
Replaces the one-table-per-integration shape that `users__slack_users` started, and makes trigger people filters actually resolve.

## One table, every provider

```
user_identities
├── user_id            → auth.users, cascade
├── organization_id    → organizations, cascade
├── provider           text          -- "github" | "google" | "slack" | …
├── external_id        text          -- the provider's stable id
├── external_scope_id  text null     -- workspace/tenant; null when global
├── handle             text null     -- display only
└── display_name       text null

UNIQUE (organization_id, provider, external_scope_id, external_id) NULLS NOT DISTINCT
```

`external_scope_id` isn't speculative: **five of seven** target providers scope their user ids. Slack, Linear, Teams and PagerDuty scope to a workspace/tenant/account; GitHub and Google don't. Without it, the same person in two Slack workspaces collides.

`NULLS NOT DISTINCT` is what makes the *global* providers dedupe — otherwise two GitHub rows with a null scope wouldn't conflict.

`provider` is text rather than the `integration_provider` enum because this also holds `google`, a sign-in provider with no connection row behind it.

## Org-scoped, and multiple accounts allowed

Matching how Linear actually behaves — **verified by experiment, not docs.** Their FAQ claims "you can only connect your GitHub account to a single Linear workspace"; in practice disconnecting in one workspace leaves the other connected, and the same account works in both. The docs are wrong.

Deliberately **no** constraint limiting a person to one account per provider, so linking a work and a personal GitHub is supported. `me` therefore resolves to a **set**, and a PR from either account is still yours — the case Vercel (strict 1:1, [long-standing complaint](https://community.vercel.com/t/still-no-support-for-multiple-github-accounts-misleading-ui-and-ux-long-standing-problem/33762)) and Linear-as-documented both get wrong.

The single constraint says only: the same external account can't be linked twice within one org.

## Matching moved from login to id

A GitHub login can be renamed; the numeric id cannot. Every sign-in record already stores the id, and payloads carry `sender.id` alongside the login. So `actorId` is what people filters compare, and `actorLogin` is display only.

## Backfill

Seeded from GitHub sign-ins so the feature works without every member clicking connect in every org — but **scoped to orgs that actually have a GitHub installation**: elsewhere no GitHub event can ever arrive, so the row would be noise.

**2,029 rows rather than 29,190.** Of 28,887 users with a GitHub sign-in, only 2,015 are in an org with an installation.

## Rehearsed

Full 80-migration chain on Postgres 16:

| | |
|---|---|
| Same account linked in two orgs | allowed |
| Two accounts (work + personal) in one org | allowed |
| Same account twice in one org | **rejected** |
| Backfill links only where an installation exists | pass |
| Backfill skips Google-only sign-ins | pass |
| Re-running the backfill | no-op |

Plus nine matching cases: `me` matching a work account, matching a *personal* account once linked, **not** matching an unlinked one, not matching when nothing is linked; specific-people by id; and a renamed login still matching because the id is stable.

## Follow-up, not in this PR

`users__slack_users` still exists — 110 rows, no readers, and now a duplicate shape. Folding those rows in and dropping it is what actually ends the pattern, but it's a data migration plus a table drop and deserves its own step.

---
## Summary by cubic
Resolves trigger people filters via a generic `user_identities` table and compares GitHub actors by numeric id. Previously `me` could not resolve and matching used login; now `me` resolves to the automation owner's linked GitHub ids and matching uses stable ids.

- Adds `user_identities` (org-scoped) with `provider`, `external_id`, optional `external_scope_id`, and a uniqueness constraint on `(organization_id, provider, external_scope_id, external_id)` using NULLS NOT DISTINCT. Multiple accounts per provider are allowed; the same external account cannot be linked twice within one org.
- Updates matching to use `actorId` and `subjectAuthorId`; `actorLogin` is display only. `apps/api` GitHub dispatch loads owner GitHub ids from `user_identities` and passes them per candidate, making `me` work. `recordAutomationEvent` now records `sender.id` and `user.id` from payloads. Code paths in `@superset/db/schema` and `packages/shared/src/automation-matching.ts` align to id-based matching.
- Backfills GitHub identities from sign-ins only for orgs with installations; rerunnable and a no-op when repeated.
- Required: run migrations `0078_user_identities` and `0079_backfill_github_identities`.
- Follow-up: fold `users__slack_users` into `user_identities` and drop the old table.

<sup>Written for commit d46e95d8da5e1b552d7d4826641f7fb71e53c74c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6544?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added organization-scoped identity records for external accounts.
  * Improved GitHub automation matching using stable account IDs rather than usernames.
  * “Me” conditions now recognize all linked GitHub identities.

* **Bug Fixes**
  * GitHub events now reliably match actors and pull request or issue authors, including accounts with changed usernames.
  * Existing GitHub identities are automatically populated for eligible organization members.


saddlepaddle:
Checked rather than assumed — there is no legacy data to migrate.

Production has **782 triggers, all `kind='schedule'`**, and **zero** GitHub triggers with an actor id list. The trigger editor (#6542) is not merged, so no GitHub trigger has ever been created through any surface.

That is also why this ordering matters: switching people matching from login to numeric id **before** the feature becomes reachable means login-based configs never exist, so the migration this finding describes never becomes necessary. Landing it after the editor would have created exactly the problem.

Verified separately that a renamed login still matches, since the id is what is compared.

saddlepaddle:
Superseded by a branch cut fresh from main — main took 0078 while this was open, and per the playbook a renumber means regenerating rather than renaming. The replacement also folds users__slack_users in, which this did not.

PR #6584, opened 2026-08-18
feat(desktop,host-service): guided add-host flow, and say when a host can't reach GitHub

Kitenite:
Settings → Hosts with no hosts says `No hosts yet.` and stops there. The install/auth/start steps live only in the docs, so adding a second machine means leaving the app to find them.

This replaces that empty state with a guide, and answers on a host's own settings page whether it can reach GitHub — before a clone needs it rather than during one.

## What's here

**Add a host** (`Settings → Hosts` when empty, `Settings → Hosts → Add host…`, or the device picker's *Add another device…*)

Four steps: install the CLI, generate a key, start the host, come back. The key is minted **only when the button is pressed** — the earlier version of this PR minted an account-scoped API key on mount, so merely navigating to a settings page created a credential. It also needed cleanup machinery to guess whether the key had really been wanted; pressing a button needs none of that. Until a key exists the paste command shows its shape with a `<your-key>` slot and copying is off.

The guide polls the host list while it's open and navigates to the new machine's page the moment it registers. The host itself is the confirmation, which beats a success screen you then have to click out of.

**GitHub readiness** on a host's settings page

`host.githubCli` (new) probes `gh auth status` on the machine. If gh is missing or signed out, a banner says so, with a button on the local host and a copyable command on a remote one — `HostServiceSection` already covers version, install source and reachability, so this only covers what that section can't know.

## What was dropped

This PR was two features. The clone-access preflight — `checkCloneAccess`, the composer clone-plan pill, setup-first workspace creates, the project setup modal — has been removed and reverted to main. It has never been executed once: it needs a second reachable host, and main has restructured the composer underneath it twice since it was written. Reverting it keeps that risk out of a change that can be checked on one machine in ten minutes.

Reverted: `CloneAccessStatus`, `ClonePlanPill`, `useCloneAccessPlan`, `ProjectSetupNotice`, the PromptGroup/NewWorkspaceScreen wiring, the v2-project setup modal, `workspace-creates`, `project.checkCloneAccess`, and the `expand-tilde` extraction that only `resolve-repo` needed.

**Reviewers: the four CodeRabbit reviews on this PR predate that split and largely discuss code that is no longer in the diff.**

## Incidental fixes the split surfaced

- `apiKey.create` no longer returns `id`. Once minting became a button press the guide stopped auto-revoking, so nothing read it — a public API change with no consumer.
- `CopyableCommand`'s `size` prop had exactly one caller, `CloneAccessStatus`.
- Removed 1.8 MB of demo GIFs and screenshots from `docs/artifacts`: five of six showed the clone flow, the sixth showed the auto-minting guide that no longer exists.
- `probeGhCli` fell back to a bare `process.env`. A launchd-started host inherits a PATH with no Homebrew, so `gh` raises `ENOENT` and the banner would tell you to install something you already have. Now goes through `getToolEnvironment()`, whose fallback repairs that PATH.
- `classifyCloneError` de-duplicates a `toCloneError` that was already inline in `onboarding/project/page.tsx`. Its three messages were hardcoded English; centralizing them is what made that worth fixing, so they're through the catalog now.
- `"Not set up on this host"` had lost its translation in all 16 locales — the branch re-keyed it, and reverting brought main's English back but not its `msgstr`. Restored from main.

## Verification

`probeGhCli` exercised directly on macOS: signed in reports `authenticated`; a missing binary raises `ENOENT` → `not_installed`; a real `gh` exiting non-zero raises code 1 → `unauthenticated`. Those are the only two things the classifier branches on.

Both typechecks clean, biome clean, desktop 3733 pass, host-service 1711 pass, i18n 157 pass, `check:i18n` exit 0 with all 16 locales at 0 missing.

**Not verified:** the guide's own loop — generate key → paste on another machine → host connects → navigate — has not been run in the app. It needs a signed-in session and a second machine to paste into.

https://claude.ai/code/session_01N82o7z1xerW5qJfKf6DLqD


Kitenite:
Follow-up commit 46642dc closes the last dead-end surfaces: OpenInWorkspaceV2 and both run-in-workspace popovers now render an inline ProjectSetupNotice when the selected project isn't set up on the chosen host, deep-linking into the auto-opened setup modal (focus=setup) instead of leaving a disabled button and a toast. Desktop typecheck + full suite (2590) green.

saddlepaddle:
Scope changed as of 8f2a67300: the clone-access preflight half has been reverted to main and this PR is now just the add-host guide plus the GitHub-readiness banner. Diff against main is 40 files (23 code + 17 catalogs), down from 80.

The four CodeRabbit reviews above were written against the previous scope and mostly point at code that is no longer in the diff.

https://claude.ai/code/session_01N82o7z1xerW5qJfKf6DLqD

PR #6586, opened 2026-08-18
feat(automations): make the webhook trigger fire — inbound bearer route, token generation, row UI

saddlepaddle:
## Summary

The `webhook` trigger kind existed on main as a row that could never fire: no inbound route, no token, no UI. This PR adds all three.

- **API** `POST /api/automations/webhook/{automationId}` — verifies `Authorization: Bearer sset_wh_…` against every enabled webhook trigger on the automation **before** reading the body (SHA-256 hash in `automation_triggers.secret_hash`, constant-time compare; 401 on miss; `sset_wh_` prefix precheck and a 300/min per-automation rate limit sit before the trigger query), records the JSON body verbatim in `automation_events` (`provider=webhook`, `event_type=webhook.received`, fresh uuid `external_event_id`), then hands off to the shared `dispatchMatchingTriggers({…, automationId})`. **No dedupe** — measured against the reference: one authenticated POST is one event and one run, `Idempotency-Key` is not honored. Body must be a JSON object/array (empty → `{}`), 400 otherwise; 413 over 1 MiB; paused automation → 400 `Automation <id> is disabled`; disabled trigger row → its token stops authenticating. If the QStash publish fails the event row is deleted and 500 returned so a sender retry re-enters cleanly.
- **Shared** `WebhookMatchableEvent` + `case "webhook"` in `triggerMatches` → `{matches:true}` (`automation-matching/webhook.ts`).
- **tRPC** `automation.rotateWebhookSecret({triggerId})` returns the token once (hash stored); `automation.setTriggerSecret({triggerId, secret})` stores a provider-issued signing secret verbatim for HMAC kinds (Circleback); `automation.get` returns `secretPrefix`/`secretRotatedAt` per trigger (never the column).
- **Desktop** row: `Webhook triggered · <URL chip, click copies> · Generate auth header` → `Copy auth header` (copies the literal `Authorization: Bearer …` line) → after reload `Regenerate auth header` (confirm dialog naming the old prefix). `EndpointChip` is a shared `TriggerSentence/components` chip (Circleback reuses it byte-identical). Uses main's `SentenceContext.triggerId`.

No migration: `secret_hash/secret_prefix/secret_rotated_at` and the `webhook` enum value were already on main. Schema comment on `secret_hash` updated (bearer kinds hash, HMAC kinds raw; never returned by the API).

## Verified

- Desktop over CDP (pre-seam-consolidation build; the rebase was mechanical): add "Webhook triggered" → Save → **Generate auth header** → **Copy auth header** (clipboard = `Authorization: Bearer sset_wh_…`); reload → **Regenerate auth header**; regenerate → old token 401, new token accepted.
- curl against the local API on the final head: no auth 401 · bad token 401 · null/scalar body 400 · 1.1 MB 413 · paused automation 400 · two identical bodies → two distinct events · same `Idempotency-Key` twice → two distinct events · 320-request bad-token burst → 293×401 + 27×429 · good token reaches the shared dispatcher (QStash rejects loopback destinations locally, so the publish itself isn't provable here — same path as the GitHub route).
- `turbo typecheck` (desktop, api, trpc, shared, db) exit 0; root `bun run lint` exit 0.

Not in scope: injecting the event payload into the run prompt (`dispatch.ts` doesn't read `automation_events` for GitHub either).

## Summary by CodeRabbit

* **New Features**
  * Added webhook trigger support for receiving authenticated JSON payloads and dispatching matching automations.
  * Added webhook endpoint display with copy-to-clipboard support, HTTP method details, and helpful status feedback.
  * Added secure webhook token generation, rotation, and one-time token display.
  * Added support for storing provider signing secrets and viewing secret metadata.
* **Security & Reliability**
  * Added authorization checks, rate limiting, payload size limits, and validation for webhook requests.
  * Webhook events are cleaned up if automation dispatch fails.

### DIFF
--- a/apps/api/src/app/api/automations/webhook/[automationId]/route.ts
+++ b/apps/api/src/app/api/automations/webhook/[automationId]/route.ts
@@ -6,8 +6,8 @@ import {
 	automationTriggers,
 } from "@superset/db/schema";
 import {
-	type MatchableEvent,
 	triggerMatches,
+	type WebhookMatchableEvent,
 } from "@superset/shared/automation-matching";
 import {
 	bearerToken,
@@ -170,20 +170,17 @@ export async function POST(
 		return Response.json({ ok: true, duplicate: true, runs: 0 });
 	}
 
-	const event: MatchableEvent = {
+	const event: WebhookMatchableEvent = {
+		provider: "webhook",
 		eventType: EVENT_TYPE,
-		repositoryId: null,
-		ref: null,
 		actorId: null,
 		actorLogin: null,
-		actorIsExternal: null,
-		labels: [],
 		body: null,
-		isFork: false,
-		subjectAuthorId: null,
 	};
 	const matched = triggers.filter(
-		(t) => t.automationEnabled && triggerMatches(t.config, event, {}).matches,
+		(t) =>
+			t.automationEnabled &&
+			triggerMatches(t.config, event, { ownerIds: [] }).matches,
 	);
 
 	console.log(
--- a/apps/api/src/app/api/automations/webhook/[automationId]/route.ts
+++ b/apps/api/src/app/api/automations/webhook/[automationId]/route.ts
@@ -5,30 +5,22 @@ import {
 	automations,
 	automationTriggers,
 } from "@superset/db/schema";
-import {
-	triggerMatches,
-	type WebhookMatchableEvent,
-} from "@superset/shared/automation-matching";
+import type { WebhookMatchableEvent } from "@superset/shared/automation-matching";
 import {
 	bearerToken,
 	WEBHOOK_TOKEN_PREFIX,
 	webhookTokenMatches,
 } from "@superset/trpc/automation-webhook-secret";
-import { Client } from "@upstash/qstash";
 import { Ratelimit } from "@upstash/ratelimit";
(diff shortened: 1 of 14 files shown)


saddlepaddle:
CodeRabbit triage (2ef2d55bc):

**Fixed**
- Non-object JSON (`null`, scalars) → 400 instead of a NOT NULL violation.
- Body cap: 1 MiB → 413.
- Rate limit: 300/min per automation (`@upstash/ratelimit`, sliding window) checked before the trigger query, plus a `sset_wh_` prefix precheck so garbage tokens never reach the DB.
- Dispatch failure: the just-recorded event is deleted in the catch, so a sender retry re-enters the full path instead of deduping into `runs: 0`.
- `organizationId` comes from the authenticating trigger row.
- Clipboard write rejections show a toast (EndpointChip + auth header).

**Kept as specified**
- Body+minute dedup fallback when no `Idempotency-Key` is sent — that's the brief's spec; identical bodies within one minute collapse. Senders that need exactly-once should send `Idempotency-Key`. Flagged to the coordinator as a product call.

**Skipped**
- Structured/`console.info` logging: matches the GitHub route's `console.log` line.
- Rate limiting / audit log on `rotateWebhookSecret`: owner-only tRPC mutation; optional hardening.

PR #6588, opened 2026-08-18
feat(mobile): "Rate Superset" settings row and App Store rating prompt after positive moments

saddlepaddle:
## Summary
- Settings → Support gets a **Rate Superset** row directly under Contact Support (last row of the group, same star + external-link treatment as the other outbound rows). Tapping opens the App Store write-review page for the live listing (`apps.apple.com/app/id6788926383?action=write-review`).
- Apple's native rating sheet (`expo-store-review`) is requested only right after Superset did something for the user: immediately when a pull request is merged from the phone; on the third session an agent finished for them.
- Never asked twice for the same build or within 90 days; Apple's own 3-per-year cap and the system-wide opt-out apply on top. `app_review_prompt_requested` fires in PostHog so we can see it working.

## Why / Context
We want more 5-star reviews without an in-app banner. Well-regarded iOS apps do two things: a passive "Rate …" row next to Contact Support in Settings, and Apple's system prompt at a peak moment (Duolingo after a lesson, Halide after real usage). The system sheet is the only mechanism that actually moves the rating count and is also the least intrusive one available — it's Apple's small card, capped and user-disableable.

## How It Works
- `packages/shared/src/constants.ts` — `COMPANY.APP_STORE_URL` (id from `eas.json` `ascAppId`).
- `screens/(authenticated)/stores/appReviewStore` — persisted tally of positive moments (deduped to one per hour so an agent stopping several times while the user watches counts once), plus last-prompted time and version.
- `screens/(authenticated)/hooks/useAppReviewPrompt` — `requestAppReview("pr_merged" | "session_completed")`: applies the gates above, waits 1.5s for the screen to settle, checks `isAvailableAsync`, marks prompted, captures the PostHog event, calls `requestReview()`.
- Wired into `WorkspaceScreen` (the existing `attention === "review"` → seen transition) and `PullRequestScreen` `onMerged`.

## Manual QA Checklist
- [x] Settings renders the new row under Contact Support (screenshots: https://claude.ai/code/artifact/f8b180a9-088f-4d61-b599-baa13772bbf8)
- [x] Fresh dev client with `expo-store-review@56.0.3` boots against `expo-modules-core@56.0.23` (no DYLD missing-symbol crash)
- [ ] Rating sheet appears after merging a PR from the phone (dev build shows it every call; TestFlight never shows it; production is where Apple's cap applies)

## Testing
- `bun run lint` (root, clean)
- `bunx tsc --noEmit` in apps/mobile — no errors in touched files; total error count unchanged vs. baseline (pre-existing host-service / workspace-fs type noise)

## Design Decisions
- **Settings row opens the store URL instead of calling `requestReview()`**: Apple says not to tie the sheet to a button, since it may not show. The URL is deterministic.
- **Community moved above Contact Support**: so the rating row is adjacent to Contact Support *and* the last row of the group.

## Notes
- Native module added → dev clients need a rebuild after pulling this.

---
## Summary by cubic
Adds a “Rate Superset” row in Settings and requests Apple’s rating sheet after positive moments to drive App Store reviews without in‑app banners. Previously there was no entry point or prompt; now the row opens the write‑review page and we request the native sheet at targeted moments with strict caps and an immediate reservation to avoid double prompts.

- Adds `expo-store-review@56.0.3`. Dev clients must rebuild after pulling.
- Settings → Support: moves Community above Contact Support; adds “Rate Superset” directly under Contact Support (last in the group). Tapping opens the write‑review page for the live App Store listing.
- Requests the system rating sheet after merging a PR on device (immediate) and after the third distinct “agent finished” session. Moments dedupe within 1h.
- Never prompts again for the same app version or within 90 days; still subject to Apple’s 3/year cap and OS opt‑out. Reserves the prompt before the 1.5s settle delay to prevent back‑to‑back moments from double‑requesting.
- Wires into `PullRequestScreen` merge callback and `WorkspaceScreen` when `attention === "review"`. Dev builds show the sheet on each call; TestFlight often suppresses it; production follows Apple’s caps.
- Captures `app_review_prompt_requested` in PostHog.

<sup>Written for commit 274eb368f32abecd17082347d27cdefc8752e903. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added prompts to rate the app after a pull request is merged or a session is successfully completed.
  * Added “Rate Superset” and “Community” options in Settings.
  * Review prompts are limited by availability, app version, and a 90-day cooldown to avoid interruptions.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -95,6 +95,7 @@
 		"expo-secure-store": "56.0.4",
 		"expo-speech-recognition": "56.0.1",
 		"expo-status-bar": "56.0.4",
+		"expo-store-review": "56.0.3",
 		"expo-system-ui": "56.0.5",
 		"expo-web-browser": "56.0.6",
 		"katex": "0.17.0",
--- a/apps/mobile/screens/(authenticated)/hooks/useAppReviewPrompt/index.ts
+++ b/apps/mobile/screens/(authenticated)/hooks/useAppReviewPrompt/index.ts
@@ -1,0 +1,1 @@
+export { type AppReviewMoment, useAppReviewPrompt } from "./useAppReviewPrompt";
--- a/apps/mobile/screens/(authenticated)/hooks/useAppReviewPrompt/useAppReviewPrompt.ts
+++ b/apps/mobile/screens/(authenticated)/hooks/useAppReviewPrompt/useAppReviewPrompt.ts
@@ -16,7 +16,9 @@ const SETTLE_MS = 1500;
  * or opened a session an agent finished for them (qualifies on the third
  * distinct one). The only UI is Apple's own sheet, which may decline to show,
  * caps itself at three a year, and honours the system-wide opt-out; on top of
- * that we never ask twice about the same build or within 90 days.
+ * that we never ask twice about the same version or within 90 days. The
+ * prompted mark is written before the settle delay so two qualifying moments
+ * in quick succession can't both schedule a request.
  */
 export function useAppReviewPrompt() {
 	const posthog = usePostHog();
@@ -37,10 +39,10 @@ export function useAppReviewPrompt() {
 			) {
 				return;
 			}
+			store.markPrompted(now, version);
 			setTimeout(() => {
 				void StoreReview.isAvailableAsync().then((available) => {
 					if (!available) return;
-					useAppReviewStore.getState().markPrompted(now, version);
 					posthog.capture("app_review_prompt_requested", {
 						moment,
 						positive_moments: moments,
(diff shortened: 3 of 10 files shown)


saddlepaddle:
CodeRabbit triage:
- **Reserve before the delay** — real; fixed in 274eb36 (`markPrompted` now runs synchronously before the settle timer, so back-to-back qualifying moments can't both schedule a request).
- **`nativeBuildVersion` vs `nativeApplicationVersion`** — deliberate. App Store ratings are per *marketing* version, and Apple's guidance is "don't ask again for the same version"; several builds of 1.0.0 should count as one. Not changing.
- **Separate session count** — moot: a merge either prompts (and then the version/90-day gates block everything) or is itself gated, so it can never pre-fill the session threshold. The one-per-hour dedupe is intentional (three separate stretches of the agent finishing for you, not three stops in one sitting). Not changing.

PR #6606, opened 2026-08-18
feat: tag-bound sidebar groups (smart sections) + section→tag migration

Kitenite:
> **Stacked on #6564** (which now carries only the tag data layer) — this PR holds the tag-bound section rendering that was extracted from it, and is the starting point for the section/group rework. **Blocked on a binding UI**: today the `tagBinding` can only be set programmatically, which is exactly why it was pulled out of #6564.

## Summary
- Sections gain a nullable `tagBinding`: membership then derives from workspaces carrying that tag (explicit `sectionId` wins; a multi-tag workspace renders in exactly one group, lowest tab order). Tag-bound headers show a `#tag` chip.
- `planSectionTagMigrations`: converts manual sections to tag-bound ones — slugged tag from the section name with per-project collision suffixes, merge-not-replace member tags, idempotent, executed in an order where no member visually leaves its group mid-switch.

## What this needs before review
- Binding UI: create/edit a tag binding from the section context menu or settings (and surface cross-captured members before running a migration — a member already carrying an earlier-ordered smart group's tag lands there, by design).
- Decide the endgame: keep manual sections alongside smart ones, or converge on auto-materialized tag groups (every in-use tag renders as a group; CLI creates groups by tagging).

## Evidence (from when this rode in #6564)
- 16 unit tests (membership precedence, same-tag twins, cross-project isolation, pinned/hidden precedence, lineage-in-group, migration planner ×6)
- Live-verified end-to-end: CLI tag writes moved rows in/out of a bound group with zero renderer writes; 14-scenario edge battery ALL PASS; migration executed live with before/after captures — see the verification dossier linked from #6562.

---
## Summary by cubic
Adds tag-bound sidebar groups (“smart sections”) and a planner to migrate manual sections. Previously, sections only used explicit sectionId; now, sections with tagBinding auto-include workspaces carrying that tag, with explicit sectionId winning.

- Member resolution: explicit sectionId > tag match; multi-tag workspaces render in the first matching tag-bound section by tabOrder; duplicate tag-bound sections resolve to the lowest tabOrder; membership does not cross projects; pinned/hidden precedence is unchanged; lineage nests inside sections (untagged children re-root if a tagged parent moves).
- UI: section headers display a “#tag” chip with a tooltip when tagBinding is set.
- Data model: adds tagBinding to the local section schema (default null) and threads tags into workspace/sidebar types; no migration required.
- Migration planner: planSectionTagMigrations derives a tag from the section name, dedupes within a project with numeric suffixes, merges (does not replace) existing workspace tags, and is idempotent. Recommended execution order per section: tag members → set tagBinding → clear sectionId.

<sup>Written for commit f1f5f340573dca1019bd6c109f5eabd8e84be077. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/components/DashboardSidebarCloudSection/DashboardSidebarCloudSection.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/components/DashboardSidebarCloudSection/DashboardSidebarCloudSection.tsx
@@ -107,6 +107,7 @@ export function DashboardSidebarCloudSection({
 					lineageDepth: 0,
 					lineageChildCount: 0,
 					lineageCollapsed: false,
+					tags: [],
 				};
 			});
 	}, [cloudWorkspaces, hostWorkspaces, localStateRows]);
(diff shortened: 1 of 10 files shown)


Kitenite:
Folding back into #6564 per discussion — the tag data layer and smart-section rendering ship as one PR again.

PR #6660, opened 2026-08-19
feat(mobile): working photo/file/camera pickers, image paste, and themed system surfaces

saddlepaddle:
## What

The attachments bottom sheet's Photos, Camera, and Files rows did nothing, and images couldn't be pasted into the composer. This PR makes all four attachment paths work and match the system look:

- **Photos/Camera/Files rows work.** The old code waited for a `transitionEnd` event on a screen `router.back()` had already unmounted — Fabric drops events from unmounted views, so the launch never ran. Pickers now present *over* the sheet; the sheet dismisses only after a confirmed pick, and cancel keeps it open. (`useAfterTransitionEnd` removed — this was its only consumer.)
- **Photo picker is a themed bottom sheet.** Explicit `PAGE_SHEET` (the automatic style stacked over the formSheet sometimes hid the Select Photos / filter / search bar), and a new `config-plugins/withIosAccentColor.js` sets the iOS global AccentColor so the picker's confirm/selection controls are white instead of default blue. Build config, no runtime code.
- **Image paste via the long-press edit menu.** New local expo module `modules/paste-input`: an ExpoView above the composer Host offers `paste(_:)` on the responder chain when the clipboard holds images (the SwiftUI field only takes text). Paste appears next to Select/Select All, no permission alert, text paste untouched.
- **No location metadata leaves the app.** Image picks are re-encoded on attach on both the photo-picker and Files paths (HEIC/webp/tiff → JPEG 0.8, PNG stays PNG), stripping GPS/EXIF and ending HEIC rejections by the agent API. Falls back to original bytes only if decoding fails.
- **Polish:** picker/camera failures alert instead of failing silently; non-image attachments render as document tiles in the composer tray instead of blank squares.

## Verification

On an iOS 26 simulator with a rebuilt dev client (new native module — **a dev-client rebuild is required**; older binaries keep running, just with dead paste):

- Photos: repeated round-trips into the composer, cancel keeps the sheet, white confirm/selection controls
- Paste: menu shows Paste only with an image on the clipboard; pasting attaches; text paste unaffected
- Files: zip picked end to end (document tile in tray); HEIC via Files transcoded, no GPS in the output
- Camera: presents full screen over the sheet (simulator shutter ignores synthetic taps, so capture is device-only)
- Location: GPS-tagged HEIC (37.7749 N) picked from the library → attachment read back with ImageIO → no GPS metadata

`lint:fix` / `typecheck` / `sherif` clean; `apps/mobile` tests 3/3 (repo-wide failures are pre-existing host-service local-env suites this PR doesn't touch).

https://claude.ai/code/session_01YKPeWnLuPs2aCYCAyj3YKZ

---
## Summary by cubic
Enables working Photos/Camera/Files attachments and image paste in the composer, and themes iOS system surfaces. Previously, tapping rows closed the sheet and nothing launched; now pickers present over the sheet, dismiss only after a confirmed pick, and cancel keeps the sheet open.

- Adds `@superset/paste-input` to surface Paste for images via the field’s menu; shows only when attachments are enabled; emits pasted images as temp JPEGs with pixel dimensions; no paste alert; text paste unchanged.
- Re-encodes image attachments to strip EXIF/GPS and avoid HEIC rejections: HEIC/HEIF/WEBP/TIFF → JPEG 0.8; PNG stays PNG. Per-asset failures drop just that item; Files attaches undecodable image bytes as a plain file.
- Presents the photo picker as a page sheet and applies a white global iOS AccentColor with `withIosAccentColor` so picker and alert controls match the app theme.
- Alerts on picker/camera failures; non-image attachments render as document tiles in the tray.
- Removes `useAfterTransitionEnd`; pickers are invoked directly over the attachments sheet.
- Required: Rebuild the iOS dev client (new native module `@superset/paste-input` and `withIosAccentColor` asset changes).

<sup>Written for commit ad7b23e13e652b39d27e5eae18521490329a7d3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added image pasting directly into the mobile composer.
  * Added image attachments from Photos, camera, files, and paste actions.
  * Added support for enabling or disabling image pasting.
  * Improved attachment previews in expanded and collapsed composer views.
  * Added a configurable iOS accent color.

* **Bug Fixes**
  * Picker sheets now close only after successful selections.
  * Improved handling of cancellations, picker errors, and unavailable cameras.
  * Unsupported image formats are converted automatically, with failed conversions handled safely.


saddlepaddle:
Addressed the CodeRabbit findings in the latest commit:

- **Batch failure on one bad asset** — fixed: `imageAssetToAttachment` returns `null` on a transcode failure and both call sites (photos, camera) filter it out; the camera path no longer reports "Camera is not available" for a transcode error.
- **Paste offered with attachments disabled** — fixed natively: new `enabled` prop on `PasteInputView`; `canPerformAction(paste:)` returns false when the composer has no tray, so the menu entry disappears rather than no-oping.
- **Points vs pixels** — fixed: dimensions multiplied by `image.scale`.
- **"PNG path does not guarantee EXIF removal"** — not applicable to the picker path: expo-image-picker re-encodes PNG unconditionally via `UIImage.pngData()` regardless of `quality` (`ImageUtils.readDataAndFileExtension`, `case UTType.png.identifier`), which drops metadata. The legitimate half of the finding — the Files-row fallback returning original bytes as an image — is fixed: a failed decode now attaches as a plain file instead.

PR #5964, opened 2026-07-26
chore(desktop): remove resource_snapshot and relay_ws_* telemetry events

saddlepaddle:
## What

Removes two high-volume PostHog telemetry pipelines:

1. **Reverts #5777** (`resource_snapshot`, SUPER-1550) — the every-5-minute desktop memory sampler. At partial rollout it is already ~500k events/day (~2M/week, instantly the #3 event in the project); at full fleet adoption it projects to ~1.4M/day on a ~1.3M/day baseline, i.e. a 2x of total ingestion, decoupled from actual usage (desktops idling all weekend keep emitting).
2. **Removes the `relay_ws_degraded`/`relay_ws_recovered`/`relay_ws_access_denied` outage telemetry** from #5653 (renderer PostHog sink, `outageReporter` module, and the telemetry-only `name` option on `createRelaySocket`). ~97k events last week. The version-stamped desktop events from #5653 (`app_version` register) are kept.

## Why

These are machine-cadence events, not user actions — they inflate PostHog ingestion volume without corresponding product signal.

## Verification

- `bun run typecheck`: 35/35 tasks pass
- Biome clean on all touched files
- `packages/workspace-client` relaySocket tests: 4 pass; desktop `terminal-ws-transport` tests: 15 pass

Note: existing 1.17.x installs keep emitting both event families until they update to a release containing this change.

---

<a href="https://stagereview.app/superset-sh/superset/pull/5964">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

---
## Summary by cubic
Remove desktop memory `resource_snapshot` sampler and relay WebSocket outage telemetry (`relay_ws_*`) to cut PostHog ingestion and focus on user-signal events. Addresses SUPER-1550 by removing machine-cadence pipelines that doubled ingestion without product value.

- **Refactors**
  - Deleted desktop memory telemetry module and tests; removed `startMemoryTelemetry`/`stopMemoryTelemetry` calls (no more `resource_snapshot`).
  - Removed outage reporter and `relay_ws_*` events from `packages/workspace-client`; dropped `setRelaySocketTelemetry` and related tests.
  - Removed optional `name` from `createRelaySocket`; updated desktop/web callers.
  - Kept desktop `app_version` identify properties.

- **Migration**
[...]

PR #6697, opened 2026-08-20
feat(desktop): re-land fleet resource telemetry as a daily digest

Kitenite:
## Summary
- Re-lands the SUPER-1550 fleet memory telemetry that #5964 removed, redesigned so it cannot repeat the ingestion problem: 5-minute samples aggregate locally and upload **one `resource_digest` event per machine per UTC day** (p50/p90/max per metric), plus rare cooldown-limited `resource_alert` events.
- Extends coverage to the blind spots the old `resource_snapshot` never had: child-process trees (host-service, ptys, agents), host memory pressure, main-process event-loop delay, and window-unresponsive episodes.
- Makes renderer `unresponsive` reporting production-grade (Sentry error + digest counter). Today a frozen window in production emits no signal at all.

## Why / Context
@saddlepaddle, tagging you since #5964 was yours and its objections shaped this design.

We currently ship zero performance telemetry from the desktop app. This week a user publicly churned over freezes and memory ("Superset froze my 64GB MacBook"), and reconstructing what happened took forensic work against the residue of old 1.16/1.17 events. Their Electron RSS never passed 2GB; the likely culprit (the pre-#6352 WebGL glyph-atlas growth, and whatever their agent child processes were doing) lives entirely in dimensions we never measured. We also cannot verify from data whether the 1.21.0 perf fixes actually moved the fleet, because every build after 1.17 reports nothing.

#5964's two objections, addressed head-on:
1. **Volume**: the old sampler uploaded every raw sample, ~288 events/day/machine, projected ~1.4M/day on a ~1.3M/day baseline. The digest uploads 1 event/day/machine: at ~8k daily actives that is ~8k events/day, roughly 0.6% of baseline, and it scales with actives rather than with installed idle machines.
2. **Idle machines emitting all weekend**: sampling contributes nothing once `powerMonitor.getSystemIdleTime()` exceeds 30 minutes. An idle desktop goes silent.

## How It Works
- `apps/desktop/src/main/lib/resource-digest/` samples every ~5 min (jittered, unref'd): Electron per-process-class RSS via `app.getAppMetrics()`, child-tree totals + host memory/load via the existing `resource-metrics` collector (numeric totals only), main-process event-loop p99 via `monitorEventLoopDelay`.
- Samples accumulate in memory (bounded, a few KB); the digest flushes on UTC day rollover and on quit (partial days under 3 samples are dropped as noise).
- Threshold crossings (renderer >3GB, total >5GB, children >16GB, host memory >92%, event-loop p99 >1s, window unresponsive) emit an immediate `resource_alert` with a 6h per-kind cooldown.
- Privacy unchanged from #5777: digest payloads pass a hard numeric-key allowlist; the only string an alert carries is its kind from a fixed enum. No IDs, paths, names, or commands can ride along.
- Same gates as before: off in dev/test, honors the telemetry-enabled flag, `track()` additionally no-ops until a user is identified.

## Manual QA Checklist
- [ ] Packaged build: `resource_digest` arrives in PostHog after a day of use (telemetry is production-only, so this needs a canary/packaged run; not yet validated live)
- [ ] Quit after >15 min of use flushes a partial digest with `flushed_on_quit: 1`
- [ ] No sampling while machine idle (no digest sample growth overnight)

## Testing
- `bun test src/main/lib/resource-digest/` (8 tests: aggregation quantiles, day rollover, allowlist filtering, idle gating, quit-flush minimum, alert cooldown, unresponsive counting, disabled/idempotent start)
- Mutation check: removing the idle gate makes the idle test fail, so the suite catches regressions there
- `bun run typecheck`, Biome clean on touched files

## Design Decisions
- **Digest instead of raw samples**: raw samples let PostHog compute arbitrary percentiles, but that flexibility is exactly what cost 2x ingestion. Per-machine daily p50/p90/max still answers every question we actually asked of the old data (fleet percentiles, release regressions, leak-vs-stable, worst-case tails).
- **Reusing the `resource-metrics` collector for child trees**: it already computes per-workspace process-tree memory for the TopBar popover; the digest reads only the numeric totals. The 5-minute idle-mode call costs one `ps` snapshot, which the popover already does far more often when open.

## Follow-ups
- Report the xterm WebGL glyph-atlas size from the renderer (the #6352 cap tracks page adds already); GPU memory is invisible to every RSS metric and was the likely mechanism in this week's churn case.
- Renderer JS heap in the digest (needs an IPC hop; kept out to stay main-process-only here).
- The Sentry `org` tag on `host-service crashed` events is empty in practice, so those 7.7k SIGABRT events/14d are unattributable; worth a separate fix.

## Risks / Rollout
- Risk: low; the sampler never throws into app code (all sample errors are swallowed), timers are unref'd, and payloads are allowlist-filtered.
- Rollout: ships with the next desktop release; no server or flag dependency.
- Rollback: revert this commit; events stop at the next update, same as #5964.

---
## Summary by cubic
Reintroduces desktop fleet resource telemetry as a low-volume daily digest and ensures quit-time events reach PostHog. Old behavior: no resource telemetry and quit-time captures could drop; new behavior: 5‑minute local sampling aggregates into one `resource_digest` per machine per UTC day plus cooldown‑limited `resource_alert`s, sampling pauses on idle machines, and the quit path drains analytics before forced exit.

- Core: `apps/desktop/src/main/lib/resource-digest/` samples with jitter and aggregates p50/p90/max for Electron RSS (by class), child-process trees, host memory pressure, main-process event-loop delay, and counts window-unresponsive episodes. Flush on UTC day rollover; flush on quit only if ≥3 samples.
- Alerts: thresholds for renderer/total/children RSS, host memory (>92%), event loop p99 (>1s), and window unresponsive; 6h per-kind cooldown.
- Privacy: numeric allowlist only; alerts carry `alert_kind` and numeric sample fields; no IDs/paths/commands.
- Gating: production-only, respects the telemetry flag, and `track()` no-ops until a user is identified; timers are unref’d.
- Wiring: start in `main/index.ts` after app setup; stop on `before-quit`. `windows/main.ts` records unresponsive episodes into the digest and sends a Sentry error. Quit sequence now awaits a time-bounded analytics drain (`flushAnalytics`, 2s) before `forceExit` to avoid losing quit-flushed events.
- Tests: coverage for aggregation, idle gating, day rollover, quit-flush minimum, cooldowns, allowlist filtering, disabled/idempotent start, and quit-time analytics drain.

**Rollout**
- No migrations or server changes. Ships in the next desktop release. No action required.

<sup>Written for commit 749fa24e8401ac13c19c158d478a50c9ceaafb86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6697?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added privacy-conscious resource monitoring to track performance and system health.
  * Added daily resource summaries and threshold-based alerts.
  * Added detection and reporting for periods when the main window becomes unresponsive.
  * Monitoring respects telemetry settings, skips extended idle periods, and stops cleanly when the app quits.
  * Added reliable analytics delivery during normal application shutdown, without delaying exit indefinitely.

* **Bug Fixes**
  * Improved shutdown handling when analytics delivery encounters delays or errors.


Kitenite:
Pushed 749fa24e8: the quit sequence now drains PostHog (time-bounded 2s posthog shutdown, started while cleanup runs) before forceExit, so quit-flushed digests reach the wire instead of dying with app.exit(). The other follow-ups from the description are now tracked: #6699 (crash survivorship: checkpoint the aggregate to disk), #6700 (host-service's own RSS), #6701 (rate-limit the Sentry unresponsive capture).

saddlepaddle:
Thanks for chasing the freeze reports. We took resource telemetry out of PostHog in #5964 and I don't want to bring it back, even as a daily digest. Nobody is set up to act on fleet percentiles, and it wouldn't have shown the case that motivated this, since that one looks like GPU memory.

Here's the shape I'd merge:

1. **Renderer hangs via Sentry's built-in detection, not a hand-rolled capture.** The main-process SDK already includes `rendererEventLoopBlockIntegration` by default. The renderer just needs `eventLoopBlockIntegration()` added to its `Sentry.init`. That sends an "Application Not Responding" event with the renderer's JS stack, so we see what was blocking. We're on Electron 41, so try `captureNativeStacktrace: true` on the main side to avoid attaching the debugger, and confirm in a packaged build that stacks come through. Run these events through `createSentryEventThrottle` in a main-process `beforeSend` so a flapping renderer can't burn quota.
2. **Memory at the moment of the report.** In that same `beforeSend`, attach current `app.getAppMetrics()` totals by process type, with the GPU process broken out on its own. Apply it to the hang events and the existing `render-process-gone` report. It's synchronous, so no background sampler is needed.
3. **Local history in the log.** Move the `unresponsive`/`responsive` log lines out of the dev-only block, and write one resource line to the existing electron-log file every few minutes. If that line includes child processes, it needs the v2 surface. The current `collectResourceMetrics` call only sees v1 sessions.
4. **Remove the rest:** `resource_digest`, `resource_alert`, the percentile aggregation, the event-loop histogram, and the quit-time PostHog flush along with the quit-sequence changes.

Freeze counts per release then come straight from the Sentry issue, so there's nothing to add to PostHog. With this shape, #6699 and #6700 can be closed, and #6701 is covered by item 1.

PR #6731, opened 2026-08-21
feat: introduce pages v1

harshithmullapudi:
## Summary

- Publish a self-contained HTML file as a **Page** from the CLI, an agent, or MCP — it gets a stable URL, and republishing the same file adds a version instead of minting a second page.
- Readers **pin comments to any element** on a published page, in the web viewer or in a native desktop pane.
- A comment thread can be **handed to an agent**, which reads the anchors and replies in-thread — which is what makes a page a working surface rather than an export.
- Pages render inside a sandboxed frame with an opaque origin, so user-authored HTML never executes anywhere that holds a session.

## Why / Context

Agents produce HTML worth sharing — reports, dashboards, one-off microsites — and today there is nowhere to put it. This adds the smallest thing that makes that shareable *and* actionable: one file, one URL, a version history, and a comment loop that feeds straight back to an agent.

The unit is deliberately Claude's: **a version is one self-contained HTML file**. Not a static site, not a build output, no accompanying media library — CSS, JS, and images are inlined by whoever authors the file.

## How It Works

**Five new Neon tables**, one migration (`0091`, with `0092` relaxing an FK):

| Table | Role |
|---|---|
| `pages` | identity: slug (public URL path), title, visibility, optional version pin |
| `page_versions` | immutable; `unique(page_id, version)` doubles as the concurrent-publish guard |
| `workspace_pages` | the republish key: `(workspace_id, entry_path)` → page |
| `page_comment_threads` | one thread pinned to one place, against one **version** |
| `page_comments` | replies; soft-deleted so removing the first comment can't take the thread |

**Publish** is a single transaction — resolve the target page, mint the next version, upsert the workspace edge. The blob upload happens *outside* it, so a multi-MB body never holds a write-pool connection open. A `blobCommitted` flag distinguishes a pre-COMMIT failure (blob is certainly orphaned, delete it) from a failure at COMMIT itself (may well have landed — deleting would strip a live version of its bytes).

**Comments store an anchor, never a position.** Coordinates are a function of viewport width, font loading, and content, so the viewer re-resolves them on every layout change. A small runtime is injected into the page frame and reports geometry outward; it renders nothing, holds no comment text, and knows no author identity. Everything a reader sees is drawn by the parent, out of reach of page-authored JavaScript.

**The agent loop:** a human activates a thread for an agent (`agent_activated_at`), the desktop hands the thread — anchor path included, not just the quoted text — to a live agent session, and the reply lands as `author_kind = 'agent'` while still naming the person who dispatched it. Activation is cleared on republish, since the new version is not what anyone agreed to.

**Surfaces**

| Surface | What it does |
|---|---|
| CLI | `pages publish / list / get / pull / versions`, `pages comments list / reply / resolve` |
| MCP | the same seven operations as tools, so an agent can publish and answer comments |
| Web (`/p/<slug>`) | viewer, comment overlay, visibility menu |
| Desktop | native page pane, Pages sidebar tab, comment overlay, hand-off-to-agent menu |
| Skill | `plugins/superset/skills/page` — how to author for the sandbox, which is where most of the sharp edges are |

## Design Decisions

| Decision | Choice | Rationale |
|---|---|---|
| Assets | none — one file, inlined | The previous revision of this PR shipped a `files`/`attachments` media library with parser-based URL rewriting. It was dropped: it roughly doubled the surface area, and the sandbox already forbids the interesting cases. A page is one file. |
| `workspace_pages.workspace_id` | bare `uuid`, no FK | Only `cloud_workspaces` holds live workspace rows in Neon; local workspaces live in a machine's own SQLite. An FK would restrict the feature to cloud sandboxes. |
| Threads anchor to a **version** | `page_version_id` not-null | An anchor only means something against the bytes it was recorded on. A thread whose version is unknown can never be resolved or honestly marked orphaned. |
| Slug | title + 5 random chars, **always** | Removes the collision path entirely — no "is this taken" lookup, no retry loop. Minted once and frozen, so a retitle never moves a shared link. |
| Frame — web | `srcdoc` | Blob serves HTML as an attachment under its own `default-src 'none'` CSP, so framing the URL directly cannot work. |
| Frame — desktop | `superset-page:` scheme | A `srcdoc` document inherits the embedder's CSP, and the renderer's policy has no `'unsafe-inline'` — it silently killed both the comment runtime and the page's own scripts. Main serves the bytes under their own policy instead. |
| Sandbox | no `allow-same-origin` | Gives the frame an opaque origin: no cookies, no storage, no parent DOM. `allow-scripts` **with** `allow-same-origin` would let it strip its own sandbox, and cross-subdomain cookies mean a `*.superset.sh` host would hand user JS a live session. |
| Default visibility | `just_me` | Widening a page is a deliberate act, so nobody discovers too late that a draft was org-visible all along. |
| Publish / visibility | creator only | An `org` page is readable org-wide, but a reader who could publish to it could replace a colleague's page at its shared URL. |

## Testing

- `typecheck` clean across `db`, `shared`, `cli-framework`, `cli`, `trpc`, `auth`, `ui`, `web`, `desktop`
- `biome check` clean
- **42 unit tests** across 5 new files — slug minting, upload validation, entry-path canonicalisation, publish rules, page access
- **20 integration tests** against real Postgres — version arithmetic, concurrent-publish races, workspace resolution, cross-org and cross-member access:

  ```
  BLAXEL_API_KEY=test BLAXEL_REGION=test BLAXEL_SANDBOX_IMAGE=test \
  BLAXEL_WORKSPACE=test OPENAI_API_KEY=test \
  bun test --env-file=.env ./packages/trpc/src/router/page/publish.integration.ts
  ```

  Deliberately not named `*.test.ts` — it needs a migrated database and full env validation, so `bun test` would fail at import time in CI. The leading `./` matters or bun runs nothing.

**Manually verified:** the desktop page pane rendering, and both web 404 screens against a local dev server.

> [!IMPORTANT]
> The end-to-end publish checklist from the previous revision is **stale** — it exercised the asset-upload flow that this revision removes. Publish → republish → version history → comment → agent reply wants one clean pass through the real CLI against the real API before merge. Publishing from a **cloud** workspace is still unverified (no sandbox to hand).

## Known Limitations

- **Blob URLs are public and permanent.** Access control is enforced at the API, not at the bytes. A `just_me` page's URL stays fetchable by anyone who has seen it, and changing visibility revokes nothing. Fine for org-internal use; **not** sufficient if `just_me` must mean confidential.
- **No orphan sweep.** A blob whose transaction failed after COMMIT, or whose page was later cascaded away, is never collected.
- **Same-org squatting of a local workspace key.** A member can claim `(workspace_id, entry_path)` for a colleague's *local* workspace, so the colleague's next republish versions the squatter's page. Cross-org is closed; this needs local workspaces to have a verifiable Neon identity.
- **`everyone` visibility is in the DB enum but not offered by the API.** Serving a page without a session needs a user-content domain. Postgres cannot drop an enum value, so it stays in the type.
- **No rate limiting** on `page.publish`.
- **The viewer's 404 returns HTTP 200.** `loading.tsx` streams the shell before `notFound()` is reached. Cosmetic while the route sits behind sign-in, since no crawler ever reaches it.
- **Quote-range anchors are unimplemented.** `page_comment_anchor_kind` carries `text`, but the viewer only places element and page anchors, and the API rejects the rest.

## Follow-ups

1. **A user-content domain** (`supersetusercontent.com`-style, *not* a `superset.sh` subdomain). Closes the blob-URL gap, lifts the inherited-CSP workaround on desktop, and unlocks `everyone`.
2. The orphan sweep.
3. Rate limiting on `page.publish`.
4. `pages delete`, `pages open`, markdown pages, `shared_version` pinning UI, quote-range anchors.

## Risks / Rollout

- **Risk:** one additive migration creating five new tables and three enums. No existing table is altered, so the blast radius is limited to new objects.
- **Rollout:** migrations apply through the normal deploy flow. No backfill, no data movement. `apps/web` needs no new env var — the page URL derives from `NEXT_PUBLIC_WEB_URL`.
- **Rollback:** the feature is inert if unused. Reverting the code leaves five empty tables behind; dropping them is safe while nothing has published.

## Summary by CodeRabbit

- **New Features**
  - Added published HTML Pages with visibility controls, version history, workspace linking, and page URLs.
  - Added threaded comments with replies, editing, resolution, deletion, and agent handoff support.
  - Added Pages browsing and viewing in the desktop app, including comment tools and secure rendering.
  - Added CLI commands for publishing, listing, viewing, pulling, versioning, and managing comments.
  - Added MCP tools for page and comment workflows.
  - Added access-aware web page views with loading, error, and not-found states.
- **Documentation**
  - Added Pages guidance covering publishing, rendering, visibility, versioning, and comments.


packages/cli/src/commands/pages/pull/command.ts
```diff
@@ -0,0 +1,49 @@
...
+import { pageRefFromArg } from "../pageRef";
+
+export default command({
+	description: "Fetch a published version back to disk",
+	args: [positional("page").required().desc("Page id or slug")],
+	options: {
+		version: number()
+			.alias("v")
+			.desc("Version to fetch (defaults to the one currently served)"),
+		out: string()
```
saddlepaddle: nit: is the "out" dir something we'd like to present publicly? I was thinking the local version of an artifact is something that is more of an implementation detail vs. something that needs to be exposed as a concept.

packages/auth/src/seed-dev.ts
```diff
@@ -60,7 +74,46 @@ async function seedDevAccount(): Promise<void> {
...
+ * exist before the flow can start — without it the authorize endpoint answers
+ * `invalid_client` and login dies before the browser opens. Production has one
+ * registered out-of-band; locally there was nothing, which is why CLI login
+ * against a dev API had never worked.
+ *
+ * A public client: no secret, PKCE carries the proof instead. The redirect URIs
+ * mirror `packages/cli/src/lib/auth.ts` — one per loopback port it may bind,
+ * plus the paste-the-code fallback on the web app for when none are free.
+ */
+async function seedCliOAuthClient(): Promise<void> {
```
saddlepaddle, line 90: ah very nice catch! apologies for the disruption as well lol

packages/trpc/src/lib/upload-bytes.ts
```diff
@@ -0,0 +1,46 @@
...
+
+	const buffer = decodeBase64Content(content);
+	if (buffer.length === 0) {
+		throw new TRPCError({ code: "BAD_REQUEST", message: "File is empty" });
+	}
+	if (buffer.length > maxBytes) {
+		throw new TRPCError({
+			code: "BAD_REQUEST",
+			message: `File too large (${(buffer.length / 1024 / 1024).toFixed(2)}MB). Maximum is ${maxBytes / 1024 / 1024}MB`,
+		});
```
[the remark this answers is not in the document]
harshithmullapudi: Resolved this

packages/db/drizzle/0092_add_pages_and_page_comments.sql
```diff
@@ -0,0 +1,91 @@
+CREATE TYPE "public"."page_comment_anchor_kind" AS ENUM('element', 'text', 'page');--> statement-breakpoint
+CREATE TYPE "public"."page_comment_author_kind" AS ENUM('human', 'agent');--> statement-breakpoint
+CREATE TYPE "public"."page_visibility" AS ENUM('just_me', 'org', 'everyone');--> statement-breakpoint
+CREATE TABLE "page_comment_threads" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"page_id" uuid NOT NULL,
+	"page_version_id" uuid NOT NULL,
```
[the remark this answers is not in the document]
harshithmullapudi: It's unreachable today

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/PagePane/components/PageHandoffMenu/PageHandoffMenu.tsx
```diff
@@ -0,0 +1,148 @@
...
+					))
+				)}
+			</DropdownMenuContent>
+		</DropdownMenu>
+	);
+}
+
+// The anchor path goes to the agent too: a page is one self-contained .html file, so the
+// published DOM is the source it edits, and repeated text alone locates nothing.
+function buildPrompt(
```
saddlepaddle: nit: I recommend pulling utils out into separate files per repo guidelines

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/PagePane/PagePane.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { PageVisibilityMenu } from "./components/PageVisibilityMenu";
+import { usePageCommentStore } from "./hooks/usePageCommentStore";
+
+interface PagePaneProps {
+	data: PagePaneData;
+}
+
+// No logo or home button, unlike the web viewer: this is a pane inside the app,
+// so the chrome around it is already ours.
+function PagePaneMessage({
```
saddlepaddle: nit: we do one react component per file normally

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/PagePane/PagePane.tsx
```diff
@@ -0,0 +1,172 @@
...
+	notFound,
+}: {
+	title: string;
+	description?: string;
+	notFound?: boolean;
+}) {
+	return (
+		<div className="flex h-full w-full flex-col items-center justify-center gap-3 px-6 text-center">
+			{notFound ? (
+				<Pixel404 className="max-w-[200px] pb-3 text-foreground" />
```
saddlepaddle: probably should replace the NotFound() component

apps/web/src/app/page/[slug]/utils/getPagesAccess/getPagesAccess.ts
```diff
@@ -0,0 +1,41 @@
+import { auth } from "@superset/auth/server";
```
Comment (saddlepaddle) [inline apps/web/src/app/page/[slug]/utils/getPagesAccess/getPagesAccess.ts:1]: nit: prefer page over p

apps/web/src/app/p/[slug]/page.tsx
```diff
@@ -0,0 +1,162 @@
...
+		(error.data?.code === code || error.shape?.data?.code === code)
+	);
+}
+
+const isNotFound = (error: unknown) => hasCode(error, "NOT_FOUND");
+const isForbidden = (error: unknown) => hasCode(error, "FORBIDDEN");
+
+// Rendered inline rather than thrown: Next strips an error boundary's message in
+// production, and the organization's name is the whole point of this screen.
+function WrongOrganization({ message }: { message: string }) {
```
Comment (saddlepaddle) [inline apps/web/src/app/p/[slug]/page.tsx]: nit: subcomponents should be in separate files instead of in the main file

saddlepaddle (commented):
This viewer is 162 lines and six error branches to fetch one document and put it in an iframe, and most of that traces to one avoidable round trip. Flagging it as a rough edge rather than a blocker — the comments in here are genuinely good and several branches are load-bearing.

### The root cause: four sequential hops, one of which is pure ID lookup

```
getPagesAccess()        → auth session + PostHog flag
page.get({ slug })      → the record
page.pull({ id })       → a blob download URL
fetch(downloadUrl)      → the bytes
```

Four hops mean four failure modes mean four error branches. But `page.get` and `page.pull` are almost entirely redundant here. The viewer reads exactly five fields off `get` — `id`, `title`, `description`, `visibility`, `createdByUserId` — and `pull` already returns four of them. `publish.ts` even says why in a comment: *"Projected from the row `loadPage` already fetched, to save a second `get` round trip."*

The only reason `get` is called is that `pullPageSchema` accepts a uuid and this route has a slug. And `pull` already delegates to `loadPage`, which handles **both** id and slug — the capability is there, just not exposed through the input schema.

### The fix

1. **Let `pull` take a slug.** Widen `pullPageSchema` to the existing `pageRefSchema` shape and pass both through to `loadPage`. ~3 lines.
2. **Add `description` to `pull`'s projection.** One line — it already projects `visibility` and `createdByUserId` for exactly this reason.
3. **Delete the `page.get` call.** That removes a network round trip from every page view, plus one try/catch and one error branch. `generateMetadata` uses the same cached call.
4. **Move the flag + auth gate to middleware** so the component starts from "you're allowed here". This also fixes the 200-status 404 listed under Known Limitations — that happens because `loading.tsx` streams the shell before `notFound()` is reached, which is a symptom of gating inside the component rather than ahead of streaming.
5. **Extract the blob fetch** into `utils/getPageContent/`, next to `getPagesAccess`. Right now the `AbortSignal.timeout`, `cache: "no-store"`, `!response.ok` handling and two `console.error`s are a data-access layer inlined into a React component.
6. **Pick one error-code shape in `hasCode`.** `error.data?.code || error.shape?.data?.code` checks both; one of them is dead.

That should land around 60 lines and three branches without losing any behaviour that's actually doing work.

### Separately: three PostHog clients in one process

`getPagesAccess` stands up its own `new PostHog(...)`, copying `getAgentsUiAccess`. There's already a documented singleton at `packages/trpc/src/lib/analytics.ts` ("Singleton — all server-side product event captures go through this client"), and it's already used for flag reads, not just captures — see `getFeatureFlagPayload` in `packages/trpc/src/lib/relay-url.ts`.

It isn't importable from `apps/web` as-is, though: `@superset/trpc` has no export path for it, and adding one would pull in that package's env validation, which `apps/web` may not satisfy. So the cheap version is a single `apps/web/src/lib/posthog-server.ts` that both `getAgentsUiAccess` and `getPagesAccess` import. Low stakes — `flushAt: 1` means there's no batching to lose — but it's three places to edit when the host or key config moves.

Worth noting this mostly evaporates if the gate becomes middleware, since there'd be one call site anyway.

### What I'd leave exactly as it is

- `WrongOrganization` rendered inline rather than thrown — Next strips error-boundary messages in production and the org name is the whole point of that screen.
- No OG/Twitter tags, with the reasoning in the comment. Correct, and the kind of thing that gets "helpfully" added later.
- The `cache()` wrapper around `getPage`, and the comment explaining that `api()` caches the client and not the result.
- `not-found.tsx` copy being true of both "missing" and "not yours" — that's the NOT_FOUND-over-FORBIDDEN decision followed through properly.

apps/web/src/app/p/[slug]/utils/getPagesAccess/getPagesAccess.ts
```diff
@@ -0,0 +1,41 @@
...
+import { PostHog } from "posthog-node";
+import { cache } from "react";
+
+import { env } from "@/env";
+
+const posthog = new PostHog(env.NEXT_PUBLIC_POSTHOG_KEY, {
+	host: env.NEXT_PUBLIC_POSTHOG_HOST,
+	flushAt: 1,
+	flushInterval: 0,
+});
```
Comment (saddlepaddle) [inline apps/web/src/app/p/[slug]/utils/getPagesAccess/getPagesAccess.ts]: Worth pulling this out into `apps/web/src/lib/posthog-server.ts` and importing it in both places.

This isn't a knock on the code — it's copied from `getAgentsUiAccess`, so it followed the local convention correctly. The convention is just a duplicated client, and this makes it the third `posthog-node` instance in one process: `packages/trpc/src/lib/analytics.ts` (the documented singleton, already used for flag reads via `getFeatureFlagPayload` in `relay-url.ts`), `getAgentsUiAccess`, and now this one.

Reaching for the trpc singleton directly would be the more correct fix, but `@superset/trpc` has no export path for `lib/analytics` and adding one drags in that package's env validation, which `apps/web` may not satisfy. Not worth the deploy risk for a tidiness change. One module in `apps/web/src/lib/` keeps the env boundary clean:

```ts
// apps/web/src/lib/posthog-server.ts
import { PostHog } from "posthog-node";
import { env } from "@/env";

// One server-side client for the whole app. flushAt: 1 / flushInterval: 0 because
// Vercel functions are short-lived and would otherwise drop buffered events.
export const posthogServer = new PostHog(env.NEXT_PUBLIC_POSTHOG_KEY, {
  host: env.NEXT_PUBLIC_POSTHOG_HOST,
  flushAt: 1,
  flushInterval: 0,
});
```

Low stakes either way — with `flushAt: 1` there's no batching to lose, so the cost is drift rather than behaviour: three places to edit when the host or key config moves.

One thing to preserve if you do fold them together: this helper passes `personProperties: { email }` into `getFeatureFlag` and `getAgentsUiAccess` doesn't. That difference is load-bearing — the `pages` flag targets on an email release condition, so the property has to travel with the call to resolve before the client's `identify` is ingested. Share the *client*, not the call.

(Related: if the flag gate moves into middleware as suggested in my other comment, this collapses to a single call site anyway.)

packages/mcp/src/tools/pages/pageRef.ts
```diff
@@ -0,0 +1,11 @@
...
+ * `pageRefSchema` wants `id` or `slug` present, but a `nullish()` tool input
+ * arrives as an explicit null, which is not the same as absent — passing it
+ * through fails the refine with a confusing message.
+ */
+export function pageRef(input: { id?: string | null; slug?: string | null }) {
+	return {
+		...(input.id ? { id: input.id } : {}),
+		...(input.slug ? { slug: input.slug } : {}),
+	};
+}
```
saddlepaddle: The diagnosis in this comment is exactly right — `.nullish()` on the tool side against `.optional()` on the tRPC side, where zod rejects `null` on the type check before the refine runs. But the coercion can live in the schema after all, one level down from where you looked.

`defineTool` takes a `ZodRawShape`, so there's no object to hang a `.transform()` on. Individual fields are still fair game:

```ts
const optionalish = <T extends z.ZodTypeAny>(schema: T) =>
  schema.nullish().transform((v) => v ?? undefined);

// get.ts / versions.ts
id: optionalish(z.string().uuid()).describe("Page UUID."),
```

Parsed input is then `string | undefined`, never `null`, and can go straight to the caller.

That deletes this file — and also the eight identical spreads in `publish.ts:82-95`, which are this same null-strip inlined rather than named. There's a ninth in `automations/list.ts:23`, so a shared helper in `packages/mcp/src/` would pay for itself immediately across the package, not just for pages.

One thing to check rather than assume: `.transform()` returns a `ZodEffects`, and the SDK runs the shape through JSON Schema generation for the tool listing. It should use the input side and keep the `.describe()`, but worth eyeballing the emitted tool list once before rolling it out anywhere else.

---

Two related notes, neither of which I think belongs in this PR:

**Constraints are declared twice.** `pageRefSchema` says `slug: z.string().min(1).max(120)`, and `get.ts` and `versions.ts` each re-declare `min(1).max(120)`. Same story for publish — `filename` 255, `title` 200, `description` 2000, `label` 200, `entryPath` 1024, all mirrored from `publishPageSchema` with nothing keeping them in step. Exporting the field schemas from the router and decorating them at the MCP edge (`optionalish(pageFields.slug).describe(...)`) would give one source for the constraint and one place for the agent-facing description.

**A raw shape can't express "exactly one of id or slug."** That's a `.refine()`, so a model can call `pages_get` with neither and only find out at runtime. The structural fix is in `defineTool`: accept a `z.ZodObject`, pass `.shape` to the SDK for the listing, and validate against the full object in the handler wrapper — which would let every tool in the package express cross-field constraints. Framework change, definitely not this PR, but it's what would make this file unnecessary rather than merely tidier.

packages/mcp/src/tools/register.ts
```diff
@@ -66,6 +74,14 @@ const REGISTRARS = [
...
 	terminalsRead,
 	terminalsClose,
+	pagesList,
+	pagesGet,
+	pagesVersions,
+	pagesPull,
+	pagesPublish,
+	pagesCommentsList,
+	pagesCommentsReply,
+	pagesCommentsResolve,
```
saddlepaddle, line 84: **This is the one I'd treat as blocking.** The `pages` PostHog flag gates the UI only — the agent surface ships to 100% of users on merge.

Three places, none of them flag-aware:

- these eight registrars, unconditional
- `page` / `pageComment` mounted unconditionally in `packages/trpc/src/root.ts:34`
- `pages` as a top-level group in `superset --help` via `cli.config.ts`

I grepped for `FEATURE_FLAGS`, `featureFlag`, and `posthog` under `packages/mcp/src`, `packages/cli/src/commands/pages`, and both page routers — no hits. So the flag hides the windows and leaves the door open: every user in every org can publish via CLI and MCP the moment this lands, and every MCP client's tool list grows by eight entries whether or not that org has Pages.

That matters more than tool-list noise, because `page.publish` has no rate limit and blob bytes are permanent and never swept. The flag is doing none of the containment it looks like it's doing.

The fix mirrors what `getPagesAccess` already does — server-side flag eval, just at registration and mount time rather than per-route. Something like gating the `REGISTRARS` entries on a resolved flag for the connecting user, and the router mount on the same.

Worth saying: I don't think this is an oversight in the code so much as an unstated decision about what "behind a flag" means for an agent-facing feature. Perfectly reasonable to ship the agent surface wide on purpose — it's genuinely inert until someone publishes. But that should be a call you make out loud, not something the rollout section implies the opposite of.

packages/trpc/src/router/page-comment/page-comment.ts
```diff
@@ -0,0 +1,336 @@
...
+					threadId: input.threadId,
+					authorKind: input.agentSessionId ? "agent" : "human",
+					// Recorded either way: an agent reply is still made on someone's behalf.
+					authorUserId: userId,
+					agentSessionId: input.agentSessionId ?? null,
+					body: input.body,
+				})
+				.returning();
+
+			return { id: comment?.id };
```
saddlepaddle, line 292: Two findings here, filed together because they share a root cause: `reply` trusts the caller both about **who is speaking** and about **whether they were invited**.

### `agent_activated_at` is migrated but never used

`0091` creates `agent_activated_at` and `agent_activated_by_user_id`, and `schema.ts:1303-1304` declares them. Grepping `agentActivated|agent_activated` across every `.ts`/`.tsx` in the branch returns exactly those two declarations — nothing reads or writes them.

So the push model isn't enforced: `pages_comments_list` returns every thread on any page the agent can read, and `reply`/`resolve` act on any thread id. Hand-off is `terminal.send` typing a prompt, and the prompt's own "Only act on the thread ids listed above" is the only fence.

That's worth reconciling because the PR description states the opposite — *"a human activates a thread for an agent (`agent_activated_at`) … Activation is cleared on republish, since the new version is not what anyone agreed to."* Either behaviour is defensible, but right now the columns, the description, and the code disagree with each other, and whoever picks this up next will trust the description.

The cheap version, since the schema is already there: stamp `agent_activated_at` in the hand-off mutation, require it non-null in `reply`/`resolve`, and scope `pages_comments_list` to activated threads when the caller is an agent. Clearing it on republish falls out of `publishPage` in a line.

### `author_kind` is caller-controlled

```ts
authorKind: input.agentSessionId ? "agent" : "human",
```

`agentSessionId` is a free `z.string().min(1).max(200)`, and `0092` dropped the FK so nothing constrains it. Any org member can post a comment that renders as an agent reply with an invented session id — and conversely an agent could omit it and post as a human.

Attribution-only, so low severity, but it's exactly the attribution the hand-off loop leans on when a reader is deciding whether a thread was answered by a person or a machine.

The server already knows which it is: an MCP call arrives with `ctx.clientLabel`, a CLI call with the pane env. Deriving `author_kind` from the authenticated caller rather than a body field costs nothing and makes the column mean what it says. `agentSessionId` can stay as a descriptive field; it just shouldn't be the thing that decides the kind.

packages/trpc/src/router/page/publish.ts
```diff
@@ -0,0 +1,286 @@
...
+					organizationId,
+				});
+				await tx
+					.insert(workspacePages)
+					.values({
+						workspaceId: input.workspaceId,
+						pageId: page.id,
+						entryPath: input.entryPath,
+					})
+					.onConflictDoNothing();
```
saddlepaddle: This `onConflictDoNothing()` swallows a case that should surface, and I think the Known Limitations entry describes it backwards.

Walk it through. `resolveTargetPage` filters the workspace edge by `eq(pages.createdByUserId, userId)`. So if a colleague already holds `(workspace_id, entry_path)`:

1. the lookup misses — their page is filtered out by the creator predicate
2. `createPage` mints a **new** page for this caller
3. the `workspacePages` insert violates `workspace_pages_workspace_id_entry_path_unique`
4. `onConflictDoNothing()` eats it, and the transaction commits

The caller gets back a URL, a version 1, and a success message — with a page that is linked to no workspace, doesn't appear in the Pages tab, and can't be found by the workspace lookup again. Their *next* publish repeats all four steps and mints another one. Republishing never works, and nothing ever says why.

The README entry reads: *"A member can claim `(workspace_id, entry_path)` for a colleague's local workspace, so the colleague's next republish versions the squatter's page."* That can't happen — `assertPageWritable` would throw FORBIDDEN first. The real failure is quieter and worse: unbounded orphan page creation with a success response.

It also isn't limited to the local-workspace identity gap. Two people publishing the same entry path from a **shared cloud workspace** is enough, and that's an ordinary thing for two people on one sandbox to do.

Minimal fix is to stop swallowing it — target the conflict and let a genuine collision throw:

```ts
.onConflictDoNothing({ target: [workspacePages.workspaceId, workspacePages.pageId] })
```

so the idempotent re-link still no-ops, while an `(workspace_id, entry_path)` collision surfaces as a real error the caller can act on ("that path is already published by someone else in this workspace — pass `--page <id>` or rename"). Better than a success message for work that didn't land.

packages/ui/src/components/PageComments/providers/CommentProvider/CommentProvider.tsx
```diff
@@ -0,0 +1,268 @@
...
+	const [enabled, setEnabled] = useState(false);
+	const [draft, setDraft] = useState<CommentDraft | null>(null);
+	const [activeThreadId, setActiveThreadId] = useState<string | null>(null);
+	const [hoverRect, setHoverRect] = useState<FrameRect | null>(null);
+	const [rects, setRectState] = useState<Record<string, FrameRect | null>>({});
+	const [submitting, setSubmitting] = useState(false);
+	const [busyThreadId, setBusyThreadId] = useState<string | null>(null);
+	const [framePointerDownAt, setFramePointerDownAt] = useState(0);
+	// A ref keeps `discardDraft` stable, so the frame's pointer-down handler is not rebound per keystroke.
+	const submittingRef = useRef(false);
```
saddlepaddle, line 104: The provider is doing more than it looks like from the call sites — `toggleEnabled` clearing the draft/active/hover set, `openDraft` closing the active thread, `discardDraft` guarded by `submittingRef` so an in-flight submit can't lose unsent text, `runSubmit` keeping the composer open on failure, `runThreadAction` closing the card only on success. That's real coordination and it's thought through. None of it should be lost.

But look at what kind of state it all is: **view state around an async mutation**. Which is the one thing react-query already models — and the store deliberately throws it away before the provider sees it.

### `submitting` and `busyThreadId` are being rebuilt from scratch

Both stores hold five live mutation objects, then flatten them into bare `Promise<void>`:

```ts
addReply: async (threadId, body) => { await reply.mutateAsync({ threadId, body }); },
```

`reply.isPending` goes in the bin. So the provider recovers it with `useState` plus a ref:

- `submitting` is just `create.isPending || reply.isPending || edit.isPending`
- `busyThreadId` is just `resolve.variables?.threadId ?? remove.variables?.threadId ?? null` — react-query keeps `variables` on the mutation while it's in flight, so this is free

That removes three `useState`s, the `submittingRef`, and both `runSubmit` / `runThreadAction` wrappers. The behaviours they encode become one-liners at the call site (`if (submitting) return`, `onSuccess: () => setDraft(null)`).

### `toThreads` is duplicated byte-for-byte

`apps/web/.../usePageCommentStore.ts` and `apps/desktop/.../usePageCommentStore.ts` contain the same 25-line mapper, same `ServerThread` alias, same two doc comments. A change to `pageComment.list`'s projection needs both edited in lockstep.

Deriving from `RouterOutputs` was the right instinct — the comment *"a renamed or dropped field fails here instead of reading undefined"* is exactly right. It just wants one home. `packages/ui` can't be it (it would need `@superset/trpc`, and a component library that knows tRPC output shapes stops being transport-agnostic), so I'd export the mapper and its return type from `packages/trpc` alongside the router it maps from. `packages/ui`'s structural `CommentThread` stays the contract, still checked where the app assigns the store.

The two hooks themselves should stay separate — desktop uses the `createTRPCReact` proxy, web uses `useTRPC()` with `queryOptions`/`mutationOptions`, and abstracting over both costs more than it saves.

### The context isn't reachable-state — it's siblings under a parent that already exists

`PagePane` justifies the provider with *"the registry renders header extras in a sibling subtree, outside this provider."* That subtree — the pane registry's `renderHeaderExtras` — genuinely is unreachable, and it would justify context. But the toggle wasn't put there; it sits in `PagePane`'s own header row. Same on web, where `page.tsx` renders the header and `<main>` itself.

So both consumers are children of one component that already exists, and there are only three `useComments()` call sites — two of which read a single field (`CommentModeToggle` wants `enabled`, `PageHandoffMenu` wants `threads`). The leaves are already pure props.

There's a cost to the current shape beyond the indirection: `rects` is rewritten on a `requestAnimationFrame` driven by scroll, resize, a `ResizeObserver` and a `subtree: true` `MutationObserver`. Every `useComments()` consumer re-renders on every frame of a scroll, including the two that wanted one field.

### The shape I'd aim for

```tsx
// app parent (PagePane / page.tsx)
const store = usePageCommentStore({ pageId, version });   // now exposes submitting, busyThreadId
const [commentMode, setCommentMode] = useState(false);
const [handoffOpen, setHandoffOpen] = useState(false);

<header>
  <PageHandoffMenu threads={store.threads} open={handoffOpen} onOpenChange={setHandoffOpen} … />
  <CommentModeToggle enabled={commentMode} onToggle={setCommentMode} />
</header>
<main>
  <PageCommentsView
    html={html} store={store} user={user}
    commentMode={commentMode}
    onFramePointerDown={() => setHandoffOpen(false)}
  />
</main>
```

`draft`, `activeThreadId`, `hoverRect`, `rects` become `useState` inside `PageCommentsView` — it's their only reader, so the per-frame re-render stops at that boundary on its own. `providers/CommentProvider/` and `hooks/useFramePointerDown/` both delete; `framePointerDownAt` was only ever a counter to fake an event across the context, and a prop callback is the event.

Net: two small `useState`s and three props duplicated across the two apps, against a deleted provider, a deleted hook, a 24-member context value, and three pieces of hand-rolled async state.

### Checklist for whoever does it

These five behaviours are the spec, and all five survive the move:

1. leaving comment mode clears draft, active thread, hover
2. opening a draft closes the active thread
3. a dismissal during an in-flight submit must not discard unsent text
4. a failed create keeps the composer open with its text
5. a failed resolve/delete keeps the card open on that thread

Worth landing as a follow-up rather than in this PR — it's a refactor of working code, and the merge-blocking items are elsewhere.

saddlepaddle (approved):

PR #6742, opened 2026-08-21
perf(admin): stop blocking the dashboard shell on work it doesn't need

saddlepaddle:
admin.superset.sh takes 5-6s to load. Production traces (30d) put the server side at:

| transaction | p50 | p75 | p95 |
|---|---|---|---|
| `middleware GET` | 628ms | 940ms | 1,044ms |
| `GET /` | 881ms | 3,011ms | 3,721ms |

Two invocations, both before any HTML. This removes three things from that path.

### The middleware was doing the layout's job twice

`proxy.ts` resolved the session on every request; `(dashboard)/layout.tsx` then ran the identical session + email-domain check. admin has exactly one route, so the middleware was pure duplication.

Deleting it also drops `@superset/auth/server` — and with it better-auth, the Stripe SDK, Resend and every react-email template — out of the middleware bundle. admin serves ~40 loads/day, so nearly every visit is a cold boot and pays that import cost.

### The layout blocked on a cross-deployment hop for three strings

It awaited `trpc.user.me` over HTTP to the api deployment (411ms in traces) to fill in a name, email and avatar the session already carries. That await also meant `loading.tsx` could never render.

### The metric caches were per-instance

`mrrCache`, `mrrPendingRunId` and `cashFlowCache` were module-level variables. On Fluid Compute that's one cache per instance.

For MRR this was worse than a miss. The pending-run handle was per instance too, so the tile's 10s poll usually landed on an instance that had never heard of the run in flight — and started a **new** Stripe Sigma query rather than collecting it. It never converged, which is why the tile sat permanently on "Computing in Stripe — up to a minute on first load", burning a Sigma run per dashboard load.

Both caches now live in the Redis already stood up for rate limiting. The pending-run handle is claimed with `SET NX` so concurrent instances agree on one owner.

### Prewarming

A QStash schedule (hourly, already created) hits `/api/analytics/jobs/refresh-mrr` so the tile only ever reads a landed result. `refreshSigmaMrr` deliberately ignores the cache on its first step — otherwise an hourly job against a 12h entry would read its own previous result and never actually refresh.

Vercel Cron was the alternative and would have kept the schedule in-repo, but QStash is where the other six schedules live and `verifyQstashRequest` already exists. Vercel Cron also has no retries and logs nothing on a missed run, which is fine for this job but not a pattern worth splitting the fleet over.

## Not included

The session lookup itself still costs 7 database round trips. That fix touches `packages/auth` and every authenticated request in the product, so it's split into its own PR rather than riding along here.

## Testing

- Lint, typecheck (37/37), full test suite green
- The MRR path is exercised by the hourly schedule; the cache behaviour is Redis-backed so it survives instance recycling

---
## Summary by cubic
Speed up the admin dashboard shell by removing duplicate auth, inlining sidebar user data, and moving business-metric caching to Redis with single-owner coordination. Previously the shell blocked on middleware and a cross-deployment `user.me` call; now it renders immediately, metrics converge, and when Redis is missing MRR runs are refused instead of starting and “computing” forever.

- Remove `apps/admin/src/proxy.ts`; the layout enforces session and `COMPANY.EMAIL_DOMAIN`. The middleware bundle no longer imports `@superset/auth/server` (and transitives: `better-auth`, Stripe SDK, `Resend`, `react-email`).
- Inline `session.user` into the sidebar as a `SidebarUser` (name, email, image). The layout stops calling `trpc.user.me`, so `loading.tsx` can render and we avoid sending unused user fields to the client.
- Add Redis-backed metric cache (`packages/trpc/src/lib/metric-cache.ts` via `@upstash/redis`) and coordinate MRR runs with `SET NX`; move cash-flow caching to Redis too. Export `@superset/trpc/business-metrics` for reuse.
- Add `/api/analytics/jobs/refresh-mrr` with an hourly QStash trigger; `refreshSigmaMrr` ignores the cache on its first step so scheduled refreshes actually refresh.
- Degraded path: if the shared cache is unavailable, MRR refuses to start runs and returns “metric cache not configured” instead of looping on “computing”.

- Rollout
  - Set `KV_REST_API_URL` and `KV_REST_API_TOKEN` in prod (same Redis used for rate limiting). The QStash schedule already exists.
  - In dev without Redis: MRR endpoints report unavailable and do not start Sigma runs; cash-flow still fetches but is not shared-cached.
  - Monitor admin load time and Stripe Sigma runs; expect immediate shell render and no duplicate runs.

<sup>Written for commit baf5cb67d0f1a1014fa8b2fab4feb405d6fecc1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added automatic MRR data refresh support for analytics.
  - Added shared caching for MRR and cash-flow metrics.
  - Added configurable polling and time limits for MRR refresh operations.

- **Bug Fixes**
  - Improved coordination of analytics refreshes to prevent duplicate processing.
  - Improved recovery from interrupted or stalled refreshes.
  - Added safer fallback behavior when shared caching is unavailable.

- **Improvements**
  - Simplified dashboard user handling for more reliable sidebar display.


saddlepaddle:
Both addressed in `baf5cb67d`.

**Redis-unavailable runaway** — correct, and it was the same failure the PR exists to fix, just moved to the degraded path. `claimMetricCache` now fails closed (a claim nobody can coordinate shouldn't be granted), and `advanceSigmaMrr` checks `isMetricCacheAvailable()` up front and returns `metric cache not configured` rather than reporting `computing` forever. Verified with KV unset: 0 of 5 claims granted, 0 Sigma runs started. With Redis present, 16 polls over 8s produce exactly 1 run, which is the NX claim behaving.

**Session user projection** — done, now passes only `{ name, email, image }`.

Leaving `metric-cache.ts` in `packages/trpc/src/lib/`: it already serves both the MRR and cash-flow paths, and that directory is where this package's shared lib code lives (`posthog-client.ts`, `relay-presence.ts`).

PR #6744, opened 2026-08-21
fix(mobile): stop waiting on the workspace destroy the host already committed

saddlepaddle:
Deleting a workspace on a phone took 10-20 seconds staring at the screen you deleted from, then usually said **"Delete failed: Request timed out"** about a delete that had gone through, and left you on the workspace with **"This workspace's host is offline"** — the placeholder for a machine that was fine and still serving every other row.

## Why the wait bought nothing

`host-service` archives the workspace row **before** the dirty-worktree preflight, the teardown script and the worktree removal — step 0, commented in `workspace-cleanup.ts` as "the commit point" — and that archive is what drops the workspace from every list. Everything the mutation was still awaiting happens after the delete is already decided.

Worse, both relays cap a tRPC exchange at 30s and answer `BAD_GATEWAY: "Request timed out"`. A teardown script routinely outlives that, so the message the phone showed was the relay describing its own tunnel, for a delete that had committed.

## What changes

The row goes and the screen is handed back the moment the user confirms; the destroy runs unattended.

- **The workspace screen leaves for the list on confirm** (`dismissTo`) instead of sitting on a placeholder whose row is gone. The home row just vanishes.
- **The confirm is the whole gate.** Every destroy goes out with `force` rather than preflighting for uncommitted work first: on a phone the answer to "this worktree is dirty" is yes anyway — agent work is cheap to redo — and a second alert after a refused destroy costs more than it protects. `force` carries git consent only; the teardown script still runs.
- **A failed destroy asks the host directly whether the workspace is still there.** Gone means it committed and only the relay gave up waiting; still there is a real failure, reported without the relay's timeout wording pinned to it.
- Teardown-script failure keeps the existing behaviour — retry without it rather than prompt, since there is nobody at a phone to fix a script.

Both call sites (home row menu, workspace actions sheet) share one `useDeleteWorkspace` hook; cloud workspaces drop their row optimistically the same way, since `cloudWorkspace.delete` awaits the provider tearing the sandbox down.

The delete-in-flight spinner on the home row goes with this — the row leaves on confirm now, so there is nothing left to spin. `removeWorkspace` / `refetchHost` on the cache-ops interface were its only callers.

## Verified on device, against production

Before/after traces with screenshots and timings: https://claude.ai/code/artifact/6ea54b44-560f-446c-b86d-f3feaf839545

Both runs used a throwaway workspace whose `.superset/teardown.sh` sleeps 35s, so the destroy outlives the relay's 30s cap — the condition that produced the report.

| | on `main` | on this branch |
|---|---|---|
| back on the list | ~30s (via an error alert) | **0.6s** |
| "host is offline" placeholder | at +25s | never |
| "Request timed out" alert | at +30.5s | never |

Host-side the destroy ran to completion in the background in both runs (worktree gone at +37s / +39.6s).

`bun test`, `biome check` and `tsc --noEmit` are clean for `@superset/mobile`.

Not exercised: cloud (sandbox) workspaces — shared code path, but no sandbox was driven — and a genuinely failing destroy, where the row returns and the alert *should* fire.

https://claude.ai/code/session_01A935A7SxB3U9wKtjjAi3pZ

## Summary by CodeRabbit

* **New Features**
  * Added a unified workspace deletion flow for cloud and connected workspaces.
  * Workspace removal now updates lists immediately and refreshes them after completion.
  * Added clearer handling for offline connections, cleanup failures, and deletion errors.

* **Bug Fixes**
  * Cloud access is cleared only after successful workspace deletion.
  * Workspace rows remain usable while deletion is in progress, preventing unnecessary navigation blocking.

### DIFF
--- a/apps/mobile/hooks/useCloudWorkspaceActions/useCloudWorkspaceActions.ts
+++ b/apps/mobile/hooks/useCloudWorkspaceActions/useCloudWorkspaceActions.ts
@@ -1,7 +1,10 @@
 import { useQueryClient } from "@tanstack/react-query";
 import { useCallback } from "react";
+import type { CloudWorkspaceRow } from "@/hooks/useCloudWorkspaces";
 import { clearSandboxAccess } from "@/lib/sandbox-access";
 import { apiClient } from "@/lib/trpc/client";
+
+const LIST_KEY = ["cloud", "cloudWorkspace", "list"];
 
 /**
  * Rename and delete for cloud workspaces go to the API, not the sandbox: the
@@ -13,10 +16,7 @@ export function useCloudWorkspaceActions() {
 	const queryClient = useQueryClient();
 
 	const invalidate = useCallback(
-		() =>
-			queryClient.invalidateQueries({
-				queryKey: ["cloud", "cloudWorkspace", "list"],
-			}),
+		() => queryClient.invalidateQueries({ queryKey: LIST_KEY }),
 		[queryClient],
 	);
 
@@ -30,11 +30,21 @@ export function useCloudWorkspaceActions() {
 
 	const remove = useCallback(
 		async (id: string) => {
-			await apiClient.cloudWorkspace.delete.mutate({ id });
-			clearSandboxAccess(id);
-			await invalidate();
+			// The row goes when the user asks, not when the provider finishes
+			// tearing the sandbox down; the invalidate below is what restores it
+			// if the delete never landed.
+			queryClient.setQueriesData<CloudWorkspaceRow[]>(
+				{ queryKey: LIST_KEY },
+				(rows) => rows?.filter((row) => row.id !== id),
+			);
+			try {
+				await apiClient.cloudWorkspace.delete.mutate({ id });
+				clearSandboxAccess(id);
+			} finally {
+				await invalidate();
+			}
 		},
-		[invalidate],
+		[invalidate, queryClient],
 	);
 
 	return { rename, remove };
--- a/apps/mobile/hooks/useCloudWorkspaceItems/useCloudWorkspaceItems.ts
+++ b/apps/mobile/hooks/useCloudWorkspaceItems/useCloudWorkspaceItems.ts
@@ -149,22 +149,9 @@ export function useCloudWorkspaceItems(): CloudWorkspaceItemsValue {
 					},
 				);
 			},
-			removeWorkspace: (hostId, workspaceId) => {
-				const key = keyFor(hostId);
-				if (!key) return;
-				queryClient.setQueryData<HostWorkspaceRow[] | undefined>(key, (rows) =>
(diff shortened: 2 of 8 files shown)


saddlepaddle:
Verified on device against production — before/after traces, screenshots and timings:
https://claude.ai/code/artifact/6ea54b44-560f-446c-b86d-f3feaf839545

Both runs used a throwaway workspace whose `.superset/teardown.sh` sleeps 35s, so the destroy outlives the relay's 30s exchange cap — the condition that produced the report.

**On `main`:** host archives the row at +0.5s, phone stays on the workspace screen with no feedback until +25s, flips to "This workspace's host is offline" (the host was online throughout), then throws "Delete failed · Request timed out" at +30.5s. The worktree was actually gone at +37s.

**On this branch:** back on the workspace list at **+0.6s**; the relay's cap passes at +30s with nothing on screen; 83 consecutive frames from +2s to +49s are pixel-stable, so no alert ever drew. Host-side the destroy ran to completion in the background exactly as before.

Also captured all three confirm variants on real workspaces in each state (clean / uncommitted changes / unpushed commits) — the dirty case now takes one confirm instead of two.

Not exercised: cloud (sandbox) workspaces, and a genuinely failing destroy where the row returns and the alert *should* fire.

saddlepaddle:
Both CodeRabbit findings reviewed against the current head (they were raised against `433db184e`, before the branch was rewritten).

**"Do not force deletion after an unavailable preview" (Critical) — moot.** The `inspect` preflight, the `warning` state and the CONFLICT force-retry are all gone; the file has zero occurrences of any of them. Every destroy now sends `force` deliberately.

The underlying concern is fair and I want it on the record rather than waved away: **this design can delete uncommitted changes without warning.** That is the intended tradeoff — on a phone the answer to "this worktree is dirty" is yes anyway, agent work is cheap to redo, and a second alert after a refused destroy cost more than it protected. `force` stays git-consent-only; the teardown script still runs.

**"Do not treat `deleteInProgress` as a completed deletion" (Major) — real, not acting on it.** Accurate that we return without confirming a terminal state, so if a *concurrent* delete of the same workspace fails and un-archives, our optimistically-removed row stays hidden.

It self-heals: `useHostWorkspaces` refetches every 30s plus on focus and pull, and each refetch replaces the cached list wholesale with the host's answer — an un-archived row comes straight back. So the exposure is one poll interval, not indefinite.

The suggested fix — wait for the competing delete to reach a terminal state, then re-query and retry — is the polling machinery this PR exists to remove, and a retry would race the host's process-local `destroysInFlight` guard. Desktop's hook carries the same warning: *"Do NOT extend this to `in-progress` … retrying just races the same guard."* Reaching it at all needs two clients deleting one workspace inside the same destroy window.

Merging with these noted.

PR #6753, opened 2026-08-22
fix(host-service): remove a git-unregistered-but-left-on-disk worktree on delete

### LINKED ISSUE #6730: [bug] Workspace delete can silently orphan a worktree on disk when git unregisters it mid-failure
### Where did this happen?

Desktop app (macOS)

### Version

1.24.0-canary.20260821030527 (git 2.50.1, Apple Git-155)

### What happened?

Deleting a workspace ("Separate session and workspace loaders", branch `separate-session-workspac`) failed with:

```
Failed to remove worktree at /Users/avipeltz/.superset/worktrees/<project-id>/thin-zone
```

Root-caused this to **three stacked bugs** in the workspace-delete saga, confirmed by reading the source and reproducing the underlying git behavior directly:

1. **`git worktree remove --force --force` can partially fail while still fully unregistering the worktree.** Reproduced directly (git 2.50.1): make a nested file inside a worktree undeletable (`chflags uchg`), then run `git worktree remove --force --force <path>`. Git prints `error: failed to delete '<path>': Operation not permitted` and exits 255 — but immediately afterward, `git worktree list --porcelain` no longer lists it and the repo's `.git/worktrees/<name>` admin entry is gone entirely, even though the working directory (and the worktree's own `.git` file) are still sitting on disk. Unregistration and file deletion are **not atomic** in git's own implementation.

2. **The app doesn't account for that.** In `packages/host-service/src/workers/tasks/git.ts` (`gitWorktreeRemoveTask`), the `git worktree remove --force --force` call's error is discarded (`.catch(() => {})`), and the task's return value (`stillRegistered`) is derived solely from re-running `git worktree list --porcelain` afterward:

   ```ts
   await git
       .raw(["worktree", "remove", "--force", "--force", target])
       .catch(() => {});
   const raw = await git.raw(["worktree", "list", "--porcelain"]);
   return {
       stillRegistered: parseWorktreeList(raw).some(
           (w) => normalizeWorktreePath(w.path) === target,
       ),
   };
   ```

   In `packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts`, when `stillRegistered` comes back `false`, the code sets `worktreeRemoved = true` and moves on — it never re-checks `existsSync(local.worktreePath)` before trusting that. (Two *other* branches in the very same function — the "session" branch and the "project repo is missing" branch — already do exactly that `existsSync` + `rm(path, { recursive: true, force: true })` fallback; it just isn't wired into the common "project + repo present" path that actually ran here.)

   When `stillRegistered` *does* come back `true` (as it apparently did on the attempt that produced this toast), the code throws:

   ```ts
   throw new TRPCError({
       code: "INTERNAL_SERVER_ERROR",
       message: `Failed to remove worktree at ${local.worktreePath}`,
   });
   ```

   — with no detail, because the real git error from step 1 was already thrown away.

3. **Nothing logs the underlying git failure anywhere.** Checked `host-service.log`, the terminal-host `daemon.log`, and the Electron `main.log` — none contain any trace of the actual `git worktree remove` error. The toast text is the entire diagnostic record. The softer warnings this same saga produces (e.g. `"N terminal(s) may still be running"` from a failed PTY kill) are also only returned as strings in a `warnings` array to the renderer — never logged server-side — so even those vanish once the toast is dismissed.

**Net effect:** the specific workspace above is now stuck — git has zero record of it, but its directory (a partial sparse-checkout: `apps/api`, `apps/discord-triage`, `apps/marketing`, `apps/mobile`, `apps/relay`, `apps/relay2`, plus root files) is still on disk, and its `workspaces` DB row is still un-archived. Clicking Delete again as-is will very likely report success (git already shows nothing registered) and archive the row, permanently leaving the orphaned folder behind — nothing targets an already-unregistered path for cleanup.

**Most likely trigger** (circumstantial, not provable after the fact because of bug #3): `runDestroyPhases` step 3a (`workspace-cleanup.ts`) disposes PTY sessions tied to the workspace before step 3b removes the worktree, but a failed kill only pushes a *warning* — it doesn't block or retry step 3b. `disposeSessionsByWorkspaceId` (`terminal.ts`) only touches sessions tracked in the `terminalSessions` table, so a detached grandchild process (e.g. a dev server or install step whose parent PTY died but which wasn't fully reaped) would survive invisibly. `host-service.log` shows repeated `[workspace-fs/watch] FSEvents overflow` bursts for this exact path in the surrounding window — consistent with something still actively writing into the directory while the recursive delete ran.

### Steps to reproduce

Reliable repro of the underlying git-level bug (not the full app flow, which requires timing a race):

```bash
git init -q main-repo && cd main-repo
git config user.email a@b.com && git config user.name test
echo hi > file.txt && git add . && git commit -q -m init
git worktree add -q --no-checkout ../wt1 -b wtbranch1
cd ../wt1 && git checkout wtbranch1
mkdir -p sub && echo x > sub/lockedfile
chflags uchg sub/lockedfile   # simulate a file the recursive delete can't remove
cd ../main-repo

git worktree remove --force --force ../wt1
# -> error: failed to delete '.../wt1': Operation not permitted (exit 255)

git worktree list --porcelain   # wt1 is already gone from the list
ls .git/worktrees/               # admin dir for wt1 is gone entirely
ls ../wt1                        # directory (and its .git file) still on disk
```

In the app, this manifests whenever `git worktree remove --force --force` hits any obstruction partway through its recursive delete (locked/open file, permission issue, a live process still writing into the tree) during workspace delete.

Full write-up with timeline, exact file/line references for every claim above, the missing-`existsSync`-fallback comparison, a concrete fix checklist, and immediate remediation steps for the specific stuck workspace: see `docs/issues/workspace-delete-orphaned-worktree.md` on the `debug-workspace-delete` branch.

Suggested fixes (detailed in that doc):
- Log the real git stderr in `git.ts`'s `gitWorktreeRemoveTask` instead of `.catch(() => {})`.
- After `removeWorktree` reports `stillRegistered: false`, still check `existsSync(local.worktreePath)` before trusting `worktreeRemoved = true`; fall back to the same `rm(..., { recursive: true, force: true })` pattern the other two branches in `workspace-cleanup.ts` already use.
- Make the terminal-disposal step actually gate (or retry) the worktree-removal step instead of degrading a failed kill to a no-op warning.
- Add a regression test for "git reports not-registered after remove, but the directory still exists on disk" — the existing suite covers "still registered → throws" and "git open failure blocks delete," but not this case.

### Related issues

- #4941 — same symptom class on an older version (1.11.3): `fatal: '<path>' is not a working tree`, partial deletion left a `.vite/` folder behind. Closed without (as far as I can tell from this investigation) the swallowed-error / missing-`existsSync`-fallback root cause actually being fixed — this issue looks like a fresh recurrence of the same underlying defect on the current host-service implementation.
- #2863 — same class of report against the *old* desktop-side implementation (rename-to-`.superset-delete-*` → `git worktree prune` → detached `rm -rf`), since replaced by the current host-service `workspace-cleanup` saga this issue is about. Worth noting the failure mode survived the rewrite in a different shape.
- #5499 — the general "reconcile partially-deleted workspaces" feature request. This issue is a concrete, reproduced root cause behind that pattern (specifically the "worktree dir on disk, no registration" flavor it documents) — fixing the two items above would close one real source of the drift #5499 is asking to reconcile after the fact.
- #5631 — adjacent ask for an auto-prune/cleanup command; a good complement once orphans are reliably created less often, but doesn't address why they're created in the first place.
- #6070, #6380 — different bugs in the same `workspace-cleanup.ts` destroy saga (force-flag semantics, branch-delete ordering). No known interaction with this bug, but worth the same author's attention since a fix here touches adjacent code.

andyst-dev:
Fixes #6730

## What & why

Two fixes for the workspace-delete saga in #6730:

**1. Orphaned-worktree cleanup (`workspace-cleanup.ts`).**
`git worktree remove --force` is not atomic: it can fully **unregister** a
worktree (so it no longer appears in `git worktree list` → `stillRegistered =
false`) while still failing to delete the working directory when a nested file
is undeletable (`Operation not permitted`, e.g. `chflags uchg`). The destroy
saga trusted `stillRegistered` alone and set `worktreeRemoved = true`, then
archived the row — leaving a silently **orphaned folder** on disk with no git
record of it. The repo-missing and session branches in the same saga already do
an `existsSync` → `rm` fallback under the managed-worktrees root guard; this
wires that same fallback into the common "project + repo present" path.

**2. Logging the underlying git error (`workers/tasks/git.ts`).**
The real git failure was swallowed by `.catch(() => {})`, so the toast text was
the only diagnostic and nothing reached host-service.log. It's now captured and
logged via `console.warn` (the existing host-service pattern) with the target
path, so the cause is discoverable when a folder is left orphaned.

## How I tested it

- New unit test in `packages/host-service/test/workspace-cleanup.test.ts`:
  `removeWorktree` returns `{ stillRegistered: false }` while the worktree's
  directory exists on disk → asserts the directory is gone after destroy.
- `bun test packages/host-service/test/` → **308 pass, 0 fail** (workspace-cleanup
  suite 26/26).
- `bun x tsc --noEmit` clean on the changed files.
- Also threaded a `hostWorktreeBaseDir` option through `makeCtx` so
  `getHostWorktreeBaseDir` returns a value in the test harness (the live
  host-settings lookup was previously untestable there).

Note: #6730's point 1 (git's non-atomic unregister/delete) is a git behavior we
cannot change in-repo; this PR covers points 2 and 3.

## Checklist

- [x] `bun run lint` and `bun run typecheck` pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace cleanup for directories that remain after Git unregisters a worktree.
  * Safely removes leftover directories within the managed workspace area.
  * Preserves directories outside the managed area and warns when cleanup is skipped.
  * Cleanup status now accurately reflects whether removal succeeded.
  * Reports cleanup failures appropriately instead of indicating successful removal.
  * Added clearer warnings when forced worktree removal fails, including the affected location and error details.

### DIFF
--- a/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
@@ -536,6 +536,40 @@ async function runDestroyPhases(
 					message: `Failed to remove worktree at ${local.worktreePath}`,
 				});
 			}
+			// `git worktree remove --force` is not atomic: it can fully
+			// unregister the worktree (so it no longer appears in
+			// `git worktree list`) while still failing to delete the working
+			// directory when a nested file is undeletable (#6730). Trusting
+			// `stillRegistered` alone then leaves an orphaned folder on disk
+			// and archives the row over it. Mirror the repo-missing branch
+			// above: when git no longer tracks it but the directory survives,
+			// remove it directly under the same root guard.
+			if (existsSync(local.worktreePath)) {
+				const worktreeBaseDir =
+					project.worktreeBaseDir ?? getHostWorktreeBaseDir(ctx);
+				if (
+					!isInsideProjectWorktreesRoot(
+						local.worktreePath,
+						project.id,
+						worktreeBaseDir,
+					)
+				) {
+					warnings.push(
+						`Skipped orphaned worktree removal at ${local.worktreePath}: folder is outside the managed worktrees root`,
+					);
+				} else {
+					try {
+						await rm(local.worktreePath, { recursive: true, force: true });
+					} catch (err) {
+						const message =
+							err instanceof Error ? err.message : String(err);
+						throw new TRPCError({
+							code: "INTERNAL_SERVER_ERROR",
+							message: `Failed to remove orphaned worktree at ${local.worktreePath}: ${message}`,
+						});
+					}
+				}
+			}
 			worktreeRemoved = true;
 		}
 	}
--- a/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
@@ -567,9 +567,9 @@ async function runDestroyPhases(
 							message: `Failed to remove orphaned worktree at ${local.worktreePath}: ${message}`,
 						});
 					}
+					worktreeRemoved = true;
 				}
 			}
-			worktreeRemoved = true;
 		}
 	}
 
--- a/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
@@ -561,8 +561,7 @@ async function runDestroyPhases(
 					try {
 						await rm(local.worktreePath, { recursive: true, force: true });
 					} catch (err) {
-						const message =
-							err instanceof Error ? err.message : String(err);
+						const message = err instanceof Error ? err.message : String(err);
 						throw new TRPCError({
 							code: "INTERNAL_SERVER_ERROR",
 							message: `Failed to remove orphaned worktree at ${local.worktreePath}: ${message}`,
--- a/packages/host-service/src/workers/tasks/git.ts
+++ b/packages/host-service/src/workers/tasks/git.ts
@@ -209,10 +209,25 @@ export const gitWorktreeRemoveTask = defineWorkerTask<
 		// command's locale- and version-dependent exit text. `--force --force`
 		// also unregisters a worktree whose directory is already gone, so no
 		// separate prune (which would clobber other stale worktrees' metadata)
-		// is needed.
+		// is needed. Log the real git error though — `--force --force` can
+		// unregister while still failing to delete the folder (undeletable
+		// nested file, #6730), and without this the underlying cause is
+		// invisible to anyone debugging an orphan.
 		await git
 			.raw(["worktree", "remove", "--force", "--force", target])
-			.catch(() => {});
+			.catch((err) => {
+				const message =
+					typeof err === "object" && err !== null && "message" in err
+						? String((err as { message: unknown }).message)
+						: String(err);
+				console.warn(
+					`[host-service] git worktree remove --force failed but may still unregister`,
+					{
+						error: message,
+						target,
+					},
+				);
+			});
 		// A `worktree list` failure throws out of the task: the post-remove
 		// state is unknown and the caller must not treat it as removed.
 		const raw = await git.raw(["worktree", "list", "--porcelain"]);
--- a/packages/host-service/src/workers/tasks/git.ts
+++ b/packages/host-service/src/workers/tasks/git.ts
@@ -218,11 +218,26 @@ export const gitWorktreeRemoveTask = defineWorkerTask<
 		// command's locale- and version-dependent exit text. `--force --force`
 		// also unregisters a worktree whose directory is already gone, so no
 		// separate prune (which would clobber other stale worktrees' metadata)
-		// is needed.
+		// is needed. Log the real git error though — `--force --force` can
+		// unregister while still failing to delete the folder (undeletable
+		// nested file, #6730), and without this the underlying cause is
+		// invisible to anyone debugging an orphan.
 		reportPhase?.("worktree-remove");
 		await git
 			.raw(["worktree", "remove", "--force", "--force", target])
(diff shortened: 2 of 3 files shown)


andyst-dev:
Addressed the reviewer's point (Minor/Functional Correctness) on this PR.

**Keep `worktreeRemoved` false when the root guard skips removal.** Previously, when the orphaned-worktree folder sat outside the managed worktrees root, the root guard pushed a skip warning **and** `worktreeRemoved` was still set to `true` right after — so the destroy response claimed a folder was removed that was actually kept on disk. The flag now lives inside the removal branch only: it's `true` iff the `rm` actually ran.

Added a regression test (`orphaned worktree outside the managed root is skipped with worktreeRemoved=false`) that seeds an orphaned worktree outside `<base>/<projectId>/`, and asserts: the folder survives (`existsSync === true`), `worktreeRemoved === false`, and the skip is surfaced as a `Skipped orphaned...` warning rather than silently reported as removed.

Validation: `bun test test/workspace-cleanup.test.ts` → **28 pass, 0 fail** (the existing outside-root/inside-root cases plus the new regression).

Kitenite:
Thanks for this, and you filed the same root cause first. #6785 carries the same fix with git's error surfaced in the thrown message and `worktreeRemoved` derived from the final disk state, and I folded your server-side `console.warn` into it. It's approved now.

One note on the current diff here, for the record: the third commit moves `worktreeRemoved = true` inside the `rm` branch, so on the ordinary path where git deletes everything it stays `false`, and the integration suite asserts `true` there.

Closing as superseded by #6785. #7094 is the one I'd like to see rebased once that merges.

PR #6792, opened 2026-08-23
test(desktop): pin that a reclaimed auth-token lock cannot crash the write

saddlepaddle:
#6790 fixed a fatal, unhandled crash on the encrypted sign-in token write path, but shipped only a negative test for the lock it *fails* to acquire. The path it actually fixed -- losing a lock we already hold -- had no coverage at all.

### The regression this catches

Deleting the `onCompromised` handler from `withAuthLock` reintroduces the crash, and every one of the 18 existing tests stays green. `proper-lockfile`'s default handler is `(err) => { throw err }`, thrown from inside its own refresh timer where no caller can catch it -- which is why those Sentry reports carry no stack.

The test also covers the second half of #6790: replacing the best-effort `await release().catch(...)` with a plain `await release()` makes the same run reject with `ERELEASED`, reporting a token that *was* written as a sign-in that failed to save.

### How the condition is provoked

`proper-lockfile` arms a refresh for the lock it hands back, on a timer derived from `stale` -- `stale / 2`, so five seconds at the current 10s setting. If the entry is gone when that refresh stats it, the library marks the lock compromised. Waiting five seconds in a file that runs in under a second is not worth it, so the test intercepts that single scheduling call: it removes the lock entry the way a second desktop process reclaiming it would, then runs the library's own refresh callback directly.

The library is not mocked. Its `stat`, its ENOENT branch, `setLockAsCompromised`, the `onCompromised` dispatch and the resulting `ERELEASED` on release are all real; only the five-second delay is skipped. The test asserts it actually drove that refresh (`refreshRan`), so it cannot silently degrade into decoration if the library stops scheduling one.

It deliberately does *not* assert that a compromised lock makes the write succeed. It asserts that the write's own outcome survives: this write landed, so it is reported as landed and readable. The existing negative test -- a write that never happened still reports failure -- is untouched.

### Verification

`bunx biome check apps/desktop/src/lib/trpc/routers/auth/utils/auth-functions.test.ts`

```
Checked 1 file in 13ms. No fixes applied.
```

`cd apps/desktop && bun run typecheck` -- clean, no pre-existing errors in this worktree:

```
$ bun run generate:icons && bun run generate:routes
$ bun run scripts/generate-file-icons.ts
Generated file icons: 1067 SVGs copied, 2046 file names, 1152 extensions, 4528 folder names
$ tsr generate
$ tsc --noEmit
```

`bun test apps/desktop/src/lib/trpc/routers/auth/` -- 18 pass before, 19 pass after, same wall-clock:

```
 19 pass
 0 fail
 44 expect() calls
Ran 19 tests across 1 file. [904.00ms]
```

Full desktop suite, `cd apps/desktop && bun test`: `2799 pass, 0 fail` across 302 files.

**Proof it bites.** With the `onCompromised` line deleted from `withAuthLock`:

```
error: ENOENT: no such file or directory, stat '/var/folders/.../auth-functions-test-HzBGf8/auth-token.enc.lock'
    path: ".../auth-token.enc.lock",
 syscall: "stat",
   errno: -2,
    code: "ECOMPROMISED"

(fail) auth token storage > survives another writer reclaiming the write lock mid-write [0.98ms]

 18 pass
 1 fail
Ran 19 tests across 1 file. [900.00ms]
```

That is the uncatchable throw itself failing the test -- Bun's runner attributes an unhandled error in a timer to the running test; a `process.on("uncaughtException")` listener never sees it. Note the other 18 tests pass without the fix, which is the gap.

With `onCompromised` restored but `release().catch(...)` reduced to `await release()`:

```
error: Lock is already released
 code: "ERELEASED"
      at withAuthLock (apps/desktop/src/lib/trpc/routers/auth/utils/auth-functions.ts:68:9)

(fail) auth token storage > survives another writer reclaiming the write lock mid-write [2.12ms]

 18 pass
 1 fail
```

Restored, both lines present:

```
 19 pass
 0 fail
Ran 19 tests across 1 file. [904.00ms]
```

Run 35 consecutive times with the fix in place: 35/35 green. The ordering is structural rather than timed -- the compromise's `stat` is issued before the write's first filesystem call, so the loss is always recorded before `saveToken` resolves. No sleeps, no fixed waits, no timing-dependent assertions.

`auth-functions.ts` is unchanged; the diff is the test file only.

Refs DESKTOP-11M

https://claude.ai/code/session_01H1nPckannUVVytgNdnd3YG

---
## Summary by cubic
Ensures reclaiming the auth-token write lock mid-write cannot crash or misreport the write. Previously, a compromised lock from `proper-lockfile` could throw from its timer or yield `ERELEASED`; this pins the intended outcome.

- Runs the real `proper-lockfile` refresh immediately by intercepting its single timer, deleting the lock entry, and invoking the library callback; asserts the refresh ran. No mocks; only the 5s delay is skipped.
- Waits for the compromise to be reported by the library’s stat callback before asserting to remove scheduling races; a missing report times out.
- Verifies the write result stands: `saveToken` resolves and `loadToken` returns the token; lock loss only logs.
- Fails on regressions: removing `onCompromised` in `withAuthLock` (uncatchable throw) or replacing `release().catch(...)` with `await release()` (`ERELEASED`).

Refs DESKTOP-11M

<sup>Written for commit 79d435a87c436a897132902f15a01fa4947ec69b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6792?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for token storage when another process reclaims the write lock during saving.
  * Verified token writes complete successfully, remain readable, and handle lock-change warnings without interrupting the result.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/auth/utils/auth-functions.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/auth/utils/auth-functions.test.ts
@@ -275,14 +275,25 @@ describe("auth token storage", () => {
 			return realSetTimeout(() => undefined, 0);
 		}) as typeof globalThis.setTimeout;
 		let recordedLosses = 0;
+		let recordLoss: () => void = () => undefined;
+		const lossRecorded = new Promise<void>((resolve) => {
+			recordLoss = resolve;
+		});
 		const warnSpy = spyOn(console, "warn").mockImplementation(() => {
 			recordedLosses += 1;
+			recordLoss();
 		});
 
 		try {
 			// Rethrowing from the refresh would fail this test outright, and a
 			// release that overrode the operation's result would reject here.
 			await saveToken({ token: "token", expiresAt: "2099-01-01" });
+			// The library reports the loss from its own stat callback, which can
+			// land after the write resolves. Wait for it rather than racing it:
+			// a count read straight after the write would pass or fail on
+			// scheduling. If the loss is never reported this hangs to the test
+			// timeout, which is the failure we want.
+			await lossRecorded;
 		} finally {
 			globalThis.setTimeout = realSetTimeout;
 			warnSpy.mockRestore();
--- a/apps/desktop/src/lib/trpc/routers/auth/utils/auth-functions.test.ts
+++ b/apps/desktop/src/lib/trpc/routers/auth/utils/auth-functions.test.ts
@@ -245,6 +245,60 @@ describe("auth token storage", () => {
 			fs.rmdirSync(lockEntry);
 		}
 	});
+
+	test("survives another writer reclaiming the write lock mid-write", async () => {
+		// proper-lockfile keeps the entry it handed us fresh on a timer (stale / 2,
+		// so five seconds out) and calls onCompromised once that refresh finds the
+		// entry gone. Its default handler rethrows from inside the timer, where no
+		// caller can catch it. Run the library's own refresh as soon as the entry
+		// is gone rather than waiting five seconds for it: everything that detects
+		// and reports the loss is the real thing, only the delay is skipped.
+		const lockEntry = `${tokenFile}.lock`;
+		const realSetTimeout = globalThis.setTimeout;
+		let refreshRan = false;
+		globalThis.setTimeout = ((
+			handler: () => void,
+			delay?: number,
+			...args: unknown[]
+		) => {
+			if (typeof delay !== "number" || delay < 1_000) {
+				return realSetTimeout(handler, delay, ...args);
+			}
+			// The lock refresh, armed the moment the write took the lock. Take the
+			// entry away the way a second desktop process reclaiming it would.
+			globalThis.setTimeout = realSetTimeout;
+			refreshRan = true;
+			fs.rmdirSync(lockEntry);
+			handler();
+			// The refresh has already run; hand the library a timer it can unref
+			// and clear like the one it asked for.
+			return realSetTimeout(() => undefined, 0);
+		}) as typeof globalThis.setTimeout;
+		let recordedLosses = 0;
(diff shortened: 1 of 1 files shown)


saddlepaddle:
Reviewed both halves of the CodeRabbit finding against the code rather than the badge.

**Valid, and fixed in 79d435a:** the compromise is reported from proper-lockfile's own `stat` callback, which can land after `saveToken` resolves, so reading `recordedLosses` immediately after the write made that one assertion pass or fail on scheduling. It now awaits the report. A loss that is never reported hangs to the test timeout — the failure we want — and the regression case still fails fast (0.77ms) with the library's `ECOMPROMISED` throw rather than hanging.

**Declined, with reason:** the suggested write-start barrier, so the refresh fires mid-write rather than at acquisition. The refresh timer is armed inside `lock()`, so the interception already happens during `saveToken`, after the lock is taken — not "before saveToken starts". More importantly the defect being pinned is *`onCompromised` throwing from a timer no caller can catch*, and that mechanism is identical whether the loss lands at acquisition or mid-write. Verified empirically in both directions: with `onCompromised` present 19/19 pass; with only that line removed, 18 pass / 1 fail with the exact production signature (`ENOENT ... stat '<...>/auth-token.enc.lock'`, `code: "ECOMPROMISED"`). A barrier would add real complexity to the test without changing what it catches.

PR #6810, opened 2026-08-24
fix(desktop): show an error state instead of a blank window when organization.list fails (#6794)

### LINKED ISSUE #6794: Blank window when organization.list fails while auth is healthy
## Context

When `organization.list` fails but the auth endpoints are healthy, a returning window renders nothing at all — no message, no retry, indistinguishable from a hang. `CollectionsProvider` wraps the whole authenticated tree and returns `null` until it can resolve the window's org, and it deliberately waits for the membership list before deciding. That wait is correct; the blank screen it produces during an outage is not.

This is not the offline case. An offline launch is already handled: the session is in-memory only, so `get-session` fails and `_authenticated/layout.tsx:244` shows "You're offline" with a Retry. The bug needs `/api/auth/*` to succeed while `/api/trpc` does not.

## Steps

1. Open the app once and let a window persist its org to the window registry.
2. Make `/api/trpc` fail while `/api/auth/*` keeps serving (bad tRPC deploy, rate limit, or a DB fault in the `organizations ⋈ members` join — better-auth's `cookieCache` at `packages/auth/src/server.ts:135-138` lets sessions answer without touching those tables).
3. Relaunch.

**Expected:** an error state explaining the app can't reach the server, with a way to retry.
**Actual:** an empty window.

### Files

- `apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/CollectionsProvider.tsx:112` — `if (windowOrgId != null && organizations == null) return;` blocks initialization. The comment above it reasons "the window is showing nothing yet either way", which holds for a few hundred milliseconds and not for an outage.
- `.../CollectionsProvider.tsx:206` — `if (!contextValue) return null;` blanks the entire authenticated subtree (`_authenticated/layout.tsx:289-336`), including `<Outlet />`, the sidebar and the window-drag regions.
- `apps/desktop/src/renderer/providers/ElectronTRPCProvider/ElectronTRPCProvider.tsx:48` — `hostServiceQueryRetry` governs cloud queries too, despite the name. It only retries when `error.data == null`, so a **500 through tRPC gets zero retries**; only transport failures retry (3x, ~4.2s total).

### Approach

Keep the guard exactly as it is — it exists so a dead registry org can't pin the window to an org whose every read 403s, and that invariant is worth more than the blank screen costs. Instead, distinguish "still loading" from "gave up" at the *render* site: when the org is unresolved **and** `organization.list` is in error, render an error state with a Retry wired to the query's `refetch()` rather than `null`.

`_authenticated/layout.tsx:204-259` already has both panels to copy — the offline card and the 15s "Still restoring your session" escape hatch built on `useDelayElapsed` (`:99-102`). Neither is applied to org resolution today.

### Gotchas

- **Do not fall back to the registry org on error.** `initializedRef` (`:98`/`:116`) is one-shot: adopting an unverified org id on a transient failure means that when the query later succeeds and reveals the id is dead, nothing re-evaluates. That pins the window permanently and is strictly worse than the blank, which does recover.
- The blank is not permanent today. `refetchOnWindowFocus`/`refetchOnReconnect` are at their defaults and `organizations` is in the effect's dependency list, so it self-heals on focus or reconnect. The harm is the silence, not the persistence.
- There are no component tests for `CollectionsProvider`, and the app has exactly one React component test overall (`components/Redirect/Redirect.test.tsx`). Extracting the resolution into a pure function next to `evictInactiveOrgs.ts` would match the five existing co-located unit tests and let the decision table be tested without a DOM.
- Introduced by #5337 (`b9d851f93`), which added per-window org context.

Kitenite:
## Problem

A returning window (one with an org already in the window registry) rendered an empty, undraggable frame when `organization.list` failed but `/api/auth/*` still answered. `CollectionsProvider` refuses to adopt a registry org it can't verify against the membership list (correct), but its render site treated "still loading" and "load failed" identically and returned `null`, unmounting the whole authenticated tree. tRPC-shaped 500s get zero retries, so an API outage looked like a hang.

## Fix

- `resolveWindowOrg.ts`: the wait / unresolvable / resolved decision extracted as a pure function. The one-shot `initializedRef` semantics and the "never adopt an unverified registry org" invariant are unchanged: `unresolvable` is rendered, never adopted.
- `components/OrgResolutionScreen`: replaces the bare `null`. Spinner with a drag strip while loading; once the list has errored, or has been pending for 15s, shows a message and a Retry wired to the query's `refetch()`. Mirrors the blocking screens in `_authenticated/layout.tsx`.
- `resolveWindowOrg.test.ts`: table test covering the five cases from the issue plus two edges.

## Verification

- `bun test` on the new test: 7 pass. With the pre-fix behaviour restored (errored → `wait`), the "unresolvable" case fails, so the test guards the regression.
- `bun typecheck` in `apps/desktop`: clean.
- Biome check on the changed folder: clean.
- No CDP run: this was done in a fresh worktree without a `.env`/dev stack, so I did not drive the real app to simulate a `/api/trpc` outage. Manual repro is in the issue (break `/api/trpc` only, relaunch within the 5-minute cookie-cache window).

Fixes #6794

https://claude.ai/code/session_011XQusXtWEwq2nxN4jfMUpv

---
## Summary by cubic
Show an error screen with Retry instead of a blank window when `organization.list` fails in the desktop app. Previously `CollectionsProvider` returned null on both load and error, blanking the frameless window; now a spinner is shown while loading and an error with Retry appears after a failure or 15s.

- Refactor: extracted org resolution into a pure `resolveWindowOrg` with three outcomes (wait, unresolvable, resolved); covered by a table test for core and edge cases.
- UI: added `OrgResolutionScreen` with spinner, drag strip, and Retry wired to `organization.list` refetch; used only when the window’s org is unresolved.
- Behavior: keep the authenticated tree mounted during org switches; never adopt a registry org unless verified by the membership list; fall back to the session org if the registry org is no longer a membership; render “unresolvable” on list failure rather than adopting.
- Review/QA: focus on the one-shot init semantics (unchanged), the resolution branching, and retry wiring; simulate a transient `organization.list` outage to see spinner → error → Retry recovery; new tests pass.

<sup>Written for commit 053e49c6823adbc6700ea66b524b778d61002696. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6810?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved organization selection when opening desktop windows.
  * Preserves valid organization context and falls back to the active session when needed.
  * Handles missing or failed organization data without leaving the app blank.

* **New Features**
  * Added a loading and error screen with retry support when organization resolution is delayed or unsuccessful.
  * Added coverage for organization resolution scenarios and fallback behavior.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/CollectionsProvider.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/providers/CollectionsProvider/CollectionsProvider.tsx
@@ -24,6 +24,8 @@ import {
 	getCollections,
 	preloadCollections,
 } from "./collections";
+import { OrgResolutionScreen } from "./components/OrgResolutionScreen";
+import { resolveWindowOrg } from "./resolveWindowOrg";
 
 // Cloud query procedures take no organizationId input (the server scopes by
 // active org), so their React Query keys don't encode the org — on org switch
@@ -86,8 +88,32 @@ export function CollectionsProvider({ children }: { children: ReactNode }) {
 
 	// Account-wide ("the orgs I belong to"), so it is not affected by — and does
 	// not depend on — the org header this provider sets.
-	const { data: organizations } =
-		cloudTrpc.organization.list.useQuery(undefined);
+	const {
+		data: organizations,
+		isError: organizationsErrored,
+		refetch: refetchOrganizations,
+	} = cloudTrpc.organization.list.useQuery(undefined);
+
+	// `unresolvable` is rendered below, never adopted: initialization is
+	// one-shot, so seeding from an unverified registry id on a transient failure
+	// would pin the window permanently. Waiting keeps the blank recoverable.
+	const resolution = useMemo(
+		() =>
+			resolveWindowOrg({
+				windowOrgPending,
+				windowOrgId,
+				organizations,
+				organizationsErrored,
+				sessionOrgId,
+			}),
+		[
+			windowOrgPending,
+			windowOrgId,
+			organizations,
+			organizationsErrored,
+			sessionOrgId,
+		],
+	);
 
 	// Initialize the window's org exactly once. After this, the window's org is
 	// owned by local state (and switchOrganization); later — possibly transient —
@@ -98,24 +124,10 @@ export function CollectionsProvider({ children }: { children: ReactNode }) {
 	const initializedRef = useRef(false);
 	useEffect(() => {
 		if (initializedRef.current) return;
-		if (windowOrgPending) return;
-		// The registry's org is only preferred while it is still one the user
-		// belongs to. Leaving an organization (or having membership revoked
-		// elsewhere) leaves a dead id in the registry, and adopting it would pin
-		// the window to an org whose every read now fails. Until the membership
-		// list has loaded we cannot tell stale from valid, so wait rather than
-		// guess — the window is showing nothing yet either way.
-		const registryOrgIsStillMine =
-			windowOrgId != null &&
-			organizations != null &&
-			organizations.some((organization) => organization.id === windowOrgId);
-		if (windowOrgId != null && organizations == null) return;
-		const resolved =
-			(registryOrgIsStillMine ? windowOrgId : sessionOrgId) ?? null;
-		if (!resolved) return;
+		if (resolution.kind !== "resolved") return;
 		initializedRef.current = true;
-		setActiveOrganizationId(resolved);
-	}, [windowOrgPending, windowOrgId, sessionOrgId, organizations]);
+		setActiveOrganizationId(resolution.organizationId);
+	}, [resolution]);
 
 	// Scope this window's cloud reads to its own org, during render rather than
 	// in an effect: children below issue their first queries while this render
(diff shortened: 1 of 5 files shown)


ashbrener (commented):
I filed #6794, so I went looking for the trap rather than the happy path. It isn't here — the invariant holds.

`resolveWindowOrg` can't return `resolved` while a registry org is present and the membership list isn't: line 29 returns `unresolvable`/`wait` before anything below it runs, and `registryOrgIsStillMine` needs `organizations != null` anyway. The effect then sets `initializedRef` only on `resolved`, so an unverified id is rendered and never adopted. That was the way to get this wrong — a transient failure pinning the window to a dead org permanently, with the one-shot ref meaning nothing re-evaluates when the list comes back.

`isError` is also the right signal rather than `status`/`failureCount`: TanStack holds `pending` through the retry sequence, so the three transport retries still show the spinner and only a settled failure reaches the error screen — which lines up with `hostServiceQueryRetry` giving tRPC-shaped 500s zero retries. And self-healing survives, since the provider stays mounted and focus/reconnect refetch still resolves.

Three things, none of them correctness:

1. **The tests cover the decision function, not the line that regressed.** Reverting the render site to `return null` reintroduces #6794 with all seven still green. Separately, "waits when nothing can supply an org" isn't reachable in the app — `layout.tsx:277` redirects before the provider mounts — while the case that *is* live and untested is `organizations` present with `isError: true`. That's precisely what a later "hoist the errored check to the top" edit would break, and it would blank a working window on a failed background refetch.

2. **The spinner renders unconditionally** (`OrgResolutionScreen.tsx:27`, outside the `showRetry` block), so "Can't reach the Superset server" sits under an animation while nothing is in flight and nothing will retry until focus or reconnect. The screen this mirrors doesn't do that — `layout.tsx:250` swaps in `HiOutlineWifi`.

3. **The body copy doesn't branch.** The heading does, so the 15s-pending case reads "Still loading your organizations" directly above "Superset couldn't load your organizations." Worth noting too that an errored query keeps `status: "error"` across `refetch()` — only `fetchStatus` flips — so clicking Retry changes nothing on screen. `isFetching` from the same destructure would fix the button state and the spinner gating together.

Drag strip is correctly outside the conditional, so the error window is still movable — easy one to miss on a frameless window.

PR #6825, opened 2026-08-24
fix(desktop): add hover marquee-scroll to v1 and v2 workspace sidebar names

AviPeltz:
## Summary
- The persistent workspace sidebar is resizable down to 225px (`MIN_WORKSPACE_SIDEBAR_WIDTH`), so long workspace names routinely truncate with no way to read them short of renaming or widening the sidebar.
- PR #6787 already solved this for the v2-workspaces list rows via `WorkspaceNameMarquee`. This PR promotes that component to a shared, generic location (`apps/desktop/src/renderer/components/WorkspaceNameMarquee`) and wires it into the other two sidebars that had the same gap:
  - v1 `WorkspaceSidebar` → `WorkspaceListItem`
  - v2 `DashboardSidebar` → `DashboardSidebarExpandedWorkspaceRow`
- Each row now drives the marquee's `forceActive` off its own keyboard-focus state (row `onFocus`/`onBlur`), so tabbing through the list also reveals a truncated name — not just hovering it, matching the correctness requirement CodeRabbit flagged on #6787.
- Slowed the scroll pace slightly (`PIXELS_PER_SECOND` 45 → 32) per manual testing feedback in the running app.
- No visual changes to any of the three call sites — same font size/weight/color as before, just truncate → marquee.

## Test plan
- [x] `bunx biome check` on all changed/added files — clean
- [x] `bun run --filter=@superset/desktop typecheck` — passes
- [x] Manually verified live in the dev app: resized the sidebar toward 225px with a long workspace name, confirmed hover reveals the full name via the scroll effect and resets on mouse-leave
- [ ] Re-verify tab-focus (keyboard-only) reveal on both v1 and v2 rows in the slowed-down build

https://claude.ai/code/session_01RGwoZn3Hn9ToYAfFoWRGsr

---
## Summary by cubic
Adds hover and keyboard-focus marquee scrolling for truncated workspace names in desktop v1 and v2 sidebars, and prevents click focus from triggering it. Previously names were cut off; now they scroll on hover or when rows have keyboard-originated focus, without auto-scrolling after click selection.

- Promotes `WorkspaceNameMarquee` to `apps/desktop/src/renderer/components/WorkspaceNameMarquee` and integrates it into v1 `WorkspaceSidebar` (`WorkspaceListItem`) and v2 `DashboardSidebar` (`DashboardSidebarExpandedWorkspaceRow`); `V2WorkspaceRow` now imports the shared component.
- Adds `useFocusVisible` to drive `forceActive` only from `:focus-visible` and keep it while focus stays within the row; avoids pointer-triggered focus starting the marquee.
- Reduces scroll speed to 32 px/s; no visual changes beyond truncate-to-marquee behavior.

<sup>Written for commit 98186ee32aba3c2a237c314e9e815fbf2225de6b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6825?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Workspace names now scroll horizontally when focused, making longer names easier to read in sidebar and dashboard workspace lists.
  * Scrolling is slower for improved readability.
  * Focus-based behavior responds appropriately to keyboard navigation while preserving selected and active workspace styling.
  * Focus remains active when moving between a workspace row and its child actions.


AviPeltz:
Follow-up fix pushed: clicking a sidebar row to select/open a workspace also focuses it (`mousedown` on a `tabIndex={0}` div focuses it in Chromium), which was tripping the same `forceActive` path built for keyboard Tab navigation. Since the sidebar stays mounted after navigating (unlike the v2-workspaces list, which you navigate away from), the just-selected row's name kept auto-scrolling indefinitely just from having been clicked, not from being hovered.

Added `useFocusVisible` (`apps/desktop/src/renderer/hooks/useFocusVisible/`), which reads the native `:focus-visible` state at focus time instead of reacting to any focus/blur — browsers already exclude pointer-triggered focus from `:focus-visible`, so a mouse click no longer activates the marquee, while Tab navigation still does. It also stays active while focus moves between focusable elements within the same row (e.g. tabbing onto the "Close workspace" button) and only clears once focus actually leaves the row.

Wired into all three rows, including `V2WorkspaceRow` (the already-merged v2-workspaces list row from #6787), which had the same latent issue — just less visible there since that list unmounts on navigation.

PR #6829, opened 2026-08-24
fix: page cmd+click should open page pane, also responsiveness of the…

harshithmullapudi:
## What & why

Two unrelated fixes in the v2 workspace, both from dogfooding the Pages feature.

**1. Cmd+clicking a page URL opened the browser pane instead of the page pane.**

When an agent publishes a page it prints `<web>/page/<slug>` into the terminal. Cmd+click resolved to the `"pane"` action, which routed unconditionally through `openUrlInV2Workspace` and rendered the page as a generic web page in a `BrowserPane` — bypassing the `PagePane` viewer we built for exactly this, along with its comments, visibility menu, and handoff menu.

Terminal URL clicks now check for a page URL first: same origin as `NEXT_PUBLIC_WEB_URL`, path exactly `/page/<slug>`. Anything else — including a `/page/` path on another host — still opens the browser pane, and Cmd+Shift still opens the real browser.

Rather than pre-resolve the slug over the network at click time, `PagePane` now resolves it itself: `page.pull` already accepts `{ slug }` as well as `{ id }`, so the pane opens synchronously and the round trip happens behind the spinner it already renders. Once resolved, the pane backfills `pageId` and `title` into its own pane data, so the tab stops showing the raw slug and later opens dedupe on the real id. `PagePaneData.pageId`/`title` are now optional to allow that.

One related fix while in there: `PagePane` threw `notFound()` on a missing page, but panes have no error boundary, so that escaped to the router and unmounted the entire workspace route. It now renders a message inside the pane.

**2. The sidebar tab strip broke between breakpoints once Pages was added.**

`WorkspaceSidebar` collapses its tabs to icon-only below a hardcoded 260px. That was tuned for three tabs. With `FEATURE_FLAGS.PAGES` on there are four, which need roughly 320px labelled and ~370 once the Changes/Review badges render — so between ~280 and ~350 the strip overflowed a container with `overflow-hidden` and no `truncate`, clipping the last tab mid-label or hiding it entirely.

The threshold now scales with tab count (88px per labelled tab, 20px hysteresis) instead of being a constant:

| Tabs | Collapse below | Expand at |
|---|---|---|
| 3 (Pages off) | 264 | 284 |
| 4 (Pages on) | 352 | 372 |

That leaves the three-tab case essentially where it was, so turning the flag off doesn't start dropping labels at widths where they still fit. Labels also `truncate` now, and the buttons are allowed to shrink (the style helper's `shrink-0` was fighting the call site's `flex-1`), so any width that still overflows ellipsises rather than clipping.

## How I tested it

Automated only — see the caveat below.

- `bun run typecheck` (apps/desktop) — passes.
- `bun run lint:fix` — clean, no fixes applied.
- `bun test src/renderer` — 1585 pass. One failure, `terminal-ws-transport > forces a reconnect on the wall-clock gap after sleep/wake`, reproduces on a clean `main` (verified via `git stash`) and is unrelated to this change.
- New unit tests: `parseSupersetPageUrl` (11 cases — origin mismatch, scheme mismatch, trailing slashes, query/hash, nested paths, missing slug, percent-encoding, non-URL text) and `openPagePaneInStore` (6 cases — add vs focus, dedupe by id, by slug, and across the mixed id/slug-only combinations).

**Not verified end-to-end in the running app, and no screenshots.** I did not drive the dev build over CDP to confirm the real Cmd+click journey, and the sidebar change is a layout change I reasoned about from the Tailwind classes rather than looked at — the 88px per-tab figure is derived, not measured. The sidebar strip is worth a look while dragging across 340–380px before this lands.

**Known gap, deliberately not fixed here:** the active tab's `actions` buttons sit beside the tab strip but aren't counted in the width measurement, so a tab with two or three actions gets ~70–100px less room than the threshold assumes. Fixing that means observing the tab strip element rather than the sidebar container — small, but it shifts behavior for every tab, so it didn't belong in this PR.

## Checklist

---
## Summary by cubic
Cmd/Ctrl+clicking a Superset page URL in Terminal now opens or focuses a Page pane instead of the generic URL viewer, and the dashboard/workspace sidebars are more compact and responsive to prevent label clipping. This also adds a Pages dashboard/detail flow with thumbnails, pins, and improved sharing/comments.

- New Features
  - Detects `/page/<slug>` links on `NEXT_PUBLIC_WEB_URL` when `FEATURE_FLAGS.PAGES` is on and opens a Page pane; other links still use the in‑app URL viewer; Cmd+Shift continues to open the system browser.
  - Refactors page viewing into a shared `PageViewer` that resolves by slug or id, handles missing pages inline, and dedupes tabs via `openPagePaneInStore`.
  - Adds `/pages` and `/pages/$slug` with search, scope tabs (All, Pinned, Team, Just me), thumbnails, and pins; clicks open inline, Cmd/Ctrl+click opens in a pane; queues pane intents and restores the last active V2 workspace.
  - Reworks sidebar tabs and rows to truncate text, keep icons from shrinking, and use consistent compact heights; removes the in‑workspace Pages tab and adds a Pages section in the dashboard sidebar.
  - Updates comments to carry `offsetX/offsetY`; aligns `@superset/trpc` page-comment schema; adds a `PageHeaderBar` on web with visibility, versions, delete, and spinner loading; CLI tables hyperlink long URLs via OSC 8.

- Migration
  - `superset pages publish` now requires an anchor: pass `--page <id>` to version an existing page or `--workspace <id|name>` with an entry path; external files are keyed under `/external/<basename>` to keep stable ids.

<sup>Written for commit 6111a8e6be347aa5cd4ee13c4dcdf82c0d227f20. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a Pages dashboard with search, scope filters, thumbnails, favorites, pinning, and page actions.
  * Pages can open from the dashboard, links, or workspace panes, reusing existing panes when possible.
  * Added page sharing, visibility, version selection, deletion, and enhanced commenting controls.
  * Publishing now supports explicit workspace selection and external page paths.
* **Improvements**
  * Sidebar layouts are more compact and responsive.
  * Page loading now shows clearer loading and error states.
  * Terminal tables support clickable hyperlinks.
* **Bug Fixes**
  * Comment pins better reflect click locations and avoid overlap.
* **Tests**
  * Expanded coverage across Pages, comments, thumbnails, navigation, and publishing.


apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/components/PagesView/hooks/usePageFavorites/usePageFavorites.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export interface PageFavoritesApi {
+	favoritePageIds: string[];
+	favoritePageIdSet: ReadonlySet<string>;
+	isFavorite: (pageId: string) => boolean;
+	setFavorite: (pageId: string, favorite: boolean) => void;
+	toggleFavorite: (pageId: string) => void;
+}
+
+export function usePageFavorites(): PageFavoritesApi {
```
saddlepaddle, line 19: Should we move this to the API? Otherwise it won't be available in web / mobile ig

apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/components/PagesGrid/components/PageCard/PageCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+import {
+	DropdownMenu,
+	DropdownMenuContent,
+	DropdownMenuItem,
+	DropdownMenuTrigger,
+} from "@superset/ui/dropdown-menu";
+import { toast } from "@superset/ui/sonner";
+import { cn } from "@superset/ui/utils";
+import { formatDistanceToNowStrict } from "date-fns";
+import { Globe, Link2, Lock, MoreVertical, Pin, PinOff } from "lucide-react";
```
saddlepaddle, line 11: nit: `MoreVertical` is lucide's deprecated alias for `EllipsisVertical` — same glyph, but the app's other overflow triggers (settings members/invitations action menus) all import `EllipsisVertical`. Worth matching here (import + usage on line 95).

apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/components/DashboardSidebarPagesSection/DashboardSidebarPagesSection.tsx
```diff
@@ -0,0 +1,102 @@
...
+	useOpenPage,
+} from "renderer/routes/_authenticated/_dashboard/hooks/useOpenPage";
+import { useSidebarSectionsCollapseStore } from "renderer/stores/sidebar-sections-collapse";
+import { DashboardSidebarSectionHeader } from "../DashboardSidebarSectionHeader";
+
+interface DashboardSidebarPagesSectionProps {
+	isCollapsed?: boolean;
+}
+
+export function DashboardSidebarPagesSection({
```
saddlepaddle: non-blocking: I don't think pinned pages need a spot on the dashboard sidebar — pages feel more ephemeral than the other things that live there, and the Pinned tab in the grid already covers re-finding them. It does match what similar products do, so not asking for a change — just noting I'd lean toward dropping this section.

apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/utils/filterPages/filterPages.ts
```diff
@@ -0,0 +1,69 @@
...
+		case "team":
+			return page.visibility === "org";
+		case "mine":
+			return page.visibility === "just_me";
+		default:
+			return true;
+	}
+}
+
+export function filterPages<T extends FilterablePage>(
```
saddlepaddle, line 42: optional, non-blocking: search and scope filtering probably belong on the server eventually. `page.list` takes only an optional `workspaceId` — no search, no scope, no pagination — so the client pulls every page in the org and narrows it in memory. Fine at current volumes, and not worth doing now.

One wrinkle if anyone picks this up later: `pinned` cannot move server-side as-is, since pins live in client-only `v2UserPreferences` (localStorage). `team`/`mine`/search could move without that; pinned would need pins persisted server-side first.

apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/usePaneRegistry.tsx
```diff
@@ -587,9 +589,32 @@ export function usePaneRegistry({
 				? {
 						page: {
 							getIcon: () => <FileText className="size-3.5" />,
-							getTitle: (pane) => (pane.data as PagePaneData).title,
+							getTitle: (pane) => {
+								const data = pane.data as PagePaneData;
+								return data.title ?? data.slug;
```
saddlepaddle: nit: should maybe render nothing if no title? or just "Untitled Page"?

apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/components/PagesGrid/components/PageCard/PageCard.tsx
```diff
@@ -0,0 +1,115 @@
+import { Button } from "@superset/ui/button";
```
saddlepaddle, line 1: File placement: a number of the new directories do not follow the co-location rules in AGENTS.md ("used once -> nest under the parent's `components/`; used 2+ times -> promote to the highest shared parent"). Collecting them in one comment so this can be fixed in a single pass. Importer counts below are from grepping the whole tree; each move is a directory rename plus import updates.

**Used once, but sitting beside their only consumer**

1. `pages/components/PageCard/` — only importer is `PagesGrid` -> move to `pages/components/PagesGrid/components/PageCard/`
2. `pages/components/PageThumbnail/` — the component is only used by `PageCard` -> move to `.../PagesGrid/components/PageCard/components/PageThumbnail/`. Note `PagesGrid` imports `THUMBNAIL_ASPECT_RATIO` from it for the skeleton; extract that constant to `PagesGrid/constants.ts` so the component can nest.
3. `pages/utils/pageThumbnailCache/` — only importer is `PageThumbnail` -> move to `PageThumbnail/utils/pageThumbnailCache/`
4. `_dashboard/hooks/useActivePageSlug/` — only importer is `DashboardSidebarPagesSection` -> move to `DashboardSidebar/components/DashboardSidebarPagesSection/hooks/useActivePageSlug/`
5. `v2-workspace/$workspaceId/utils/parseSupersetPageUrl/` — only importer is `TerminalPane` -> move to `.../components/TerminalPane/utils/parseSupersetPageUrl/`

**Nested under a parent that does not use them**

6. `PagePane/components/PagePaneTitle/` and `PagePane/components/PagePaneHeaderExtras/` — neither is imported by `PagePane`. Both are used by `usePaneRegistry.tsx`, which currently reaches past `PagePane` to get them (usePaneRegistry.tsx:68-69). Move both up to `usePaneRegistry/components/`, as siblings of `PagePane`.

**Promoted above their actual shared parent**

7. `renderer/hooks/usePageFavorites/` — both importers (`DashboardSidebarPagesSection`, `pages/page.tsx`) are under `_dashboard/` -> move to `_dashboard/hooks/usePageFavorites/`. AGENTS.md treats the global location as a last resort.
8. `PageViewer/utils/toThreads/` — also imported by `_dashboard/hooks/usePageHeaderData`, so an outside module reaches into `PageViewer` internals -> promote to `_dashboard/utils/toThreads/`.
9. `packages/ui/.../PageComments/utils/pinLayout/` — all three importers (`PageCommentsView`, `CommentPopover`, `CommentBubble`) live under `PageCommentsView/` -> move to `PageCommentsView/utils/pinLayout/`.

For contrast, these are already placed correctly and should not move: `PageViewer`, `useOpenPage`, `usePageHeaderData`, `openPagePaneInStore`, `filterPages`, `DashboardSidebarPagesSection`, `PageViewerMessage`, `CommentModeButton`, `CommentBubble`, `relativeTime`, `sharedVersion`, and web's `PageHeaderBar`.

apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/$slug/page.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { Spinner } from "@superset/ui/spinner";
+import { createFileRoute, useNavigate } from "@tanstack/react-router";
+import { ArrowLeft } from "lucide-react";
+import { useFeatureFlagEnabled } from "posthog-js/react";
+import { useState } from "react";
+import { Redirect } from "renderer/components/Redirect";
+import { PageViewer } from "renderer/routes/_authenticated/_dashboard/components/PageViewer";
+import { usePageHeaderData } from "renderer/routes/_authenticated/_dashboard/hooks/usePageHeaderData";
+
+export const Route = createFileRoute("/_authenticated/_dashboard/pages/$slug/")(
```
saddlepaddle, line 7: Route param naming: `$slug` is the only `$slug` route in the desktop app. Every other dynamic route is keyed on the id — `$taskId`, `$automationId`, `$projectId`, `$workspaceId`, `$agentId`, `$hostId`, `$teamId` — with natural keys (`$pluginName`, `$prNumber`) only where there is no id.

Tasks is the precedent worth copying: the route is `tasks/$taskId`, and the resolver `task.byIdOrSlug` accepts either — it UUID-tests the param and falls back to a slug lookup (task.ts:450). So the param is named for the id, callers that have an id pass one, and a slug still works. `page.pull` already takes `{ id }` or `{ slug }`, so most of that is in place here.

Concretely I would rename this route to `pages/$pageId` and have it resolve either form. That matters beyond naming: every in-app caller already holds the id — `page.list` returns it and `useOpenPage` receives `{ id?, slug, title? }` — and then discards it to navigate by slug, so the id gets re-derived over the network. The only entry point that genuinely starts without an id is a Cmd+click on a page URL in a terminal.

Not asking for it in this PR. Two notes for whenever it happens: `useActivePageSlug` would become an id comparison, which is more robust than matching slugs across panes; and `PagePane` should keep tolerating a slug, since resolving before opening would reintroduce exactly the click-time round trip this PR deliberately avoids.

apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/components/PageThumbnail/PageThumbnail.tsx
```diff
@@ -0,0 +1,107 @@
...
+					cache: "force-cache",
+				});
+				if (!response.ok) {
+					throw new Error(`Preview failed to load (${response.status})`);
+				}
+				return response.text();
+			}),
+	});
+
+	const isLoading = isVisible && (pull.isPending || thumbnail.isPending);
```
saddlepaddle: Bug: a failed `page.pull` leaves this card spinning forever instead of falling back to the `FileText` placeholder.

When `pull` errors, `downloadUrl` stays undefined, so the `thumbnail` query never enables. In react-query v5 (5.101.1 here) a disabled query that has never fetched sits at `status: "pending"`, so `thumbnail.isPending` is permanently `true` — and `isLoading` on this line stays `true` for the life of the card. `pull.isError` is never consulted, so there is no path to the placeholder.

Reproduced while reviewing: in a dev environment whose page rows point at blobs from a different blob store, `page.pull` throws `NOT_FOUND` ("Page content is not available") and every card in the grid spins indefinitely. A page published into that same environment renders fine, so the two states are indistinguishable in the UI — a permanently broken page looks exactly like a slow one.

The same `version > 0` guard has a second instance of this: a page row with no versions also leaves the thumbnail query disabled and the spinner running.

`PageViewer` already gets this right — it checks `pull.error` and renders `PageViewerMessage` (PageViewer.tsx:97). Something equivalent here, or just gating on whether the thumbnail query is actually enabled:

```ts
const thumbnailEnabled = Boolean(downloadUrl) && version > 0;
const isLoading =
    isVisible && (pull.isPending || (thumbnailEnabled && thumbnail.isPending));
```

which covers the error case and the no-versions case together.

packages/ui/src/components/PageComments/components/PageHeader/utils/relativeTime/relativeTime.ts
```diff
@@ -0,0 +1,7 @@
+import { formatDistanceToNowStrict } from "date-fns";
+
+export function relativeTime(value: Date | string): string {
```
saddlepaddle: Half of this wrapper is redundant, the other half is load-bearing — worth splitting the difference rather than keeping or deleting it wholesale.

Redundant: the `value instanceof Date ? value : new Date(value)` normalization. date-fns v4 accepts `Date | number | string` directly — I checked against the pinned 4.1.0, and `formatDistanceToNowStrict("2026-08-01T00:00:00Z", { addSuffix: true })` returns "24 days ago" with no coercion needed.

Load-bearing: the invalid-date guard. Raw `formatDistanceToNowStrict(new Date("garbage"))` throws `RangeError: Invalid time value` rather than returning anything. Given this PR already fixes a case where a thrown error escaped a pane and unmounted the workspace route (the `notFound()` change in `PagePane`), a header that throws on one malformed timestamp is the same failure shape. Keep the guard.

Related inconsistency while here: `CommentPopover.tsx:172` formats timestamps with `formatDistanceToNowStrict(comment.createdAt, { addSuffix: true })` inline — same package, same component tree, same options, no guard. Its `createdAt` is a `number` so it cannot hit the throw today, but two spellings of one thing in one package will drift. Either have `CommentPopover` use `relativeTime` (which would move it up to `PageComments/utils/`, since the consumers would no longer all live under `PageHeader/`), or leave a note saying why the header formats differently.

packages/ui/src/components/PageComments/components/PageHeader/utils/sharedVersion/sharedVersion.ts
```diff
@@ -0,0 +1,13 @@
+export function nextSharedVersion(
+	picked: number,
+	latestVersion: number | null,
+): number | null {
+	return picked === latestVersion ? null : picked;
+}
+
+export function servedVersion(
```
saddlepaddle: Agreed, and the server already does most of this — the rule `sharedVersion ?? latestVersion` now has four implementations:

- `page.get` already returns a computed `servedVersion` field (page.ts:210)
- `page.pull` applies the same fallback internally to pick which version to serve (page.ts:356) but returns only the raw `sharedVersion` + `latestVersion` (page.ts:406)
- this `servedVersion()` helper
- and its callers in `PageTitleMenu` / `PagePaneTitle`

Since `usePageHeaderData` reads from `page.pull`, the cheapest fix is to add `servedVersion` to `pull`s response the same way `get` does it, and delete `servedVersion()` here. Worth keeping it distinct from the existing `version` field in that response — `version` is the version you are currently looking at, `servedVersion` is what the share link hands to a visitor, and they differ whenever an explicit version is requested.

`nextSharedVersion` is the write direction, so a response field cannot replace it — but it has a staleness bug that moving it server-side would fix. The collapse-to-null decision compares the picked version against the client's `latestVersion`. If someone publishes v6 while the header still believes latest is v5, and the user picks v5, the client sends `null` — "follow latest" — and the share link starts serving v6, which is the opposite of what they chose. If `setSharedVersion` took the picked number and did the comparison against a freshly read latest (page.ts:277 currently stores `input.version` as-is), it would correctly pin v5.

Both are follow-ups rather than blockers, but the second one is a real behavior bug, not just tidiness.

apps/desktop/src/renderer/routes/_authenticated/_dashboard/pages/components/PagesView/hooks/usePageFavorites/usePageFavorites.ts
harshithmullapudi: yeah true I made this for quick so that we understand if it's useful maybe we find it useful we can move to the database

AviPeltz: hey

PR #6831, opened 2026-08-24
fix(mobile): give the terminal screen its back swipe back

saddlepaddle:
Swiping in from the left edge did nothing on a workspace screen. Two separate causes, both verified on an iOS 26.5 simulator by A/B-ing a throwaway probe route with Maestro edge swipes.

## What was wrong

**`fullScreenGestureEnabled: false` disables the back swipe entirely on iOS 26** — not just the full-screen part of it. The system's back gesture there is the content pop, `react-native-screens` gates it on that flag (`isFullScreenSwipeEffectivelyEnabled` defaults to YES on 26), and the old edge recognizer is no longer a fallback. Verified: a pushed screen with **no WebView at all** and the flag set does not pop from an edge swipe; the identical screen without the flag pops.

**The WebView swallows every touch over its own area** — plain HTML or the xterm page, touch handlers or not, gesture enabled or not. So even with the flag gone, a swipe starting over the terminal never begins. An invisible 20pt strip keeps a finger's width of the left edge native, which is all UIKit needs.

`collapsable={false}` on that strip is load-bearing: with no background and no handlers, Fabric flattens the view away, no native view exists, and the touch reaches WebKit again. That is why a first attempt at the strip looked correct and did nothing.

## Verified

Probe route pushed on the same stack, driven with Maestro:

| screen content | full-screen flag | edge swipe pops |
| --- | --- | --- |
| plain RN view | default | yes |
| plain RN view | `false` | **no** |
| WebView (plain or xterm) | `false` | no |
| WebView | default | no |
| WebView + flattened strip | default | no |
| WebView + `collapsable={false}` strip | default | **yes** (from 4pt, 8pt, 16pt) |

Width was never the problem — 20pt is enough.

## Cost

The leftmost 20pt of the terminal stops taking taps, long-press-select and scroll drags. Roughly the first three character columns, and about the strip iOS already reserves in native apps.

Drags further right stay the terminal's, because WebKit still owns those touches — so enabling the gesture cannot pop the screen out from under a drag over output.

## Not in this PR

`files-changed`, `file` and `pull-request/[pullRequestId]/index` still set the same flag, so by the first finding they have **no back swipe on iOS 26** either. They are native-content screens, so dropping the flag should restore it fully, but I have not driven those screens — left for a follow-up.

https://claude.ai/code/session_01JwmrtU7LZ5GrN5TrfG1J7n

---
## Summary by cubic
Restores the iOS back swipe on the Workspace terminal screen by removing `fullScreenGestureEnabled: false` and reserving a 20pt native edge above the `WebView`. On iOS 26, edge swipes now pop as expected without interrupting terminal drags; the edge also dismisses the keyboard and is hidden from VoiceOver.

- Removes `fullScreenGestureEnabled: false` from the Workspace stack options and the screen’s `headerOptions`.
- Adds an absolute 20pt-wide `Pressable` over the left edge to keep it native and dismiss the keyboard on tap; sets `accessible={false}` so it’s not announced by VoiceOver.
- Side effect: the leftmost ~20pt no longer passes taps/selection/scroll to the terminal; drags further right remain terminal-only.
- Follow-up: `files-changed`, `file`, and `pull-request` screens still set the flag and have no back swipe on iOS 26.

<sup>Written for commit 75595a1e4a20a018e37de7b1d55aef525ccf743a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6831?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored the native iOS edge-swipe gesture for navigating back from workspace screens.
  * Improved back-swipe responsiveness when starting the gesture over terminal content.
  * Preserved keyboard dismissal when tapping the terminal’s left edge.
  * Improved VoiceOver behavior by directing keyboard-dismissal actions through the accessible backdrop control.

### DIFF
--- a/apps/mobile/app/(authenticated)/_layout.tsx
+++ b/apps/mobile/app/(authenticated)/_layout.tsx
@@ -61,7 +61,6 @@ export default function AuthenticatedLayout() {
 					headerBackButtonDisplayMode: "minimal",
 					headerShadowVisible: false,
 					title: "Workspace",
-					fullScreenGestureEnabled: false,
 				}}
 			/>
 			<Stack.Screen
--- a/apps/mobile/screens/(authenticated)/workspace/[id]/WorkspaceScreen/WorkspaceScreen.tsx
+++ b/apps/mobile/screens/(authenticated)/workspace/[id]/WorkspaceScreen/WorkspaceScreen.tsx
@@ -60,11 +60,13 @@ import { WorkspacePlaceholder } from "./components/WorkspacePlaceholder";
 
 const NOTICE_MS = 1500;
 
+// No fullScreenGestureEnabled: false here. On iOS 26 the system's back swipe
+// IS the full-screen one, and react-native-screens reads that flag as "no back
+// gesture at all" rather than "edge only" — the old edge recognizer is gone.
 const headerOptions = {
 	headerShown: true,
 	headerBackButtonDisplayMode: "minimal",
 	headerShadowVisible: false,
-	fullScreenGestureEnabled: false,
 } as const;
 
 const PENDING_CREATE_POLL_MS = 2_000;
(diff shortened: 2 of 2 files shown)


saddlepaddle:
Addressed the review comment in d3ff1918f: the edge strip is now a `Pressable` carrying the same `composerRef.blur()` as the backdrop it sits above, so a tap in that 20pt dismisses the keyboard again. A Pressable also can't be flattened away, which is the property the strip needs, so `collapsable={false}` is no longer necessary. Re-verified on the simulator — swipes starting 4pt and 12pt in still pop the screen.

PR #6858, opened 2026-08-25
feat(desktop,host-service): forward remote workspace ports to the local machine

Ymirke:
**Links**
- ExecPlan: `plans/done/20260824-1849-remote-port-forwarding.md`
- Docs already said this was on the roadmap: `apps/docs/content/docs/remote-workspaces.mdx` ("Ports on a remote host")

## Summary
- Ports of a remote workspace now reach `localhost:<port>` on the user's machine. Selecting the workspace starts the forwards; leaving it stops them.
- A taken local port is reported as busy with two choices: stop the local process (only when Superset started it) or forward to another port. The desktop never remaps silently.
- Host-service only forwards ports the port scanner attributes to the requested workspace. Other services on the host stay unreachable.

## Why / Context
The ports sidebar already detects ports on a remote host, but `localhost:PORT` on the desktop pointed at nothing; users had to open an SSH tunnel or a Tailscale route by hand. The docs called forwarding a roadmap item. This is a first, self-contained version of it: relay transport only, desktop only.

## How It Works
- **Host-service** gets a `/tcp/:port` WebSocket route (`packages/host-service/src/ports/tcp-forward-route.ts`) behind the existing `wsAuth`. It checks `portManager.getPortsByWorkspace(workspaceId)` and then splices bytes to `net.connect(127.0.0.1, port)`. Frames are binary and under the relay's 1 MiB limit; a slow peer pauses the socket.
- **relay2 needs no change.** It already splices any `/hosts/:hostId/*` WebSocket to the host verbatim through the per-stream dial-back.
- **Desktop main** adds a `ForwardTransport` interface with one implementation, `RelayForwardTransport` (`apps/desktop/src/main/lib/port-forward/`). `PortForwardManager.sync()` is the single entry point: it stops forwards outside the requested set and starts the missing ones on the same port number. Each accepted local TCP connection becomes one relay stream. Exposed as the `portForwards` tRPC router; listeners are dropped on `before-quit`.
- **Renderer**: `RemotePortForwarder` (mounted in the dashboard layout) syncs the selected v2 workspace's remote ports, debounced 200 ms. `PortForwardsProvider` mirrors main's state; `usePortOpenActions` opens forwarded remote ports at their local port; rows show `3000 · forwarded`, `3000 → localhost:54321`, `3000 · local port busy` with the two actions. `setJwt` pushes the relay JWT to main, which has no auth client of its own.

## Manual QA Checklist
Done against a real remote host (Ubuntu, host-service 1.24.2 + this branch's bundle, relay2) from a dev desktop signed in to the production API:
- [x] Open a remote workspace: its ports forward (`43104 active`, `8765 active`, `3000 busy` because a local process owned 3000)
- [x] `curl http://localhost:43104/` on the Mac returns the page served on the host (HTTP 200, ~1 s per new connection)
- [x] Click a forwarded row: opens `localhost:<port>` per the Settings → Links → Ports preference
- [x] "Use another port" on the busy row: label becomes `3000 → localhost:61559`
- [x] Leave the workspace: all forwards stop, `lsof` shows the local port released; return: they restart
- [x] Ask for a port the workspace did not start: host-service closes with `1008 port not owned by workspace`
- [x] A host on the v1 relay: rows show "Host relay does not support port forwarding (protocol v1)" (probe of `/health` for `proto: 2`)
- [ ] Not done: "Stop local <process>" path with a local Superset-owned port (the local collision in the test was a non-Superset process, so only "Use another port" was offered)
- [ ] Not done: WebSocket/HMR through the forward (the test server was a plain HTTP server; the tunnel is protocol-agnostic and the relay path already carries binary frames)

## Testing
- `bun run typecheck` (desktop, host-service)
- `bun run lint` (Biome on all changed files)
- `bun test apps/desktop/src/main/lib/port-forward` (7 tests: sync, switch, busy with owner, ephemeral retry, kill-and-restart, failed probe, stopAll)
- `bun test` for `deriveForwardSyncInput` and `formatPortRowLabel` (10 tests)
- `bun run test:integration:ports` in `packages/host-service` (8 tests under `node --test`, since `@hono/node-ws` needs a real Node server): echo both ways, early frames, ownership refusal, missing workspaceId, refused upstream, text frames, teardown both directions

## Design Decisions
- **Raw TCP tunnel, not an HTTP proxy**: dev servers use WebSockets for hot reload; an HTTP proxy would need separate upgrade handling and the relay's HTTP exchange buffers whole bodies. A byte tunnel makes HTTP, WebSocket, gRPC and databases behave the same.
- **Transport interface with relay first**: the relay works for every host with no setup. A Direct transport (Tailscale, WireGuard, VPN, LAN — no vendor plugin, just an address the host advertises) and an SSH transport plug into the same interface later. Decisions D-2 and D-8 in the plan record the intended auth model for Direct (user JWT, same access check as the relay).
- **Only ports attributed to the workspace**: same rule `ports.kill` uses today; blocks `127.0.0.1:5432`, `6379`, metadata endpoints unless the workspace runs them.
- **Same port number, never a silent remap**: apps talk to each other by port number; the user must see and choose.
- **Buffer from accept in main**: under Bun, bytes on an accepted socket are lost until a consumer is attached (even after `pause()`); Node keeps them. A bounded buffer from the first tick is also the right shape in production, since a browser sends its request right after connect.

## Known Limitations
- Each new local TCP connection costs one relay dial-back (~1 s). Keep-alive connections are fine; tools that open a connection per request feel it.
- A refused upstream on the host (e.g. a loopback firewall) shows as a reset connection locally, and can surface as the relay's generic "Host did not answer" instead of the host's reason.
- Relay protocol v1 hosts are not supported (v1 sends client frames as text only). The v1 relay is deprecated and the cutover to relay2 is in progress.
- The relay sees the forwarded bytes, as it does for terminals today.

## Follow-ups
- Direct transport (host-service listens on advertised interfaces with user-JWT auth; desktop prefers direct, falls back to relay)
- SSH transport (`ssh -N -L` with the user's own config)
- Per-connection failure reporting on the row; pre-dial one spare stream per forward to hide the dial-back latency
- CLI: `superset ports forward`

## Risks / Rollout
- Risk: the host route is new attack surface. It is gated on the host-service secret (relay dial-back) and on port ownership; it cannot reach ports the workspace did not start.
- Rollout: desktop and host-service must both carry this change for forwarding to work; older host-services return 404 on `/tcp/*`, which the desktop shows as an error on the row. No data or schema changes.
- Rollback: revert the three commits; nothing persists.

https://claude.ai/code/session_01NxTU235jowe5cpVsG47JqW

## Summary by CodeRabbit

* **New Features**
  * Added automatic port forwarding for remote workspaces.
  * Forwarded ports use the same local port when available, with alternate mappings supported.
  * Added shared tunnels that update automatically when switching workspaces.
  * Added status indicators for active, busy, and failed forwards.
  * Added controls to stop conflicting local processes or retry on another port.
* **Bug Fixes**
  * Improved cleanup and recovery when forwarding sessions or connections close unexpectedly.
* **Documentation**
  * Documented setup requirements, supported hosts, forwarding behavior, TCP limitations, and security considerations.


Ymirke:
Hey guys!

I got very excited about superset with your remote workspaces feature. For whatever reason this hasn't been developed towards quickly enough by the startups in this space IMO. 

I really wish you supported **automatic port forwarding**, but excited you have this listed in the docs and roadmap. 

To me this is a really important to validate features/get a feel for them before merging a PR, so it is key for my workflow. 

This PR is a rough implementation, but could be a good POC for testing out the Raw TCP tunnel approach for this feature.

_Sidenote 1: remote workspaces feature has slow keyboard input. SSH is pretty snappy (client renders inputs immediately I guess), but I assume this is something that could be replicated?._

_Sidenote 2: there seems to be some kind of performance issue when creating new workspaces randomly, my pretty new macbook pro turned up the fans quite loud, not sure if it is a dev only thing._

saddlepaddle:
Hey @Ymirke — really solid PR. The transport seam, the ownership rule, and the no-silent-remap UX all held up under close review, and they're all still here. We pushed one commit on top as a **proposal**: 3b31008. If you disagree with the direction, push back — it's structured to be easy to drop.

**What it changes:** the host leg. `/tcp/:port` (one relay dial-back per TCP connection) becomes `/fwd?workspaceId=` — one WebSocket per (host, workspace), every forwarded connection a numbered stream inside it. Framing lives in `@superset/shared/port-forward-mux` (OPEN/OPENED/OPEN_FAIL/DATA/EOF/CLOSE/WINDOW + session HELLO/PING/PONG, 256 KiB frames, 1 MiB per-stream credit windows). Your `PortForwardManager`, the renderer, the busy-port flow, and the per-OPEN ownership check are all unchanged — `RelayForwardTransport` still hands the manager a plain `Duplex`, the mux lives inside it. The relay still needs zero changes: the session is spliced verbatim like a terminal stream, and dial tickets keep their single-use lifetime.

**Why reshape before merge:** wire protocols ship once. Host-services update slowly, so the moment `/tcp/:port` reaches a release, the desktop has to keep speaking it long after a replacement. Doing it now means exactly one protocol ever exists — and your "~1 s per new connection" known-limitation goes away instead of getting documented.

**Also folded in from review:** a race where a sync landing mid-`probe` orphaned a bound local listener (reproduced, now a regression test); a transient stream failure permanently wedging a forward in `error` (a later success now restores `active`); per-window wanted-sets with union reconcile so two windows don't tear down each other's forwards; the ports-provider gate now requires the selected workspace to be remote; docs version floor corrected. One inherited bot finding we verified as **wrong**: `net.Server.close()` releases the listening handle synchronously, so the suggested awaited-close refactor isn't needed.

**Measured end-to-end through production relay2** (Linux host running this branch's bundle): cold session dial 991 ms, now paid once per workspace; warm stream open p50 166 ms (~1 RTT) vs ~1 s per connection before; single-stream throughput 5–8.9 MB/s — the 1.9 MB/s previously measured on this path was per-frame overhead, and frame coalescing lifts it; 26.5 MB/s aggregate across 4 streams; and terminal-class echo RTT held at p50 73 ms / p95 175 ms *while* a forward pushed 82 MB through the same host tunnel — no contention.

Tests: `bun test src/main/lib/port-forward` (18), `bun run test:integration:ports` (14, under Node like the other WS suites), `packages/shared` framing tests (11), full desktop suite green.

saddlepaddle (approved):
Reviewed line-by-line across the reshape: mux protocol + endpoints tested at three layers, measured end-to-end through production relay2 (throughput 2.6–4.7× the per-connection design, no terminal-latency contention under saturated forwards), and verified in the desktop app including Vite HMR over the tunnel. CodeRabbit's follow-up findings addressed. Approving.

saddlepaddle:
Merging to catch the next desktop + host-service release — CI is green on the current head and the branch is synced with main (including the i18n sweep; the busy-port actions are Lingui-ized with ja/zh-CN catalog entries). @Ymirke thank you for a genuinely excellent contribution — the transport seam you designed is what made the mux reshape possible without touching your manager or UX. Happy to iterate post-merge if anything about the reshape doesn't sit right.

PR #6875, opened 2026-08-26
feat: public agent usage leaderboard

harshithmullapudi:
## What & why

An opt-in public leaderboard for agent usage, so engineers can see how much
agent work they're running relative to everyone else.

**What gets published.** Token counts, model names, session counts and a USD
estimate — never repo names, file paths or prompts. Opt-in is explicit, the
handle is chosen by the user, and leaving deletes every published row rather
than hiding it (FK cascade from `leaderboard_participants`). The join dialog
states that the account name is shown publicly alongside the handle, because
the public API serves it.

**The pieces:**

- `packages/db` — migration `0094` adds `leaderboard_participants`,
  `leaderboard_daily` and `leaderboard_daily_factory`, plus a
  `leaderboard_visibility` enum. Generated with `drizzle-kit`, not hand-edited.
- `packages/trpc` — `leaderboard` router: `join` / `leave` / `setVisibility` /
  `publish` / `standings` / `previewRank` / `me`, rate-limited per user through
  Upstash. Response shapes live in a db-free `types.ts` so consumers can import
  them as types without pulling in the db client.
- `packages/host-service` — builds the publish payload from local transcripts.
  The scan-and-parse half of `computeUsageHistory` is extracted into
  `entries.ts` so the leaderboard payload and usage history share one
  implementation instead of two copies of the profile-discovery logic.
- `apps/api` — public, cached, CORS-open `GET /api/leaderboard`,
  `/api/leaderboard/stats` and `/api/leaderboard/u/[handle]`.
- `apps/desktop` — opt-in from Settings → Account or the Usage tab, a first-run
  prompt, and a 2h auto-publish loop that skips the round-trip when the payload
  hash is unchanged.
- `apps/marketing` — `/leaderboard`, `/stats` and `/user/[handle]` (with OG
  images), all server-rendered on a 5 min revalidate.

**Factory tiers.** Participants get a 1–4 tier over a rolling 30-day window,
gated by the weakest of four axes (width, depth, output, sustain) so that
improving an axis already ahead of the others moves nothing. Tier is
deliberately not re-sliced by the selected period — it is already a 30-day
figure, and re-windowing it would report a number that means nothing.

## How I tested it

- `bun run lint` — clean across 6533 files (CI fails on warnings, so this is
  the gating check).
- `bun run typecheck` — passes for every package. The one failure on `main`
  (`@superset/api` can't resolve `resend`) predates this branch and is
  untouched by it.
- `bun test` — 73 new tests across 6 files covering tier computation and
  progress, period/day-range resolution, day bucketing, factory-day rollups,
  the auto-publish schedule and the first-run flag. The persisted-key registry
  test passes, so the two new localStorage keys are properly allowlisted.
- `bunx sherif` — 0 errors.

**Not yet done:** no CDP screenshots of the desktop opt-in flow or the three
marketing pages. Everything user-visible here is unverified in a running app —
worth a look before this merges.

## Review fixes

The first review pass turned up three issues, all fixed in `61834e0d0`:

- **Deleted accounts leaked into public aggregates.** `getStats` didn't filter
  `users.deletedAt` the way `getStandings` and `getParticipant` do, so a
  deleted account's tokens still counted toward the public totals and the tier
  distribution. The windowed participant count inside `getStandings` had the
  same gap, which additionally made `total`/`hasMore` disagree with the `rows`
  returned beside them.
- **Unbounded scan from the public API.** `resolveWindow` now clamps an
  explicit `from`/`to` to `MAX_WINDOW_DAYS` (366), so an unauthenticated caller
  can no longer ask for a full-table scan of `leaderboard_daily`. Clamping
  rather than rejecting matches how `limit`/`offset` are already handled on
  these routes.
- **Handle race returned the wrong error.** A handle claimed between `join`'s
  pre-check and its insert raised a raw Postgres unique violation; it now
  surfaces as `CONFLICT`, same as the pre-check path. The detection walks the
  error's cause chain because drizzle wraps driver errors.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added public leaderboard pages with rankings, statistics, filters, tier progress, token breakdowns, and participant profiles.
  - Added profile sharing, social links, avatars, and social preview images.
  - Added desktop leaderboard opt-in, privacy controls, suggested handles, rank previews, first-run prompts, and automatic usage publishing.
  - Added tier badges, progress visuals, responsive statistics, and improved usage formatting.

- **Bug Fixes**
  - Improved token formatting around unit-rounding boundaries.
  - Added graceful handling for unavailable statistics and missing profiles.

- **Tests**
  - Added coverage for leaderboard aggregation, scheduling, tiers, date ranges, persistence, and formatting.


packages/trpc/src/router/leaderboard/leaderboard.ts
```diff
@@ -0,0 +1,562 @@
...
+					day.cacheWrite1h +
+					day.output,
+				usdEstimate: day.usdEstimate.toFixed(6),
+				approximate: day.approximate,
+				sessions: day.sessions,
+			}));
+
+			await db
+				.insert(leaderboardDaily)
+				.values(rows)
```
saddlepaddle: A factory-only payload (`days: []`, non-empty `factoryDays` — `groupFactoryDays` emits these for merged-PR days with no transcript usage in the window) gets past the both-empty guard at L392 and hits `.values([])` here, which drizzle throws on ("values() must be called with at least one value"). The factory insert below guards `length > 0`; this one needs the same, otherwise every auto-publish cycle 500s for those users until new transcript usage lands.

packages/trpc/src/router/leaderboard/leaderboard.ts
```diff
@@ -0,0 +1,562 @@
...
+		? sql`and p.user_id <> ${excludeUserId}`
+		: sql``;
+
+	if (!range) {
+		const rows = await db.execute<{ ahead: number; total: number }>(sql`
+			select
+				count(*) filter (where p.tokens > ${tokens})::int as ahead,
+				count(*)::int as total
+			from leaderboard_participants p
+			where p.visibility = 'public' and p.revoked_at is null
```
saddlepaddle: Both `rankFor` branches skip the `users.deleted_at` filter that `getStandings`/`getParticipant` apply, so a soft-deleted user still counts as "ahead" in `previewRank`/`me` and desktop ranks disagree with the published board. Worth extracting one shared eligibility predicate so these can't drift.

packages/trpc/src/router/leaderboard/queries.ts
```diff
@@ -0,0 +1,459 @@
...
+				isNull(users.deletedAt),
+			),
+		)
+		.groupBy(
+			leaderboardDaily.userId,
+			leaderboardParticipants.handle,
+			leaderboardParticipants.tier,
+			users.name,
+		)
+		.orderBy(desc(byCost ? spend : total))
```
saddlepaddle: Single-aggregate sort with limit/offset paging: ties can order differently between requests, so a participant straddling a page boundary can appear on both pages (duplicate React key — `LeaderboardTable` keys on `row.handle`) or drop out. Needs a stable tiebreaker, e.g. `.orderBy(desc(...), leaderboardDaily.userId)`.

packages/trpc/src/router/leaderboard/queries.ts
```diff
@@ -0,0 +1,459 @@
...
+			sessions: Number(participant.sessions),
+			approximate: participant.approximate,
+		},
+		window: {
+			range,
+			tokens: Number(windowTotals?.tokens ?? 0),
+			usd: String(windowTotals?.usd ?? "0"),
+			sessions: Number(windowTotals?.sessions ?? 0),
+		},
+		rank: Number(ranked?.ahead ?? 0) + 1,
```
saddlepaddle, line 457: `ranked` only counts `tokens > 0` participants, so a joined-but-unpublished user (tokens = 0) gets `ahead = N`, `rank = N + 1`, `total = N` — the profile page and OG card render "#N+1 of N" (and "#1 of 0" on an empty board). Either count the subject into `total` or floor `rank` at `total`.

apps/desktop/src/renderer/hooks/useLeaderboardOptIn/useLeaderboardOptIn.ts
```diff
@@ -0,0 +1,81 @@
...
+		async (handle: string) => {
+			setJoining(true);
+			try {
+				await apiTrpcClient.leaderboard.join.mutate({
+					handle,
+					visibility: "public",
+				});
+
+				if (activeHostUrl && machineId) {
+					const { days } = await publishUsage(
```
saddlepaddle: `publishUsage` failing here (host offline, older host-service without `usage.leaderboardPayload`) lands in the same catch as `join.mutate`: we toast "Couldn't join", return false, and skip `membership.refetch()` — but the participant row was already created. The user believes the join failed while they're actually enrolled, and auto-publish starts once the membership query refetches. Commit the join success (toast + refetch) before the publish attempt, with a separate error path for the publish.

apps/desktop/src/renderer/hooks/useLeaderboardJoinPreview/useLeaderboardJoinPreview.ts
```diff
@@ -0,0 +1,57 @@
...
+					day.uncachedInput +
+					day.cachedInput +
+					day.cacheWrite5m +
+					day.cacheWrite1h +
+					day.output,
+				0,
+			);
+			const providers = [...new Set(payload.days.map((day) => day.provider))];
+			const rank = await apiTrpcClient.leaderboard.previewRank.query({
+				period: "month",
```
saddlepaddle: The teaser sums the user's last 30 days but ranks with `period: "month"` (calendar month-to-date), so early in a month it compares the user's 30 days against everyone else's 1–2 days — an inflated "You'd be #N of M" that drops sharply after joining. Should be `period: "30d"` to match the copy and the board default.

apps/desktop/src/renderer/routes/_authenticated/components/LeaderboardAutoPublish/hooks/useLeaderboardAutoPublish/autoPublishSchedule.ts
```diff
@@ -0,0 +1,52 @@
...
+	lastPayloadHash: string | null;
+}
+
+export const INITIAL_AUTO_PUBLISH_STATE: AutoPublishState = {
+	lastPublishedAt: 0,
+	lastPayloadHash: null,
+};
+
+export function isPublishDue(state: AutoPublishState, now: number): boolean {
+	return now - state.lastPublishedAt >= PUBLISH_INTERVAL_MS;
```
saddlepaddle, line 22: If `lastPublishedAt` is in the future (publish landed while the clock was fast, then NTP corrects back), this stays false until wall-clock catches up — auto-publish stalls for the whole skew. Treat `lastPublishedAt > now` as due. Note the test "fires when a clock change puts the last publish in the future" asserts `toBe(false)`: the name describes the fix, the assertion pins the bug.

apps/api/src/app/api/leaderboard/route.ts
```diff
@@ -0,0 +1,31 @@
...
+} from "@/lib/leaderboardResponse";
+
+export const dynamic = "force-dynamic";
+
+export async function GET(request: Request): Promise<Response> {
+	const limited = await rateLimited(request);
+	if (limited) return limited;
+
+	const params = new URL(request.url).searchParams;
+	const result = await getStandings({
```
saddlepaddle: Nothing catches throws from `getStandings`, so a db error — or an out-of-range date like `?periodStart=2026-99-99`, which passes `parseDayKey`'s shape regex and blows up in the query layer — returns a bare 500 with no CORS headers; the marketing board's cross-origin fetch only sees an opaque TypeError. Same shape on stats and u/[handle]. Wrap all three handlers so errors go through the CORS-wrapped response helper, and reject invalid calendar dates in `parseDayKey`.

apps/marketing/src/app/leaderboard/components/LeaderboardBoard/LeaderboardBoard.tsx
```diff
@@ -0,0 +1,183 @@
...
+			).then((response) => (response.ok ? response.json() : null)),
+			fetch(
+				`${apiUrl}/api/leaderboard/stats?${buildStandingsQuery(selection)}`,
+				{
+					signal: controller.signal,
+				},
+			).then((response) => (response.ok ? response.json() : null)),
+		])
+			.then(([nextStandings, nextStats]) => {
+				setStandings(nextStandings);
```
saddlepaddle: A non-OK response resolves to null (L59) and this unconditionally replaces populated standings; with `isLoading={loading && !standings}` the table then shows the "Nobody has joined the board yet" empty state. A tab flip during a 429 from the shared 60/min/IP limiter blanks a populated board. Keep previous standings on failure (only set when non-null), same for stats.

apps/marketing/src/app/user/[handle]/page.tsx
```diff
@@ -0,0 +1,187 @@
...
+}
+
+export default async function UserProfilePage({ params }: PageProps) {
+	const { handle } = await params;
+	const profile = await fetchParticipant(handle, { period: "all" });
+
+	if (!profile) notFound();
+
+	const colors = buildModelColors([profile.models]);
+	const shareUrl = `${env.NEXT_PUBLIC_WEB_URL.replace(/\/$/, "")}/user/${profile.handle}`;
```
Comment (saddlepaddle) [inline apps/marketing/src/app/user/[handle]/page.tsx]: `NEXT_PUBLIC_WEB_URL` is the dashboard origin (used as `dashboardUrl` in CTAButtons) and has no `/user/[handle]` route — so Copy link, the X post, and LinkedIn all share a 404. Should be `COMPANY.MARKETING_URL`, matching the canonical/OG URLs.

apps/marketing/src/app/stats/components/StatsBody/StatsBody.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+			<div className="grid gap-6 md:grid-cols-2">
+				<Panel title="Popular models" meta="by users">
+					<ModelBars rows={toUserRows(models.byUsers)} colors={colors} />
+				</Panel>
+				<Panel title="Top models" meta="by spend">
+					<ModelBars rows={toSpendRows(models.bySpend)} colors={colors} />
+				</Panel>
+				<Panel title="Model volume" meta="by tokens" className="md:col-span-2">
+					<ModelBars rows={toTokenRows(models.bySpend)} colors={colors} />
```
saddlepaddle: This panel is "by tokens" but renders `models.bySpend` — a list `getStats` selects and orders by summed usd (top 20 by spend). A cheap high-volume model drops out entirely past the top-20-by-spend cut, and the bars render in spend order with token widths, so they're visibly non-monotonic. Needs a tokens-ordered aggregation from `getStats`.

apps/marketing/src/app/leaderboard/utils/fetchLeaderboard/fetchLeaderboard.ts
```diff
@@ -0,0 +1,134 @@
...
+		from: options.from,
+		to: options.to,
+	});
+
+	try {
+		const response = await fetch(
+			`${env.NEXT_PUBLIC_API_URL}/api/leaderboard/u/${encodeURIComponent(handle)}?${query}`,
+			{ next: { revalidate: REVALIDATE_SECONDS } },
+		);
+		if (!response.ok) return null;
```
saddlepaddle: Every failure — 429 from the shared limiter, 500, network — collapses to null, and `user/[handle]/page.tsx` turns null into `notFound()`, so ISR can cache a 404 for a live profile for the whole revalidate window. Only a real 404 should return null; transient errors should throw so the stale page survives revalidation.

saddlepaddle (commented):
Review pass (Claude Code) — inline comments on the clear-cut issues. A few design-level findings (tier hysteresis, usd bounds, rate-limiter failure mode) to follow after another pass.

apps/api/src/lib/leaderboardResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+		},
+	});
+}
+
+const redis =
+	env.KV_REST_API_URL && env.KV_REST_API_TOKEN
+		? new Redis({ url: env.KV_REST_API_URL, token: env.KV_REST_API_TOKEN })
+		: null;
+
+const limiter = redis
```
saddlepaddle: This ternary can't take the null branch: apps/api's env requires both KV vars (`env.ts:58`), unlike `packages/trpc` where they're `.optional()` — looks copied from the trpc-side pattern (`metric-cache.ts`, trpc `leaderboard.ts:29`), where the guard is legitimate. `new Redis()` also never fails at construction (plain HTTP client — failures happen per-call in `.limit()`), so there's nothing to guard here.

Convention elsewhere in apps/api (mcp, support/lookup, automations + circleback webhooks) is an unconditional module-level limiter in the route file that uses it, with a per-surface prefix, and the handler shaping its own 429. `Ratelimit` instances are stateless, so per-route instances sharing this prefix keep the shared 60/min budget — inlining is behavior-neutral. With the limiter moved into the routes, this file goes back to being leaderboard response/param helpers.

apps/api/src/lib/leaderboardResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+	if (!limiter) return null;
+	const { success } = await limiter.limit(clientKey(request));
+	if (success) return null;
+	return Response.json(
+		{ error: "Rate limit exceeded" },
+		{ status: 429, headers: { "Access-Control-Allow-Origin": "*" } },
+	);
+}
+
+export function parsePeriod(
```
saddlepaddle: Convention nit: routes elsewhere in apps/api parse query params inline with zod (support/lookup: `safeParse` → 400) rather than hand-rolled `parseX` helpers. Since all three routes take the identical window param set, one shared schema collapses these four functions and is also the natural home for the real-calendar-date validation flagged on the routes:

```ts
const dayKey = z.string().refine(isRealDayKey).optional().catch(undefined);

export const windowQuery = z.object({
	period: z.enum(LEADERBOARD_PERIODS).catch("30d"),
	periodStart: dayKey,
	from: dayKey,
	to: dayKey,
});
// route: windowQuery.parse(Object.fromEntries(new URL(request.url).searchParams))
```

`.catch()` preserves today's coerce-to-default behavior; swapping to `safeParse` → 400 is the stricter option if we'd rather reject garbage on a public API — either way the choice becomes explicit. `publicJson` is worth keeping (it pins the CORS + CDN cache policy the three routes must share) — consider giving it an error sibling so the 404 in u/[handle] and the error paths stop hand-rolling/omitting the CORS header.

apps/api/src/app/api/leaderboard/u/[handle]/route.ts
```diff
@@ -0,0 +1,38 @@
+import { getParticipant } from "@superset/trpc/leaderboard-queries";
```
Comment (saddlepaddle) [inline apps/api/src/app/api/leaderboard/u/[handle]/route.ts]: Naming: `/api/leaderboard/u/[handle]` → `/api/leaderboard/user/[handle]`. No other apps/api route abbreviates a segment, and the marketing page this feeds is already the spelled-out `/user/[handle]`. Cheap while unmerged — the only consumer is `fetchLeaderboard.ts:125`.

apps/desktop/src/renderer/components/LeaderboardJoinDialog/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface LeaderboardPreview {
```
saddlepaddle: `rank`/`total` re-declare half of `previewRank`'s output by hand (which already drifted — the server also returns `percentile`). The local view-model is fine since `tokens`/`providers` are computed in the hook, but the server half should be inferred, matching the `RouterOutputs` idiom used elsewhere in the renderer (e.g. `useCloudWorkspaces`):

```ts
import type { RouterOutputs } from "@superset/trpc";

export type LeaderboardPreview = Pick<
	RouterOutputs["leaderboard"]["previewRank"],
	"rank" | "total"
> & {
	tokens: number;
	providers: string[];
};
```

apps/desktop/src/renderer/lib/leaderboard/publishUsage.ts
```diff
@@ -0,0 +1,66 @@
...
+	cacheWrite5m: number;
+	cacheWrite1h: number;
+	output: number;
+	reasoningOutput: number;
+	usdEstimate: number;
+	approximate: boolean;
+	sessions: number;
+}
+
+export interface LeaderboardPayload {
```
saddlepaddle, line 17: Same pattern, stronger case: `LeaderboardPayloadDay`/`LeaderboardFactoryDay`/`LeaderboardPayload` hand-copy the publish contract that already exists as the typed tRPC input — a field added to `publishSchema` has to be re-copied here by hand. Since the payload is forwarded straight into `leaderboard.publish.mutate`, derive it instead:

```ts
import type { RouterInputs } from "@superset/trpc";

type PublishInput = RouterInputs["leaderboard"]["publish"];
export type LeaderboardPayloadDay = PublishInput["days"][number];
export type LeaderboardFactoryDay = PublishInput["factoryDays"][number];
```

Related: `payloadVersion: 2` is hard-coded at the call site (L52) — worth a shared constant next to `publishSchema` so the version lives in one place.

apps/desktop/src/renderer/routes/_authenticated/components/LeaderboardAutoPublish/hooks/useLeaderboardAutoPublish/useLeaderboardAutoPublish.ts
```diff
@@ -0,0 +1,67 @@
...
+			console.warn("[leaderboard] auto-publish failed (will retry):", error);
+		} finally {
+			inFlight.current = false;
+		}
+	}, [optedIn, activeHostUrl, machineId]);
+
+	useEffect(() => {
+		if (!optedIn || !activeHostUrl || !machineId) return;
+
+		void maybePublish();
```
saddlepaddle, line 54: The join backfill never seeds auto-publish state, so immediately after a user joins this fires with `lastPublishedAt: 0` → "due" → rebuilds and re-publishes the same 30-day window `publishUsage` just uploaded (`publishWindowDays` returns MAX_WINDOW_DAYS when lastPublishedAt is 0). Harmless data-wise (server upserts) but a redundant full 30-day transcript scan + upload on every join.

Fix: when join's `publishUsage` succeeds in `useLeaderboardOptIn`, seed the state:

```ts
writeAutoPublishState({ lastPublishedAt: Date.now(), lastPayloadHash: null });
```

`lastPayloadHash: null` is fine — the next due cycle (2h later) rebuilds a small window and publishes normally.

apps/marketing/src/app/utils/fetchLeaderboard/fetchLeaderboard.ts
```diff
@@ -0,0 +1,134 @@
+import type { LeaderboardPeriod } from "@superset/trpc/leaderboard-periods";
```
saddlepaddle, line 2: Proposal: fold the public read path into the standing tRPC router instead of a parallel REST surface. This file is a hand-rolled client (`buildQuery` serialization + unvalidated `response.json() as ...` casts) against three hand-parsed routes — the whole boundary is untyped mirror maintenance.

Shape:
1. Public procedures on the leaderboard router (`publicProcedure`, IP-keyed `enforce`, zod input schema replacing the hand parsers):
```ts
public: {
  standings: publicProcedure.input(standingsQuery).query(({ input }) => getStandings(input)),
  stats: publicProcedure.input(windowQuery).query(({ input }) => getStats(input)),
  participant: publicProcedure.input(participantQuery).query(({ input }) => getParticipant(input.handle, input)),
}
```
2. CDN caching via `responseMeta` on the api adapter (`apps/api/src/app/api/trpc/[trpc]/route.ts` — GET is already exported):
```ts
responseMeta({ paths, type, errors }) {
  if (type === "query" && errors.length === 0 && paths?.every((p) => p.startsWith("leaderboard.public"))) {
    return { headers: { "cache-control": "public, s-maxage=300, stale-while-revalidate=86400" } };
  }
  return {};
}
```
3. Marketing uses a tRPC client with `httpBatchLink({ methodOverride: "GET" })` (stable cacheable URLs) and a custom fetch threading `next: { revalidate }` for the ISR pages. End-to-end inferred types; the `leaderboard-types` mirror shrinks to whatever the components still name.

Deletes: the three `/api/leaderboard*` routes, `leaderboardResponse.ts` (parsers, `publicJson`, the misplaced limiter — my two earlier comments there become moot), and this file. Also structurally fixes the bare-500-without-CORS class, since errors flow through the shared handler.

One caveat to get right: the `responseMeta` predicate must be airtight — cache headers only when *every* path in the batch is an anonymous public read; a miss here CDN-caches session-varying responses. Keeping the public procedures under a distinct `leaderboard.public.*` prefix makes that auditable. Also confirm the marketing origin is CORS-allowed on `/api/trpc` (`proxy.ts` getCorsHeaders).

apps/marketing/src/app/components/StatStrip/StatStrip.tsx
```diff
@@ -0,0 +1,35 @@
+interface Stat {
```
saddlepaddle, line 1: Structure: this and several siblings are cross-page but parked under `/leaderboard`, so `/stats` and `/user/[handle]` reach into a sibling route's directory (`@/app/leaderboard/components/...`). Per AGENTS.md co-location rule 2 (used by 2+ pages → highest shared parent), promote to `src/app/components/` — the app's existing shared home (CTAButtons, Footer, …):

- `StatStrip`, `TierTube`, `TokenSplitBar`, `ModelBars` (with its `buildModelColors`/`to*Rows` exports) → `src/app/components/`
- `utils/formatUsage`, `utils/avatarUrl`, `utils/fetchLeaderboard` → a shared utils home beside them

Single-page pieces (`LeaderboardBoard`, `LeaderboardTable`, `MetricTabs`, `RangeTabs`, `TierBadge`, stats' `Panel`) are correctly nested and stay put.

apps/marketing/src/app/leaderboard/components/MetricTabs/MetricTabs.tsx
```diff
@@ -0,0 +1,43 @@
...
+			{METRICS.map((metric) => {
+				const active = metric.id === value;
+				return (
+					<button
+						key={metric.id}
+						type="button"
+						role="tab"
+						aria-selected={active}
+						onClick={() => onChange(metric.id)}
+						className={`px-5 py-2 text-xs font-mono uppercase tracking-wider transition-colors ${
```
saddlepaddle: These pills and `RangeTabs`' preset pills are the same control rendered two different ways in the same toolbar: here active = `bg-brand/10 text-brand` inside one bordered container; in RangeTabs active = per-button `border-brand text-brand bg-brand/5`. A11y also diverges — this has `role="tablist"`/`role="tab"`/`aria-selected`, RangeTabs' pills have none. Worth one tiny shared pill-tabs component (local to the page, or `@superset/ui` ToggleGroup restyled) so both rows get one style + one a11y story.

apps/marketing/src/app/leaderboard/components/RangeTabs/utils/formatRangeLabel/formatRangeLabel.ts
```diff
@@ -0,0 +1,10 @@
+import type { DateRange } from "react-day-picker";
+
+export function formatRangeLabel(range: DateRange | undefined): string {
```
saddlepaddle: Near-verbatim copy of the private `formatRangeLabel` already inside `starchart/components/DateRangePicker` (L23–35) — second consumer → one shared helper instead of a copy. Standardize it on `date-fns` while at it: it's the house library (declared in 7 packages, 20 importing files) and marketing is the odd one out — starchart currently imports it *undeclared* via hoisting.

1. Add `"date-fns": "4.1.0"` to `apps/marketing/package.json`
2. One shared helper (e.g. `src/app/utils/formatRangeLabel/`) built on `format(date, "MMM d")`, used by both RangeTabs and starchart's DateRangePicker
3. Delete this `utils/formatRangeLabel/` folder and starchart's private copy

apps/marketing/src/app/leaderboard/utils/formatUsage/formatUsage.ts
```diff
@@ -0,0 +1,40 @@
+export function formatTokens(tokens: number): string {
```
saddlepaddle: Two things here:

1. The date halves have date-fns builtins (house library): `formatDayRange` → `format(parseISO(day), "MMM d")` (parseISO on a date-only key preserves the calendar day, so the UTC pinning dance goes away) and `dayCount` → `differenceInCalendarDays(parseISO(to), parseISO(from)) + 1`.

2. `formatTokens` has already diverged from the desktop copy (`LeaderboardJoinDialog/utils/formatTokens`): that one has no T tier, 1 decimal for B vs 2 here, and no locale grouping — so the join dialog and the public profile render the same user's tokens differently today. The number formatters aren't date-fns territory; the fix is one shared implementation in `packages/shared` (subpath export, like `@superset/shared/billing`) consumed by marketing, the OG image, and the desktop dialog — delete both copies.

apps/marketing/src/app/user/[handle]/utils/getInterBold/getInterBold.ts
```diff
@@ -0,0 +1,17 @@
+const INTER_BOLD_URL =
+	"https://fonts.gstatic.com/s/inter/v18/UcCO3FwrK3iLTeHuS_nVMrMxCp50SjIw2boKoduKmMEVuFuYMZhrib2Bg-4.ttf";
+
+let interBoldPromise: Promise<ArrayBuffer> | null = null;
+
+export function getInterBold(): Promise<ArrayBuffer> {
+	if (!interBoldPromise) {
+		interBoldPromise = fetch(INTER_BOLD_URL).then((res) => {
```
Comment (saddlepaddle) [inline apps/marketing/src/app/user/[handle]/utils/getInterBold/getInterBold.ts]: This runtime fetch to fonts.gstatic.com sits on the OG request path: if Google Fonts is slow/unreachable — or this pinned `/v18/...ttf` hash is retired — `getInterBold()` rejects and the whole opengraph-image route 500s, so scrapers cache "no image" for exactly the share links this feature exists for. There's no graceful degradation possible either: satori requires at least one font, so a catch can't render fontless.

Fix: commit the Inter 700 `.ttf` (OFL-licensed, subset if size matters) next to the route and read it locally — the pattern from the Next/Vercel og docs:

```ts
import { readFile } from "node:fs/promises";
import { join } from "node:path";

const interBold = readFile(join(process.cwd(), "src/app/user/[handle]/assets/Inter-Bold.ttf"));
```

Kills the cross-service dependency, the 500 mode, and the per-cold-start refetch in one move.

apps/marketing/next.config.ts
```diff
@@ -27,6 +27,10 @@ const config: NextConfig = {
 				protocol: "https",
 				hostname: "unavatar.io",
 			},
+			{
+				protocol: "https",
```
saddlepaddle: This allowlist entry is dead — both `avatarUrl` consumers (`DeveloperAvatar`, the profile header) render with `unoptimized`, so nothing routes dicebear through `/_next/image` — and it couldn't work anyway: these URLs are `/svg`, which the optimizer rejects without `dangerouslyAllowSVG` (correctly unset). Meanwhile it's a standing liability: dicebear is generative, every distinct `seed`/param combo is a new source image, so this entry lets anyone mint unbounded unique transformations through our optimizer (billed per transformation + cache pollution). Delete the entry; keep the `unoptimized` hotlinking.

Optional follow-up: `@dicebear/core` can generate these in-process (deterministic from seed), which would also remove the third-party runtime dependency from the public board and stop sending handles to dicebear — but that's polish, not required.

apps/marketing/src/app/leaderboard/components/LeaderboardBoard/components/LeaderboardTable/LeaderboardTable.tsx
```diff
@@ -0,0 +1,134 @@
+import Link from "next/link";
```
saddlepaddle, line 1: Structure pass — full sweep of the PR's added files against the AGENTS.md co-location rules (complements the StatStrip cross-page comment). Every claim below is importer-verified.

**Single-consumer components parked at page level — nest under their only importer, `LeaderboardBoard`:**
- `leaderboard/components/LeaderboardTable/` → `LeaderboardBoard/components/LeaderboardTable/` (sole importer `LeaderboardBoard.tsx:14`; `DeveloperAvatar` moves along, its nesting is already right)
- `leaderboard/components/MetricTabs/` → `LeaderboardBoard/components/MetricTabs/` (sole importer `LeaderboardBoard.tsx:15`)
- `leaderboard/components/RangeTabs/` → `LeaderboardBoard/components/RangeTabs/` (both importers in the subtree: `LeaderboardBoard.tsx:16`, `LeaderboardBoard/utils/buildStandingsQuery.ts:2`)

**Follow-on to the StatStrip/TierTube promotion:** `TierBadge` moves to `app/components/TierBadge/` in the same change — its importers (`TierTube.tsx:1` via TIER_NAMES, `LeaderboardTable.tsx:7`) will span `app/` once TierTube is promoted, so `leaderboard/components/` stops being the shared parent.

**Desktop:**
- `LeaderboardJoinDialog/utils/formatTokens/` — sole importer is `RankTeaser.tsx:2`, and it's also a *third* drifted copy: pre-existing `usage/components/UsageHistorySection/utils/formatUsage` renders `850K` (Math.round) where this renders `850.0K` (toFixed(1)). Fold all three into the `packages/shared` formatTokens proposed in the formatUsage comment, and pick one rendering.
- `renderer/hooks/useLeaderboardAutoPublish/` → `routes/_authenticated/components/LeaderboardAutoPublish/hooks/` (sole importer `LeaderboardAutoPublish.tsx:1`; exact same-directory precedent: `AgentHooks/hooks/`, `GlobalBrowserLifecycle/hooks/`). `autoPublishSchedule`/`autoPublishState` + tests move with it. In the same pass, move `useLeaderboardOptIn` and `useLeaderboardJoinPreview` to `routes/_authenticated/hooks/` — every consumer is `_authenticated`-scoped and that directory holds exactly this kind of hook — so the trio stays together.

Everything else in the PR verified correctly placed: `LeaderboardBoard`, `buildStandingsQuery`, `StatsBody` + `Panel`, `Unavailable`, `OgStat`, `ShareButtons`, `getInterBold`, `constants.ts`, `RankTeaser`, `types.ts`, `lib/leaderboard`, `LeaderboardAutoPublish`/`LeaderboardFirstRunDialog`/`LeaderboardSection`/`LeaderboardPrompt`, and the join dialog's home in `renderer/components/`.

packages/trpc/src/router/leaderboard/schema.ts
```diff
@@ -0,0 +1,88 @@
+import { z } from "zod";
+import { LEADERBOARD_PERIODS } from "./periods";
+
+const dayKey = z.string().regex(/^\d{4}-\d{2}-\d{2}$/, "Expected YYYY-MM-DD");
+
+const tokenCount = z.number().int().min(0).max(Number.MAX_SAFE_INTEGER);
```
saddlepaddle: Integrity: the payload is entirely client-asserted and this cap makes spoofing free — `tokenCount` allows `MAX_SAFE_INTEGER` per field per row, so one curl with a session token claims 9 quadrillion tokens and takes #1. Two structural gaps compound it: `hostId` is a free-form client string forming part of the upsert key, so even a plausible per-day cap can be multiplied by fabricating hostIds; and the only moderation lever, `revokedAt`, is both loud (`requireParticipant` rejects the publish, telling the actor they're caught) and self-reversible (join's `onConflictDoUpdate` resets `revokedAt: null`, so a revoked user re-joins to un-ban themselves).

Minimum hardening for launch:
1. Cap `tokenCount` at a generous-but-sane per-day value (e.g. 50B) — also protects the `numeric`/bigint sums (see the usd overflow thread).
2. Bound hostIds per user (schema-level `hostId` allowlist isn't possible, but a per-user distinct-host cap at publish time is).
3. Add a `flaggedAt` column the board queries exclude but publish ignores — the shadow lever: accept the payload with a 200, never count it. Wire cheap ingest heuristics to set it (per-day tokens above threshold, usd/token ratio implausible for the named model, fresh account straight to top-10) rather than rejecting.
4. Split admin revocation from self-leave so `join` can only clear the latter.

Rate limiting (30 publishes/h/user) protects infra, not integrity — none of the above is covered by it.

packages/trpc/src/router/leaderboard/schema.ts
```diff
@@ -0,0 +1,88 @@
...
+	day: dayKey,
+	sessions: z.number().int().min(0).max(100_000),
+
+	parallelSessions: z.number().min(0).max(10_000),
+	agentPrsMerged: z.number().int().min(0).max(10_000),
+});
+
+export type PublishFactoryDay = z.infer<typeof publishFactoryDaySchema>;
+
+export const PUBLISH_MAX_DAYS = 5_000;
```
saddlepaddle: Two problems with this cap. It counts rows (day × provider × model), not days — the legitimate max is the 30-day join backfill × plausible combos, i.e. hundreds of rows, so 5,000 is ~10x oversized; and the top of the range is unusable anyway, since >~4,369 rows × 15 columns exceeds Postgres's 65,535 bind-parameter wire limit and the insert throws. Meanwhile nothing bounds the day *values*: `dayKey` is shape-only, so a client can publish arbitrarily backdated or even future days, which pollute `dayRangeStart/End`, all-time totals, and upcoming windows (a future-dated row pre-loads tomorrow's board).

Fix together: reject days outside `[today - 35, today]` server-side in publish, and drop `PUBLISH_MAX_DAYS` to ~2,000 (36 days × ~55 combos, comfortably under the wire limit — which also closes the bind-param crash).

packages/trpc/src/router/leaderboard/schema.ts
```diff
@@ -0,0 +1,88 @@
...
+export const metricSchema = z.enum(["tokens", "cost"]);
+
+export const standingsSchema = z.object({
+	period: periodSchema.default("30d"),
+	periodStart: dayKey.optional(),
+	from: dayKey.optional(),
+	to: dayKey.optional(),
+	metric: metricSchema.default("tokens"),
+	limit: z.number().int().min(1).max(100).default(50),
+	offset: z.number().int().min(0).max(100_000).default(0),
```
saddlepaddle: `offset` up to 100,000 lets the anonymous public API request arbitrarily deep scans of a grouped aggregate — OFFSET is O(n) in Postgres, so offset=100000 walks and discards 100k grouped rows per request, 60×/min/IP under the shared limiter. The board only pages sequentially via Load more; nobody legitimately reaches page 2,000. Drop to ~1,000 here and in the api route's `parseBoundedInt(params.get("offset"), 0, 0, 100_000)` mirror (both places), and revisit with keyset pagination only if the board ever actually grows past it.

apps/marketing/src/app/utils/fetchLeaderboard/fetchLeaderboard.ts
harshithmullapudi: Digging into the CORS caveat you flagged at the end — it's the blocker, so we're keeping REST for this PR and I'll take the three in-place comments instead (per-route limiter, shared zod `windowQuery`, `/u/` → `/user/`).

Marketing hits the leaderboard on two paths and they diverge:

- **Server-side (ISR)** — `fetchStandings`/`fetchStats`/`fetchParticipant` run during revalidation, server-to-server, no CORS. A tRPC client works fine here.
- **Browser-side** — `LeaderboardBoard` refetches on every metric/range change and `loadMore` paginates, both from the browser, cross-origin. That works today only because the REST handlers return `Access-Control-Allow-Origin: *`.

`/api/trpc` doesn't. `proxy.ts:16-24` reflects the origin only for `NEXT_PUBLIC_WEB_URL`, `NEXT_PUBLIC_ADMIN_URL`, `NEXT_PUBLIC_DESKTOP_URL` and trusted Vercel previews — marketing isn't on that list, so post-fold the board would server-render fine and then fail on every interaction, which is the same opaque-TypeError mode you flagged on `LeaderboardBoard.tsx`.

Adding it is one line, but `proxy.ts:36` sets `Access-Control-Allow-Credentials: true`, so `*` isn't available — it means putting the public marketing origin on the same *credentialed* allowlist the dashboard and desktop use, on an endpoint serving every session-bearing procedure in `appRouter`. That's a wider blast radius than the fold buys back, given the bare-500-without-CORS class you were structurally fixing is already handled by a `publicRoute` wrapper + `publicError` (the `publicJson` error sibling you suggested on the other thread).

Filing the fold as a follow-up with the allowlist decision as its first question. Also worth noting `createContext` does a session lookup per request, which the public reads would start paying.

packages/trpc/src/router/leaderboard/leaderboard.ts
```diff
@@ -0,0 +1,562 @@
...
+			prefix: "ratelimit:leaderboard:preview",
+		})
+	: null;
+
+async function enforce(
+	limiter: Ratelimit | null,
+	key: string,
+	message: string,
+): Promise<void> {
+	if (!limiter) return;
```
saddlepaddle, line 82: Per review discussion: fail-closed is the accepted policy for the write path (join/publish/preview) — but make it deliberate. This `await limiter.limit()` is unwrapped, so an Upstash outage surfaces as an unhandled 500, not a policy decision. Wrap it: on limiter *error* (not just `success: false`), throw the same `TOO_MANY_REQUESTS` (or `SERVICE_UNAVAILABLE`) so callers see a clean, retryable signal. For the anonymous GET surface (the api routes today, `leaderboard.public.*` after the consolidation), recommend the opposite — fail open: they're CDN-cached reads and a Redis blip shouldn't blank the public board. Either way, no limiter outage should ever produce an unhandled 500.

packages/trpc/src/router/leaderboard/schema.ts
saddlepaddle: Prioritization from review discussion — the cheap subset that covers most of the risk, all launch-sized:

1. Per-row caps: `tokenCount` → e.g. `.max(50_000_000_000)` per day-row field, `usdEstimate` → `.max(10_000)` (nobody spends $1M/day/model; 10k × 2,000 rows also keeps `recomputeTotals` inside `numeric(14,6)`). Pure zod number edits.
2. Day bounds `[today − 35, today]` — already flagged on `PUBLISH_MAX_DAYS`, doubles as anti-abuse (kills backdating/future-loading).
3. Distinct-host cap: one `count(distinct host_id)` per publish, reject past ~20. Closes the hostId multiplication.
4. The quiet lever: one `flaggedAt` (or `bannedAt`) column that `join` never touches and `requireParticipant` deliberately does NOT check — publishes keep succeeding — added only to the board-eligibility predicate (one line if the shared where-clause from the deletedAt thread lands). Manual `UPDATE` is enough moderation tooling at launch scale.

Deferred without guilt: automated heuristics/scoring — at launch, eyeballing the top 20 plus the manual flag column covers it.

packages/trpc/src/router/leaderboard/tier.ts
```diff
@@ -0,0 +1,194 @@
...
+
+	const dayTier =
+		promotable > previousTier
+			? promotable
+			: held < previousTier
+				? held
+				: previousTier;
+
+	const sustain = sustainTier(activeDays);
+	const tier = Math.min(dayTier, sustain) as Tier;
```
saddlepaddle, line 142: Low priority — flagged for awareness, deliberately not a launch blocker (per review discussion it may even be engagement-positive; see below).

Mechanism: the stored tier is `min(dayTier, sustainTier)`, and recompute feeds that capped value back as `previousTier`. So a user holding tier 4 via the demote hold (share ≥ 0.4) who dips below the sustain floor (e.g. a vacation) gets stored at 2 — and on return, re-reaching 4 requires the full promote share (≥ 0.6), not the 0.4 hold they'd been living on. The transient dip converts the hysteresis into a permanent demotion: the 40% hold silently becomes a 60% re-climb.

Product double edge: a re-climb can drive engagement (grinding a tier back is a loop), but this one is opaque — the drop is triggered by *absence*, not performance, lands after a vacation with no explanation, and the re-climb threshold is invisible. Motivating loops are legible; this reads as random loss.

If/when it's worth fixing: persist the pre-cap `dayTier` as the hysteresis state (`previousTier` input) and apply the sustain cap only to the stored/displayed tier — the hold then survives activity dips, and the sustain axis stays a display-time cap instead of leaking into the state machine.

apps/marketing/src/app/utils/fetchLeaderboard/fetchLeaderboard.ts
harshithmullapudi: Correcting my earlier reply on this thread: we went with your proposal after all. `NEXT_PUBLIC_MARKETING_URL` turned out to be already required by `packages/auth/src/env.ts:40` (which apps/api imports) and already in the `turbo.jsonc` env allowlist, so the CORS half was a one-line addition to `getAllowedOrigins`, not the widening I made it out to be. It is a real widening of the credentialed allowlist — an XSS on the marketing site becomes an API-session risk — but that is the property the dashboard origin already has, and superset.sh is our own mostly-static Next app. I overweighted it.

Shipped as described:

- `leaderboard.public.{standings,stats,participant}` on `publicProcedure`, IP-keyed `enforce` at 60/min reusing the `ratelimit:leaderboard:public` prefix, zod inputs (`windowSchema`, `standingsSchema`, `participantSchema`) replacing the hand parsers. `participant` throws `NOT_FOUND` instead of returning null.
- `responseMeta` on the api adapter gated exactly as you warned: bails unless `type === "query"`, `errors.length === 0`, `paths` is non-empty, and *every* path starts with `leaderboard.public.` — the non-empty check matters because `[].every()` is `true`. Stats keeps its 3600s TTL when the whole batch is stats, otherwise 300s. Added `vary: origin` since `proxy.ts` reflects the origin per-request.
- Marketing uses `createTRPCClient` + `httpBatchLink` with superjson and a custom fetch threading `next: { revalidate: 300 }`. Note `methodOverride` only accepts `"POST"` in @trpc/client 11.16 — queries are already GET with the input in the query string, so no override is needed for cacheable URLs.
- Deleted: all three `/api/leaderboard*` routes, `leaderboardResponse.ts`, and the hand-rolled `fetchLeaderboard` client. `LeaderboardBoard` no longer takes an `apiUrl` prop.
- `onError` now also suppresses `NOT_FOUND` for `leaderboard.public.participant`, since an unknown handle is normal crawler traffic.

Your three other comments on that surface are moot as you predicted; replying on each.

apps/marketing/src/app/components/TokenSplitBar/TokenSplitBar.tsx
```diff
@@ -0,0 +1,65 @@
...
+	const segments: Segment[] = [
+		{ label: "Input", tokens: split.uncachedInput, color: "#d25611" },
+		{ label: "Output", tokens: split.output, color: "#c19a5b" },
+		{ label: "Cache read", tokens: split.cachedInput, color: "#6b8ca3" },
+		{
+			label: "Cache write",
+			tokens: split.cacheWrite5m + split.cacheWrite1h,
+			color: "#7a9e7e",
+		},
+	];
```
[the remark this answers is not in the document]
harshithmullapudi: Checked this one against the parser and I do not think it holds — `reasoningOutput` is a **subset** of `output`, not a sibling, so adding it as its own segment would double-count.

`packages/host-service/src/trpc/router/usage/history/parse.ts`:

- L185-186 (Anthropic): `output: num(usage.output_tokens)`, `reasoningOutput: num(usage.output_tokens_details?.thinking_tokens)` — thinking tokens are a breakdown *of* `output_tokens`.
- L292 (OpenAI): `output: num(usage.output_tokens)`, `reasoningOutput: num(usage.reasoning_output_tokens)` — same relationship.

That is also why `leaderboard.ts` computes `tokens` as `uncachedInput + cachedInput + cacheWrite5m + cacheWrite1h + output` with `reasoningOutput` deliberately excluded; including it would inflate every total.

So the premise that "reasoning-only usage is shown as no usage" is not reachable: reasoning-only usage still carries `output_tokens > 0` and renders inside the Output segment. Same for the join preview — adding `day.reasoningOutput` to that sum would make the previewed rank inflated relative to the server`s `tokens` column, which is the bug the `period: "month"` → `"30d"` fix on the sibling thread was closing.

Leaving both as-is. Happy to reopen if the subset relationship is wrong for a provider I have not checked.

apps/api/src/lib/leaderboardResponse.ts
harshithmullapudi: Moot as you called it — we took the tRPC fold, so `leaderboardResponse.ts` is deleted entirely along with the limiter that lived in it. The rate limit is now IP-keyed `enforce` inside `leaderboard.public.*`, reusing the same `ratelimit:leaderboard:public` prefix and 60/min budget, next to the existing publish/join/preview limiters in the router.

apps/api/src/lib/leaderboardResponse.ts
harshithmullapudi: Superseded by the fold — the hand-rolled `parseX` helpers are gone with `leaderboardResponse.ts`. The shared shape you sketched now lives as `windowSchema` in `packages/trpc/src/router/leaderboard/schema.ts`, with `standingsSchema` and `participantSchema` extending it, validated by tRPC at the procedure boundary. Real-calendar-date validation went in as a shared `isDayKey` in `periods.ts` (regex + `Date` round-trip) used by the `dayKey` refine, so the `2026-13-45` → `RangeError` path you flagged on the routes is closed at the schema.

Your `publicJson` note landed too: I had added a `publicError` sibling and a `publicRoute` wrapper before the fold, and the fold subsumed both — errors now flow through the shared tRPC handler with CORS from `proxy.ts`, which is the structural fix you were after.

apps/api/src/app/api/leaderboard/u/[handle]/route.ts
Comment (harshithmullapudi) [reply apps/api/src/app/api/leaderboard/u/[handle]/route.ts]: Renamed to `/api/leaderboard/user/[handle]` first, then the whole route went away with the fold — the profile read is now `leaderboard.public.participant`. Good catch either way; the abbreviation would have shipped.

packages/trpc/src/router/leaderboard/schema.ts
harshithmullapudi: Both taken. `PUBLISH_MAX_DAYS` → 2,000 with a comment recording that it counts rows and why the number is what it is (36 days x ~55 combos, and the 15-columns-per-row bind-parameter ceiling you worked out).

Day values are now bounded in `publish` via `assertDaysInWindow`, applied to both `days` and `factoryDays`, with `PUBLISH_WINDOW_DAYS = 35`.

One deviation: I allowed one day of *forward* slack rather than a hard `<= today`. Day keys are UTC (`toISOString().slice(0, 10)` in `groupEntriesByDay`), so the client and server should agree — but if a publish crosses a UTC midnight with the host clock slightly ahead of the API`s, a legitimately-today row would be rejected and take the whole publish with it. A one-day window still blocks the pre-loading you flagged (a spoofer gains at most the current day, which they can already write), while a strict bound would fail real publishes for seconds around midnight every day. Happy to tighten if you would rather eat that.

Rejecting the whole payload rather than filtering the offending rows was deliberate: silently dropping them would hide exactly the case worth seeing.

packages/trpc/src/router/leaderboard/schema.ts
harshithmullapudi: Done — `offset` max is now 1,000, with a comment on why (O(n) OFFSET over a grouped aggregate on an anonymous endpoint; the board only pages sequentially). The `parseBoundedInt` mirror you asked me to change in the api route is gone entirely — that route was deleted in the tRPC consolidation, so the zod schema is now the single place this is bounded.

packages/trpc/src/router/leaderboard/leaderboard.ts
harshithmullapudi: Implemented as recommended, split both ways.

`enforce` (write path — join/publish/preview) now wraps `limiter.limit()` and turns a limiter *error* into `SERVICE_UNAVAILABLE` with the same message, logged. Fail-closed, but as a stated policy rather than an unhandled 500.

`enforcePublicRead` (`leaderboard.public.*`) fails open: on limiter error it logs and returns, so a Redis blip cannot blank the public board. It still throws `TOO_MANY_REQUESTS` on a genuine `success: false`. Timing note — this landed alongside the consolidation, so the anonymous GET surface you mentioned is the tRPC public procedures now, not the api routes.

packages/trpc/src/router/leaderboard/tier.ts
harshithmullapudi: Agreed on the mechanism and leaving it open rather than resolving — this is a state-machine change to how `previousTier` is persisted, which is beyond the bug/structure scope for this PR, and you have marked it non-blocking.

Recording the fix so it is not lost: persist the pre-cap `dayTier` as the hysteresis input and apply the sustain cap only to the stored/displayed tier, so the 0.4 demote hold survives an activity dip instead of silently becoming a 0.6 re-climb.

Your framing of why it reads badly is the part worth keeping — the drop is triggered by absence rather than performance, and the re-climb threshold is invisible. Will raise it as a follow-up before the tier system is user-visible enough for anyone to hit a vacation.

apps/marketing/src/app/leaderboard/components/LeaderboardBoard/components/LeaderboardTable/LeaderboardTable.tsx
harshithmullapudi: Done, all of it. Also promoted `leaderboardClient` (added by the tRPC consolidation) to `app/utils/` alongside the others so it did not become a fresh instance of the same problem.

Nesting under `LeaderboardBoard`: `LeaderboardTable` (with `DeveloperAvatar`), `MetricTabs`, `RangeTabs`. `TierBadge` went to `app/components/` with the `TierTube` promotion, as you predicted it would need to.

Desktop: `useLeaderboardAutoPublish` → `LeaderboardAutoPublish/hooks/` with `autoPublishSchedule`/`autoPublishState` and the test; `useLeaderboardOptIn` and `useLeaderboardJoinPreview` → `_authenticated/hooks/`. One thing your sweep did not mention that the move required: `persisted-key-registry.test-data.ts:146` pins the writer path for `leaderboard-auto-publish-v1`, so it needed updating in the same change or CI would have failed the unregistered-writer check.

`LeaderboardJoinDialog/utils/formatTokens` is deleted — folded into `@superset/shared/format-tokens` along with the marketing and usage-history copies, replying on that thread with which rendering won.

Converted the cross-directory imports to `@/app/...` rather than deep relative paths while moving them, so the next move does not churn call sites.

apps/marketing/src/app/components/StatStrip/StatStrip.tsx
harshithmullapudi: Done as part of the full sweep — `StatStrip`, `TierTube`, `TokenSplitBar`, `ModelBars` (with `buildModelColors`/`to*Rows`) and `TierBadge` are in `src/app/components/`, and `formatUsage`, `avatarUrl`, `fetchLeaderboard`, `formatRangeLabel` and `leaderboardClient` are in `src/app/utils/`. `/stats` and `/user/[handle]` no longer reach into `@/app/leaderboard/...`. Details on the single-consumer half in the LeaderboardTable thread.

apps/marketing/src/app/leaderboard/components/MetricTabs/MetricTabs.tsx
harshithmullapudi: Added `LeaderboardBoard/components/PillTabs/` and rebuilt both rows on it, so there is one style and one a11y story.

I standardised on the RangeTabs treatment (individually-bordered pills) rather than the MetricTabs one (segments in a single bordered container), because RangeTabs has to wrap and sit inline with the Custom popover trigger — which already renders as a bordered pill — so a shared segmented container would have broken that layout. MetricTabs therefore changes visually; that seemed like the point.

`PillTabs` owns the `role="tablist"`/`aria-label` wrapper and renders each option as `role="tab"` with `aria-selected`, and takes `children` so the Custom popover trigger sits inside the same tablist with `role="tab"`/`aria-selected={customActive}` — previously it had no tab semantics at all. Active state is driven by `value: T | null`, so RangeTabs passes `customActive ? null : value.period` instead of needing a predicate.

packages/trpc/src/router/leaderboard/leaderboard.ts
```diff
@@ -0,0 +1,673 @@
...
+		}),
+
+	leave: protectedProcedure.mutation(async ({ ctx }) => {
+		const userId = ctx.session.user.id;
+		await db
+			.delete(leaderboardDaily)
+			.where(eq(leaderboardDaily.userId, userId));
+		await db
+			.delete(leaderboardDailyFactory)
+			.where(eq(leaderboardDailyFactory.userId, userId));
```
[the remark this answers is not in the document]
harshithmullapudi: Fixed — `leave` now runs all three deletes inside `dbWs.transaction`. Confirmed the premise: `packages/db/src/client.ts:18` builds `db` on `drizzle-orm/neon-http` while `dbWs` (L24) is the `Pool`-backed client, and `dbWs.transaction` is the existing pattern in `accept-invitation-endpoint.ts:158` and `cloud-workspace/provision.ts`.

Not adding the rollback test you suggested: there is no DB-backed test harness for this router (no existing test file for `leaderboard.ts`), so it would mean standing one up for a single assertion. Flagging rather than silently skipping.

apps/desktop/src/renderer/routes/_authenticated/hooks/useLeaderboardOptIn/useLeaderboardOptIn.ts
```diff
@@ -0,0 +1,97 @@
...
+				if (activeHostUrl && machineId) {
+					try {
+						published = (
+							await publishUsage(activeHostUrl, machineId, BACKFILL_DAYS)
+						).days;
+						writeAutoPublishState({
+							handle,
+							lastPublishedAt: Date.now(),
+							lastPayloadHash: null,
+						});
```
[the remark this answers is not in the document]
harshithmullapudi: Fixed. Added `clearAutoPublishState()` and call it in `leave` right after the mutation succeeds, before the membership refetch.

Worth noting this is the mirror of the handle-scoping change on the other thread: that one stops a *different* handle inheriting stale state, but as you spotted it does nothing when the same handle rejoins, because the stored handle still matches. Clearing on leave closes that half — a later enrollment now starts from `lastPublishedAt: 0`, which `publishWindowDays` treats as a full 30-day backfill.

packages/shared/src/format-tokens.ts
```diff
@@ -0,0 +1,8 @@
+/** "1.24T", "13.9B", "4.2M", "850K", "312" */
+export function formatTokens(tokens: number): string {
+	if (tokens >= 1e12) return `${(tokens / 1e12).toFixed(2)}T`;
+	if (tokens >= 1e9) return `${(tokens / 1e9).toFixed(1)}B`;
+	if (tokens >= 1e6) return `${(tokens / 1e6).toFixed(1)}M`;
+	if (tokens >= 1e3) return `${Math.round(tokens / 1e3)}K`;
```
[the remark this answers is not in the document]
harshithmullapudi: Real bug in code I had just written, thanks. Both cases confirmed: `999_999` gave `1000K` and `999_999_999` gave `1000.0M`.

Rewrote `formatTokens` over a `UNITS` table that re-checks after rounding — if the scaled value rounds to 1000 or more it steps up to the next unit, so those now read `1.0M` and `1.0B`. Added the boundary tests you asked for, including `999_999_999_999` → `1.00T` and the non-carrying neighbours (`1_000` → `1K`, `999_400` → `999K`) so a future change cannot fix one direction by breaking the other.

packages/trpc/src/router/leaderboard/schema.ts
harshithmullapudi: Taken, with the DB half folded into `0094` rather than shipped as a new migration — `0094` only exists on this unmerged branch (`origin/main` does not contain it), so it has never been applied and the column joins the original `CREATE TABLE`. Regenerated via `drizzle-kit generate` after deleting the old file/snapshot/journal entry rather than hand-editing the SQL; diffed the result against the original to confirm the only changes are the intended ones.

Against your four points:

1. **Token cap** — `MAX_TOKENS_PER_ROW_FIELD = 50e9` per field per row. Sized so a full payload (2,000 rows x 5 fields) stays far under the `bigint` rollup at `leaderboard.ts:178`. While checking this I found the overflow is not just the `usd` one you referenced: at `MAX_SAFE_INTEGER` per field, 2,000 rows summed to ~9e19 against a `bigint` ceiling of ~9.22e18, so the rollup `UPDATE` was a reachable 500, not only an integrity problem. Also capped `usdEstimate` at $10k/row, and widened `usd` to `numeric(20,6)` (and `usd_estimate` to `numeric(14,6)`) in `0094` — `numeric(14,6)` topped out near $99.99M, which a year of legitimate rows was closer to than I was comfortable with.
2. **Host cap** — `MAX_HOSTS_PER_USER = 10`, counted as distinct `hostId`s excluding the incoming one, enforced in publish.
3. **`flagged_at`** — added to `leaderboard_participants` and folded into the `onTheBoard` predicate, which covers standings, stats, tier distribution and profiles in one place, plus `p.flagged_at is null` in `rankFor`. `me` deliberately does not filter on it, so the participant still sees themselves enrolled and publish still returns 200 — the shadow property you wanted. The ingest heuristics that would *set* it are not wired up; the lever exists but nothing pulls it yet.
4. **Split revocation from self-leave** — turned out to need no schema change. `leave` deletes the participant row outright, so `revoked_at` is only ever set by an admin; the entire fix was removing `revokedAt: null` from `join`'s `onConflictDoUpdate`. A revoked user rejoining no longer un-bans themselves.

Leaving this thread open until the flagging heuristics land, since point 3 is only half-delivered.

packages/trpc/src/router/leaderboard/leaderboard.ts
```diff
@@ -0,0 +1,700 @@
...
+			),
+		);
+
+	if (Number(seen?.hosts ?? 0) >= MAX_HOSTS_PER_USER) {
+		throw new TRPCError({
+			code: "BAD_REQUEST",
+			message: "Too many machines publishing for this account.",
+		});
+	}
+}
```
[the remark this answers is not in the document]
harshithmullapudi: Both halves confirmed and fixed.

**Factory bypass** — verified: `leaderboard_daily_factory` carries `host_id` and it is part of `leaderboard_daily_factory_identity_key`, so a factory-only payload wrote a host ID my check never counted. `enforceHostBudget` now counts the union:

```sql
select count(*)::int as hosts from (
  select host_id from leaderboard_daily        where user_id = $1 and host_id <> $2
  union
  select host_id from leaderboard_daily_factory where user_id = $1 and host_id <> $2
) hosts
```

**Race** — the check and both inserts now run inside one `dbWs.transaction`, and the transaction opens with `select 1 from leaderboard_participants where user_id = $1 for update`. The row lock is what actually serialises it; a transaction alone would not, since READ COMMITTED lets concurrent publishes both observe a count under the cap before either writes. `recomputeTotals`/`recomputeTier` stay outside — they are idempotent recomputes over committed state.

Note `db` is the neon-http client and cannot transact, so this uses `dbWs`, same as the `leave` fix on the other thread.

packages/trpc/src/router/leaderboard/schema.ts
```diff
@@ -0,0 +1,118 @@
+import { z } from "zod";
+import { isDayKey, LEADERBOARD_PERIODS } from "./periods";
+
+const dayKey = z.string().refine(isDayKey, "Expected a real YYYY-MM-DD date");
+
+// Per field, per (day, provider, model, host) row. Generous next to real usage,
+// but low enough that a full payload cannot overflow the bigint rollup on
+// leaderboard_participants.
+export const MAX_TOKENS_PER_ROW_FIELD = 50_000_000_000;
```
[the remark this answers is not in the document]
harshithmullapudi: Confirmed the arithmetic — 2,000 rows x 5 fields x 50e9 is 5e14 per payload, and since `provider`/`model` are free-form the upsert key is unbounded, so ~19 payloads clear `MAX_SAFE_INTEGER`. Day keys and host IDs are bounded now, but those two are not.

Took the second option you offered — preserve precision rather than add a cumulative cap. `recomputeTotals` was `select sum(...)::bigint` followed by `Number(agg.tokens)` in the update, so the exact sum was rounded in JS before it ever reached the column. It is now a single statement:

```sql
update leaderboard_participants p set tokens = t.tokens, ... 
from (select coalesce(sum(d.tokens), 0)::bigint as tokens, ... 
      from leaderboard_daily d where d.user_id = $1) t
where p.user_id = $1
```

The arithmetic never leaves Postgres, which fixes it for all seven bigint columns at once rather than just `tokens` — `uncached_input`, `cached_input`, both cache-write columns, `output` and `reasoning_output` had the same conversion. Ordering and comparisons were already exact in SQL, so standings are correct even at values a JS number could not hold.

Deliberately not adding a cumulative cap on top: a legitimate ceiling is guesswork, and a hard reject would break a real heavy user before it inconvenienced anyone fabricating model names. Detecting implausible accumulation belongs with the ingest heuristics behind `flagged_at`, which is the one thread I have left open (3859049801).

apps/marketing/src/app/utils/formatUsage/formatUsage.ts
```diff
@@ -0,0 +1,27 @@
...
+	if (!Number.isFinite(value)) return "$0";
+	if (value >= 1e6) return `$${(value / 1e6).toFixed(2)}M`;
+	if (value >= 1e3) return `$${(value / 1e3).toFixed(1)}K`;
+	return `$${value.toFixed(2)}`;
+}
+
+export function formatCount(value: number): string {
+	if (value >= 1e6) return `${(value / 1e6).toFixed(1)}M`;
+	if (value >= 1e3) return `${(value / 1e3).toFixed(1)}K`;
+	return value.toLocaleString("en-US");
```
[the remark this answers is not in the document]
harshithmullapudi: Both confirmed — `formatUsd(999_999)` gave `$1000.0K` and `formatCount(999_999)` gave `1000.0K`.

Rather than add the carry check to each, I extracted it: `@superset/shared/format-scaled` takes a unit table and a base formatter, picks the largest unit the value clears, then re-checks the *rounded* result and steps up if it reached 1000. `formatTokens`, `formatUsd` and `formatCount` all run through it now, so there is one implementation instead of the fourth hand-written copy — which is how these three drifted apart in the first place (you flagged the same bug in `formatTokens` two reviews ago).

`formatUsd(999_999)` → `$1.00M`, `formatCount(999_999)` → `1.0M`.

Boundary tests added for both formatters and for the helper, covering the carry cases, the neighbours that must *not* carry (`1_000` → `1.0K`, `999_400` → `999.4K`), and the top unit having nothing to carry into. `apps/marketing` had no test setup at all, so this also needed `bun-types` in its tsconfig and devDependencies, matching desktop and shared.

One case deliberately left alone: below the smallest unit the value goes to the base formatter untouched, so `formatUsd(999.999)` renders `$1000.00` rather than `$1.0K`. Unscaled, but not misleading the way `1000K` is.

saddlepaddle (approved):
Reviewed all 47 review threads and spot-verified the fixes against 0c7a8a1 — resolutions are real. Two threads deliberately left open as follow-ups (flagging heuristics, tier hysteresis), both non-blocking.

PR #6878, opened 2026-08-26
feat(desktop): redesign browser overflow menu, always-visible Chrome import banner

AviPeltz:
## Summary
- The in-app browser pane's "..." menu now matches Chrome's own page menu — grouped sections instead of a flat list, with **Find in page**, **Print**, **Zoom** (+/- with live %), and a real **device toolbar** (viewport emulation via Electron's `enableDeviceEmulation`, with rotate) added alongside the regrouped existing items (Take a screenshot, Hard reload, Copy URL, Open in Browser).
- New **Downloads** tracking end-to-end: a `will-download` listener, a `downloads` local-db table + migration, and a dialog with cancel/open/show-in-folder — verified live with a real in-flight download (progress → cancel → cleanup).
- **Signed-in sites** (cookie-domain list — this app has no credential vault, so "logins" are session cookies, not passwords) and **History** search now filter+cap instead of rendering/searching unbounded lists; a real profile with 1000+ cookie domains was visibly slow to open before this.
- The Chrome/Chromium import banner now shows above any loaded page (not just the blank new-tab state), is dismissed per-install rather than per-pane (the pane component is reused unkeyed across tab switches, which was both resetting and bleeding the dismissal), and names the actually-detected browser instead of hardcoding "Chrome".
- Fixed a Radix DropdownMenu + Dialog interaction where opening History/Downloads/Clear-data/Import from a menu item flashed the dialog open and immediately closed it.

## Known limitation
`webview.findInPage()` never fires `found-in-page` in this app's environment (confirmed via a raw API call bypassing all app code — likely an Electron/platform gap, not app-code wiring). The find bar degrades honestly (no fake match counts) instead of pretending to work; tracked separately for investigation.

## Test plan
- [x] `bunx tsc --noEmit -p apps/desktop/tsconfig.json` — clean on all touched files
- [x] `bunx biome check` — clean on all touched files
- [x] `bun test` — `browser-manager.test.ts`, `browserRuntimeRegistry.test.ts`, persisted-keys registry test all pass
- [x] Verified live via CDP against a running dev build: import banner on a live page, empty-page copy, full menu redesign, zoom, device toolbar + rotate + close, History/Downloads/Clear-browsing-data dialogs opening and staying open, Signed-in sites search+cap (1015 → 20 rendered rows), a real download's full progress→cancel→cleanup lifecycle
- [ ] Manual smoke test of Print / Copy URL / Open in Browser (not exercised live — they pop native OS dialogs / the user's default browser / overwrite the clipboard)

https://claude.ai/code/session_017dQtKEkDMveWJW1Fj3jztf

## Summary by CodeRabbit

* **New Features**
  * Added browser downloads with progress tracking, cancellation, file opening, folder access, and history management.
  * Added screenshot capture, previews, file access, and screenshot management.
  * Added find-in-page, printing, zoom controls, device emulation, and responsive device presets.
  * Added browser history search and browsing-data management.
  * Added signed-in site management with per-domain cookie clearing.
  * Added a persistent browser import banner with dismissible prompts.
  * Browser history searches can now return up to 200 results.
* **Bug Fixes**
  * Improved import success reporting and handling of partial or unsuccessful imports.
  * Downloads now recover gracefully after application restarts.


AviPeltz:
Reviewed CodeRabbit's findings from the latest pass. 3 were already resolved by the prior fix commit (findNext direction, savePath→id resolution, Promise.allSettled) and CodeRabbit's own bot already marked those "Addressed in commit fb214ea".

The remaining 4 were all valid and are fixed in 69a2d5399:

- **Zoom row + "forget site" were keyboard-unreachable** — both were plain buttons/nested-button-in-item that Radix's dropdown never gives keyboard focus to (arrow keys only move between registered menu items, and Tab closes the menu). Zoom is now one `DropdownMenuItem` with Left/Right/Enter driving it; forget-site now fires from the item's `onSelect` so Enter/Space on the row works, not just a mouse click on the icon.
- **Failed `clearCookiesForDomain` was silently swallowed** — a site that failed to "forget" just sat there with zero feedback. Now shows an error toast, matching the pattern already used in `ClearBrowsingDataDialog`.
- **`DEVICE_PRESETS`/`DEFAULT_DEVICE_PRESET` lived under `DeviceToolbar/` but were consumed by 3 files** (`DeviceToolbar.tsx`, `BrowserPane.tsx`, `deviceToolbarStore.ts`) — moved to `BrowserPane`'s existing shared `constants.ts` per the repo's co-location convention.
- **`DownloadsDialog`'s `DownloadRow` interface duplicated `SelectDownload`** — swapped to import the generated type directly from `@superset/local-db` so the two can't drift.

Typecheck and biome are clean on all touched files; existing browser-pane tests still pass.

AviPeltz:
Checked the latest CodeRabbit review (5028389703) — it covers the commits up through 69a2d539 and had 3 "outside diff range" findings not caught by the earlier line-level pass. All 3 were valid and are fixed in 3292058:

- **Device toolbar clamped the emulated viewport to the pane size** — `maxWidth`/`maxHeight: 100%` on the device placeholder shrank a preset that didn't fit rather than letting the existing `overflow-auto` container scroll to it, so the page could report running at (e.g.) iPad Air dimensions while actually rendering smaller. Removed the clamp.
- **A zero-result import still counted as "imported something"** — a history/cookie import that completed but wrote 0 records permanently dismissed the banner and showed a success toast for a run that changed nothing. Now `importedSomething` requires `result.imported > 0`, consistent with the earlier Keychain-denied fix.
- **Concurrent downloads could race onto the same save path** — `resolveSavePath` only checked the filesystem, but Electron doesn't create the file until the transfer starts writing, so two downloads picked close together could resolve to the same "next available" name and one would clobber the other. Now reserves the path in memory for the download's lifetime and releases it on completion.

Note: the review's 6 line-level findings from the prior pass are still shown as unresolved threads even though CodeRabbit's own bot already marked them "✅ Addressed" in comments — that's just it not auto-collapsing, not a re-flag.

Typecheck, biome, and existing tests are clean. Not pushed yet — let me know if you want this round pushed too.

PR #6897, opened 2026-08-26
fix(host-service): stop a busy pty-daemon from being orphaned by a fresh spawn

Kitenite:
Ships the parked orphaned-daemon fix (was sitting unpushed on `fix/ptyd-orphaned-by-unlink-spawn`) now that SUPER-2043 gives it field evidence: Clinton's `ps` shows a v1.23.0 `pty-daemon.js` + `terminal-host.js` still running days later alongside the v1.24.2 app. Companion to #6895 (the renderer-side freeze fix) — this one is the machine-wide CPU-leak half of that report.

## The bug

`Server.listen()` in pty-daemon unlinks the socket path unconditionally before binding. POSIX does not close the incumbent listener on unlink: it stays bound to a now-unreachable inode, keeps its entire PTY tree (agents, dev servers, watchers) alive and burning CPU, and raises no `EADDRINUSE` — so nothing downstream ever notices. The only guard is adoption, and `tryAdoptFromSocket` conceded whenever the hello probe failed, even though it had just proven the socket connectable one line earlier. A CPU-starved daemon is exactly the one that misses the 3s probe budget, making the leak self-reinforcing: high CPU → probe timeout → spawn over the live daemon → one more tree burning CPU → the next probe is likelier to time out.

## The fix

- **`tryAdoptFromSocket`**: when the ordinary 3s probe comes back silent on a socket that was accepting, escalate — 8s total, 4s per attempt — before conceding. The per-attempt cap matters as much as the total: retrying a 1.5s attempt can never adopt a daemon that uniformly needs longer than 1.5s per connection.
- **`runUpdate`**: always write the successor manifest. Skipping it when the predecessor had none (socket-adopted) left the successor invisible to the next boot's `tryAdopt`, forcing the very socket-adopt path that orphans.
- **`spawn`**: log `pty_daemon_spawn_over_live_socket` when spawning while a still-accepting socket holds the path. The event names the hazardous *condition*, not the orphaning outcome — once the daemon-side bind-refusal fix (in flight separately) lands, the same condition ends with the child standing down, and the event keeps marking the adopt-then-spawn disagreement worth investigating.

## What changed vs. the first cut (the reason it was parked)

The original version added a ~17s worst-case stall to terminal readiness when a daemon is wedged-but-accepting (1s connect check + 3s probe + 1s re-check + 12s escalation). This version bounds that at **~11s**:

- escalated budget 12s → **8s** (still two full 4s attempts);
- the serial `isSocketConnectable` re-check is gone — the escalated retry loop itself observes per-attempt connects and **stops immediately when connects are refused** (`stopWhenNoListener`), so a daemon that dies mid-probe costs ~50ms of escalation instead of the whole budget;
- the handoff path keeps the old retry-through-refused behavior it depends on (predecessor-exit → successor-bind gap).

The stall only ever applies to the wedged-daemon case, where the pre-fix behavior was strictly worse: instant spawn-over, orphaning the incumbent and all its sessions forever. Boot is unaffected (`ensure()` is fire-and-track); terminal attaches surface the wait as the existing "Terminal not ready — retrying" state.

## Tests

- node-test regression (from the original commit): a manifest-less listener answering hello after 2.5s is adopted, not spawned over — asserted by pid + incumbent still listening. Mutation-checked (fails with escalation disabled).
- New unit tests for `stopWhenNoListener`: refused connects end the escalated loop immediately; silent-but-accepting listeners still get the whole budget; without the flag the handoff behavior (retry through refused) is preserved; per-attempt cap adopts a hello slower than the default attempt. Mutation-checked.
- Full `DaemonSupervisor.test.ts` 57/57; `DaemonSupervisor.node-test.ts` 18/19 — the one failure ("auto-update defers when live sessions exist") reproduces identically on pristine origin/main (stale since #5833, deliberately changed behavior) and is untouched here.

---
## Summary by cubic
Fixes a busy pty-daemon being orphaned by a fresh spawn. Old behavior: when the hello probe timed out, adoption conceded and `Server.listen()` unlinked the socket path, leaving the incumbent daemon bound to an unreachable inode with its whole PTY tree alive and burning CPU. New behavior: adoption escalates to an 8s budget with a 4s per-attempt cap before conceding, and the successor manifest is written even when the predecessor had none.

**Behavior and testing**
- Escalation logs `pty_daemon_socket_adopt_escalated` and stops only when a connect is definitively refused (ECONNREFUSED/ENOENT), so a daemon that dies mid-probe costs about one attempt instead of the whole budget; a timed-out connect still pending is treated as a flooded-but-live daemon and keeps the remaining budget.
- Worst-case terminal readiness stall behind a wedged-but-accepting daemon is ~11s; boot is unaffected and terminal attaches surface the wait as "Terminal not ready — retrying".
- Spawning over a still-accepting socket logs `pty_daemon_spawn_over_live_socket`.
- The pre-existing `auto-update defers when live sessions exist` node-test failure reproduces identically on `main` and is untouched.

<sup>Written for commit ddbd06705040df786d38e9d4bc554a07f096c711. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6897?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved daemon recovery when responses are delayed, helping healthy background processes remain available instead of being unnecessarily replaced.
  - Prevented active daemon sockets from being overwritten without recording the event.
  - Ensured successor daemon state is consistently written during updates.
  - Improved handling of unavailable sockets by stopping recovery attempts promptly when no daemon is listening.
- **Tests**
  - Added coverage for delayed responses, unavailable sockets, and adopting existing daemons under degraded conditions.

## Relationship to the daemon-side structural fix (in flight)

A structural version — `Server.listen()` connect-probing and refusing to unlink a live socket, plus spawn-time socket-ownership verification — is being prepared separately and supersedes the escalation as the root-cause fix. This PR still earns its keep alongside it: the daemon-side refusal only runs in the newly bundled binary, so the supervisor-side guard here is what covers an older installed app version that launches later against the shared socket, and any path where a new host-service meets a socket it cannot identify. The rebase plan is agreed with that work's author (escalation trims to one 4s attempt once the structural guard lands).

### DIFF
--- a/packages/host-service/src/daemon/DaemonSupervisor.node-test.ts
+++ b/packages/host-service/src/daemon/DaemonSupervisor.node-test.ts
@@ -22,7 +22,7 @@ import {
 	encodeFrame,
 	FrameDecoder,
 } from "@superset/pty-daemon/protocol";
-import { DaemonSupervisor } from "./DaemonSupervisor.ts";
+import { DaemonSupervisor, ptyDaemonSocketPath } from "./DaemonSupervisor.ts";
 import { EXPECTED_DAEMON_VERSION } from "./expected-version.ts";
 import {
 	type PtyDaemonManifest,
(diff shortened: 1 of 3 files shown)


Kitenite:
Re CodeRabbit's outside-diff finding ("do not spawn over a connectable socket"): agreed on the end-state, deliberately out of scope here. Refusing to spawn entirely needs the daemon-side structural fix — `Server.listen()` connect-probing and refusing to unlink a live socket, spawn-time ownership verification, and a forceRestart path that can kill an unadopted socket holder (otherwise a truly deadlocked daemon strands terminals with no recovery). That fix exists and is being rebased over this PR by a parallel workstream; until it lands, spawning after the exhausted escalation is the intentional interim behavior, and `pty_daemon_spawn_over_live_socket` telemetry marks every occurrence.

PR #6904, opened 2026-08-27
feat(i18n): Lingui 6 foundation across desktop, web, and mobile (Phase 0)

Kitenite:
## What

Phase 0 of the internationalization strategy (`plans/20260826-i18n-strategy.md`, full research + phasing in the plan doc): one shared message catalog wired into every build pipeline, proving the library decision before any bulk migration.

- **`packages/i18n`**: Lingui 6.6.0, `.po` catalog (committed compiled `messages.ts`), `SUPPORTED_LOCALES`/`resolveLocale`/`inferLocale`, shared `I18nProvider`.
- **Wired into all five pipelines** with 14 real messages (incl. one ICU plural):
  - desktop renderer: Babel macro plugin via `@vitejs/plugin-react` (theme-import toasts)
  - desktop main: plain `i18n._({id, message})` descriptors — the extractor picks these up, no Babel needed in main (tray labels)
  - web / marketing / docs: `@lingui/swc-plugin` via `experimental.swcPlugins` under Next 16 Turbopack
  - mobile: macro plugin in `babel.config.js`, device locale from the already-installed `expo-localization`
- **First-load language inference**: persisted setting (Phase 1) → OS/browser locale → `en`. Desktop main uses `app.getPreferredSystemLanguages()`, renderer/web `navigator.languages`, mobile `getLocales()`.
- **Enforcement**: `lingui extract` on `pretypecheck` (drift fails typecheck), new **i18n catalogs** step in the CI lint job (`extract --clean` + `compile --strict` + clean `git diff`).
- Root `lingui.config.ts` re-exports the package config: the Babel/Metro plugins discover config by upward search from each app's cwd and never reach `packages/i18n` otherwise.

## Verification

- turbo typecheck 38/38 · biome lint clean · sherif clean · turbo test 19/19
- docs `next build` passes end-to-end (proves the SWC plugin under Next 16.2 Turbopack, incl. prerendering the macro-using Sidebar)
- desktop `electron-vite` build: compiled message ids present in renderer bundle, zero residual `@lingui/react/macro` imports
- mobile `expo export`: Hermes bundle contains compiled ids, no macro imports
- marketing dev server boots clean with the SWC plugin + `proxy.ts` (the js-lingui#2423 Next 16 failure mode does not reproduce on 6.6.0/16.2.11)
- web/marketing local builds fail on env (`neon()`, t3-env) identically **without** this change; their Turbopack compile stages pass

## Notes

[...]

PR #6905, opened 2026-08-27
feat(i18n): locale-aware formatting layer (Phase 1)

Kitenite:
Stacked on #6904. First Phase 1 PR from `plans/20260826-i18n-strategy.md`.

## What

- **`@superset/i18n/format`**: memoized `Intl` wrappers reading the active locale from the shared i18n instance — `formatNumber`, `formatPercent`, `formatCompactNumber`, `formatCurrency`, `formatPrice` (Stripe minor units), `formatDate`, `formatDateTime`. Formatter instances are cached per locale+options since several call sites run in render paths.
- **Replaces every client-side hardcoded `"en-US"` formatter**:
  - `packages/ui` ai-elements `context.tsx` — 10 sites (percent / compact / currency)
  - `apps/mobile` ai-elements fork — 10 sites
  - desktop billing: three duplicate local cents→currency helpers (`usePaymentFailedCard`, `BillingOverview`, `RecentInvoices`) collapse into `formatPrice`
  - six settings `toLocaleDateString("en-US", …)` sites (organization, api-keys, teams ×2, members ×2)
  - usage `formatUsd` and RecentInvoices' English-month date-fns pattern

## Deliberately not touched

- `packages/shared/rrule.ts` machine-readable date keys (correctly pinned)
- `format-tokens` (shared with the CLI, which stays English for now)
- marketing date sites (English-only until Phase 6 localized routes)
- server-side `packages/auth` `formatPrice` (needs the stored user locale from the persistence PR; used in Stripe/checkout server strings)
- `speechRecognition.lang` (follows the language setting once it exists)

## Verification

turbo typecheck 38/38 · biome clean · sherif clean · turbo test 19/19. Behavior for English users is unchanged: the active locale is `en` everywhere today, and `Intl` output for `en` matches the previous hardcoded `en-US` output.

---
## Summary by cubic
Adds `@superset/i18n/format`, an `Intl` wrapper layer that formats numbers, currencies, and dates using the active locale from the shared i18n instance, and replaces every client-side hardcoded `"en-US"` formatter with it. English users see no behavior change — the active locale is `en` today, and `Intl` output for `en` matches the previous hardcoded output.

**Changes**
- Exposes `formatNumber`, `formatPercent`, `formatCompactNumber`, `formatCurrency`, `formatPrice` (Stripe minor units), `formatDate`, and `formatDateTime`; helpers are plain per-call `Intl` constructions since construction cost is negligible at current call rates.
- Makes `I18nProvider` remount its subtree when the locale changes so plain formatter calls re-render everywhere without new hooks.
- Replaces the 10 hardcoded formatters in `packages/ui`'s ai-elements `context.tsx` plus its 10-site fork in `apps/mobile`.
- Collapses three duplicate cents→currency helpers in desktop billing into `formatPrice`, swaps six settings `toLocaleDateString("en-US", …)` sites, and moves usage `formatUsd` and RecentInvoices' English-only date-fns pattern onto `formatCurrency`/`formatDate`.

**Deliberately not touched**
- Machine-readable `rrule` date keys stay pinned.
- `format-tokens`, marketing dates, and `speechRecognition.lang` stay English until later phases.
- Server-side `formatPrice` in `packages/auth` stays hardcoded until the stored user locale lands.

<sup>Written for commit 378b7ce5db292c273bf65c6ea3e5fd0be864e413. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6905?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/i18n/src/format/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export function getActiveLocale(): string {
+	return i18n.locale || DEFAULT_LOCALE;
+}
+
+// Intl constructors are expensive; several call sites run in render paths.
+const numberFormatCache = new Map<string, Intl.NumberFormat>();
+const dateFormatCache = new Map<string, Intl.DateTimeFormat>();
+
+function numberFormatter(
```
saddlepaddle: nit: i think we can skip these centralized formatters as it causes obfuscation, I think lingui should provide a hook for getting a locale -> we should be able to pass that locale right? I also suspect the perf implications are not as bad as stated, can you have claude check?

Kitenite: Checked — you're right on perf. Measured (100k iterations, bun): `Intl.NumberFormat` construction ~5µs, `DateTimeFormat` ~19µs, cached `.format()` ~0.2µs. So construction is 25–100x the format call but microseconds absolute, negligible at our call rates. Dropped the caches in 378b7ce5d; the helpers are now one-line `new Intl.X(getActiveLocale(), opts).format(v)` wrappers.

On skipping the central helpers entirely: Lingui's hook (`useLingui()` → `i18n.locale`) works for components, but a chunk of the call sites aren't components — `formatUsage.ts`, module-level helpers, and later the Electron main process — and this PR also exists because three files had grown identical private cents→currency helpers. So I kept the thin shared wrappers as the one place currency/date shapes live, and fixed the actual correctness gap your hook suggestion points at from the other side: plain `i18n.locale` reads aren't subscribed to locale changes, so `I18nProvider` now keys its subtree by active locale and remounts on change (rare event, and arguably the UX you want on a language switch anyway). Net: call sites stay plain function imports, locale flow is provider-owned.

If you'd still rather see explicit locale at call sites, the shape I'd go to is a `createFormat(locale)` factory + `useFormat()` hook — happy to switch, it's mechanical from here.

PR #6937, opened 2026-08-28
fix(mobile): move workspace search into a sheet with the native search bar

saddlepaddle:
## Summary
- Home loses the in-page search pill; a native magnifying-glass button in the header opens a full-height **Search** sheet.
- The sheet uses the native iOS search bar (`Stack.SearchBar`, a `UISearchController`) on its own header. In that presentation context it activates properly — field shrinks, Liquid Glass ✕ cancel appears — and results stay tappable, unlike the root screen where iOS 26 swallows every touch (#6659).
- Matching (name, branch, project, session titles) and pin/recency ordering are unchanged, scoped to the active host or Cloud as before.

## Why / Context
#6659 moved search in-page because a `UISearchController` on the root screen on iOS 26 lays an invisible view over the content that eats every touch (unfixed upstream through react-native-screens 4.27). The in-page pill is not native and scrolls away with the list. Giving search its own sheet sidesteps the bug: on the sheet's own header the same controller works.

## How It Works
- `HomeScreen` renders a right `Stack.Toolbar` button (hidden while the selected host is offline, mirroring the list it searched) that pushes `/(authenticated)/(home)/search`, a `formSheet` route with a full-height detent and native title.
- `SearchScreen` renders `Stack.SearchBar` (stacked placement, bar and nav bar always visible) plus a `FlatList` with automatic content insets as the sole content child. The filtering logic moved out of `HomeScreen` unchanged; the sheet reads the same host/cloud/terminal queries, so it adds no requests.
- Tapping a result calls `router.back()` then `router.push(workspace)` — a push alone would land behind the modal.
- `autoFocus` is Android-only in react-native-screens; the sheet focuses the bar through its ref on the stack's `transitionEnd` event.

## Manual QA
Done on an iPhone 17 Pro simulator, iOS 26.0, against production data:
- [x] Header magnifier opens the sheet; the bar is focused on open with the controller active (field shrinks, glass ✕ appears)
- [x] Typing filters live with the "N results on this host" row; session titles match too
- [x] Tapping a result dismisses the sheet and opens the workspace; back returns to home with no sheet left behind
- [x] Glass ✕ clears the search; header ✕ closes the sheet while search is active
- [x] Magnifier hidden when the selected host is offline

Not verified by hand: Cloud scope (flag-gated) beyond code parity with the previous filter.

## Testing
- `bun run lint`
- `bun run typecheck` — 0 errors in `apps/mobile`; the 18 remaining are pre-existing in `packages/port-scanner`, `pty-daemon`, `workspace-fs` and identical on the base commit
- `bun run test` in `apps/mobile` (47 pass)

## Design Decisions
- **Sheet's own header, not a nested `_layout` Stack**: inside a nested Stack the `UISearchController` never activates (no cancel button, looks like a plain field; `obscureBackground`/`hideNavigationBar` have no effect). On the sheet's own header it activates and touches still work — the difference from #6659 is the presentation context.
- **Lean result rows instead of `WorkspaceRow`**: search is find-and-open; the full row needs pull-request, diff-stat and attention plumbing that would duplicate most of `HomeScreen`.
- **Scope**: the sheet searches what the list shows (selected host or Cloud); other hosts aren't in memory.

## Known Limitations
- react-native-screens logs "FormSheet with ScrollView expects at most 2 subviews. Got 5" in dev for this shape (toolbar + search bar + list); the layout renders correctly.
- UIKit labels the glass cancel button "Close", the same accessibility label as the header ✕.

https://claude.ai/code/session_01Ck1GRWn2c2mjHBE3e2yaZ1

---
## Summary by cubic
Moves workspace search from the in-page pill on Home into a full-height search sheet with the native iOS search bar. The native bar works on the sheet's own header, where the same controller on the root screen swallowed every touch on iOS 26 (#6659).

- Home now opens the sheet from a native magnifying-glass toolbar button instead of the in-page pill.
- Matching (name, branch, project, session titles) and pin/recency ordering are unchanged, scoped to the active host or Cloud.
- Tapping a result closes the sheet, then opens the workspace.
- The magnifier is hidden while the selected host is offline.
- The search bar is focused when the sheet opens.
- In dev, `react-native-screens` logs a subview-count warning for the sheet; the layout renders correctly.

<sup>Written for commit 74a0f82e5e4e58a681cc17cf920313108ad95dba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated workspace search screen accessible from the home screen.
  * Search across workspace names, branches, projects, and terminal sessions.
  * Results are sorted by pinned status and recent activity, with result counts and empty states.
  * Selecting a result opens the corresponding workspace.
  * Search is unavailable when the selected host is offline.

### DIFF
--- a/apps/mobile/app/(authenticated)/(home)/_layout.tsx
+++ b/apps/mobile/app/(authenticated)/(home)/_layout.tsx
@@ -10,6 +10,15 @@ export default function HomeLayout() {
 		>
 			<Stack.Screen name="index" options={{ title: "" }} />
 			<Stack.Screen
+				name="search"
+				options={{
+					presentation: "formSheet",
+					title: "Search",
+					sheetAllowedDetents: [1.0],
+					sheetGrabberVisible: true,
+				}}
+			/>
+			<Stack.Screen
 				name="filter"
 				options={{
 					presentation: "formSheet",
--- a/apps/mobile/app/(authenticated)/(home)/search.tsx
+++ b/apps/mobile/app/(authenticated)/(home)/search.tsx
@@ -1,0 +1,3 @@
+import { SearchScreen } from "@/screens/(authenticated)/(home)/search";
+
+export default SearchScreen;
--- a/apps/mobile/screens/(authenticated)/(home)/home/HomeScreen.tsx
+++ b/apps/mobile/screens/(authenticated)/(home)/home/HomeScreen.tsx
@@ -2,13 +2,10 @@ import { LegendList } from "@legendapp/list/react-native";
 import { useQueryClient } from "@tanstack/react-query";
 import { isAfter } from "date-fns";
 import * as Haptics from "expo-haptics";
-import { useFocusEffect, useRouter } from "expo-router";
-import { Search } from "lucide-react-native";
+import { Stack, useFocusEffect, useRouter } from "expo-router";
 import { useCallback, useMemo, useState } from "react";
 import { RefreshControl, useWindowDimensions, View } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
-import { Icon } from "@/components/ui/icon";
-import { Input } from "@/components/ui/input";
 import { Text } from "@/components/ui/text";
 import {
 	type CloudWorkspaceStatus,
@@ -81,7 +78,6 @@ type HomeListItem =
 			workspace: HostWorkspaceItem;
 			cloudStatus?: CloudWorkspaceStatus;
 	  }
-	| { kind: "note"; label: string }
 	| { kind: "hostOffline"; hostName: string };
 
 function homeListItemKey(item: HomeListItem): string {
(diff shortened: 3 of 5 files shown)


saddlepaddle:
Read both CodeRabbit comments; merging without changes:

- **Cloud project names in Cloud scope** — parity with the in-page search this replaces (`matchesQuery` used the same host-project map). Cloud is flag-gated; worth doing when Cloud rows get project metadata of their own.
- **`glassHeaderOptions` for the sheet** — that config is a transparent/blur header and does not provide a close button; the sheets using it (`CommitsSheet`, `JumpToFileSheet`) render their own `Stack.Toolbar` ✕ as this one does. The sheets in this same `(home)` layout (organizations, filter, new-session) use the plain header, which is what this route matches. A transparent header behind an active `UISearchController` would also need the touch/activation verification redone.

PR #6939, opened 2026-08-28
fix(desktop,ui,mobile): three paste defects found auditing the paste surfaces

Kitenite:
Stacked on #6935 (base: `fix-multipaste`). That PR is a no-op refactor; this
one is the behaviour fixes the same audit turned up. Three independent
commits, reviewable separately.

## 1. The terminal rich input silently ate pasted files

`TerminalRichInput` renders `PromptInput` + `TiptapPromptEditor` with **no
attachment tray**. The editor's paste handler claimed the event and pushed the
files into state that nothing rendered and `handleSubmit` never read, so:

- pasting a screenshot into the terminal composer did nothing at all
- a clipboard carrying an image *and* text lost the text as well

Now the tray renders, and submit writes the attachments into the worktree's
`.superset/attachments` (the shared convention, so they self-ignore) and appends
their worktree-relative paths to the prompt. A PTY only takes bytes, so a path
is how a pasted image reaches the agent, and it matches what a file drop onto
the terminal already does. Attachments with no typed text are a valid
submission. An upload failure rejects the submit, which `PromptInput` already
restores the composer from, so a failed attach never costs the draft.

Both editors also stop swallowing text on a mixed image+text paste: attach the
files, then leave the paste unhandled so the text still lands.

**Verified over CDP in the running app**, real pasteboard, real Cmd+V, real PTY:

| Step | Result |
|---|---|
| Paste screenshot into rich input | tray shows `image.png` (before: nothing) |
| Paste image + text | text lands in editor **and** image attaches |
| Submit | PTY receives `describe this screenshot .superset/attachments/image.png .superset/attachments/image_1.png` |
| Files on disk | `image.png` (260B), `image_1.png` (351B), collision-probed |
| `.gitignore` | written with `*`; `git status` stays clean |
| Composer | text and tray both cleared |

## 2. `accept` / `maxFiles` / `maxFileSize` were dead props on every v2 composer

`usePromptInputAttachments` prefers the provider context, and `PromptInput`
provided **no** attachments context at all in provider mode — it returned
`inner` bare. So every consumer got the controller's raw `add`, which does no
filtering. The constraints and `onError` were silently inert whenever a
`PromptInputProvider` was present, which is every v2 composer.

Concretely: the new-workspace screen asks for `maxFiles={5}` and a 10MB cap and
enforced neither. A paste of fifty files, or a single 2GB one, went straight in.

Extracts the validation into `applyAttachmentConstraints` so both storage modes
share it, routes the provider path through it, and re-provides the validated
context to `PromptInput`'s subtree. State still lives in the controller.
9 unit tests cover capacity, per-file size, accept matching, and which
condition reports which error code.

## 3. `apps/mobile/modules/paste-input` was dead

Superseded by the composer module's own `paste(itemProviders:)` when the
composer became a native view (#6761). Nothing has imported
`@superset/paste-input` since; the only reference left was a tsconfig path
alias, so the Swift view and its JS wrapper were both unreachable. Removed.

## Testing

- `bun test` across `packages/ui`, `packages/chat-ui`, and the TerminalPane
  tree: 82 pass
- New: 9 tests for `applyAttachmentConstraints`, 5 for attachment paths in
  `prepareTerminalSubmission`
- `bun run --cwd apps/desktop typecheck`, `packages/ui`, `packages/chat-ui`
- `bunx biome check` over `packages`, `apps/desktop/src`, `apps/mobile`
- `bunx sherif`

## Risks

- (1) changes what submitting the terminal composer does when attachments are
  present. It previously did nothing with them, so there is no prior behaviour
  to regress; the text-only path is untouched.
- (2) starts enforcing limits that were declared but inert. Composers that were
  relying on the absence of enforcement would now see `onError`; the only
  in-repo callers are the ones that set the limits deliberately.
- Rollback: revert per commit, they are independent.

---
## Summary by cubic
Fixes three paste defects: terminal-pasted files now attach and reach the agent, composer attachment limits actually enforce, and the dead mobile `paste-input` module is gone. Two behavior changes — terminal submits with attachments and enforced constraints — are independent and revert separately.

**Bug Fixes**
- Terminal rich input had no attachment tray, so pasting a screenshot did nothing and a clipboard carrying image + text lost the text. The tray now renders, files are written to the worktree's `.superset/attachments` (self-ignored like file drops), and their paths append to the prompt; attachments alone are a valid submission and a failed upload restores the draft.
- `accept`, `maxFiles`, `maxFileSize`, and `onError` were dead props on every v2 composer because provider-backed composers read the controller's raw `add` unvalidated. Constraints now register into the controller and are enforced by `applyAttachmentConstraints` at the single `add` every caller funnels through, so the new-workspace screens' 5-file/10MB caps and the terminal's 10-file/50MB cap finally apply; partial rejections report too instead of dropping files silently.
- Removed `apps/mobile/modules/paste-input` — never imported since the native composer's own `paste(itemProviders:)` superseded it, and a tsconfig alias was the only reference left.

<sup>Written for commit 61d6c4a55e87211065a58455ccbadda74c44c601. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6939?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Terminal prompts now support attaching multiple files, including file-only submissions.
  - Uploaded attachments are available to terminal commands with safe path handling.
  - Attachment chips are displayed in the terminal composer.

- **Improvements**
  - Pasting images alongside text now preserves both the attachment and pasted text.
  - Attachment validation consistently enforces file type, size, and count limits.

- **Bug Fixes**
  - Attachment upload errors now appear as clear notifications.
  - Prevented pasted text from being unintentionally discarded when files are included.


Kitenite:
Verification note on commit 3 (the mobile module removal), since the three commits were **not** verified to the same depth:

**Done**
- Static reachability: nothing imports `@superset/paste-input` or references `PasteInputView` / `requireNativeView("PasteInput")` outside the module. The only reference was its tsconfig path alias.
- `tsc --noEmit` before and after, on the same tree: **141 errors both ways**, byte-identical set, none paste-related. (Those 141 are pre-existing debt in `apps/mobile` on `main` — worth its own ticket, mobile typecheck is evidently not gated.)
- No committed `ios/` directory and no `Podfile.lock` in the repo, so the native project is prebuild-generated. Removing a local autolinked Expo module means it simply stops being linked on the next prebuild.

**Not done**
- No iOS build, no simulator run. The residual risk is native autolinking, not TypeScript, and a build is the only thing that would actually close it.

Commits 1 and 2 were verified end to end over CDP against the running desktop app. If anyone wants commit 3 held back until someone runs an iOS build, it is independent and can be dropped from this PR without touching the other two.

Kitenite:
**Update: the mobile commit is now built and verified, so disregard the "not done" line in my previous comment.**

`apps/mobile/AGENTS.md` already documents the recipe for this ("Iterating on a native module? Build its own pod scheme, not the app"), so there was no reason to leave it unbuilt.

Before/after on the same tree, restoring the module and removing it again:

| Check | Before (module present) | After (removed) |
|---|---|---|
| `expo-modules-autolinking search -p ios` | `@superset/paste-input` resolved, `PasteInputModule` registered | module absent, `PasteInputModule` occurrences: **0** |
| `Podfile.lock` | `PasteInput` pod | **0** hits |
| `ExpoModulesProvider.swift` | 4 local modules | `AlertPromptModule`, `AttachmentsSheetModule`, `ComposerModule` |
| Workspace schemes | `PasteInput` scheme present | absent |
| `tsc --noEmit` | 141 errors | 141 errors, identical set |

Builds after removal, iOS 26.5 simulator SDK:

- `expo prebuild --platform ios --clean` + `pod install` — **exit 0**
- `xcodebuild -scheme Composer` (the AGENTS.md fast path) — **exit 0**, only pre-existing `ComposerDictation.swift` Swift-6 warnings
- `xcodebuild -scheme Superset -configuration Debug -sdk iphonesimulator` (full app target) — **exit 0**, 0 compile errors, 0 `PasteInput` references anywhere in the build output

Worth noting for a separate ticket: `apps/mobile` has **141 pre-existing `tsc` errors on `main`**, so mobile typecheck is evidently not gated in CI.

Kitenite:
**Final update: the app was built *and run*.** My earlier comment said "built"; that was not the same as running it, so here is the run.

iPhone 17 Pro simulator, iOS 26.5, the Debug build from the previous comment:

- `simctl install` + `simctl launch` — launched, pid stayed alive (no crash from the removed native module)
- Connected the dev client to Metro and loaded the real JS bundle
- App booted to its normal signed-in UI: `Organization` header, `Search workspaces`, `No projects on this host yet`
- Metro log shows the app's own runtime: `Application Opened`, `$screen_name: "/"`, PostHog feature flags fetched, `$device_name: Simulator iOS`, `$os_version: 26.5`
- **0** occurrences of `Unable to resolve` / `Cannot find native module` / `Invariant Violation` / `PasteInput` anywhere in the Metro output

So the module removal is verified at every level it can be: autolinking manifest, Podfile.lock, generated `ExpoModulesProvider.swift`, pod scheme build, full app build, and now a running app.

Kitenite:
**Correction: commit 2 as originally pushed did not work, and I only found that by testing it in the app.** Fixed in `a12c216`.

The original approach re-provided a validated attachments context from inside `PromptInput`. But the screens that add files call `useProviderAttachments()` in their **own component body, above `<PromptInput>`** (e.g. `NewWorkspaceScreen.tsx:141` vs `<PromptInput>` at line 739), so they never saw that context and kept using the controller's raw `add`.

The unit tests passed the whole time because they only covered the pure validator, never the wiring. Pasting 7 files into the new-workspace composer still attached all 7.

**Now:** the constraints are registered into the controller (`__registerConstraints`, matching the existing `__registerFileInput` / `__registerTextarea` pattern) and validated inside the provider's `add` — the single point every caller funnels through regardless of where it took the context. `onError` is also wired on the three composers that declare limits, so truncation is no longer silent.

Verified in the running app over CDP, real pasteboard, real Cmd+V:

| Case | Before fix | After fix |
|---|---|---|
| 7 files, `maxFiles={5}` | 7 attached, no toast | **5 attached** + "Too many files. Some were not added." |
| 12MB file, `maxFileSize={10MB}` | attached | **0 attached** + "All files exceed the maximum size." |
| 1 valid + 1 oversized | both attached | **1 attached** (oversized dropped) |

Known limitation, carried over deliberately from the original `addLocal`: a *partially* oversized batch drops the offending files silently — `max_file_size` only fires when the entire batch is too large. That is the pre-existing semantics and my extracted validator preserves it exactly (there is a unit test pinning it). Worth a follow-up, not a silent behaviour change here.

Kitenite:
Two updates.

**1. Partial rejections now error instead of dropping files silently** (`a2f047e`)

`accept` and `maxFileSize` only fired `onError` when the *entire* batch was rejected. A mixed paste attached some files and discarded the rest with no feedback. Both now report on a partial rejection too, with wording that distinguishes it from the all-rejected case.

Verified in the app: pasting a 12MB file alongside a valid one attaches the valid one and toasts *"Some files exceed the maximum size and were not added."*

**2. The terminal rich input before/after, actually reproduced**

My earlier description of the pre-fix behaviour was read off the code, not measured. Reproduced properly by reverting the three files to `f7dea882d~1` in the running app and pasting the same payloads:

| Paste | Before | After |
|---|---|---|
| Screenshot only | tray cards **0**, editor empty — paste vanished | tray shows `image.png` |
| Image + text | tray cards **0** *and* `editorText: ""` — **the text was swallowed too** | text lands in editor **and** image attaches |

That confirms both halves of the bug, including the text loss, against the real app rather than by inspection.

Every change in this PR and #6935 now has a measured before and after in the running app.

PR #6941, opened 2026-08-28
fix(ui): keep the maxFiles budget in step within a single tick

Kitenite:
Follow-up to #6939. CodeRabbit raised this on that PR, but the review landed after it was merged, so the fix needs its own PR.

## The bug

`attachmentCountRef.current = attachmentFiles.length` is a **render-time** assignment. Two `add()` calls before the next render therefore both budget against the pre-add count, so each admits up to the full remaining capacity and together they can exceed `maxFiles`.

```
maxFiles = 5, tray empty
add(3 files) -> sees count 0, capacity 5, admits 3
add(3 files) -> still sees count 0 (no render yet), admits 3
                                              => 6 attachments
```

## The fix

Advance the ref as each batch is admitted. The render-time assignment still re-syncs it from state afterwards, so removals and clears stay authoritative.

## Why not CodeRabbit's suggestion

It proposed moving validation inside the `setAttachmentFiles((prev) => ...)` updater and using `prev.length`. That is the right instinct for freshness, but `applyAttachmentConstraints` calls `onError`, and React invokes state updaters twice under StrictMode — every rejection would fire two toasts. Correcting the budget in place keeps the side effect outside the updater.

## Testing

- `bun test packages/ui/src` — 50 pass
- `bun run --cwd packages/ui typecheck`, `bunx biome check`

**Not verified in the running app.** Re-running the in-app cap test needs the new-workspace composer, and the workspace I had been testing in is mid-onboarding with no project after an unrelated `setup.sh --force` (its Neon branch had been deleted by the branch-cleanup sweep). The earlier in-app checks — 7 files against `maxFiles={5}` attaching 5 with a toast, 12MB against a 10MB cap attaching none — were run against the code this sits on top of, and this change only tightens the count between two calls in one tick.

---
## Summary by cubic
Fixes a same-tick race where attachment count checks used a render-time snapshot, letting consecutive `add()` calls collectively exceed `maxFiles` and a `clear(); add(...)` reject files that now fit.

All list mutations now funnel through a synchronous mirror ref that updates the list and count before returning, and `takeFiles()` reads that mirror too — so a submit right after an add no longer clears files it never handed back. Validation stays outside the React state updater because StrictMode invokes updaters twice and would double the error toast.

<sup>Written for commit bf08454d64b017ca14fb6b7decb472bae246b43e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file attachment validation when adding multiple batches quickly.
  * Attachment limits are now tracked accurately across consecutive additions, preventing invalid uploads caused by outdated counts.
  * Multiple add actions in the same moment now consistently respect the configured attachment constraints.
  * Attachment counts now remain synchronized with the selected files for more reliable upload behavior.
  * Improved consistency when removing attachments or navigating away while files are selected.

### DIFF
--- a/packages/ui/src/components/ai-elements/prompt-input.tsx
+++ b/packages/ui/src/components/ai-elements/prompt-input.tsx
@@ -290,6 +290,8 @@ export function PromptInputProvider({
 		[],
 	);
 	// Read at call time so `add` stays stable while still seeing the live count.
+	// Re-synced from state on every render, and advanced by `add` itself so two
+	// calls in the same tick do not both budget against the pre-add count.
 	const attachmentCountRef = useRef(0);
 	attachmentCountRef.current = attachmentFiles.length;
 
@@ -309,6 +311,10 @@ export function PromptInputProvider({
 			if (incoming.length === 0) {
 				return;
 			}
+			// Keep the budget in step within this tick. Validation deliberately
+			// stays outside the updater: it calls onError, and React invokes
+			// updaters twice under StrictMode, which would double every toast.
+			attachmentCountRef.current += incoming.length;
 
 			setAttachmentFiles((prev) =>
 				prev.concat(
--- a/packages/ui/src/components/ai-elements/prompt-input.tsx
+++ b/packages/ui/src/components/ai-elements/prompt-input.tsx
@@ -262,16 +262,35 @@ export function PromptInputProvider({
 		attachmentsStore?.get ?? getEmptyAttachments,
 	);
 	const attachmentFiles = attachmentsStore ? storeFiles : localFiles;
+	// A synchronous mirror of the list. React defers state updaters to render,
+	// so without this a same-tick `clear(); add(...)` (or two adds in a row)
+	// would budget against a list that no longer reflects what just happened.
+	// The render assignment below re-syncs it from the authoritative state.
+	const attachmentFilesRef =
+		useRef<PromptInputAttachmentItem[]>(attachmentFiles);
+	attachmentFilesRef.current = attachmentFiles;
+	const attachmentCountRef = useRef(0);
+	attachmentCountRef.current = attachmentFiles.length;
+
+	// Every list mutation funnels through here, so this is the one place the
+	// mirror has to be kept honest: add, remove, clear, takeFiles and setFiles
+	// all land their new length before the call returns.
 	const setAttachmentFiles = useCallback(
 		(
 			updater: (
 				prev: PromptInputAttachmentItem[],
 			) => PromptInputAttachmentItem[],
 		) => {
+			const base = attachmentsStore
+				? attachmentsStore.get()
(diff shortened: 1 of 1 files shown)


Kitenite:
**Now verified in the running app — disregard the "not verified" caveat in the description.**

Reproduced against the code currently on `main`, then against this PR, in the same session: two `add()` calls dispatched in one tick (three files each) into the new-workspace composer, which declares `maxFiles={5}`.

| Code under test | Attachments in the tray |
|---|---|
| `main` today (no fix) | **6** — cap exceeded |
| This PR | **5** — cap holds |

Restored the fix afterwards and re-ran to confirm: 5 again.

So the finding was real on merged code, and the one-line ref advance closes it.

Kitenite:
Second finding addressed in `47d81ea` — thanks, it was correct.

Rather than patch each mutator, I made the single funnel synchronous: `setAttachmentFiles` keeps a mirror of the list, computes the next value from it, and lands both the mirror and the count before returning. `add()` no longer advances the count itself, and `remove` / `clear` / `takeFiles` / `setFiles` are correct for free. The render-time assignment still re-syncs from the authoritative state.

Verified in the running app:

| Sequence | Result |
|---|---|
| Two adds in one tick, 3 each, `maxFiles={5}` | **5** (cap holds) |
| Clear the tray | **0** (removals tracked) |
| Fresh batch of 5 after clearing | **5** admitted in full — the stale-count bug would have rejected these |

Kitenite:
Third finding fixed in the latest commit — this one was the most serious of the three, and the analysis was exactly right.

`takeFiles` read the render-time ref while `setAttachmentFiles` had moved on to the synchronous mirror, so a submit in the same tick as an add returned the pre-add list *and then cleared the files it never handed back*. `NewWorkspaceModal/PromptGroup.tsx` calls `takeFiles()` synchronously on submit, so a fast paste-then-send could silently drop attachments.

It now reads the mirror. That made the second ref redundant, so both are collapsed into one, and the unmount blob-URL cleanup uses it too.

**Verification note, stated plainly:** this one is covered by typecheck and the unit suite, but *not* by an in-app run. Exercising it means actually submitting the v1 new-workspace form, which creates a real workspace and worktree — not something I wanted to do just to observe a ref read. The synchronous mirror it now reads is the same one I verified in-app in the previous commit (clear the tray, then a fresh batch of 5 is admitted in full).

PR #6945, opened 2026-08-28
fix(host-service): dedupe replayed Codex token events across rollout files

hairez:
## What & why

Codex 0.144 and 0.145 replay the parent thread's entire `token_count` history into forked and subagent rollout files, re-timestamped to the spawn instant. `parseCodexLogFile` only skips consecutive identical deltas within a single file, so the same API call gets counted once per descendant file. On heavily forked days that inflates reported Codex usage by up to 126x.

`parseClaudeLogFile` already handles the equivalent problem, deduping globally on `message.id + requestId` through a Map shared across every file. This gives the Codex path the same treatment, with three differences that matter:

- The key is the `last_token_usage` delta paired with the thread's running `total_token_usage`. The delta alone would merge two genuinely distinct turns that happened to cost the same; the pair will not.
- The key is claimed before the cutoff check rather than at the push site. Many replayed events have their originals outside the window, so a dedupe that only sees post-cutoff events cannot recognise them. The Claude path can claim its key after the cutoff because its rewrites keep the original timestamp instead of re-stamping to the spawn instant.
- Codex files are visited in rollout-chronological order, so the earliest file carrying an event keeps it. Sorting is on the filename's start timestamp rather than mtime, since a long-running parent outlives the fork it spawned.

Counts are normalised with `?? 0` before keying, because pre-0.145 records omit `cache_write_input_tokens` while the replayed copies re-serialise it as `0`.

Deliberately not fixed by excluding forked files: those carry a large share of genuinely new daily usage, so dropping them undercounts badly.

## How I tested it

Against `ccusage@20.0.20 codex daily -z UTC` over 30 days of my own rollout history: every day with usage now matches exactly on output, cache-read and uncached-input. Before, the worst day reported 126x its real cache reads. No day moved down, and days that were already correct are unchanged.

`codex-replay.test.ts` writes real temp rollouts: a parent with three turns, plus two subagents that each replay all three at their spawn instant and then take a turn of their own. The fixture is built so the first and third parent turns have identical deltas, so keying on the delta alone would silently drop one, and the parent omits `cache_write_input_tokens` while the replays serialise it as `0`. Files are handed to the parser in reverse-chronological order, so any correct ordering in the result comes from the production sort.

Two tests: that a replayed history is counted once on the day the thread spent it, and that a replay of turns older than the window adds nothing. Each assertion was mutation-tested, so none is vacuous:

| change that should break it | caught by |
|---|---|
| per-file dedupe set instead of shared | both tests |
| claim the key after the cutoff check | the out-of-window test only |
| drop the `?? 0` normalisation | both tests |
| visit files unsorted | both; the parent's day disappears entirely |

The cutoff row is the useful one: only the out-of-window case exposes it, which is why a version that claims the key at the push site still passes a casual suite while undercounting.

## Checklist

---
## Summary by cubic
Fixes Codex usage over-counting when forks and subagents replay the parent thread's entire token history, so the same API call is now counted once per day instead of once per descendant file.

- Dedupes replayed events across rollout files using the per-turn delta paired with the thread's running total, so identical-cost turns remain distinct.
- Claims the dedupe key before the cutoff check so replays of turns older than the window are not counted as fresh usage.
- Visits rollout files in start-time order so the earliest file carrying an event keeps it.
- Normalizes missing cache-write fields so pre-0.145 records match their replayed copies.

<sup>Written for commit a0a5c0e897e778a4b94d84ed8d11f61d53e00f3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6945?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex usage history accuracy by preventing replayed turns from being counted multiple times.
  * Correctly attributes usage to the original thread, including when activity is replayed in forks or subagents.
  * Ensures older replayed turns do not incorrectly increase usage totals within the selected period.

* **Tests**
  * Added coverage for forked Codex rollouts, turn deduplication, sorting, and cutoff handling.

### DIFF
--- a/packages/host-service/src/trpc/router/usage/history/codex-replay.test.ts
+++ b/packages/host-service/src/trpc/router/usage/history/codex-replay.test.ts
@@ -1,0 +1,200 @@
+import { afterAll, describe, expect, test } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { collectLogFiles, sortCodexFiles } from "./logs";
+import { parseCodexLogFile, type UsageLogEntry } from "./parse";
+
+const root = mkdtempSync(join(tmpdir(), "codex-replay-"));
+afterAll(() => rmSync(root, { recursive: true, force: true }));
+
+interface Turn {
+	input: number;
+	cached: number;
+	output: number;
+	reasoning: number;
+}
+
+interface Rollout {
+	/** Session start / spawn instant, and the name the rollout is stored under. */
+	start: string;
+	/** Parent turns the fork re-emits, all stamped at `start`. */
+	replayed?: Turn[];
+	/** Turns this thread ran itself, at their own timestamps. */
+	own?: Array<{ at: string; turn: Turn }>;
+	/** Pre-0.145 rollouts leave `cache_write_input_tokens` out entirely. */
+	omitCacheWrite?: boolean;
+}
+
+function counts(turn: Turn, omitCacheWrite: boolean) {
+	return {
+		input_tokens: turn.input,
+		cached_input_tokens: turn.cached,
+		...(omitCacheWrite ? {} : { cache_write_input_tokens: 0 }),
+		output_tokens: turn.output,
+		reasoning_output_tokens: turn.reasoning,
+		total_tokens: turn.input + turn.output,
+	};
+}
+
+let sequence = 0;
+
+function writeRollout(dir: string, rollout: Rollout): void {
+	const omitCacheWrite = rollout.omitCacheWrite ?? false;
+	const replayed = rollout.replayed ?? [];
+	const own = rollout.own ?? [];
+	const lines: string[] = [
+		JSON.stringify({
+			timestamp: rollout.start,
+			type: "session_meta",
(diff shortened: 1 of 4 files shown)


Kitenite:
Thanks for this, the write-up and the mutation-tested fixture were genuinely useful. I verified it against my own rollouts and it matches ccusage exactly on the worst days.

Closing rather than merging though: the replay only happens in Codex 0.144 through 0.146. 0.147 and later stopped copying the parent's token_count history into fork and subagent rollouts, and the affected files age out of the 31-day scan window on their own. So this would be shipping a dedupe layer, plus its small cross-thread collision risk on identical first turns, for data that's already disappearing.

If the replay comes back in a future Codex version, this is the branch to revive. Keying on the lineage root from session_meta (forked_from_id / parent_thread_id) instead of raw counts would close the collision gap at that point.

PR #6951, opened 2026-08-28
perf(i18n): load translation catalogs on demand

Kitenite:
Every translation catalog was statically imported, so each surface bundled all of them.

At three locales that is ~1.1 MB of translations in every bundle. Scaling the locale list — which is the next step for i18n — makes it much worse: at eighteen locales it would be roughly **10 MB** of catalog in every bundle.

## What changed

- **English stays bundled.** It is the fallback for any message a translation is missing, so it has to be present and synchronous.
- **Every other catalog loads on demand**, through a spelled-out import map rather than a template, because bundlers only follow `import()` calls they can resolve statically.
- **`initI18n` keeps its synchronous signature** for call sites that cannot await — module scope, and anything that needs a locale active before first paint. If the catalog is not loaded, it activates English and swaps in when the import resolves; Lingui emits a `change` event on that second activation, which `I18nProvider` already subscribes to.
- **`initI18nAsync` is the awaited form.** Three call sites need it: the Electron boot path, the RSC entry point, and `applyAppLanguage`.

That last one is a correctness fix rather than a nicety: `applyAppLanguage` rebuilds the application and tray menus, which resolve their labels once at build time. Without awaiting, it would rebuild them *before* the catalog arrived and render them in English.

## Verification

- `bun run --cwd packages/i18n check` (extract + `--strict` + drift) passes
- Typecheck clean: desktop, web, marketing
- Desktop suite 2953/2953, i18n suite 89/89, biome clean

No behaviour change for users at the current three locales; this is the groundwork the locale expansion needs.

---
## Summary by cubic
Loads translation catalogs on demand instead of bundling every locale into each surface, cutting ~1.1 MB per bundle today and avoiding ~10 MB as the locale list grows; English stays bundled as the synchronous fallback.

- `initI18n` stays synchronous: it activates English first, swaps in the lazy-loaded catalog when it resolves, and degrades to English with a log if the import fails.
- `initI18nAsync` awaits the catalog in the Electron boot path and `applyAppLanguage`, so application and tray menus don't rebuild in English.
- `initServerI18n` stays synchronous for the RSC slot; new `preloadServerLocale` loads a non-default catalog ahead of time.

<sup>Written for commit e866150e831cf3a8120e3fb25ab60b80a13b8af9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6951?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a setting to automatically copy terminal selections, with persistent enable/disable controls.

- **Bug Fixes**
  - Improved language switching so translated application and tray menus are ready before display.
  - Ensured locale resources finish loading before language changes complete.
  - Improved server-side initialization for non-default languages.

- **Performance**
  - Reduced initial loading by loading additional language catalogs only when needed.

- **Documentation**
  - Clarified supported locale and language fallback behavior.
  - Updated locale tests for unsupported-language handling.

### DIFF
--- a/apps/desktop/src/lib/trpc/routers/settings/index.ts
+++ b/apps/desktop/src/lib/trpc/routers/settings/index.ts
@@ -625,7 +625,7 @@ export const createSettingsRouter = () => {
 
 		setLanguage: publicProcedure
 			.input(z.object({ language: z.string().nullable() }))
-			.mutation(({ input }) => {
+			.mutation(async ({ input }) => {
 				const value =
 					input.language === null || input.language === "auto"
 						? null
@@ -646,7 +646,8 @@ export const createSettingsRouter = () => {
 					.run();
 				// The application and tray menus resolve their labels when they are
 				// built, so they need an explicit rebuild on a language change.
-				applyAppLanguage(value);
+				// Awaited: the catalog for the new locale loads on demand.
+				await applyAppLanguage(value);
 			}),
 
 		getSelectedRingtoneId: publicProcedure.query(() => {
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -6,7 +6,7 @@ import {
 	writeSharedDisabledAgentIds,
 	writeSharedDisabledSkillIds,
 } from "@superset/agent-setup";
-import { i18n, initI18n } from "@superset/i18n";
+import { i18n, initI18nAsync } from "@superset/i18n";
 import { settings } from "@superset/local-db";
 import { app, dialog, Notification, net, protocol, session } from "electron";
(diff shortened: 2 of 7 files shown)


Kitenite:
Addressed the unhandled-rejection comment: a failed lazy catalog import now logs and stays on English instead of firing an unhandled rejection.

PR #6984, opened 2026-08-29
feat(i18n): translate new strings automatically on the PR

Kitenite:
## Why

Sixteen locales sit behind `compile --strict`, so **every PR that adds an English string owes sixteen translations before CI passes**. #6952 already paid this tax by hand. This closes the loop on the PR itself, keeping strict-mode's guarantee (no locale ever ships partially translated) without making "add a button" cost a translation errand.

## How

**`packages/i18n/scripts/translate-missing.ts`** — finds every empty `msgstr`, translates each locale's gap in one Claude request anchored to ~40 of that catalog's *own existing translations* (register and terminology come from the catalog, not a hand-maintained glossary), then validates every fill against the English source: placeholders, tag markers, ICU structure, and the one/few/many/other requirement for ru/pl/cs. **A fill that fails validation stays empty** — the strict gate then fails loudly rather than a broken string shipping.

**`Translate Catalogs` workflow** — on same-repo PRs: extract → count gaps (entry-aware; the PO header is `msgid ""` and must not count) → translate → re-verify `compile --strict` + the stale check → push the commit back. The push uses a GitHub App token because `GITHUB_TOKEN` pushes trigger no workflows and would leave the PR checkless on its newest commit. Fork PRs (no secrets) surface their gap in the existing catalog audit for a maintainer.

Machine fills are a floor, not a ceiling — reword freely in the same PR; the stale-check keys exemptions to exact English text so later edits re-check.

## Verification

- Validator unit tests (8): dropped/renamed placeholders, unbalanced tags, removed ICU, Slavic plurals missing few/many all rejected; faithful translations and reworded ICU prose accepted
- Counter tested against real catalogs: baseline 0, one blanked entry → 1, restored → 0; skips the `en-only-changes.txt` exemption file
- Workflow YAML validated; catalog gate, i18n tests (116), and desktop typecheck green

The live API path first exercises on the next PR that adds a string — the strict compile after the translator means its failure mode is a red check, never a silent merge.

https://claude.ai/code/session_012U97YWVsDvZwQkBaSGEmor

---
## Summary by cubic
Adds a workflow that fills missing locale translations on the PR itself, so adding an English string no longer requires manual translation in all sixteen locales before `compile --strict` passes.

**New Features**
- The `Translate Catalogs` workflow extracts catalogs, translates empty entries with Claude, verifies `compile --strict`, and pushes fills back to the PR branch.
- `scripts/translate-missing.ts` anchors each locale's fills on that catalog's existing translations and validates placeholders, tag markers, ICU structure, and one/few/many/other for `ru`/`pl`/`cs`; fills that fail stay empty so the strict gate fails loudly.
- Also runnable locally with `bun run --cwd packages/i18n translate`.

**Rollout notes**
- Pushes use a least-privilege GitHub App token (`contents:write`) passed only to the push step, since `GITHUB_TOKEN` pushes trigger no workflows and would leave the PR checkless.
- Commit and push run with hooks disabled, the token absent from the commit env, and the head ref reaching the shell only as a quoted environment variable, since a branch name can contain shell metacharacters.
- Checkout persists no credentials and fetches full history so the stale-translation check can resolve its merge base.
- Fork PRs surface their gap in the existing catalog audit for a maintainer.

<sup>Written for commit ac750f2ab26ad73075e7f03cf7a6ca1b671802c9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/6984?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automated translation of missing locale catalog entries during pull requests.
  * Added local translation support with dry-run and locale-specific options.
  * Added validation to preserve placeholders, tags, ICU formatting, and locale-specific plural rules.
  * Added strict checks that reject invalid translations and identify unresolved entries.
* **Documentation**
  * Documented the automated translation workflow and local command.
* **Tests**
  * Added coverage for translation integrity and ICU formatting validation.

### DIFF
--- a/.github/workflows/translate-catalogs.yml
+++ b/.github/workflows/translate-catalogs.yml
@@ -1,0 +1,105 @@
+name: Translate Catalogs
+
+# Every enabled locale is gated by `lingui compile --strict`, so a PR that
+# adds or changes an English string owes a translation in all of them. This
+# workflow closes that gap on the PR itself: when the extracted catalogs have
+# untranslated entries, it fills them with Claude, verifies the strict gate,
+# and pushes the fills back to the PR branch.
+#
+# The push uses a GitHub App token rather than GITHUB_TOKEN because pushes
+# made with GITHUB_TOKEN do not trigger workflows — the PR would be stuck
+# with no checks on its newest commit.
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+concurrency:
+  group: translate-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  translate:
+    name: Fill missing translations
+    # Fork PRs have no secrets; their catalog gap surfaces in the Lint job's
+    # catalog audit instead, for a maintainer to fill.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Mint app token
+        id: app-token
+        uses: actions/create-github-app-token@df432ceedc7162793a195dd1713ff69aefc7379e # v2.0.6
+        with:
+          app-id: ${{ secrets.GH_APP_ID }}
+          private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+
+      - name: Checkout PR head
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Setup Bun
+        id: setup-bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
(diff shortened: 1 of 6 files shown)


Kitenite:
Took all three workflow comments: app token scoped to contents:write, checkout no longer persists credentials (extract and the translator run repo-controlled code, so the token is passed only to the push step), and fetch-depth: 0 so the stale check can resolve its merge base.

Kitenite:
Both follow-ups taken: commit runs token-free with hooks disabled (PR steps can write .git/hooks), and the push quotes the head ref and repository from env instead of interpolating — a branch named `$(...)` must stay data.

PR #7011, opened 2026-08-30
test(agent-setup): pin Claude system-default account switching

alldevlo:
## What & why

Pins the generated Claude wrapper's real execution path for a Usage-tab switch from a secondary profile back to the system-default login.

The lower-level resolver already covered the environment mutation in isolation. This integration regression now creates the actual wrapper, starts it with a stale Superset-injected `CLAUDE_CONFIG_DIR`, publishes an empty default-account pointer, and proves the real Claude binary receives `CLAUDE_CONFIG_DIR` unset. That protects Keychain-backed macOS default logins from accidentally staying pinned to a signed-out profile.

## How I tested it

- `bun run lint`
- `bun run typecheck`
- `bun test packages/agent-setup/src/agent-wrappers.test.ts --test-name-pattern 'runs the generated wrapper on the system default after a profile switch'`
- `bun test packages/agent-setup/src/default-account-resolver.test.ts`

The full `agent-wrappers.test.ts` file also reaches and passes the new case, but two pre-existing OpenCode dynamic-import cases fail because their temporary modules disappear before the second and third imports. This change does not touch those cases.

## Checklist

---
## Summary by cubic
Adds an integration test that verifies the generated Claude wrapper un-sets `CLAUDE_CONFIG_DIR` when switching back to the system-default login after a profile switch, preventing Keychain-backed macOS defaults from staying pinned to a signed-out profile.

<sup>Written for commit b91ce74a388fed5a49d9b63185fbcf8235461a0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7011?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Claude now correctly uses the system’s default configuration after switching profiles, preventing stale profile settings from affecting generated wrappers.

* **Tests**
  * Added coverage to verify the corrected Claude configuration behavior.

### DIFF
--- a/packages/agent-setup/src/agent-wrappers.test.ts
+++ b/packages/agent-setup/src/agent-wrappers.test.ts
@@ -58,6 +58,7 @@ const {
 	buildCodexWrapperExecLine,
 	buildCopilotWrapperExecLine,
 	buildWrapperScript,
+	createClaudeWrapper,
 	createClaudeSettingsJson,
 	createCodexHooksJson,
 	createCodexWrapper,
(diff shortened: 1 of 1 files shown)


Kitenite:
Thanks for taking the time on this. Closing without merging: the resolver's unset path is already covered by `default-account-resolver.test.ts`, and this adds an end-to-end copy of the same assertion without a corresponding fix, so it does not earn its keep in the suite. If you hit an actual account-switch bug in the app, an issue with repro steps would be very welcome.

PR #7035, opened 2026-08-31
feat(desktop): preview image diffs in SCM pane

theblondealex:
## Summary

- Render inline previews for supported image files in the desktop Changes pane, using the same placeholder layout as the deferred-diff and binary states (icon, text line, Open file).
- Reuse the existing image extension and MIME-type helpers and the workspace `filesystem.readFile` procedure with a 10MB cap.
- Keep the Open file action and the binary-file fallback for unsupported, deleted, failed, or oversized files.
- Refetch a preview when git reports that path changed; unrelated edits leave previews alone.
- Draw transparent images on the same checkerboard the file pane's ImageView uses.

## How to run and test

1. Run `./.superset/setup.local.sh` once in a workspace.
2. Run `bun run dev`.
3. Open a workspace with staged or unstaged image changes.
4. Open the Changes pane and select the image diff.
5. Confirm PNG/JPEG previews render inline and Open file still works.
6. Replace the image on disk and confirm the preview updates; delete a tracked image and confirm it falls back to "Binary file hidden".

## Validation

- `bun run lint`
- `bun run --cwd apps/desktop typecheck`
- `bun run check:i18n` (one new string, "Loading image…", translated in all 16 locales)
- `bun test` for the DiffPane folder
- Verified in the dev desktop app: added and modified PNGs preview, an on-disk replacement refetches, a deleted image falls back to the binary placeholder without Open file.

## Image example

image

## Scope note

The preview reads the current worktree bytes, so it is shown for working-tree and against-base diffs only. Commit refs keep the plain binary placeholder, since the worktree may no longer hold the commit's version of the file. A true old-vs-new visual comparison is left for a follow-up.

https://claude.ai/code/session_01NAmSwAu95MUWMncyMeBTfa


Kitenite:
Merged \`main\` into this branch (it had conflicted with #7092 / #7102 / #7139) and pushed one follow-up commit so the placeholder matches the Changes pane as it is on main now:

- \`<Trans>\` keyed by English text (#7092), same layout as the other placeholders (icon + selectable text line + Open file), and the deferred-diff spinner while loading.
- Reused the existing "File is too large to preview" string for the 10MB cap; "Loading image…" is the one new string and is translated in all 16 locales.
- Commit refs keep the plain binary placeholder: the preview reads the working tree, which is not the commit's version of the file.
- \`git:changed\` invalidation is scoped to the reported paths so an unrelated edit doesn't refetch every preview.
- Transparent images sit on the same checkerboard the file pane's ImageView uses.

Verified in the dev app: added + modified PNGs preview, a replaced image refetches on its own, a deleted image falls back to "Binary file hidden" without Open file.

theblondealex:
awesome!! glad it made it into main

PR #7044, opened 2026-08-31
fix(accounts): stop tests clobbering the real account pointers, and give Codex session sharing

Kitenite:
Reported as "logins aren't working as expected and sessions access aren't transferred when switching account, same with codex". It turned out not to be a Usage-tab bug. The tab was showing the truth and agents were running somewhere else entirely.

## What was happening

`agents.test.ts`'s `buildTerminalAgentLaunch default account env` block called `setDefaultAccountSelection` with a throwaway in-memory DB but no `SUPERSET_HOME_DIR` isolation. `syncDefaultAccountPointer` resolves the real home, so running the host-service suite wrote the developer's live `~/.superset/state/`:

| pointer | value the tests left behind |
|---|---|
| `default-codex-home` | `$TMPDIR` (from `existingDir = tmpdir()`) |
| `default-claude-config-dir` | empty (the last case sets claude to `null`) |

Every Superset terminal and agent launch injects that codex pointer as `CODEX_HOME`. On the reporting machine:

- `~/.superset/state/default-codex-home` was `/var/folders/…/T` while the org DB's `host_settings` row was null for both. The UI writes DB and pointer together, so the two disagreeing is what proved this was a direct pointer write rather than anything a user did.
- Live terminals were exporting `CODEX_HOME=/var/folders/…/T`.
- `$TMPDIR/auth.json` existed with a valid token: a real `codex login`, done because Codex looked signed out, landing in a directory macOS periodically purges.
- `$TMPDIR/sessions` held 3 rollouts against `~/.codex/sessions`' 1697. That is the missing session access.

The Claude pointer being blanked is the other half: a selected Claude account silently reverts to the system login.

## Changes

**1. Tests can no longer write the real `~/.superset`** (`scripts/test-preload.ts`)

Every test run gets its own Superset home, wired in from the root `bunfig.toml` and re-entered by the host-service preload, since a package-local `bun test` reads that bunfig rather than the root one.

The override is unconditional and that is the point. Superset terminals already export `SUPERSET_HOME_DIR=~/.superset`, and that is where we run tests, so the variable is not unset-and-forgotten, it arrives already aimed at the real home. A `??=` default is a no-op in exactly the environment that needs it. I found this because the canary caught my first attempt.

`test/superset-home-isolation.test.ts` is that canary and fails if a run ever loses the preload. The offending describe block also gets a per-test home, which makes those cases independent rather than sharing one pointer file.

**2. We no longer trust our own injected `CODEX_HOME`** (`resolveAmbientCodexHome`)

`discoverCodexHomes` and `defaultCodexHome` both read `process.env.CODEX_HOME` to decide which home Codex uses with no override, and anything started from a Superset terminal inherits our injected value, a host-service restarted from one included. That is circular: the selected profile masquerades as the system default, so `fetchCodexAccounts` labels it `selection: null` and dedupes the real default away, and provisioning shares config out of the profile into itself.

It also let an unvalidated env value become a selectable account. The default entry skips the `auth.json` check every other candidate gets, so `setDefaultAccount`'s "only accept a discovered login" guard vouched for it. A user-set `CODEX_HOME` is still honoured; only one equal to `SUPERSET_DEFAULT_CODEX_HOME` is ignored.

**3. Codex gets session sharing** (`shareCodexSessionState`)

Claude profiles had `shareClaudeSessionState`; Codex had nothing, so a switch stranded every rollout in the previous account's home and `codex resume` came up empty. This is the second half of the report and a real gap independent of the pointer bug.

`session-share.ts` is now provider-agnostic via `SessionShareSpec`. Codex shares `sessions/`, `archived_sessions/`, `shell_snapshots/` and `history.jsonl`. Rollouts are keyed by date, so two accounts interleave in one tree exactly as two same-day sessions already do. `auth.json` stays per-account; `config.toml` / `AGENTS.md` / `prompts/` stay with agent-setup's ledger since they are rewritten tmp-then-rename and would fork through a symlink.

## Verification

- host-service **1447 pass, 0 fail** (145 files); agent-setup **272 pass, 0 fail**. Typecheck, biome and sherif clean.
- A full host-service run now leaves the real pointers byte-identical with unchanged mtimes. Before this, the same run rewrote them.
- The guard is load-bearing: with the preload stubbed out, the canary fails.
- The new tests can fail: emptying `CODEX_SHARE.dirs` reds 2 of the Codex share tests, and restoring the old unconditional env trust reds the circularity test.

## Not covered here

- A default pointer naming a dir that matches no discovered account is invisible in the UI. `getQuota` marks `isDefault` by equality, so a stale value shows zero accounts as default with no way to see or clear it. That is why the tab showed a healthy `~/.codex` with no default marker.
- If the pointer file is absent, every org host-service races `migrateDefaultAccountPointer` with its own legacy `host_settings` value and first-to-link wins host-wide.
- `profile-remove.ts:26` still protects the raw `process.env.CODEX_HOME`, which errs safe but can block removing a legitimate profile.

Anyone who ran the host-service suite has the bad pointers on disk. Re-pick the Claude account in the Usage tab; Codex returns to `~/.codex` once `default-codex-home` is empty or repointed.

---
## Summary by cubic
Stops host-service tests from clobbering the real `~/.superset` account pointers, which left Codex signed out with no session history, and lets Codex sessions survive an account switch.

**Bug Fixes**
- Every test run gets an isolated `SUPERSET_HOME_DIR` via a preload script; a canary test fails if that isolation is lost.
- Superset's own injected `CODEX_HOME` is no longer trusted as the system default — the real ambient home is preserved in `SUPERSET_AMBIENT_CODEX_HOME` and restored when the system default is selected, so the selected profile can't masquerade as it.
- Removing a Codex profile Superset injected is no longer blocked as if it were the system default.
- Session sharing no longer defaults its target home, and only swaps a session tree when both homes sit on one filesystem.
- History merges keep a cursor, so a late append through an fd opened before the swap drains on the next provision, in capture-generation order; a log the CLI replaced outright is re-linked.
- If you ran the host-service suite, `~/.superset/state/default-codex-home` may point at a temp dir; clearing it (or re-picking the Claude account in the Usage tab) returns Codex to `~/.codex`.

**New Features**
- `shareCodexSessionState` symlink-merges `sessions/`, `archived_sessions/`, `shell_snapshots/`, and `history.jsonl` into the default Codex home so `codex resume` works across accounts, matching Claude's existing sharing.

<sup>Written for commit a301de2ba5170eb2b82cf02232121cbf1b022703. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7044?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added Codex session sharing, preserving rollout data and prompt history across configured accounts.
  * Improved Codex home detection with support for user overrides, custom defaults, and safer fallbacks.

* **Bug Fixes**
  * Preserved credentials and existing history during account setup.
  * Prevented unsafe session merges across different filesystems.
  * Restored session data reliably after cross-filesystem recovery.
  * Isolated automated tests from the local Superset home directory and cleaned up temporary data.

* **Tests**
  * Added coverage for Codex home resolution, session sharing, account provisioning, profile removal, and test isolation.

### DIFF
--- a/apps/desktop/test-setup.ts
+++ b/apps/desktop/test-setup.ts
@@ -9,6 +9,7 @@
  * DO NOT mock internal code here - tests should use real implementations
  * or mock at the individual test level when necessary.
  */
+import "../../scripts/test-preload.ts";
 import { beforeEach, mock } from "bun:test";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -1,3 +1,8 @@
+[test]
+# Isolates SUPERSET_HOME_DIR so a test can never write the developer's real
+# ~/.superset. See scripts/test-preload.ts for what went wrong without it.
+preload = ["./scripts/test-preload.ts"]
+
 [install]
 linker = "isolated" # Prevent hoisting from resolving `path-scurry@2` to `lru-cache@6` (missing `LRUCache`), which breaks the `@superset/desktop` postinstall native rebuild.
 exact = true
--- a/packages/agent-setup/src/agent-wrappers-claude-codex-opencode.ts
+++ b/packages/agent-setup/src/agent-wrappers-claude-codex-opencode.ts
@@ -278,8 +278,11 @@ export function createCodexWrapper(): void {
 	const notifyPath = getNotifyScriptPath();
 	const script = buildWrapperScript(
 		"codex",
-		buildDefaultAccountResolver("CODEX_HOME", "default-codex-home") +
-			buildCodexWrapperExecLine(notifyPath),
+		buildDefaultAccountResolver(
+			"CODEX_HOME",
+			"default-codex-home",
+			"SUPERSET_AMBIENT_CODEX_HOME",
+		) + buildCodexWrapperExecLine(notifyPath),
 		{ agentId: "codex" },
 	);
 	createWrapper("codex", script);
--- a/packages/agent-setup/src/agent-wrappers-common.ts
+++ b/packages/agent-setup/src/agent-wrappers-common.ts
@@ -125,8 +125,10 @@ export function buildDefaultAccountResolver(
   superset_default_account="$(cat ${pointer} 2>/dev/null)"
   if [ -n "$superset_default_account" ] && [ -d "$superset_default_account" ]; then
     export ${envVar}="$superset_default_account"
+    export SUPERSET_DEFAULT_${envVar}="$superset_default_account"
   else
     unset ${envVar}
+    unset SUPERSET_DEFAULT_${envVar}
   fi
 fi
 
--- a/packages/agent-setup/src/agent-wrappers-common.ts
+++ b/packages/agent-setup/src/agent-wrappers-common.ts
@@ -117,8 +117,19 @@ function buildRealBinaryResolver(): string {
 export function buildDefaultAccountResolver(
 	envVar: string,
 	pointerName: string,
+	ambientEnvVar?: string,
 ): string {
 	const pointer = `"$SUPERSET_HOME_DIR/state/${pointerName}"`;
+	const restoreSystemDefault = ambientEnvVar
+		? `if [ -n "\${${ambientEnvVar}}" ]; then
+    export ${envVar}="\${${ambientEnvVar}}"
+    export SUPERSET_DEFAULT_${envVar}="\${${ambientEnvVar}}"
+  else
+    unset ${envVar}
+    unset SUPERSET_DEFAULT_${envVar}
+  fi`
+		: `unset ${envVar}
+  unset SUPERSET_DEFAULT_${envVar}`;
 	return `if [ -n "$SUPERSET_TERMINAL_ID" ] && [ -n "$SUPERSET_HOME_DIR" ] \\
   && { [ -z "\${${envVar}}" ] || [ "\${${envVar}}" = "\${SUPERSET_DEFAULT_${envVar}}" ]; } \\
   && [ -f ${pointer} ]; then
@@ -127,8 +138,7 @@ export function buildDefaultAccountResolver(
     export ${envVar}="$superset_default_account"
     export SUPERSET_DEFAULT_${envVar}="$superset_default_account"
   else
-    unset ${envVar}
-    unset SUPERSET_DEFAULT_${envVar}
+    ${restoreSystemDefault}
   fi
 fi
 
--- a/packages/agent-setup/src/ambient-codex-home.test.ts
+++ b/packages/agent-setup/src/ambient-codex-home.test.ts
@@ -1,4 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync, symlinkSync } from "node:fs";
+import { tmpdir } from "node:os";
 import path from "node:path";
 import { resolveAmbientCodexHome } from "./provider-profiles";
 
@@ -6,16 +8,19 @@ describe("resolveAmbientCodexHome", () => {
 	const HOME = "/home/tester";
 	let previousCodexHome: string | undefined;
 	let previousInjected: string | undefined;
+	let previousAmbient: string | undefined;
 
 	beforeEach(() => {
 		previousCodexHome = process.env.CODEX_HOME;
 		previousInjected = process.env.SUPERSET_DEFAULT_CODEX_HOME;
+		previousAmbient = process.env.SUPERSET_AMBIENT_CODEX_HOME;
 	});
 
 	afterEach(() => {
 		for (const [key, value] of [
 			["CODEX_HOME", previousCodexHome],
 			["SUPERSET_DEFAULT_CODEX_HOME", previousInjected],
+			["SUPERSET_AMBIENT_CODEX_HOME", previousAmbient],
 		] as const) {
 			if (value === undefined) delete process.env[key];
 			else process.env[key] = value;
@@ -25,12 +30,14 @@ describe("resolveAmbientCodexHome", () => {
 	it("falls back to ~/.codex when nothing is set", () => {
 		delete process.env.CODEX_HOME;
 		delete process.env.SUPERSET_DEFAULT_CODEX_HOME;
+		delete process.env.SUPERSET_AMBIENT_CODEX_HOME;
 		expect(resolveAmbientCodexHome(HOME)).toBe(path.join(HOME, ".codex"));
 	});
 
 	it("honours a CODEX_HOME the user set themselves", () => {
 		process.env.CODEX_HOME = "/home/tester/my-codex";
 		delete process.env.SUPERSET_DEFAULT_CODEX_HOME;
+		delete process.env.SUPERSET_AMBIENT_CODEX_HOME;
 		expect(resolveAmbientCodexHome(HOME)).toBe("/home/tester/my-codex");
 	});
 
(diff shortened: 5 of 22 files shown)


Kitenite:
## CDP verified in the dev app

Driven over CDP against a dev build of this branch (own port; the two other worktrees' dev apps were left alone).

**Setup.** A second Codex profile `~/.codex-verify` with a dummy token (no real credentials copied), seeded with its own rollout and history so the merge had something to move.

**Result of clicking "Make default" on it:**

| surface | before | after |
|---|---|---|
| `sessions` | real dir | symlink to `~/.codex/sessions` |
| `archived_sessions` | absent | symlink to `~/.codex/archived_sessions` |
| `shell_snapshots` | absent | symlink to `~/.codex/shell_snapshots` |
| `history.jsonl` | real file | symlink to `~/.codex/history.jsonl` |
| `auth.json` | real file | **still a real per-account file** |

The payoff, which is the reported bug:

```
sessions visible from the NEW account (~/.codex-verify): 1702
sessions in the system default        (~/.codex):        1702
earliest year visible from the new account: 2025
```

Before this change that first number was 0. The second account's pre-existing rollout was merged into the shared tree, and its history line was appended rather than dropped (9615 -> 9616 lines).

Also confirmed in the UI: the default marker moved between the two Codex cards, and the existing RestartSessionsDialog fired correctly for Codex ("Restart running Codex agents?"). Dismissed with Not now.

**Cleanup:** the verification profile was removed by unlinking each symlink first so nothing could traverse into `~/.codex`, the merged rollout was deleted, and `history.jsonl` was restored byte-exact (md5 match). `~/.codex` verified intact afterwards: auth.json present, no stray symlinks, no `.superset-merge` debris.

**Not covered by this run:** the ambient-`CODEX_HOME` circularity. Turbo does not pass that variable through to the dev host-service, so the condition could not be reproduced in-app; it stays covered by the unit tests in `ambient-codex-home.test.ts`, which fail if the guard is reverted.

Kitenite:
## CDP verification, part 2: the Claude path

The first CDP run only covered Codex. Since this branch **refactors the shared `session-share.ts` code that Claude also uses**, the Claude path carried real regression risk and was untested. Now covered.

Throwaway profile `~/.claude-verify` (synthetic `.claude.json`, no real credentials), seeded with one transcript and a history line, then **Make default** in the Usage tab:

| check | result |
|---|---|
| all 10 `CLAUDE_SHARE` dirs + `history.jsonl` | symlinked into `~/.claude` |
| seeded transcript | **moved** into `~/.claude/projects/-verify-project/` |
| reachable back through the profile symlink | yes |
| transcripts visible from the new account | 1 -> 1480 |
| history line merged | yes |
| `.claude.json` (identity) | stayed a real per-profile file |

So both providers now go through the generalized `SessionShareSpec` with verified behavior, including the data-moving merge case that the Codex run did not exercise (that profile had nothing to merge).

### Verification coverage, stated plainly

| change | how verified |
|---|---|
| Codex session sharing | CDP, real account switch |
| Claude session sharing (refactor regression) | CDP, real account switch |
| `mainHome` required-parameter refactor | CDP (both runs exercised the refactored code) + typecheck |
| Test-isolation guard | filesystem before/after on a full suite run, canary test, preload stubbed out to confirm it fails |
| Ambient-`CODEX_HOME` circularity | unit + mutation test only. Turbo does not pass the variable to the dev host-service, so the condition is not reproducible in-app |

Cleanup: both throwaway profiles removed by unlinking each symlink first so nothing could traverse into the real homes; merged artifacts deleted; `~/.claude` verified at 1479 transcripts, unchanged from before the run.

Kitenite:
## CDP verification, part 3: the circularity bug, reproduced A/B

The previous note said this one was not reproducible in-app because turbo does not pass `CODEX_HOME` to the dev host-service. That was a tooling limit, not a real one. Adding the variable to `globalPassThroughEnv` locally (reverted, not committed) makes the condition reachable, and the host-service then inherits it:

```
CODEX_HOME=~/.codex-circular
SUPERSET_DEFAULT_CODEX_HOME=~/.codex-circular
```

That is precisely what happens to a host-service restarted from a Superset terminal. Same build, same dummy second home, guard toggled:

| | **with** the guard | **without** it |
|---|---|---|
| holds the default slot (checked radio) | `~/.codex`, real PRO quota | **`~/.codex-circular`** |
| demoted to a secondary profile | `~/.codex-circular` | **`~/.codex`**, the real account |
| "Switch sign-in" command offered | `CODEX_HOME=…/.codex-circular codex login` | bare **`codex login`** |

The roles invert completely. The third row is the damaging part: the broken build offers bare `codex login` as the way to fix the circular entry, so following the UI would re-authenticate the user's **real** `~/.codex` while the interface claims to be repairing a different account.

This closes the last untested row. Every change on the branch now has in-app evidence:

| change | verification |
|---|---|
| Codex session sharing | CDP, real account switch |
| Claude session sharing (refactor regression risk) | CDP, real account switch, including the data-moving merge |
| `mainHome` required-param refactor | CDP, both runs built from the refactored code |
| Ambient-`CODEX_HOME` circularity | **CDP A/B, guard toggled** + unit and mutation tests |
| Test-isolation guard | filesystem before/after on a full suite run, canary test, preload stubbed to prove it fails |

`turbo.jsonc` and `provider-profiles.ts` restored; working tree matches the four commits. agent-setup 272 pass / 0 fail, host-service 1447 pass / 0 fail.

Kitenite:
Addressed all six findings from the latest Cubic review in `491565241`:

- load the repository isolation preload from desktop/UI package-local Bun setups
- make the isolation canary validate before importing/calling the pointer writer
- allow `~/.codex` to be a secondary profile when the genuine main `CODEX_HOME` is custom
- restore an interrupted pending session tree when cross-filesystem recovery cannot complete
- preserve the genuine ambient Codex home separately from Superset's injected selection
- canonicalize injected-path comparisons, including symlink aliases
- keep the wrapper's provenance twin synchronized when pointers change

CDP A/B on this exact branch with an isolated synthetic HOME and a genuine custom `CODEX_HOME`:

**Before (old provider-home exclusion restored temporarily):** clicking **Make default** on `~/.codex` changed the radio/pointer and showed the restart dialog, but `~/.codex/sessions` stayed a real directory and its seeded rollout did not merge into the custom main.

**After:** the same click made `~/.codex/sessions` and `history.jsonl` symlinks to the custom main; both seeded rollouts were present there; `auth.json` remained a real per-account file with identical contents; the restart dialog fired and was dismissed with **Not now**.

Verification:
- focused new/affected tests: 58 pass, 0 fail
- agent-setup full suite: pass
- host-service full suite: 1458 pass, 8 todo, 0 fail
- desktop full suite: 3238 pass, 0 fail
- agent-setup + host-service typecheck: pass after final canonicalization change
- targeted Biome + `git diff --check`: pass
- package-local preload runs did not leave test-home directories behind

Kitenite:
Follow-up review fixes are in `4af45cd79`:

- When the system-default account is a user-defined Codex home, the launch overlay now sets both `CODEX_HOME` and `SUPERSET_AMBIENT_CODEX_HOME`. This preserves actual Codex behavior while retaining nested host-service provenance.
- The secondary-home test now uses a temporary profile and compares the canonical result, so macOS `/var` → `/private/var` and symlinked homes cannot make it fail spuriously.
- The isolation preload records its generated home with a process-local global symbol. The canary requires that marker, canonicalizes both paths, and rejects canonical `~/.superset` before importing or invoking the pointer writer.

Verification after these edits:
- focused host-service tests: 32 pass, 0 fail
- agent-setup + host-service typecheck: pass
- targeted Biome + `git diff --check`: pass

Kitenite:
Resolved the final Cubic P2 in `a301de2ba`, including the return path the suggested one-line change did not cover.

The host launch overlay now marks the ambient custom `CODEX_HOME` as Superset-injected. The Codex wrapper receives `SUPERSET_AMBIENT_CODEX_HOME` and uses it when the pointer returns to system default. This preserves both directions for an already-running terminal:

1. custom ambient default → selected profile
2. selected profile → custom ambient default

Claude retains the existing unset-on-system-default behavior.

Verification:
- wrapper + ambient tests: 18 pass, 0 fail
- host default-account + isolation tests: 8 pass, 0 fail
- agent-setup + host-service typecheck: pass
- targeted Biome + `git diff --check`: pass

PR #7047, opened 2026-08-31
feat(marketing): rebuild /download as a real downloads page

Kitenite:
Rebuilds `/download` from a single-purpose auto-download interstitial into a browsable downloads page, in the shape peers ship it (Linear, Zed, Cursor).

## What changed

The site header and footer are restored (the `Header` bailed on `/download`, which left the page orphaned behind a lone logo). Above the fold: a status band with the detected platform, a spec block, and a platform-aware button. Then surface cards for Desktop, CLI and Mobile. Then the release catalog.

The catalog is live GitHub data, not hardcoded. It reads `desktop-v*` tags and groups each release's assets into per-OS columns with real sizes, latest expanded and older versions collapsed behind `<details>`, so it works with no JS. Three asset classes are filtered out: update manifests, the unversioned `latest` aliases (they would double every row), and the `-mac.zip` archives, which exist for electron-updater and are not what a person should install. If GitHub is unreachable the section degrades to a link rather than an empty shell.

The card visuals are drawn in code rather than captured. A screenshot on an evergreen page pins a version number, a shell prompt and real worktree paths into a page that outlives all three. Desktop reuses `AppMockup`, the same component the front page hero uses, so it tracks the real UI automatically.

## Two bugs fixed along the way

**Linux was told it was unsupported.** Linux desktop has shipped since February (AppImage, with auto-update, per the `2026-02-16` changelog entry), but the page still said "Superset is Mac-only for now" and offered a waitlist. Linux now auto-downloads like macOS, and the waitlist is Windows-only, which matches the assets we actually publish. No Windows installer has ever been released despite `electron-builder.ts` configuring NSIS.

**The site's fonts never applied.** `globals.css` declared `--font-family-sans` / `--font-family-mono`, but Tailwind v4's namespace is `--font-*`, so every page silently fell back to the system stack while Inter and IBM Plex Mono were downloaded and unused. The `font-optical-sizing` and `cv01`/`ss03` tuning was dead for the same reason. This is a site-wide visual change; I checked home, pricing, changelog and download.

## Notes for review

- `robots: index: false` is kept, because the page still fires the download on arrival and should not be a cold search landing. That does mean the catalog is not indexable. Cursor's page does not auto-download, which is how theirs stays indexed. Happy to change this, it is a conversion-path decision.
- 22 English strings need translations; `translate-catalogs.yml` fills them on this PR.

## Follow-ups, not in this PR

- 28 compare pages still say "macOS now; Windows and Linux coming".
- The CLI's own TUI footer prints `docs.superset.sh/cli`, which 404s. Only `/cli/getting-started` and `/cli/cli-reference` exist. Fixed the marketing link here; the CLI needs its own fix.

## Testing

9 tests cover the asset mapper against the verbatim `desktop-v1.25.1` asset list, including the stable asset keys. Verified they fail when the mapper breaks. Lint, typecheck and tests pass. Verified rendering over CDP; no hydration errors.

https://claude.ai/code/session_01DUN9fyJmJKwAkzgW55Efbe

---
## Summary by cubic
Rebuilds `/download` from a single auto-download interstitial into a browsable downloads page with a live release catalog. The header and footer are restored, surface cards cover Desktop, CLI, and Mobile, and the catalog groups each `desktop-v*` release's assets into per-OS columns with real sizes — latest expanded, older collapsed behind `<details>`, works without JS, falls back to a link when GitHub is unreachable; the page still fires a download on arrival, so `robots: index: false` stays.

The status band now also lists version, download size, and publish date for the visitor's platform, using the same catalog data. The three mock bands share one visual treatment, and the brand color returns to the section eyebrow, "Latest" tag, and hover states. Fixes four bugs the rebuild surfaced: Linux was wrongly told it's unsupported, the site's fonts never loaded, release dates caused a hydration mismatch across timezones, and auto-download fired before platform detection resolved.

**Bug Fixes**
- Linux desktop has shipped since February, but the page said "Mac-only" and offered a waitlist; Linux now auto-downloads, and the waitlist is Windows-only, matching the assets actually published. The ja and zh-CN waitlist copy is refreshed to the same Windows-only wording.
- `globals.css` used `--font-family-*` variables while Tailwind v4 needs `--font-*`, so every page silently fell back to the system stack; fonts now apply site-wide.
- Release dates were rendered with the visitor's timezone on the client and UTC on the server (Vercel), which React reported as a hydration mismatch; dates are now pinned to UTC via a shared helper.
- Auto-download started before platform detection resolved and sent everyone the Apple Silicon build; it now waits for detection, and the GitHub release fetch is capped at 5s so a slow API can't stall the page.

The 22 new strings ship translated across all enabled locales.

<sup>Written for commit 86346d4eaa6b62b2d97c8a0e68f3bd2549eb2bfa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7047?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Redesigned the download page with platform-specific downloads, release details, file sizes, and release notes.
  * Added expandable release history for macOS, Windows, and Linux assets.
  * Added CLI installation guidance with copy-to-clipboard support and a terminal preview.
  * Added desktop, CLI, and mobile preview cards; mobile is marked “Coming soon.”
  * Download behavior adapts to the detected platform, including Windows availability messaging.
  * The site header now appears on every route.

* **Localization**
  * Added translated download-page content across supported locales.


Kitenite:
Both CodeRabbit findings triaged.

**`Platform.Unknown` treated as downloadable — real, fixed in 2ef2b54a8.**

Confirmed and it was a regression this PR introduced. The interstitial it replaced gated on `isMacPlatform()`, which is `false` for `Unknown`, so it correctly waited for detection. I swapped that for `hasDesktopBuild()`, which answers `true` for `Unknown` so the download *button* always has something to offer. That fallback is fine for a link somebody clicks and wrong for an automatic redirect: the effect ran on first render, latched `firedRef`, and scheduled the Apple Silicon build for everyone. An Intel Mac would have got an unusable binary and Linux a `.dmg`.

Took the "gate the interstitial until detection completes" option: `hasDesktopBuild()` still backs the button, `shouldAutoDownload()` additionally requires a resolved platform. Also cancels the timer on re-run and latches `firedRef` inside the callback, so a platform resolving mid-flight cannot leave a stale URL pending. Regression test added as suggested.

**Linux asset name — incorrect, no change made.**

The claim is that the workflow produces `Superset-x64.AppImage` while the constant asks for `Superset-x86_64.AppImage`. It's the other way round. electron-builder maps `x64` to `x86_64` in the AppImage artifact name, so `artifactName: superset-${version}-${arch}.${ext}` yields `superset-1.25.1-x86_64.AppImage`, and the workflow's `sed` extracts `x86_64`, producing `Superset-x86_64.AppImage`.

Verified against the live release rather than by reading:

```
$ gh release view desktop-v1.25.1 --json assets   # AppImage assets
["superset-1.25.1-x86_64.AppImage", "Superset-x86_64.AppImage"]

$ curl -sIL -o /dev/null -w '%{http_code}' .../download/Superset-x86_64.AppImage
200
$ curl -sIL -o /dev/null -w '%{http_code}' .../download/Superset-x64.AppImage
404
```

The suggested name is the one that 404s, so the constant stays as is.

PR #7049, opened 2026-08-31
fix(browser): keep sign-in popups in the pane's session

Kitenite:
**Links**
- Issue: [SUPER-1272](https://linear.app/superset-sh/issue/SUPER-1272/browser-doesnt-work-well-with-popups) (reported in Discord by nitinl)

## Summary
- "Sign in with Google" never completed in the in-app browser. Popup OAuth flows now open as real popup windows on the pane's own session, instead of being flattened into a split pane or kicked out to the system browser.
- Two separate defects on the same path, both required for the flow to work end to end.

## Why / Context

The reported symptom is Google's cookie mismatch page. The underlying cause is that a `window.open` sign-in popup was never a popup:

**Defect 1: popups were denied outright.** `setWindowOpenHandler` denied *every* `window.open` and re-emitted the URL, which both renderer surfaces opened as a new split pane. A detached pane has no `window.opener` and no window name, so popup OAuth (Firebase `signInWithPopup`, Google Identity Services, Auth0) can never hand its result back to the page that started it. It also denied `about:blank` popups, which is how several auth libraries open the window before navigating it, so `window.open` returned `null` and tripped `auth/popup-blocked`.

**Defect 2: allowed popups got kicked to Chrome.** The app-wide guard in `electron-app/factories/app/setup.ts` sends any http(s) `will-navigate` in a non-webview `webContents` to the system browser. A popup is a `BrowserWindow`, not a webview, so fixing defect 1 alone made things look fixed while the sign-in URL opened in **Chrome**: the popup sat on `about:blank` forever while the user signed in in a different browser, landing the cookies in a different jar than the pane that started the flow. This is the half that actually produces the reported "cookie mismatch" behaviour, and it is invisible to unit tests.

## How It Works

- `resolveWindowOpen` gates on Chromium's own disposition. `new-window` (a scripted `window.open(url, name, "width=...")`) returns `action: "allow"`, so Electron creates a real window that keeps `window.opener`, the window name, and `window.close()`. Tab dispositions (`target="_blank"`) keep the existing split-pane behaviour, so that UX is unchanged.
- Popups inherit the opener's session. `partition` is deliberately never set in `overrideBrowserWindowOptions`, which is what keeps the popup's cookies in the pane's jar.
- Geometry is left to Electron, which already parses `width`/`height`/`x`/`y` from the `features` string and is outranked by `overrideBrowserWindowOptions`. Hand-rolling it would override Chromium rather than defer to it.
- A `WeakSet` of pane-popup `webContents` (`popup-window.ts`) lets the app-wide guard skip these windows. The check lives inside the `will-navigate` listener rather than at `web-contents-created`, because the mark is added on `did-create-window`, which Electron fires after creation but before the popup's first navigation.
- Popups get the pane's scheme guard and the same window-open policy, so the nested consent window Google opens mid-flow stays a popup too, and they load remote content with no preload, no Node, sandboxed.

## Manual QA Checklist

Verified over CDP against the dev app on this branch (renderer `localhost:4885`, dedicated debug port), driving a real mouse click on a real "Sign in with Google" button inside the guest webview. Baseline captured by reverting to the old deny-all path and repeating the identical journey.

| Observation | Before (deny-all) | After |
|---|---|---|
| `window.open` | `BLOCKED (returned null)` | returned a handle |
| `postMessage` to opener | none | `auth-ok:authpopup` delivered |
| Real popup window | 0 | 1, `name="authpopup"`, `opener=true` |
| Split panes created | popup flattened into a 2nd pane | none |
| Who fetched the OAuth URL | n/a | the app itself, not the system browser |
| Popup geometry | n/a | `outer=500x600`, matching the `features` string |

- [x] Popup sign-in completes: opener receives the callback
- [x] Popup opens on the pane's session (no `partition` override)
- [x] `target="_blank"` link still opens as a split pane (unchanged)
- [x] `about:blank` popup is allowed rather than returning null
- [x] Disallowed schemes (`file:`) still denied, popup or not
- [x] App launches without errors, pane restores across restart

Not covered: a real Google account sign-in end to end. See Known Limitations.

## Update: widened after testing a real account

Testing against a live Deel account found the disposition gate alone was not enough. Deel opens Google sign-in with a bare `window.open(url)` (no name, no features). Measured in Electron 41:

| call shape | disposition | frameName | features |
|---|---|---|---|
| `window.open(url)` (Deel) | `foreground-tab` | `''` | `''` |
| `window.open(url, name, "width=...")` (Firebase) | `new-window` | `'authpopup'` | `location=yes,...` |
| `window.open(url, name)` | `foreground-tab` | `'authwin'` | `''` |
| `<a target="_blank">` | `foreground-tab` | `''` | `''` |

Rows 1 and 4 are **indistinguishable**, so disposition cannot decide on its own. Gating only on `new-window` sent Google's account chooser to a split pane, where it lost `window.opener` and Deel's callback bounced back to `/login`.

The gate now also matches the shape of the URL: an OAuth 2.0 authorization request carries `client_id` + `redirect_uri` + `response_type` (RFC 6749 4.1.1). Keyed on those parameters rather than provider hostnames, so it covers any identity provider with no vendor allowlist to maintain. An ordinary `target="_blank"` link has none of them and still opens as a split pane, so that behaviour is unchanged.

**Verified end to end with a real Google account against `app.deel.com`:** the account chooser opens as a real popup window, the handshake completes, the popup closes itself, and the pane lands authenticated on Deel.

Worth noting for the open UA question below: Google raised **no** "browser or app may not be secure" wall in this flow, despite the pane presenting an `Electron/41.10.3` user agent.

## Edge-case sweep

Now a committed, repeatable smoke test: `apps/desktop/scripts/cdp-smoke-popups.ts`, alongside the existing `cdp-smoke-integrations.ts`. It serves its own fixture, drives thirteen `window.open` shapes through a live browser pane, and asserts popup count, opener identity, cookie-jar sharing, the returned handle, and pane count for each. Latest run: 13/13, 7 postMessage callbacks delivered.

```
case               popups  opener  jar   handle  panes  verdict
noArgs             1       true    -     HANDLE  +0     ok
blankFragment      1       true    -     HANDLE  +0     ok
oauthBare          1       true    true  HANDLE  +0     ok
oauthBlankTarget   1       true    true  HANDLE  +0     ok
oauthNamed         1       true    true  HANDLE  +0     ok
withFeatures       1       true    true  HANDLE  +0     ok
oidcHybrid         1       true    true  HANDLE  +0     ok
noopener           1       false   true  NULL    +0     ok
nested             2       true    true  HANDLE  +0     ok
selfClosing        0       -       -     HANDLE  +0     ok
blankLink          0       -       -     null    +1     ok
plainScriptedOpen  0       -       -     NULL    +1     ok
disallowedScheme   0       -       -     NULL    +0     ok
```

The `jar` column is the one that matters most: the popup reads a cookie the pane set, a direct check of the property whose absence was the reported symptom. `blankLink` and `disallowedScheme` are pinned as expected-negative so a future change cannot quietly turn every link into an OS window or let `file:` through.

Nine real `window.open` shapes driven through the running app. One misbehaved and is fixed; the rest now have measurements behind them.

| shape | handle | popup | `window.opener` | panes |
|---|---|---|---|---|
| `window.open()` (no args) | handle | 1 | intact | unchanged |
| `window.open("about:blank#state")` | handle | 1 | intact | unchanged (was: null + stray pane) |
| `window.open(oauthUrl, "_blank")` | handle | 1 | intact | unchanged |
| `window.open(oauthUrl, "name")` | handle | 1 | intact | unchanged |
| `window.open(oauthUrl, n, "noopener")` | null | 1 | none | unchanged (spec-correct) |
| OIDC hybrid `response_type=code id_token` | handle | 1 | intact | unchanged |
| nested popup opened from a popup | handle | 2 | intact | unchanged |
| `window.close()` from a popup | handle | 0 | n/a | unchanged |
| plain `<a target="_blank">` | n/a | 0 | n/a | split pane |

## Testing
- `bun run typecheck`
- `bun run lint` (6882 files, clean)
- `bun test src/main` (647 pass, 0 fail; 12 new tests across `popup-window.test.ts` and `browser-manager.test.ts`)
- `bun turbo run build --filter=@superset/desktop`
- `bun run --cwd packages/i18n check` (no new strings, no catalog changes)

The new tests are pinned in both directions: forcing `isPopupDisposition` to `false` (the old behaviour) fails exactly the three popup tests while the split-pane tests keep passing.

## Design Decisions
- **Gate on disposition, not on URL patterns or window name.** `new-window` is Chromium's own "this is a popup" determination, so it matches what a real browser does and needs no allowlist of auth providers. This is also what Ferdium landed for the same bug ([ferdium-app#2412](https://github.com/ferdium/ferdium-app/pull/2412), after [#1420](https://github.com/ferdium/ferdium-app/issues/1420) "oAuth window opens in browser"), arrived at independently here and cross-checked afterwards.
- **Exempt via a registry rather than loosening the global guard.** The app-wide "external URLs go to the system browser" rule is worth keeping for everything else; only pane popups need the carve-out.
- **One commit, not two.** Either half alone leaves the flow broken, so splitting would only produce a knowingly-broken intermediate revision.

## Known Limitations
- Tested end to end with a real Google account against Deel (see above). Not exercised against every provider; the OAuth detection is parameter-based, so a provider whose authorization endpoint omits `response_type` or `redirect_uri` would still fall through to the split-pane path.
- **A bare `window.open(url)` to a non-sign-in URL still returns `null` while the URL opens as a split pane** (`plainScriptedOpen` above). The page is told its popup was blocked even though we did open something, so a site that calls `w.focus()` on the result will throw. Pre-existing rather than new: before this PR *every* `window.open` behaved this way, sign-in included.

  I looked at closing it properly and do not think it can be done within this PR's scope. Four routes, all dead ends:

  1. **`createWindow` on `WindowOpenHandlerResponse`** — the right answer in principle, since a pane's webview could *be* the opened window and keep both the opener and the split-pane UX. It must return a `WebContents` synchronously, and panes are renderer-owned `<webview>` elements created asynchronously, so main cannot produce one in time. Would require moving panes to main-process `WebContentsView`s.
  2. **A preload that patches `window.open` in the guest** — with `contextIsolation` on, a preload runs in an isolated world and cannot touch the page's `window.open`. Turning it off for arbitrary web content is not a trade worth making.
  3. **Smuggling a marker through `frameName` or `features` from a main-world injection** — mutates page-observable state (`window.name`), risks flipping Chromium's own disposition computation, and any page that captures `window.open` first defeats it.
  4. **Treating every `foreground-tab` as a popup** — turns every `target="_blank"` link into an OS window, which is a worse regression than the bug.

  So it stays a documented gap, pinned by the smoke test so it cannot silently change, rather than a hack. The real fix is the panes-in-main architecture change.
- **Separate, still open: the "This browser or app may not be secure" wall.** The app sets no user-agent override, so panes present the default Electron UA and Google may reject it (`disallowed_useragent`). That is a different failure from this one and is deliberately out of scope. It did not trigger in the Deel flow, so it may be narrower than feared. Worth noting for whoever picks it up: UA spoofing is reportedly no longer a fix ([ferdium-app#2324](https://github.com/ferdium/ferdium-app/issues/2324), Jan 2026: "changing the user agent seems to no longer work no matter the value"), so the real answer is likely the system browser plus a loopback OAuth redirect. Our popup is now a real `BrowserWindow` rather than a webview, which may change Google's verdict on its own; untested.

## Follow-ups
- Optional polish: Ferdium also sets `parent: mainWindow` so a sign-in window cannot get lost behind the app. Skipped here because `BrowserManager` has no handle to the owning window and `outlivesOpener: false` already ties the popup's lifetime to its opener.

## Risks / Rollout
- Risk: low and contained to the in-app browser pane. The behaviour change is scoped to `window.open` with a popup disposition; `target="_blank"` links, the address bar, and CDP navigation are untouched. Popups keep the pane's scheme guard, so they cannot reach `file:`/`chrome:`.
- Rollout: ships with the desktop app, no flag, no migration, no API coupling.
- Rollback: revert this commit. Popups return to opening as split panes.

https://claude.ai/code/session_01QQV2Jd8VXMYgjRFQSeRNJH

---
## Summary by cubic
Fixes sign-in popups in the in-app browser so OAuth flows like "Sign in with Google" complete in the pane's own session instead of failing on Google's cookie mismatch page (SUPER-1272).

- Scripted `window.open` popups now open as real popup windows, keeping `window.opener`, the window name, and the opener's cookie session; a bare `window.open(url)` of an OAuth authorization URL is detected by parameter shape (`client_id`, `redirect_uri`, `response_type`, with `response_type` validated against spec-defined values) rather than by disposition, which cannot distinguish it from a `target="_blank"` link.
- Popups are exempt from the app-wide rule that sends navigations to the system browser, which previously landed sign-in cookies in a different browser's cookie jar.
- `about:` popups are allowed so auth libraries can open a window before navigating it; popups inherit the pane's scheme guard, and `file:` and `chrome:` URLs stay denied.
- Ordinary `target="_blank"` links still open as split panes.

**Testing**
- Adds `cdp-smoke-popups.ts`, a dependency-free smoke test driving thirteen `window.open` shapes through a live pane, asserting popup count, opener identity, cookie-jar sharing, returned handle, exact opener callbacks, and pane count. Its fail-safe cleanup closes each test-created pane through its own header action and verifies current target deltas, so a missed target cannot close a pre-existing pane and recreated webviews cannot escape.

<sup>Written for commit 04fc2efda658007e9782531088b21d21c63ce858. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7049?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Browser panes now support sign-in and other compatible popup flows directly within the app.
  * Popups preserve the originating pane’s session, opener relationship, and navigation protections.
  * Blank-page, nested-popup, and supported OAuth flows are handled consistently.

* **Bug Fixes**
  * Prevented valid popup navigation from being incorrectly redirected to the system browser.
  * Continued blocking unsupported tab-opening behaviors and disallowed URL schemes.

### DIFF
--- a/apps/desktop/scripts/cdp-smoke-popups.ts
+++ b/apps/desktop/scripts/cdp-smoke-popups.ts
@@ -391,150 +391,193 @@ async function main() {
 						.filter((target) => target.type === "webview")
 						.map((target) => target.id),
 				);
-				let actionError: string | null = null;
-				await g.eval(`window.__handle = null; ${script}; 1`).catch((error) => {
-					actionError = error instanceof Error ? error.message : String(error);
-				});
-				await sleep(
-					OBSERVE_MS ??
-						(name === "nested" || name === "selfClosing" ? 3000 : 2000),
-				);
+				const createdPopups = new Map<string, CdpTarget>();
+				const createdPanes = new Map<string, CdpTarget>();
+				try {
+					let actionError: string | null = null;
+					await g
+						.eval(`window.__handle = null; ${script}; 1`)
+						.catch((error) => {
+							actionError =
+								error instanceof Error ? error.message : String(error);
+						});
+					await sleep(
+						OBSERVE_MS ??
+							(name === "nested" || name === "selfClosing" ? 3000 : 2000),
+					);
 
-				const after = await targets();
-				const live = after.filter(
-					(target) => target.type === "page" && !pageIdsBefore.has(target.id),
-				);
-				const createdPanes = after.filter(
-					(target) =>
-						target.type === "webview" && !paneIdsBefore.has(target.id),
-				);
-				const opener: Array<boolean | null> = [];
-				const jar: Array<boolean | null> = [];
-				const windowNames: Array<string | null> = [];
-				const observationErrors: string[] = [];
-				for (const p of live) {
-					if (!p.webSocketDebuggerUrl) {
-						observationErrors.push(`popup ${p.id} has no debugger URL`);
-						continue;
+					const after = await targets();
+					const live = after.filter(
+						(target) => target.type === "page" && !pageIdsBefore.has(target.id),
+					);
+					const livePanes = after.filter(
+						(target) =>
+							target.type === "webview" && !paneIdsBefore.has(target.id),
+					);
+					for (const target of live) createdPopups.set(target.id, target);
+					for (const target of livePanes) createdPanes.set(target.id, target);
+					const opener: Array<boolean | null> = [];
+					const jar: Array<boolean | null> = [];
+					const windowNames: Array<string | null> = [];
+					const observationErrors: string[] = [];
+					for (const p of live) {
+						if (!p.webSocketDebuggerUrl) {
+							observationErrors.push(`popup ${p.id} has no debugger URL`);
+							continue;
+						}
+						const c = await Cdp.connect(p.webSocketDebuggerUrl).catch(
+							(error) => {
+								observationErrors.push(`popup ${p.id}: ${String(error)}`);
+								return null;
+							},
+						);
+						if (!c) continue;
+						try {
+							await c.send("Runtime.enable");
+							// Probe the popup directly. This also covers about:blank, which never
+							// loads the fixture's window.__info snapshot.
+							opener.push(
+								await c.eval<boolean>("!!window.opener").catch(() => null),
+							);
+							jar.push(
+								await c
+									.eval<boolean>('document.cookie.includes("paneprobe=1")')
+									.catch(() => null),
+							);
+							windowNames.push(
+								await c.eval<string>("window.name").catch(() => null),
+							);
+						} finally {
+							c.close();
+						}
 					}
-					const c = await Cdp.connect(p.webSocketDebuggerUrl).catch((error) => {
-						observationErrors.push(`popup ${p.id}: ${String(error)}`);
-						return null;
-					});
-					if (!c) continue;
-					try {
-						await c.send("Runtime.enable");
-						// Probe the popup directly. This also covers about:blank, which never
-						// loads the fixture's window.__info snapshot.
-						opener.push(
-							await c.eval<boolean>("!!window.opener").catch(() => null),
-						);
-						jar.push(
-							await c
-								.eval<boolean>('document.cookie.includes("paneprobe=1")')
-								.catch(() => null),
-						);
-						windowNames.push(
-							await c.eval<string>("window.name").catch(() => null),
-						);
-					} finally {
-						c.close();
-					}
-				}
-				const handle = await g
-					.eval<string>("String(window.__handle)")
-					.catch(() => "?");
+					const handle = await g
+						.eval<string>("String(window.__handle)")
+						.catch(() => "?");
 
-				const bad: string[] = [];
-				if (actionError) bad.push(`action failed: ${actionError}`);
-				bad.push(...observationErrors);
-				if (live.length !== want.popups)
-					bad.push(`popups ${live.length}!=${want.popups}`);
-				if (createdPanes.length !== want.panes)
-					bad.push(`panes +${createdPanes.length}!=+${want.panes}`);
-				if (want.opener !== undefined) {
-					if (
-						opener.length !== live.length ||
-						opener.some((value) => value !== want.opener)
-					) {
-						bad.push(`opener ${formatObservations(opener)}!=${want.opener}`);
+					const bad: string[] = [];
+					if (actionError) bad.push(`action failed: ${actionError}`);
+					bad.push(...observationErrors);
+					if (live.length !== want.popups)
+						bad.push(`popups ${live.length}!=${want.popups}`);
+					if (livePanes.length !== want.panes)
+						bad.push(`panes +${livePanes.length}!=+${want.panes}`);
+					if (want.opener !== undefined) {
+						if (
+							opener.length !== live.length ||
+							opener.some((value) => value !== want.opener)
+						) {
+							bad.push(`opener ${formatObservations(opener)}!=${want.opener}`);
+						}
 					}
-				}
-				if (want.jar !== undefined) {
-					if (
-						jar.length !== live.length ||
-						jar.some((value) => value !== want.jar)
-					) {
-						bad.push(`jar ${formatObservations(jar)}!=${want.jar}`);
+					if (want.jar !== undefined) {
+						if (
+							jar.length !== live.length ||
+							jar.some((value) => value !== want.jar)
+						) {
+							bad.push(`jar ${formatObservations(jar)}!=${want.jar}`);
+						}
 					}
-				}
-				if (want.names) {
-					const gotNames = windowNames
-						.filter((value): value is string => value != null)
-						.sort();
-					const wantedNames = [...want.names].sort();
-					if (
-						gotNames.length !== live.length ||
-						JSON.stringify(gotNames) !== JSON.stringify(wantedNames)
-					) {
-						bad.push(
-							`names ${JSON.stringify(windowNames)}!=${JSON.stringify(want.names)}`,
-						);
+					if (want.names) {
+						const gotNames = windowNames
+							.filter((value): value is string => value != null)
+							.sort();
+						const wantedNames = [...want.names].sort();
+						if (
+							gotNames.length !== live.length ||
+							JSON.stringify(gotNames) !== JSON.stringify(wantedNames)
+						) {
+							bad.push(
+								`names ${JSON.stringify(windowNames)}!=${JSON.stringify(want.names)}`,
+							);
+						}
 					}
-				}
-				if (want.handle !== undefined) {
-					const got = handle === "HANDLE";
-					if (got !== want.handle) bad.push(`handle ${handle}`);
-				}
-				if (bad.length)
-					failures.push(`${name}: ${bad.join(", ")} (${want.note})`);
+					if (want.handle !== undefined) {
+						const got = handle === "HANDLE";
+						if (got !== want.handle) bad.push(`handle ${handle}`);
+					}
+					if (bad.length)
+						failures.push(`${name}: ${bad.join(", ")} (${want.note})`);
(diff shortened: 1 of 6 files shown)


Kitenite:
Both review findings addressed in 92047aaac, one accepted as-is and one accepted in substance but not in remedy. A third issue came out of the reviewer's own research and is fixed too.

**1. `about:blank` popups — accepted, was a real bug.**

Confirmed against the code: `window.open("about:blank")` passes no features, so Chromium reports `foreground-tab`, which fell through the disposition gate and got denied. `window.open` then returns `null`, which an auth library that opens-then-navigates cannot distinguish from a blocked popup. My own comment on that branch claimed the case was "handled above", which was only true for `new-window` — the comment was wrong, not just the code. Blank opens are now popups, and the `about:blank` branch in `resolveWindowOpen` is gone because it became unreachable.

**2. `isOAuthAuthorizationUrl` too broad — accepted in substance, different remedy.**

The concern is fair: presence-only checks would pull an unrelated URL carrying those three parameter names out of the split-pane path. Fixed by constraining `response_type` to the values the specs actually define (RFC 6749 §3.1.1 plus the OIDC hybrid combinations, space-delimited and order-independent), and requiring http(s).

I did not take the suggested remedy of also requiring a recognized authorization endpoint. A provider-hostname allowlist goes stale and, more importantly, misses self-hosted identity providers (Keycloak, Authentik, an Okta custom domain). The error cost here is asymmetric: a false positive opens a link as a window instead of a split pane, which is cosmetic, while a false negative makes sign-in silently impossible, which is the bug this PR exists to fix. Constraining the value space removes the realistic false positive without taking on that risk. Regression test added for the exact case raised, a non-authorization URL carrying all three parameter names.

**3. `webPreferences` pinning — found via the review's own research, fixed.**

The Electron behaviour quoted in the analysis is the important part: if a child's `webPreferences` (notably `sandbox`) diverge from the opener's process, Electron isolates the child and `window.opener` comes back `null`. My `overrideBrowserWindowOptions` pinned `sandbox: true` and friends, which bought nothing — Electron already inherits the opener's security preferences and refuses to relax them — while carrying exactly that hazard against the one property this popup exists to preserve. Removed; the popup now inherits.

**Verification.** Re-ran all four shapes in the running app after the changes:

| case | popup | `window.opener` | panes |
|---|---|---|---|
| `about:blank` then navigate | 1 | intact | unchanged |
| bare `window.open(OAuth URL)` | 1 | intact | unchanged |
| `window.open(url, name, "width=...")` | 1 | intact | unchanged |
| plain `<a target="_blank">` | 0 | n/a | split pane, as before |

658 tests pass (5 new), typecheck and lint clean.

PR #7050, opened 2026-08-31
feat(ui): adopt the new design system across desktop and shared packages

harshithmullapudi:
Ports the palette, type scale and control-height tokens into `packages/ui` (`theme.css` for colour, `design-tokens.css` for scales/sizing/radii/fonts), reskins the theme engine onto them, moves control geometry onto Zed's model, and then fixes everything that never made it onto the tokens.

## The palette

`applyUIColors` writes UIColors as inline styles on `:root`, which beats any stylesheet rule, so the built-in light and dark themes carry the same values as `theme.css` — terminal ANSI and editor syntax colours included. Monokai, theme import and the system light/dark pairing are untouched.

This also fixes `hsl(var(--token))` call sites that broke outright: the tokens are full `oklch()` values rather than HSL channels, so those wrapped into invalid colours and rendered nothing.

## Geometry

Icon buttons were a full step larger than their text counterparts — a `size="sm"` button rendered at 24px beside a `size="icon"` at 36px. Icon sizes are now square at the matching text height, and every variant carries `border border-transparent` so switching variant no longer shifts a button by a pixel.

Controls are borderless with a hover fill; borders belong to containers. Dividers inside a surface use `--border-variant`, boundaries between surfaces keep full-strength `--border`. Elevation comes from the background ladder rather than shadows on inline controls, and radius drops to a 4px grid. Heights move onto `--btn-h-*`, `--input-h` and `--header-height`, which shrink on desktop and grow to touch targets under 768px.

## What wasn't on the tokens

**The sidebar.** `--sidebar-*` were still stale shadcn defaults, so the rail painted `#18181b`, blue-tinted, against a neutral `#191919` everywhere else. Nothing reads that family directly, so it now aliases the palette. Monokai and the marketplace themes keep their own authored sidebar colours.

**Status colours.** 233 class strings across 95 files were raw Tailwind palette utilities no theme could reach; they resolve through `success` / `warning` / `destructive` / `primary` now, with light/dark pairs collapsing to one token. Merged-PR and closed-issue purple had nowhere to land, so the status ramp is exposed as utilities and that state uses `status-1`. Diff addition/deletion colours stop branching on theme in JS.

Third-party brand colours are deliberately left literal: template gallery banners, the Slack logo, ringtone gradients, Paywall accents paired with hand-picked hex gradients, and the macOS traffic-light dots.

**Hover.** Four competing conventions. `--muted` is a 19% black wash in light, which is what made hover states read as grey blobs. The Files/Changes/Review tabs were worse than inconsistent — inactive tabs rested on `bg-border/30` and hovered to `bg-border/20`, so pointing at one made it *fade* instead of light up. Both tab strips read `--fill-hover` / `--fill-selected` now.

**Light mode** put the terminal on pure white while every surface around it was `--background` — a contrast dark mode doesn't have, because there the two are identical. The terminal surface tracks `--background` in both.

**Sidebar chrome.** The rail's right edge was drawn twice, once by `ResizablePanel` and once by the sidebar itself, so the seam rendered 2px. Pane frames drop from a 2px active ring to 1px. The presets strip and the collapsed rail's headroom strip get the bottom border they were missing, so the tab bar underline no longer breaks at the seam. List rows run full-bleed and square on a 16px indent grid, and the nav block — which sat 8px further in than the section headers below it — joins the same rail.

## Known gaps

- `bg-accent` (106 sites) and `grayAlpha-100` (44) survive as hover fills. Both are visually within a couple of percent of `--fill-hover`, but `bg-accent` is entangled with the `focus:` and `data-[highlighted]:` states in the shadcn menu primitives, so converting only the hover half would desync mouse from keyboard.
- `--success`, `--background-2/3` and `--border-variant` are stylesheet-only — the theme engine doesn't write them, so a custom theme can't drive the full surface ladder. Pre-existing, not introduced here.

## Testing

`lint`, `typecheck` and `packages/i18n check` all pass. Behaviour, props and logic are unchanged throughout — this is presentation only.

Full test suite is 211 failures against a **212** baseline on a clean tree; those are pre-existing parallel git integration tests, unrelated to this change.

Not verified over CDP: the running Electron had no `--remote-debugging-port` open and restarting it would have killed the active session. Visual claims here come from measuring screenshots of the running app plus reading the compiled values.

https://claude.ai/code/session_01QX5uz46zUJPqSgqTc5JRph

---
## Summary by cubic
Adopts the new design system tokens across the desktop app and `packages/ui`, so the palette, type scale, and control geometry now resolve through theme values instead of hardcoded Tailwind utilities, and custom themes can drive the full UI; this also fixes `hsl(var(--token))` call sites that rendered invalid colors, because the tokens are full `oklch()` values, not HSL channels. Presentation only — behavior and logic are unchanged, though secondary buttons move from `outline` to `ghost` and icon buttons square up to text height.

**What changed**
- Ports the palette, type scale, and control-height tokens into `packages/ui` (`theme.css` for colors, `design-tokens.css` for scales, sizing, radii, and fonts) and reskins the theme engine onto them; Monokai, theme import, and system light/dark pairing are untouched.
- Icon buttons are now square at the matching text height, every variant carries `border border-transparent` so switching variants doesn't shift a button by a pixel, and secondary buttons drop `outline` for `ghost`.
- Controls are borderless with a hover fill; dividers inside a surface use `--border-variant`, while boundaries between surfaces keep full-strength `--border`.
- Heights move onto `--btn-h-*`, `--input-h`, and `--header-height`, which shrink on desktop and grow to touch targets under 768px; radius drops to a 4px grid.
- 233 class strings across 95 files switch from raw Tailwind palette utilities to `success`, `warning`, `destructive`, `primary`, and `status-1`, and diff colors stop branching on theme in JS.
- The sidebar now aliases the palette instead of stale shadcn defaults; Monokai and marketplace themes keep their authored sidebar colors.
- The Files/Changes/Review tabs read `--fill-hover` / `--fill-selected`, the light-mode terminal tracks `--background`, and the rail seam, pane rings, and nav-block indentation are fixed.
- Third-party brand colors (Slack logo, template banners, ringtone gradients, Paywall accents, macOS traffic lights) stay literal.

**Known gaps**
- `bg-accent` (106 sites) and `grayAlpha-100` (44) survive as hover fills; `bg-accent` is entangled with menu primitive `focus`/`data-[highlighted]` states, and converting only the hover half would desync mouse from keyboard.
- `--success`, `--background-2/3`, and `--border-variant` are stylesheet-only — the theme engine doesn't write them, so a custom theme can't drive the full surface ladder.

`lint`, `typecheck`, and `packages/i18n check` pass. The full test suite is 211 failures against a 212 baseline on a clean tree; those are pre-existing parallel git integration tests, unrelated to this change.

<sup>Written for commit ed7878bed7c227ec3f0c4f623f02eb85ebf4e7f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7050?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/desktop/src/renderer/components/MarkdownEditor/components/CodeBlockView/CodeBlockView.tsx
```diff
@@ -49,7 +49,7 @@ export function CodeBlockView({
 					<DropdownMenuTrigger asChild>
 						<button
 							type="button"
-							className="flex items-center gap-1 h-6 px-2 text-xs bg-background/80 backdrop-blur border border-border rounded hover:bg-accent transition-colors"
+							className="flex items-center gap-1 h-[var(--btn-h-sm)] px-2 text-xs bg-background/80 backdrop-blur rounded hover:bg-grayAlpha-200 transition-colors"
```
saddlepaddle, line 52: intentional grayAlpha?

apps/desktop/src/renderer/screens/main/components/WorkspaceSidebar/WorkspaceListItem/WorkspaceStatusBadge.tsx
```diff
@@ -25,22 +25,22 @@ export function WorkspaceStatusBadge({
...
 			icon: (
 				<LuGitMerge
-					className={cn(iconClass, "text-purple-500")}
+					className={cn(iconClass, "text-status-1")}
 					strokeWidth={STROKE_WIDTH}
 				/>
 			),
-			bgColor: "bg-purple-500/10",
-			hoverBgColor: "hover:bg-purple-500/30",
+			bgColor: "bg-status-1/10",
```
saddlepaddle, line 42: what is status-1 in terms of a color type?

apps/desktop/src/renderer/screens/main/components/WorkspaceView/RightSidebar/ChangesView/components/FileList/FileListGroupedVirtualized.tsx
```diff
@@ -203,7 +203,7 @@ export function FileListGroupedVirtualized({
 									{null}
 								</FolderRow>
 							) : (
-								<div className="ml-1.5 border-l border-border pl-0.5">
+								<div className="ml-1.5 border-l border-border-variant pl-0.5">
```
saddlepaddle, line 206: nit: another strange key

Kitenite:
I reviewed this on top of latest main and captured the before/after here:

https://app.superset.sh/page/pr-7050-design-system-before-after-df6az0

The implementation is largely sound programmatically—desktop/shared UI typechecks and changed-file lint pass—but visually, I think the app looks significantly worse after this change. The smaller controls, flatter surfaces, and weaker separation between elements make the interface harder to scan. The settings and dark/light comparisons on the page show this particularly clearly.

I don't think we should ship this visual direction as-is. Please revisit the control sizing, surface contrast, and visual hierarchy before moving forward.

There are also two concrete issues to address: keyboard-focused inputs lose their visible focus indicator (reproduced with matched before/after evidence), and the theme suite goes from 29/29 passing on main to 28/29 after integration. The page includes the exact SHAs, measurements, and verification limits.

harshithmullapudi:
Thanks for the detailed pass, but I think we're judging this against the wrong goal.

The point of this PR isn't to finalise the design. It's to get consistency — so that everything resolves through one set of tokens and later we have *one* place to make design changes instead of 233 class strings spread across 95 files. The visual direction is completely up for debate, and that debate gets a lot cheaper once this lands.

On a few of the comments — I don't think it makes sense to lean on Claude as the judge of the visual direction. If you asked it about Codex's UI it would say much the same thing. Most of the designs that are currently popular among dev tools, Claude probably wouldn't approve of either, and I'm not expecting it to.

My eye may not be great either 😄 but I'd still rather go with that, and ideally with an actual designer 😂

The two concrete findings are separate and fair — the lost focus indicator on keyboard-focused inputs and the theme suite going 29/29 → 28/29. I'll fix both.

And yeah — this PR was really meant as a reference for @AviPeltz, a starting point for us to align on the design from.

saddlepaddle:
Hey @harshithmullapudi, sorry for letting this sit so long. We're not prioritizing the design-system migration right now, so I'm moving this to draft rather than trying to land it.

It's not wasted: consolidating onto one set of tokens is the right direction, and this stays here as the reference for when we pick the design work back up with @AviPeltz. When we do, we'll likely start with a tokens-only pass that keeps today's look, and take the visual changes separately. Apologies again for the slow turnaround on this one.

PR #7054, opened 2026-09-01
fix(marketing): stop inferring Mac CPU arch from the WebGL renderer

### LINKED ISSUE #7036: Website downloads x64 by default
### Where did this happen?

Web app (app.superset.sh)

### Version

N/A

### What happened?

Accessed your website via a M5 MBP. Clicked download. It downloaded the x64 instead of arm version of app. Not sure if x64 is universal, but it was confusing nonetheless. Had to go to github to grab amd specifically.

### Steps to reproduce

Accessed your website via a M5 MBP. Clicked download. It downloaded the x64 instead of arm version of app. Not sure if x64 is universal, but it was confusing nonetheless. Had to go to github to grab amd specifically.

_No response_

Kitenite:
Fixes #7036

## Summary

- `detectMacArch` read the WebGL unmasked renderer string and treated any string without `apple` in it as proof of an Intel Mac. Apple Silicon visitors whose browser masks or software-renders that string were handed `Superset-x64.dmg`.
- Replaced with UA Client Hints (`navigator.userAgentData`). When the browser exposes nothing (Safari, Firefox) the page defaults to Apple Silicon, labels the pill plain "macOS" instead of claiming a detection, and links the Intel build one click below the button.
- `download_started` / `download_manual_clicked` carry `archSource` (`client-hint` / `default` / `not-mac`), and a new `download_arch_switched` event fires when someone takes the Intel link, so PostHog can size how often the default is wrong.
- Added the first tests for this module, including the negative case that pins the regression.

## Why / Context

The reporter was on an M5 MacBook Pro and got the Intel build, then had to go to GitHub Releases to fix it by hand.

Inferring a CPU property from a GPU string is wrong regardless of this bug: the two aren't related, and the renderer string is absent under software rendering and forged by fingerprint-resistant browsers and extensions. Every genuine failure path already fell back to Apple Silicon; only a *present but wrong* renderer string produced Intel.

`/download` is the only surface that can serve the x64 build. `apps/web` and `apps/docs` hardcode arm64.

## How It Works

`detectMacArch` now:

1. Asks `navigator.userAgentData.getHighEntropyValues(["architecture"])`, which reports the CPU directly in Chromium and is unaffected by GPU state.
2. Normalises the answer by prefix (`normalizeArch`), so `x86_64` and `arm64` map correctly rather than falling through.
3. Returns Apple Silicon with `archSource: "default"` when there is no signal at all.

The interstitial treats `archSource: "default"` as unconfirmed: the platform pill reads "macOS", and a line under the button says the browser didn't report a chip and links `Superset-x64.dmg` for Intel Macs. The `ReleaseList` from #7047 still lists every asset below.

## Design Decisions

- **Why default unconfirmed to Apple Silicon, not x64.** An intermediate commit on this branch defaulted to x64 because it runs on either chip via Rosetta. PostHog for the last 30 days says that mis-serves more people than it rescues:

  | Browser | Mac downloads | Resolved Intel under the old code |
  |---|---|---|
  | Chrome + Edge + Opera | 4,110 | 232 |
  | Safari + Firefox | 626 | 4 |

  Safari and Firefox have no client hint, so an x64 default sends all ~626/month to the Intel build, and nearly all of them are Apple Silicon. Three things make that landing worse than "degraded but runs": a fresh Apple Silicon Mac has no Rosetta installed, so first launch is an install prompt; the auto-updater keeps the user on x64 for every future release; and Apple has said Rosetta narrows after macOS 27. An arm64 build on an Intel Mac fails loudly with a "can't be opened" dialog, and the note on the page already points at the fix.
- **Why not invert the WebGL check to require a positive `intel|amd|radeon` match**: it would fix the reported case but keep a heuristic that misfires on exactly the privacy-configured browsers that cause the problem. Deleting it removes the failure mode instead of narrowing it.
- **Why prefix matching in `normalizeArch`**: exact `=== "x86"` would send an engine reporting `x86_64` to the Apple Silicon default, which is this same silent-wrong-build bug pointed the other way. I can't test a real Intel Mac, so the assumption is made cheap to be wrong about.
- **Why no `Sec-CH-UA-Arch` server tier**: it covers no browser `navigator.userAgentData` doesn't, and it makes the HTML vary on a request header while Next silently drops the `Vary` that would make that safe to cache.
- **Why not a universal binary**: it removes the guess entirely, and Cursor, Slack and Obsidian ship universal. But VS Code's universal build is 1.74x its arm64 build, which would put Superset near 1GB, paid by every Apple Silicon user on every download and auto-update. Worth a deliberate decision, not this PR.

## Manual QA Checklist

Reproduced and verified on real Apple Silicon (M5 Max), driving Chrome over CDP against the page with downloads denied.

Before the fix, against production `superset.sh/download`:

- [x] Chrome with SwiftShader software rendering requested `Superset-x64.dmg`
- [x] Chrome with a masked renderer string (`Mesa OffScreen`) requested `Superset-x64.dmg`
- [x] Stock Chrome requested `Superset-arm64.dmg` (control)

After the fix, against this branch:

- [x] SwiftShader: `Superset-arm64.dmg`, page labels it "macOS · Apple Silicon"
- [x] Stock Chrome: `Superset-arm64.dmg` (control)
- [x] Client hint `x86`: `Superset-x64.dmg`, page labels it "macOS · Intel"
- [x] Client hint `x86_64`: `Superset-x64.dmg`
- [x] Client hint `arm64`: `Superset-arm64.dmg`
- [x] Unrecognised value (`wasm`): `Superset-arm64.dmg` (safe default)
- [x] Real browsers against the final commit (`bun` drivers over CDP / WebDriver BiDi on throwaway profiles, downloads denied or github.com blocked):

  | Browser | Arch signal | Redirected to | Pill | Intel note |
  |---|---|---|---|---|
  | Chrome 152 stock | client hint `arm` | `Superset-arm64.dmg` | macOS · Apple Silicon | no |
  | Chrome 152 `--use-angle=swiftshader` (renderer = SwiftShader, the old Intel trigger) | client hint `arm` | `Superset-arm64.dmg` | macOS · Apple Silicon | no |
  | Chrome 152 under Rosetta (`arch -x86_64`, chrome://version says "x86_64 translated") | client hint `arm` | `Superset-arm64.dmg` | macOS · Apple Silicon | no |
  | Comet (Chromium 152) | client hint `arm` | `Superset-arm64.dmg` | macOS · Apple Silicon | no |
  | Firefox 154 stock | none | `Superset-arm64.dmg` | macOS | yes, links `Superset-x64.dmg` |
  | Firefox 154 `privacy.resistFingerprinting` (renderer = "Mozilla", the old Intel trigger) | none | `Superset-arm64.dmg` | macOS | yes |

- [ ] Safari 26.6: not driven. WebDriver needs `sudo safaridriver --enable` on this machine. Safari has no `userAgentData` (WebKit standards-positions #70 is still open), so it takes the same path Firefox exercised above.

Also confirmed the healthy path was never broken: a port of the old `detectMacArch` resolved correctly in Chrome, Safari 26.6, Firefox 148, Arc and Comet on this machine, which is why this only ever bit a subset of visitors.

## Testing

- `bun run --cwd apps/marketing typecheck`
- `bun test --cwd apps/marketing` (85 pass, 6 files)
- `bunx biome check apps/marketing`
- `bun run --cwd packages/i18n check` (strict compile 0 missing across 17 locales, stale audit clean)

Confirmed the tests can fail: reintroducing the Intel default turns 4 red, and reverting `normalizeArch` to exact matching turns 6 red.

## Known Limitations

- **Intel Macs on Safari and Firefox get the Apple Silicon default.** Those browsers expose no architecture signal. The old WebGL check did not classify them correctly either (Safari masks the renderer as "Apple GPU" on every Mac), so this is not a regression, but the recovery is now explicit: the page says the chip wasn't detected and links the Intel build.
- **The Intel branch is verified with a synthetic client hint, not real Intel hardware.** Prefix normalisation is the hedge.
- **This may not be what the reporter hit.** Every stock browser on my M5 Max resolves correctly, so I could only reproduce under software rendering or a masked renderer. Their report says `app.superset.sh`, which hardcodes arm64 and can't produce x64. The bug fixed here is real and proven; whether it's theirs is unconfirmed. `archSource` in PostHog will size it.
- **The 16 non-English translations of the new note were written by hand in this PR**, not by the translate workflow. Reword freely.
- **iPad Safari is classified as a Mac.** iPadOS has sent the desktop `Macintosh` user agent by default since iPadOS 13, so an iPad visitor hits `detectMacArch`, gets the Apple Silicon default, and auto-downloads a DMG instead of seeing the "open this on your desktop" link. Not a regression (the WebGL path did the same via "Apple GPU"), and PostHog can't size it because it files iPads under Mac OS X too. The standard tell is `navigator.platform === "MacIntel" && navigator.maxTouchPoints > 1`, which Claude, Cursor, Obsidian and Postman all use. One-line follow-up.

## Follow-ups

- `apps/web/src/app/(agents)/agents/page.tsx:73`, `apps/web/.../settings/billing/page.tsx:63` and `apps/docs/.../DownloadButton.tsx` hardcode arm64 with no escape hatch. Routing them at `superset.sh/download` would keep this logic in one place.
- The desktop app could detect a wrong-arch install at runtime from `process.arch` plus a Rosetta check, which catches every path including builds grabbed from GitHub.

https://claude.ai/code/session_01YU7cRGNVocDp2PwUCKbzU7
https://claude.ai/code/session_01DyrTQocncE9XV6tHTXe3J8


Kitenite:
Re the two cubic P2/P3 notes: the fallback direction is now Apple Silicon again (f29b25e0e7), with the data behind that choice in the description's Design Decisions. The test comment and PR body are updated to match what ships.

PR #7092, opened 2026-09-02
refactor(i18n): make the English text the message id

saddlepaddle:
## What & why

Follows #7091 (merged). Moves every surface to Lingui's default mode, where the English text is the message id, and deletes the machinery explicit ids required.

With explicit ids, editing English under a stable id left every translation in place and non-empty, so `compile --strict` saw nothing wrong and the old wording shipped. A custom stale-translation audit (plus an exemptions file and a full-history checkout in CI) existed only to catch that. With the text as the id, an English edit creates a new entry that is empty in every locale and the strict compile catches it natively. Identical English also collapses into one entry: 6,883 catalog entries become 4,894.

What changed:

- **Codemod over 971 source files.** `<Trans id="…">` → `<Trans>`; `t({ id, message })` and `msg({ id, message })` drop the id; runtime `i18n._({ id, message })` becomes `i18n._(msg({ message }))`, with `values` spread into the descriptor where present, since Lingui 6 types `_` as taking a descriptor alone. Three hand-built descriptor objects that were never extracted become `msg()` calls and are translated for the first time. Descriptors whose message is a variable keep their runtime id; they were never in the catalog.
- **Catalogs re-keyed by English text.** A carry-over script mapped every old id to its text and voted where several ids shared one text: 78,096 translations carried over, 703 votes had disagreeing old translations (mostly the same word translated inconsistently, e.g. German "Home" as "Start" under five ids and "Home" under one), ties preferring a translation that is not just the English. 13 strings that first entered the catalog here were translated by hand across all 16 locales.
- **Contexts where one English word meant two things.** "View" (app menu vs. button), "Plan" (billing vs. agent mode), "Exploring" (roadmap status vs. agent activity), "Open" (pull-request status vs. action), "Home" (directory vs. navigation), and "Unknown" (a person vs. a branch or task, which French genders differently) carry a `context` and keep their separate translations.
- **Every build runs the macro.** The renderer, the Next apps, and mobile already did. The Electron main process gets `linguiMacroPlugin` in `apps/desktop/vite/helpers.ts`; the bun-compiled CLI, host-service, and desktop `bundle:cli` get a Bun bundler plugin exported from `@superset/i18n/bun-plugin` (forwarded through a new `plugins` option in cli-framework); the API's Next config gets the same SWC plugin its siblings use. `apps/desktop/scripts/verify-single-flight.ts` mocks the macro alongside its electron mocks.
- **Deleted**: `check-stale-translations.ts` and its test, `locales/en-only-changes.txt`, the `fetch-depth: 0` checkout in CI's Lint job. `check` is now extract + strict compile.
- **One test shim for `@lingui/core/macro`.** `bun test` runs uncompiled source, where the macro entry point throws. `packages/i18n/test-setup.ts` derives ids with Lingui's own `generateMessageId`, so tests that load a real catalog get real translations; shared, cli, marketing, and mobile preload it by relative path. Desktop and ui keep their existing React-macro shims. `@lingui/core` is declared where the macro is now imported (`packages/shared`, `apps/docs`).
- **Catalog integrity test.** `packages/i18n/test/catalog-integrity.test.ts` checks every translation keeps its `{placeholders}`, `<0>…</0>` tag markers, and ICU plural/select shape. Strict compile only checks that an entry exists; this is the validation the deleted translator used to do for machine fills, now applied to hand-written ones. It found two messages whose Japanese, Chinese, and French translations had lost their second line on main (the merge confirmation's "This merges into {2}…" and the sign-in legal text's Terms and Privacy links); those are fixed here, plus the German versions that were still English.
- Docs: `AGENTS.md` and `packages/i18n/README.md` describe the text-as-id workflow and when to add a `context`.

The 120k-line catalog diff is the re-keying itself; the compiled `messages.ts` files keep the same message texts under hashed ids.

## How I tested it

- `bun run check:i18n`: strict compile passes with zero missing entries across 16 locales, and a second run leaves the catalogs unchanged.
- `bun run typecheck` across the monorepo: 0 errors. `bun run lint`: clean. `bun run test` (turbo): all packages pass, including desktop (3,286), shared (920), ui (60), i18n (100).
- The main-process plugin run directly over `src/main/lib/menu.ts` emits 21 hashed ids and leaves no `msg()` calls.
- German spot checks after carry-over: plain "View" → Ansehen, menu "View" → Ansicht; plain "Plan" → Plan, billing "Plan" → Tarif; "Restart" → Neu starten (the tie-break no longer picks the untranslated copy).

## Checklist

https://claude.ai/code/session_013UiC9jnCiE3bpgg6YzPje1


Kitenite:
Thanks for pushing on this. I'd like to hold this one and take #7091 on its own, with #7121 going in first. The short version of why:

- **The conflict problem lives in the committed `messages.ts`, and this PR keeps committing it.** Both PRs that were CONFLICTING this week conflicted on exactly those 17 one-line files and nothing else. #7121 gitignores them and compiles at install/build, and a 3-string PR drops from 34 locale files to 17. Please rebase #7091 on top of it once it merges; the `.po` conflicts regenerate away and the `origins: false` change is a nice complement.
- **Text-as-id turns homonyms into collisions.** "Open" (status vs action), "Home" (directory vs nav), "Unknown" (person vs branch). The six `context`s here cover the ones found so far, and the bots flagged more; 7 review threads are still open. With explicit ids this class of bug cannot exist.
- **Every bundler now has to run the macro**, including a new Babel-based Bun plugin for the CLI, host-service, and the desktop main process. That is the same build-surface that produced three CI gaps on #7121 (release workflow, CLI workflow, turbo ordering). I'd rather not widen it right as we are shrinking it.
- **Merge hazard.** 1,031 files; against main today it conflicts on 5 source files plus all 34 catalogs, and every in-flight PR that touches a string will conflict with it. There is no committed codemod to rerun after a rebase.

The parts I do want from here: the catalog-integrity test (placeholders, tags, ICU shape) and the `bunx --bun … --workers 1` invocation from #7091, which I reproduced: under Bun's worker pool `compile --strict` exits 0 on a missing translation. If you want to split the integrity test out, it can land independently of the id question.

Happy to talk through the stale-translation angle if you think the audit is too much to carry; the dedupe (6,883 to 4,894) is the strongest argument for this, and I'd rather weigh it deliberately than take it inside a 1,000-file diff.

PR #7094, opened 2026-09-02
fix(host-service): remove worktree with a native rm before git unregisters (#6887)

### LINKED ISSUE #6887: [bug] Workspace delete times out at 120s — git's single-threaded recursive delete can't finish a node_modules-heavy worktree
### Where did this happen?

Desktop app (macOS)

### Version

1.24.2 (git 2.50.1, Apple Git-155), Darwin arm64

### What happened?

Deleting a workspace fails with:

```
Failed to verify worktree removal at ~/.superset/worktrees/<project-id>/<worktree>:
[host-worker] Task "git/removeWorktree" timed out after 120000ms
```

This is **not** a hang, and not #6730's swallowed-error case — the removal is genuinely
still running when the 120 s task timeout fires. `gitWorktreeRemoveTask` lets **git** do
the recursive delete, and git's `remove_dir_recursively` is single-threaded.

Numbers from one affected worktree (a pnpm/Turborepo JS monorepo):

| | |
|---|---|
| filesystem entries in the worktree | 161,521 |
| …inside `node_modules` | 140,157 (**87%**) |
| directories | 17,972 |
| workspaces on this machine carrying `node_modules` | 22 of 26 |

Stripping *just* the `node_modules` trees with a **4-way parallel** `rm -rf` took **56 s**
on this machine. Git's single-threaded walk over the same set, plus the remaining ~21k
entries, does not fit in 120 s.

Environment factors that make it much worse, and which are ordinary for a heavy
multi-workspace user: 10-core / 16 GB machine, ~26 workspaces open, load average ~18,
swap fully consumed (5.7 GB of 6 GB), disk 96% full. Cold metadata I/O is the
bottleneck — `du -sh ~/.superset/worktrees` did not complete in 10 minutes.

Ruled out as causes: stale `.git` lock files, hooks, submodules, `core.fsmonitor`,
nested repos, immutable flags. `git worktree list --porcelain` returns in 50 ms. Git
itself is healthy, and a synthetic worktree of 160k files in only 3.2k directories
removes in 19 s — directory count and cold cache are what push the real tree over.

### Why raising the timeout isn't the fix

The timer is armed at **dispatch**, not enqueue — `WorkerTaskRunner.drainQueue` sets
`task.timeoutHandle` when a slot picks the task up — so worker-pool queueing is already
correctly excluded. The 120 s is real, single-threaded delete work.

### Steps to reproduce

1. Open a workspace on a large JS monorepo and install dependencies, so the worktree
   carries ~150k files across many nested `node_modules`.
2. Have enough other workspaces open that memory/IO are under pressure and the tree's
   metadata is cold.
3. Delete the workspace.
4. After 120 s, the toast above appears. The workspace correctly un-archives and stays
   visible — but it cannot be deleted, and retrying hits the same wall.

### Suggested fix

`gitWorktreeRemoveTask` doesn't need git to do the deleting. The task's own comment
already notes that `--force --force` **also unregisters a worktree whose directory is
already gone**. So:

1. `rm -rf` the directory first with a parallel/native delete (or `fs.rm` with
   concurrency) — measurably 4–8× faster than git's serial walk on the same tree.
2. Then call `git worktree remove --force --force` purely to unregister — near-instant.
3. Then verify with `git worktree list --porcelain`, exactly as today.

Additionally:

- Scale the timeout with tree size, or emit progress, rather than a flat 120 s cap.
- Distinguish "still working" from "failed" in the toast. Today a timeout reads as a
  hard failure, with no hint that the operation was progressing or that it would
  succeed against a smaller tree.

### Workaround

Delete `node_modules` before deleting the workspace. It is 87% of the inodes, and
removing it drops the tree to ~21k entries, after which the delete finishes in under a
second.

```bash
find <worktree> -maxdepth 4 -name node_modules -type d -prune -print0 \
  | xargs -0 -P4 rm -rf
```

### Related issues

- **#6730** — same command, different failure mode (git unregisters mid-failure and the
  app trusts it). Both trace back to `gitWorktreeRemoveTask` letting git own the
  recursive delete; doing the `rm -rf` in-app first shrinks the surface for that bug too.
- **#5631** — auto-prune / storage cleanup. A "reclaim `node_modules` from idle
  workspaces" action would make workspaces deletable again. Worth noting for that issue
  that with pnpm the win is **inodes, not bytes**: every `node_modules` file here is a
  hardlink with a link count of 85, shared across all workspaces and a single 8.9 GB
  content-addressable store, so stripping nine workspaces freed ~0 disk while cutting
  each tree from 161k to 21k entries. `du` over the worktree root double-counts those
  hardlinks badly. It's the metadata volume that breaks delete, not the disk usage.
- **#6717** — process/scaling cost with open worktree count; same "many workspaces on
  one machine" pressure.

andyst-dev:
Fixes #6887

## What & why

Deleting a worktree lets **git own the recursive delete** — `git worktree remove --force --force` calls git's `remove_dir_recursively`, a single-threaded walk that cannot finish a node_modules-heavy tree (161k+ entries, pnpm hardlinks) inside the task's 120 s budget. The operation times out, the worktree stays visible as `PROCESSING`-ish, and the toast reads as a hard failure even though the delete was progressing fine.

**This PR stops delegating the delete to git.** `gitWorktreeRemoveTask` now:
1. Removes the directory with the **native Node recursive `rm`** (`force`, so a missing or concurrently-cleaned path is a no-op) — measurably several times faster than git's serial walk on the same tree.
2. Then calls `git worktree remove --force --force` purely to **unregister** the now-empty worktree — near-instant.
3. Then checks `git worktree list --porcelain`, exactly as today.

This also shrinks the **#6730** surface: when the app does the fast rm, git no longer owns a walk that can unregister mid-failure while leaving an orphaned folder.

## How I tested it

- New regression tests (`packages/host-service/src/workers/tasks/git-remove.test.ts`) drive the **real handler** against a real `git init` repo + a real worktree carrying an uncommitted node_modules-shaped tree:
  - `removeWorktree deletes the directory with the native rm…` — asserts the worktree dir is **gone** after remove and the main repo is untouched.
  - `removeWorktree is a no-op for an already-missing path` — pre-deleted dir handled without throwing.
- `bun test src/workers/tasks/git-remove.test.ts src/workers/host-worker-pool.test.ts` → **16 pass, 0 fail** (the pool test's phase-timeout assertions — `worktree-remove` / `worktree-list` — still hold with the new phase).
- `bun x tsc --noEmit -p tsconfig.json` → clean.
- `biome check` on the two changed files → clean.

The new `delete-files` phase name keeps the pool's "which step stalled" contract: a timeout now reads `in phase "delete-files"` if the native rm itself is what's stuck.

## Checklist

- [x] `bun run lint` and `bun run typecheck` pass

---
## Summary by cubic
Fixes #6887 by removing a worktree directory with Node's native recursive `rm` before asking Git to unregister it. Previously, Git performed the recursive delete and could time out on `node_modules`-heavy trees, leaving the worktree in a processing state and showing a failure; Git now only unregisters the emptied worktree, while missing paths remain safe no-ops.

**Bug Fixes**

- Native deletion only runs when the caller confirms the path is inside the project's managed worktrees root; otherwise git's own removal is kept.
- Reports a new `delete-files` phase so pool timeouts identify whether native deletion stalls.
- Adds real-repository regression tests for heavy uncommitted trees, already-missing directories, and the git-only fallback.

<sup>Written for commit 98782ec72e3ddd242895ca361e5fd2e350cb2602. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7094?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Worktree removal now reliably deletes the associated directory and its contents, including deeply nested uncommitted files.
  * Worktree registry entries are removed while preserving the main repository.
  * Missing worktree paths are handled without errors.
  * Cleanup is restricted to managed worktree locations, helping prevent unintended deletion of files outside the workspace.

### DIFF
--- a/packages/host-service/src/trpc/router/workspace-cleanup/git-ops.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/git-ops.ts
@@ -61,6 +61,10 @@ export const cleanupGitOps = {
 		repoPath: string;
 		worktreePath: string;
 		gitEnv: GitTaskEnv;
+		/** Confirmed inside the managed worktrees root by the caller before it
+		 * is set — the worker will native-rm the directory first (see
+		 * `gitWorktreeRemoveTask`). */
+		nativeRm?: boolean;
 	}): Promise<{ stillRegistered: boolean; removeError?: string }> {
 		// Generous timeout: removal recursively deletes the worktree
 		// directory, which can take a while for large trees (node_modules
--- a/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
+++ b/packages/host-service/src/trpc/router/workspace-cleanup/workspace-cleanup.ts
@@ -526,11 +526,27 @@ async function runDestroyPhases(
 			let stillRegistered = true;
 			let removeError: string | undefined;
 			try {
-				({ stillRegistered, removeError } = await cleanupGitOps.removeWorktree({
-					repoPath: project.repoPath,
-					worktreePath: local.worktreePath,
-					gitEnv: repoGitEnv,
-				}));
+				// The fast native rm only runs when the caller can prove the
+				// path is inside this project's managed worktrees root — a
+				// stale or corrupt `worktreePath` row (e.g. a symlink out of
+				// the root) must never reach the worker's recursive delete
+				// (#6785). Same `isInsideProjectWorktreesRoot` gate the
+				// disk-recheck fallback below uses.
+				const worktreeBaseDir =
+					project.worktreeBaseDir ?? getHostWorktreeBaseDir(ctx);
+				const nativeRm = isInsideProjectWorktreesRoot(
+					local.worktreePath,
+					project.id,
+					worktreeBaseDir,
+				);
+				({ stillRegistered, removeError } = await cleanupGitOps.removeWorktree(
+					{
+						repoPath: project.repoPath,
+						worktreePath: local.worktreePath,
+						gitEnv: repoGitEnv,
+						nativeRm,
+					},
+				));
 			} catch (err) {
 				const message = err instanceof Error ? err.message : String(err);
 				throw new TRPCError({
--- a/packages/host-service/src/workers/tasks/git-remove.test.ts
+++ b/packages/host-service/src/workers/tasks/git-remove.test.ts
@@ -1,0 +1,132 @@
+import { afterEach, expect, test } from "bun:test";
+import { execFileSync } from "node:child_process";
+import {
+	existsSync,
+	mkdirSync,
+	mkdtempSync,
+	readdirSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { type GitTaskEnv, gitWorktreeRemoveTask } from "./git";
+
+const tempDirs: string[] = [];
+
+function git(cwd: string, ...args: string[]) {
+	execFileSync("git", args, { cwd, stdio: "pipe" });
+}
+
+/** A real git repo with a worktree carrying a deep node_modules-shaped tree. */
+function makeRepoWithHeavyWorktree(): { repo: string; worktree: string } {
+	const repo = mkdtempSync(join(tmpdir(), "rm-by-hand-repo-"));
+	tempDirs.push(repo);
+	git(repo, "init", "-q", "-b", "main");
+	git(repo, "config", "user.email", "test@test.dev");
+	git(repo, "config", "user.name", "test");
+	writeFileSync(join(repo, "readme.md"), "root\n");
+	git(repo, "add", ".");
+	git(repo, "commit", "-qm", "init");
+
+	const worktree = mkdtempSync(join(tmpdir(), "rm-by-hand-worktree-"));
+	tempDirs.push(worktree);
+	// A genuinely separate worktree dir, not a symlink.
+	rmSync(worktree, { recursive: true, force: true });
+	git(
+		repo,
+		"-c",
+		"worktree.pruneExpire=never",
+		"worktree",
+		"add",
+		"-b",
+		"feature/heavy",
+		worktree,
+	);
+
+	// node_modules-shaped payload: many nested dirs on disk in the worktree.
+	// They don't need to be committed — the handler's job is to remove the
+	// *directory*, heavy or not. `--force --force` clears uncommitted changes.
+	const nodeModules = join(worktree, "node_modules");
+	mkdirSync(join(nodeModules, "pkg-a", "deep"), { recursive: true });
+	mkdirSync(join(nodeModules, "pkg-b"), { recursive: true });
+	for (let i = 0; i < 50; i++) {
+		writeFileSync(join(nodeModules, "pkg-a", `file-${i}.js`), "// x\n");
+	}
+	writeFileSync(join(worktree, "worktree-only.txt"), "uncommitted\n");
+
+	return { repo, worktree };
+}
+
+afterEach(() => {
+	for (const d of tempDirs.splice(0)) {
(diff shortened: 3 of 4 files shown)


Kitenite:
Ran this end-to-end and I don't think it should land as-is.

The `rm` at `git.ts:278` runs before git has said whether the path is a worktree. Today git is the guard; now a stale or corrupt `worktreePath` row is an unguarded `rm -rf`. Repro'd it in the app: a row pointing (via symlink) outside `~/.superset/worktrees` got deleted with one click and reported success. Main left it alone. Needs the same `isInsideProjectWorktreesRoot` gate the sibling branches in `workspace-cleanup.ts` already use, decided at the call site.

Also the gain is ~1.2–1.7x on a warm cache (101k files: git 5.1–5.7 s, node rm 3.0–4.9 s), not "several times", and the 120 s budget / silent wait from #6887 are unchanged. #6785 already covers the git-error capture and guarded fallback for #6730, so I'd merge that first and fix #6887 with a progress heartbeat.

Kitenite:
Update: #6785 is approved and covers the #6730 side (disk re-check after git's verdict, git's error kept, root-guarded fallback `rm`). Once it merges, please rebase this on top of it. The only textual conflict is the comment block above the worktree-remove phase in `git.ts`.

Two things to carry over in the rebase:

1. Put the native `rm` behind `isInsideProjectWorktreesRoot` (from `workspace-creation/shared/worktree-paths.ts`), decided at the call site in `workspace-cleanup.ts` the same way the sibling branches do. That closes the unguarded delete I reproduced above.
2. Keep #6785's `removeError` capture rather than the `.catch(() => {})`.

With the guard in place I'm fine taking the speedup even at the 1.2 to 1.7x I measured. The 120 s budget and the silent wait from #6887 can be a separate change.

andyst-dev:
Rebased on top of #6785 (now merged) and rebuilt per your review. Head `98782ec72`, mergeable.

What changed, addressing both points:

1. **Native `rm` behind `isInsideProjectWorktreesRoot`, decided at the call site.** The worker (`gitWorktreeRemoveTask`) now takes an opt-in `nativeRm` input; the destroy saga computes `isInsideProjectWorktreesRoot(local.worktreePath, project.id, worktreeBaseDir)` right before the `removeWorktree` call and passes `nativeRm: true` only when the path is provably inside the managed root. A stale/corrupt `worktreePath` row (e.g. the symlink-out-of-root repro) never reaches the recursive delete — it falls back to git-only removal, and the existing guarded `isMissingPath` disk-recheck + warning path still owns the outside-root case. The worker has no project-root knowledge and can't set this flag itself.
2. **#6785's `removeError` capture kept** — the `git worktree remove` error is still captured into `removeError` and surfaced by the caller; no `.catch(() => {})` anywhere.

Validation (real handler, real git repo + worktree carrying an uncommitted `node_modules`-shaped tree):
- `git-remove.test.ts` 3/3: native-rm path runs (asserted via the `delete-files` phase, so it discriminates the new code from the git-only fallback) and removes the dir; git-only fallback still unregisters; an already-missing path is a no-op.
- `test/workspace-cleanup.test.ts` 31/31 green (destroy saga, incl. the outside-root warning and still-registered error paths).
- `tsc -p tsconfig.json --noEmit`: no errors in the changed files (the two `@mastra/core/*` resolution errors are pre-existing and unrelated).

The 120s budget / silent-wait from #6887 are intentionally left for a separate change, as you noted.

PR #7101, opened 2026-09-02
fix(desktop): keep sidebar usable after project close

cu9hue:
## What & why

Fixed the app getting stuck after closing the active project. The sidebar is freezing if a user closes an active project.

It now removes that project from the sidebar cache and clears the stale workspace selection before navigating away, so the sidebar stays usable.

## How I tested it

I've replicated the bug, then verified it goes away after the fix.

## Checklist

---
## Summary by cubic
Closing the active project no longer leaves the desktop sidebar frozen. The sidebar now removes all of the closed project's workspace groups from its cache and clears a stale workspace selection before navigating to the workspace view.

**Behavior changes**
- If closing fails, the previous sidebar cache and workspace selection are restored and the user is sent back to the previous workspace.
- Adds test coverage for removing all workspace groups for a closed project.

<sup>Written for commit 64c255ce3c6a1be35cb0c2d882b6a41e9cbd72b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the close-project experience with immediate sidebar updates and navigation to the workspace view.
  * Added recovery handling to restore the previous project and workspace if closing fails.
  * Prevented closed projects from remaining visible in grouped workspace listings.
  * Improved handling of workspace-loading failures during project closure.
  * Ensured the last viewed workspace is cleared only when it belongs to the closed project.

### DIFF
--- a/apps/desktop/src/renderer/lib/persisted-keys/persisted-key-registry.test-data.ts
+++ b/apps/desktop/src/renderer/lib/persisted-keys/persisted-key-registry.test-data.ts
@@ -186,6 +186,10 @@ export const PERSISTED_KEY_REGISTRY: ReadonlyArray<
 		["lastViewedWorkspaceId"],
 	],
 	[
+		"src/renderer/screens/main/components/WorkspaceSidebar/ProjectSection/ProjectHeader.tsx",
+		["lastViewedWorkspaceId"],
+	],
+	[
 		"src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/TerminalPane/richInputOpenStore.ts",
 		["superset.terminalRichInputOpen"],
 	],
--- a/apps/desktop/src/renderer/react-query/workspaces/utils/workspace-removal.test.ts
+++ b/apps/desktop/src/renderer/react-query/workspaces/utils/workspace-removal.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "bun:test";
 import {
 	getWorkspaceFocusTargetAfterRemoval,
+	removeProjectFromGroups,
 	removeWorkspaceFromGroups,
 } from "./workspace-removal";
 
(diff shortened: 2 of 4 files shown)


Kitenite:
Thanks for digging into this. I reproduced the journey on main and the freeze is already fixed there by #7108, which this branch predates.

What I ran: a dev build, v1 sidebar, right-click the project header → Close Project → confirm, driven with real mouse events over CDP. Same script against main and against a checkout with this branch's `@radix-ui/react-context-menu` version.

| After confirming the dialog | main (context-menu 2.2.16) | this branch's base (2.3.7) |
|---|---|---|
| `document.body.style.pointerEvents` | `""` | `"none"` |
| dialogs / overlays left in the DOM | 0 | 0 |
| real click on another workspace row | navigates | nothing happens |

That second column is the freeze you hit. 2.3.7 pulls in a second copy of `@radix-ui/react-dismissable-layer`: the context menu's copy locks the body, the alert dialog's copy records `none` as the value to restore, and restores it on close. Nothing in this diff touches body pointer-events, so I can't see how it would clear that. If you rebase onto main the freeze should be gone without these changes.

What's left in the PR is an optimistic close that navigates away before the terminals are killed. It's a reasonable idea, but `ProjectHeader` is on the v1 sidebar, which we're winding down, so I'd rather not expand it. For the record, two things would also need fixing before it could land: the rethrow in `onMutate` turns a failed `workspaces.get` lookup into a failed close where main proceeds, and biome format fails on the `.catch` block (run `bunx biome check` from `apps/desktop`).

One thing I did notice on main: after the close, the route stays on the deleted workspace id and shows the empty state instead of jumping to the next workspace. Minor, and separate from the freeze.

Closing this one. 1.25.2 carries #7108. If you still see the sidebar lock up on a build that includes it, please reopen with the steps and I'll take another look.

PR #7116, opened 2026-09-02
feat(mobile): move the workspace PR link beside the quick keys

saddlepaddle:
## What & why

The pull request chip sat at the head of the session tab strip. It moves to its own square ahead of the quick-key bar: a peer of the bar rather than a key in it, sharing its fill, corner and height so the row reads as one strip with a break in it.

Three decisions in the diff:

- **Inserted, not reserved.** A reserved slot was built first and read as a dead control at the head of the row on every workspace without a pull request. So the bar has the whole row until a pull request exists, and the keys slide over once when it arrives, inside the same growth transaction everything else in the cluster uses.
- **The bar fills its row on every phone.** Nine keys are narrower than a Pro Max and wider than an iPhone 17 once the chip has taken its share, so hugging the keys left a stub of empty row after the bar and filling the row left empty panel after the last key. A small custom `Layout` (`QuickKeyRow`) splits spare width evenly between the keys, hairlines excluded, and falls back to natural width plus scrolling when there is none. Flexible frames do not work here: inside a horizontal scroll view `maxWidth: .infinity` means infinite, not the viewport.
- **Renamed end to end** since it no longer belongs to the strip: `ComposerSessionAction` → `ComposerQuickKeysAction` (own file), prop `quickKeysAction`, event `onQuickKeysActionPress`. The strip is back to tabs plus its trailing controls.

Routing, tints, accessibility labels and the bundled marks are unchanged. Desktop is untouched.

## How I tested it

Screenshots from an iPhone 17 Pro Max simulator (iOS 26) against prod, with and without a pull request, are on the Superset page: https://app.superset.sh/page/pr-button-left-of-the-key-bar-7f8jo2

- [x] Open PR: emerald mark in a 36pt square ahead of the bar, bar narrows and scrolls beside it with the edge fade
- [x] No PR: no chip, bar spans the row edge to edge with the keys spread evenly and `^C` fully visible
- [x] Copy Selection (select mode) becomes one full-width key as a side effect of the fill
- [x] `xcodebuild -scheme Composer` and a full `-scheme Superset` Debug build, no errors
- [x] `biome check`, mobile `tsc --noEmit`, `bun test` (66 pass)

Not exercised live: the slide when a pull request first appears. It is the same mechanism a new session tab uses (view inserted into the row under the module's `growth` transaction), read off the code rather than recorded.

## Checklist

---
## Summary by cubic
Moves the workspace pull request link from the front of the session tabs to its own square beside the quick-key bar. The link appears only when a pull request exists; when it appears, the keys animate over and narrow, while the bar fills the available phone width and scrolls when needed.

**Changes**

- Distributes spare width evenly across key buttons while keeping dividers at their natural size.
- Renames the native and React Native APIs to `quickKeysAction` and `onQuickKeysActionPress`.
- Keeps the control fixed while the quick keys scroll and preserves its icon, tint, accessibility label, and routing.
- Leaves desktop behavior unchanged.

<sup>Written for commit 0205ae3f0057ff912791fd713f5fe6e05fa2e59c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7116?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an optional action control beside the composer’s quick keys.
  - Supports custom icons, accessibility labels, and pull-request status colors.
  - Quick-key buttons now expand to use available horizontal space.

- **Changes**
  - Moved the pull-request action from the session-tab strip to the quick-key row.
  - The quick-key row remains visible when an action is available, even without quick-key items.
  - Updated related naming and interaction labels to reflect the new placement.

### DIFF
--- a/apps/mobile/modules/composer/ios/ComposerModel.swift
+++ b/apps/mobile/modules/composer/ios/ComposerModel.swift
@@ -60,9 +60,9 @@ final class ComposerModel {
   /// has no catalog. See `ComposerSessionTabLabels`.
   var sessionTabLabels = ComposerSessionTabLabels()
 
-  /// The strip's leading control. Nil on every surface with nothing to link
-  /// to, which is all of them but the workspace terminal.
-  var sessionAction: ComposerSessionAction?
+  /// The control beside the quick keys. Nil on every surface with nothing to
+  /// link to, which is all of them but the workspace terminal.
+  var quickKeysAction: ComposerQuickKeysAction?
 
   /// What the active agent can run behind `/` (or `$`). Empty hides the
   /// suggestion panel entirely — a plain shell, an agent without discovery,
@@ -123,8 +123,9 @@ final class ComposerModel {
   @ObservationIgnored var onSessionTabCopyId: ((String) -> Void)?
   @ObservationIgnored var onNewSessionPress: (() -> Void)?
   @ObservationIgnored var onAllSessionsPress: (() -> Void)?
-  /// The leading control was pressed. Where it goes is React Native's to know.
-  @ObservationIgnored var onSessionActionPress: (() -> Void)?
+  /// The control beside the quick keys was pressed. Where it goes is React
+  /// Native's to know.
+  @ObservationIgnored var onQuickKeysActionPress: (() -> Void)?
   /// Files and images pasted into the field, already written to disk. The tray
   /// lives in React Native, so the composer hands over URIs and lets it add
   /// them the same way the pickers do.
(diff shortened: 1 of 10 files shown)


saddlepaddle:
Re the CodeRabbit nitpick (non-remounting regression test for `quickKeysAction` → null → action): skipping. The contract lives in the native view (`ComposerAnchorView`'s prop setter and `ComposerModel`), which has no test host here — `bun test` never instantiates Expo views, and the repo has no XCTest target for the composer module. The clearing path was verified by screenshotting the transition on the simulator, the same way #7038 caught the original undefined-vs-null bug.

PR #7121, opened 2026-09-03
build(i18n): compile message catalogs at install and build instead of committing them

Kitenite:
## Why

`packages/i18n/locales/*/messages.ts` was committed as one ~600 KB line per locale. Git cannot three-way merge a one-line file, so any two PRs that each added a string conflicted on all seventeen of them. Both PRs marked CONFLICTING at the time (#7118, #7102) conflicted on exactly those 17 files and nothing else; their `.po` sources auto-merge cleanly. A three-string PR (#7113) touched 34 locale files and 316 lines.

## What

- `locales/*/messages.ts` is gitignored and removed from the tree. Only the `.po` catalogs are committed.
- `lingui compile --strict` becomes the i18n package's `build` script (0.3s). The `check` and `translate` scripts call it.
- **Every consumer compiles the catalogs itself**, so no production path depends on an install hook:
  - desktop: `predev`, `pretypecheck`, and `precompile:app`. The release and canary workflow (`build-desktop.yml`) installs with `--ignore-scripts` and calls `bun run compile:app` directly, so this hook is what makes releases work.
  - web, marketing, docs, admin: `prebuild`. `vercel build` runs `bun run build` (confirmed in the preview deploy run log), so the hook runs for both preview and production, which are both prebuilt in Actions.
  - mobile: `predev`, `preios`, `preandroid`, and `eas-build-post-install`.
- Root postinstall also runs the build (before its `CI` early exit), so a fresh clone has the files. The same run log shows `vercel build` re-runs `bun install` with scripts.
- Turbo: apps' `build`/`typecheck`/`test` already depend on `^build`; a package-level `packages/i18n/turbo.json` makes the i18n package's own `typecheck`/`test` depend on its `build` and declares the outputs so the cache restores them; root `dev` depends on `@superset/i18n#build`; root `build:types` depends on `^build` (host-service's `tsc` walks into the i18n package through `@superset/shared`, and CI's Typecheck job raced it against the catalog compile once).
- The CLI compiles the catalogs in `prebuild` and `prebuild:dist`: `build-cli.yml` installs Linux deps with `--ignore-scripts` and calls `build:dist` directly.
- README and AGENTS.md updated.

## Rebasing open PRs that touched catalogs

A branch that modified `messages.ts` hits a one-time modify/delete conflict on those 17 files. Resolve with:

```
git rm packages/i18n/locales/*/messages.ts
```

The `.po` changes merge on their own. This branch was itself rebased over #7118 that way.

## Verified

Every build below ran with the 17 compiled files deleted first, so the hook under test is what produced them. Test suites were compared against a clean checkout of `main`.

| Surface | Command | Result |
|---|---|---|
| all packages | `bun run typecheck --force` | pass, catalogs regenerated by the turbo dependency |
| all packages | `bun run lint` | pass |
| i18n | `bun run --cwd packages/i18n check` | pass |
| desktop | `bun run --cwd apps/desktop build` (full electron-builder package) | pass in 7m; `precompile:app` ran; DMG and .app produced; asar contains the catalog chunks |
| web | `dotenv -e ../../.env -- bun run build` | pass (first attempt stalled on a Google Fonts fetch inside Turbopack, unrelated; rerun passed) |
| marketing | same | pass |
| docs | `bun run build` | pass |
| admin | `dotenv -e ../../.env -- bun run build` | pass |
| mobile | `bun run eas-build-post-install && expo export --platform ios` | pass; the Hermes bundle contains the translated strings |
| tests, main baseline | `bun turbo run test --force` in a clean checkout of main | 22/22 tasks pass |
| tests, this branch | `bun turbo run test --force` in a clean checkout at this commit | desktop 3311 pass / 0 fail; i18n 119 pass |

**Runtime, not just build.** Served the built marketing site with `next start`: `/ja/pricing` renders the same Japanese text density as production (105 CJK characters, identical page length). Started the desktop dev stack (`bun dev:desktop`, where turbo restored the catalogs from cache before the renderer came up) and drove it over CDP with real mouse input: Settings gear, Appearance, Language select, 日本語. The document language flipped to `ja`, the page re-rendered fully in Japanese (設定 / 外観 / 言語, screenshot captured), and choosing 自動(システム) restored English. No exceptions, no `window.onerror`, no `unhandledrejection` across the whole journey. The Expo export of the mobile bundle contains the translated strings (UTF-16 in the Hermes bytecode).

One detour worth knowing: on the first desktop run the app went blank after the switch back to Auto. That was this worktree's `.env` missing the R2/Cloudflare variables the API requires, so every tRPC call was an HTML 500 and the collections provider could not remount (a plain reload was blank in English too). With the placeholders added the full journey passes. Pre-existing dev-env issue, unrelated to catalogs.

Without the `.env`, the Next builds compile successfully and then fail at page-data collection on missing env vars, the same as on main. In my own Superset worktree the desktop task under turbo showed 35 git-fixture failures (`git rev-parse HEAD` returning empty in a temp repo); the same files pass when run directly there and the whole task passes through turbo at this exact commit in a clean checkout, so that is the local environment, not this change.

https://claude.ai/code/session_014AGKZoTGQeb5vnpnRhnmL8

## Summary by CodeRabbit

* **Improvements**
  * Internationalization catalogs are generated automatically during installation, development, builds, and distribution workflows.
  * Applications and tools consistently build the latest translation assets before use.
  * Generated translation files are no longer committed, reducing merge conflicts and streamlining localization sources.

* **Documentation**
  * Updated contributor and package documentation with the revised translation catalog workflow and build commands.


AviPeltz:
dont merge yet pls

PR #7153, opened 2026-09-04
fix(desktop): honour "Auto-run on workspace creation" terminal scripts on v2

### LINKED ISSUE #7129: "Auto-run on workspace creation" for terminal scripts does nothing on v2
### Where did this happen?

Desktop app (macOS)

### Version

1.25.1

### What happened?

I have a terminal script with "Auto-run on workspace creation" enabled. It runs on v1. Since my machine flipped to v2 it never runs — I create a workspace and the script's terminal simply never appears. The toggle is still there in Settings and still shows as on.

As far as I can tell the trigger has no reader on v2. The only code that acts on `applyOnWorkspaceCreated` is v1's init procedure — `apps/desktop/src/lib/trpc/routers/workspaces/procedures/init.ts:211-214`:
```
const defaultPresets = getPresetsForTrigger("applyOnWorkspaceCreated", project.id);
```
and everything that consumes it (`WorkspaceInitEffects.tsx`, the `workspace-init` store, `useCreateWorkspace.ts`) is under `screens/main` / the Electron tRPC router, i.e. the v1 surface.

On v2 the port looks half-finished. `useV2PresetExecution.ts:85-87` declares the selector for both triggers:
```
function selectAutoApplyPresets(presets, field: "applyOnWorkspaceCreated" | "applyOnNewTab")
```
but line 143 is its only call site, and it passes "applyOnNewTab". So new-tab auto-run works and workspace-creation auto-run has a type signature and nothing else. The create payload agrees: `useSubmitWorkspace` sends `agents`, `namingPrompt` and `command` but no presets, and `appendLaunchesToPaneLayout` builds tabs only from the terminals and agents the server returns.

The setting survives the v1→v2 migration (`renderer/lib/v1-migration/presets.ts` carries the field over) and the editor still renders the toggle (`PresetEditorDialog.tsx:725-745`), so it reads as a working feature that silently does nothing. Either honour it on v2 or hide the toggle there.
However it would be nice to keep this very helpful feature.

### Steps to reproduce

1. Settings → Terminal → a terminal script → enable "Auto-run on workspace creation". Leave "Auto-run on new tab" off.
2. Create a workspace, agent set to "No agent" (so the only terminal that could appear is the script's).
3. The workspace opens without the script's terminal; the script never runs.
4. For contrast, enable "Auto-run on new tab" on the same script and open a new tab — that trigger fires correctly.

Nothing in the console — there is no failing call, the trigger is simply never consulted.

Workaround for anyone hitting this: put the command in the project's setup commands instead. `runSetup` defaults on in the v2 create input, so the host-service spawns a setup terminal for it on every create. Per-workspace, `superset workspaces create --command "…"` also works.

Kitenite:
**Links**
- Issue: https://github.com/superset-sh/superset/issues/7129

## Summary
- Terminal scripts tagged "Auto-run on workspace creation" now run on v2. The toggle persisted and displayed as on, but v2 only ever wired the sibling "Auto-run on new tab" trigger, so nothing consumed the flag.
- The v1 trigger selector moves to `shared/preset-trigger-selection.ts` so both surfaces resolve targeted-over-global scripts the same way; the v2 hook's private copy is removed.

## Why / Context
Since the v1→v2 flip, `applyOnWorkspaceCreated` had no reader on v2: every consumer lived under `screens/main` and the Electron routers. The v1→v2 migration copies the field and the editor still renders the toggle, so users saw a working-looking setting that silently did nothing (#7129).

## How It Works
Scripts live in `v2TerminalPresets`, a renderer localStorage collection the host service cannot read, so the trigger stays renderer-side:
- `queueWorkspaceCreationPresets` runs in the create `.then` once the host result lands. Gated on `projectId !== null && alreadyExists === false` (the same gate the star-nag counter uses), it writes the matching script ids to a new `pendingCreationPresetIds` field on the workspace's `v2WorkspaceLocalState` row. Reopened branches and project-less sessions never queue.
- `useRunWorkspaceCreationPresets` on the v2 workspace page drains that queue once on `isLayoutReady`. It clears the queue before running (one-shot even across a mid-run remount), drops scripts deleted since the create or resolving to no commands, and calls the existing `executePreset` serially. Split-pane, sequential, and new-tab-per-command keep their normal semantics; a split-pane script joins the agent tab the create seeded.
- The workspace page suspends behind `WorkspaceCreatingState` until the create transaction settles, so the queue is present before the page's effects run; the hook also live-queries the row in case that ordering ever changes.

Persisted-state note: the new field is a short array of ids on an existing row, cleared on first open, so it adds no unbounded growth to the localStorage blob.

## Manual QA Checklist
Driven end to end over CDP in the dev app with real mouse input on the New Workspace form, a throwaway repo imported as a project, and terminal buffers read back through the terminal registry:
- [x] Baseline on main: create with "No agent" and a creation-tagged script → workspace opens empty (bug reproduced)
- [x] With fix: same steps → script tab opens and the command runs
- [x] Two creation scripts (one with two commands) → tabs open in presets-bar order, one tab per command for the multi-command script
- [x] Switch away and back → nothing re-runs, queue stays empty
- [x] Split-pane ("Current tab") script + Claude agent with a prompt → script pane joins the agent tab
- [x] Sequential script with two commands → one terminal, commands joined with `&&`
- [x] Global + demo-targeted + other-project-targeted scripts → only the demo-targeted one runs
- [x] "Auto-run on new tab" only → nothing on create, fires from `+ → Terminal` (covers the moved selector)
- [x] "Open workspace" on an existing branch → same workspace, no re-run (`alreadyExists` gate)
- [x] No console errors or warnings during the runs

## Testing
- `bun run typecheck` (apps/desktop)
- Biome on changed files
- `bun test` for `shared/preset-trigger-selection`, `queueWorkspaceCreationPresets`, `useRunWorkspaceCreationPresets`, and the local-state schema heal (57 pass; the ordering test was flipped once to confirm it fails)

## Known Limitations
- CLI and automation creates do not trigger these scripts: no renderer is present at create time. Honouring them there would mean shipping resolved script commands in the create payload, which flattens execution modes. Project-less sessions are also excluded, matching v1.
- v1's unused `settings.getWorkspaceCreationPresets` procedure is left as is; it belongs to the v1 surface.

https://claude.ai/code/session_01YCcGF74n8GEiAv5kSv77Zr

---
## Summary by cubic
Fixes #7129 by making the "Auto-run on workspace creation" terminal preset toggle actually run scripts on v2 workspaces. Previously the toggle was persisted and shown as enabled, but v2 only wired "Auto-run on new tab", so the setting silently did nothing.

**Notes**
- Scripts run once when a newly created workspace first opens; reopening the workspace or re-rendering mid-run won't re-run them.
- Queued scripts run serially in presets-bar order, one script throwing doesn't skip the rest, and each runs in a fresh tab so a script never lands in the agent's terminal.
- Workspaces that adopt an existing worktree or were created with "Run setup" off don't queue these scripts; neither do reopened branches, project-less sessions, or CLI/automation creates.
- Moved v1's trigger selector to `shared/preset-trigger-selection.ts` so v1 and v2 resolve targeted-over-global presets the same way.

<sup>Written for commit 701c4b7ed346fbe87b357f80f17d22e1154501c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Workspace creation presets can now be automatically queued and run when a new workspace opens.
  * Presets are selected for the relevant project, ordered by tab position, and global presets are used as a fallback.
  * Presets run once after the workspace layout is ready.

* **Bug Fixes**
  * Deleted or empty presets are skipped without interrupting execution.
  * A failed preset no longer prevents subsequent queued presets from running.
  * Presets now apply only when their configured trigger matches.


Kitenite:
Responses to the bot reviews:

- **Greptile P1 (failure discards queue):** `executePreset` catches and toasts its own errors, so the loop never aborted in practice, but the point stands for an unexpected throw. Pushed a per-preset guard so one failure cannot skip the scripts after it. Retrying on the next open is deliberately not done: a script whose command ran but whose pane creation failed would run twice.
- **CodeRabbit (agent-config readiness):** `resolvePresetLaunchCommands` falls back to the stored command snapshot when the linked agent has not loaded. A preset is only dropped when the agent list is unloaded *and* its snapshot is empty, and imports always write a snapshot. Same behaviour as the "new tab" trigger and the presets bar; leaving as is.
- **CodeRabbit (move queue to SQLite):** the field is a handful of ids on the per-workspace row that already exists for the pane layout, and it is cleared on first open, so it adds no unbounded growth. Presets themselves live in localStorage, so a host-side queue would still need the renderer to fill it. Leaving as is.

Kitenite:
Follow-up from a deeper review pass (pushed as the third commit):

- **Sequential/split-pane creation scripts targeted the agent tab.** The create seeds the agent tab as active, so a sequential script was typed into the agent's PTY as a prompt. Creation scripts now always open in a fresh tab. Re-verified over CDP with Claude selected: the script gets its own tab and the agent buffer never sees the command.
- **"Import worktrees" queued creation scripts.** Adopting an existing worktree reports `alreadyExists: false`; creates carrying a `worktreePath` or `runSetup: false` no longer queue.
- **Queue drain guards against a deleted row** (TanStack DB `update` throws on a missing key).

Considered and left as follow-ups rather than blockers: resolving the tagged set at open time instead of at create time (v1 semantics; needs a broader refactor), duplicate drains when two windows open the same workspace mid-create, the pre-existing auto-adopt race for multi-session scripts, and the lost-settle-event recovery path, which degrades to the old "no script" behaviour.

PR #7190, opened 2026-09-05
fix(desktop): reap the per-workspace dev app profiles Superset leaks

saddlepaddle:
## Why

`apps/desktop/src/main/index.ts` renames the app in dev so several worktrees are distinguishable:

```ts
if (IS_DEV) {
  const workspaceName = resolveDevWorkspaceName();
  if (workspaceName) app.setName(`Superset (${workspaceName})`);
}
```

Renaming an Electron app also moves `app.getPath("userData")`, so **every workspace the dev app is launched from mints its own Chromium profile directory** under `~/Library/Application Support`. Nothing ever removed them.

Measured on Satya's Mac on 2026-09-04: **347 `Superset (<workspace>)` directories totalling 89.3 GB**, of which **339 (83.7 GB) had gone untouched for more than three days**; the oldest dated to February, and most pointed at worktrees that no longer exist. Individual profiles run from a few hundred MB to 1.8 GB (Chromium `Cache`, `Code Cache`, `GPUCache`, `Local Storage`, `IndexedDB`, `Service Worker`).

**This is not a live incident.** Those 339 were deleted by hand; `~/Library/Application Support` now holds 8 `Superset*` directories totalling 5.7 GB — the plain `Superset` daily driver at 2.9 GB plus seven profiles all touched since 2026-09-01. This PR is purely about stopping the leak recurring.

## What

**1. Workspace delete (the primary ask).** The workspace's profile is removed at the **end of `runDestroyPhases` (step 5)**, past every point that can roll the delete back. It swallows its own failures — reclaiming disk must never fail a delete.

> Placing this in `runTeardown` (step 2) was the brief's suggestion and my first implementation, and it was wrong: every throw between step 2 and the end un-archives the workspace and returns it to the user, so a *failed* delete handed back a live workspace with its logins and browser storage already wiped. Both bot reviewers caught it independently. Step 5 also covers the two cases step 2 missed — the `skipTeardown` force-retry path and project-less session workspaces.

**2. Dev-only startup sweep (belt and braces).** `apps/desktop/src/main/dev-app-profile-sweep.ts`, modelled on `network-logger-sweep.ts` — same single-purpose module, same off-the-hot-path async fs callbacks, same best-effort logging. It removes a sibling profile only when it is not the running app's, not an installed build's, untouched for **14 days**, and **not held by a live Chromium process**. A packaged build returns immediately and sweeps nothing.

**3. Shared name derivation.** `@superset/shared/dev-app-profile` is the single definition of the directory name, so the desktop (which mints profiles) and host-service (which reaps them) cannot drift. `app.setName` uses it too, and refuses a derived name that is not a single path segment — Electron would otherwise nest `userData` in a directory neither the sweep nor teardown could ever reap.

Only ~180 of the added lines are non-test, non-comment code; the rest is tests and the reasoning below.

## Guards

The plain `Superset`, `Superset Dev` and `Superset Canary` profiles belong to installed apps — one is a daily driver at 2.9 GB — and are refused by an explicit deny set on top of the `Superset (` prefix that cannot match them anyway. A workspace name carrying a path separator or `..` is refused so no derived path can escape the profiles directory.

**Age alone does not prove a profile is idle.** I measured directory mtime against the newest file inside every profile on the machine and found no lag at all (max 0.00 days), but that does not prove an app running *continuously* past the threshold keeps its root fresh — and the cost of being wrong is deleting a live profile. Chromium's `SingletonLock` is a symlink to `<hostname>-<pid>`, so the owning process is probed directly and vetoes removal. Format verified against the real profiles here (`Satyas-MacBook-Pro.local-78859` — the hostname contains hyphens, so the pid comes from the last one). It fails closed: a lock that cannot be read or parsed counts as in use, which at worst strands one directory that a real delete still reaps.

**Why the directory test is a prefix test rather than a `Superset (…)` shape test.** Workspace names have carried newlines and stray characters, so the machine held directories like `Superset (satyapatel@satyas-mbp:~` and `Superset (saddlepaddle` — no closing paren, and just as stranded. A shape test would have skipped them forever. The bare `Superset (` prefix still cannot match an installed build, so nothing is given up. Both forms are pinned in tests.

**Why the 14-day threshold and the stranded-or-not predicate live in the shared module** rather than at the top of the sweep module, as the brief suggested: testing them from `apps/desktop` needs `mock.module("electron", …)`, which in Bun leaks across the whole run — a partial electron mock in a new test file broke 68 unrelated desktop tests. Keeping the decision electron-free avoids that; the sweep module stays pure IO.

**What deleting a profile costs.** Only Chromium state — caches, `Local Storage`, `Service Worker`, `Preferences`. The auth token, `local.db`, `app-state.json` and window geometry live in `~/.superset/`, so a reaped workspace re-syncs rather than losing anything.

## Known gap

`removeDevAppProfile`'s `appDataDir` defaults to the real profiles directory, so a future test that drives a delete with a plausible workspace name would reap a real profile. I instrumented the function and re-ran the whole host-service suite: all six calls it makes land in `/var/folders/.../host-service-appdata-*` temp directories, none on the real path. The parameter carries a warning.

## Verification

- `packages/shared`: 951 pass / 0 fail — name derivation, the installed-build guard, the path-escape guard, per-platform `appData` resolution including empty variables, and the 14-day boundary (13.99d keeps, 14d reclaims).
- `packages/host-service`: 1569 pass / 0 fail.
- `apps/desktop` main: 686 pass / 0 fail (baseline unchanged).
- End-to-end against a seeded fake Application Support, exercising the real readdir/stat/readlink/rm path: stale and truncated-name profiles reaped, and a dead-pid lock reaped; a 400-day-old profile with a **live** lock kept, an unparseable lock kept, plus fresh, installed (`Superset` / `Superset Canary`), foreign (`Slack`) and the running app's own profile all kept.
- `typecheck` clean in all three packages; `biome check` clean repo-wide; `sherif` clean; `check:i18n` reports no new or missing messages (no user-facing strings — the only additions are `console.warn` diagnostics, matching the network-log sweep).

**Not done:** the manual two-real-dev-apps walkthrough (launch from two worktrees, delete one workspace through the UI, backdate a third profile). I exercised the same decision and IO path against a seeded fixture instead, rather than booting two Electron instances.

https://claude.ai/code/session_01KBGRsxUo1ZYUj4RxUfJDiS

## Summary by CodeRabbit

* **New Features**
  * Development builds use separate browser profiles for each workspace.
  * Stale development profiles are cleaned up automatically while active profiles remain protected.
  * Workspace-specific profiles are removed when a workspace is deleted.

* **Bug Fixes**
  * Improved profile-name validation prevents accidental removal of installed-app or unrelated profiles.
  * Workspace deletion preserves profiles shared with another active workspace.
  * Profiles with active browser locks, or when cleanup cannot safely complete, are no longer removed.

### DIFF
--- a/apps/desktop/src/main/dev-app-profile-sweep.ts
+++ b/apps/desktop/src/main/dev-app-profile-sweep.ts
@@ -1,0 +1,63 @@
+import fs from "node:fs";
+import path from "node:path";
+import {
+	isDevAppProfileDirName,
+	isStrandedDevAppProfile,
+} from "@superset/shared/dev-app-profile";
+import { app } from "electron";
+
+const IS_DEV = process.env.NODE_ENV === "development";
+
+/**
+ * Dev builds rename the app to `Superset (<workspace name>)` so several
+ * worktrees are distinguishable, which moves userData and mints a Chromium
+ * profile per workspace — 89.3GB across 347 directories on one machine, none
+ * of them ever removed.
+ *
+ * Workspace teardown now reaps the profile of a workspace deleted through the
+ * app. This sweep catches everything else: worktrees removed by hand, deletes
+ * that failed midway, and profiles minted by builds predating that step. What
+ * counts as stranded (and how old is old enough) lives in
+ * `@superset/shared/dev-app-profile` alongside the name derivation.
+ */
+export function sweepDevAppProfiles(): void {
+	// Packaged builds never rename themselves, so they have no per-workspace
+	// profiles — and must never go looking for directories to delete.
+	if (!IS_DEV) return;
+
+	const userData = app.getPath("userData");
+	const currentProfile = path.basename(userData);
+	const profilesDir = path.dirname(userData);
+	const now = Date.now();
+
+	fs.readdir(profilesDir, (readError, names) => {
+		if (readError) {
+			console.warn("[main] dev app profile sweep failed:", readError);
+			return;
+		}
+		// Name test first so the stats below touch only our own profiles rather
+		// than every application's data directory.
+		for (const name of names.filter(isDevAppProfileDirName)) {
+			const target = path.join(profilesDir, name);
+			fs.stat(target, (statError, stats) => {
+				if (statError || !stats.isDirectory()) return;
+				const stranded = isStrandedDevAppProfile({
+					name,
+					currentProfile,
+					mtimeMs: stats.mtimeMs,
+					now,
+				});
+				if (!stranded) return;
+				fs.rm(target, { recursive: true, force: true }, (rmError) => {
+					if (rmError) {
+						console.warn(
+							"[main] dev app profile sweep failed:",
+							target,
+							rmError,
+						);
+					}
+				});
+			});
+		}
+	});
+}
--- a/apps/desktop/src/main/dev-app-profile-sweep.ts
+++ b/apps/desktop/src/main/dev-app-profile-sweep.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 import {
 	isDevAppProfileDirName,
+	isProfileLockHeld,
 	isStrandedDevAppProfile,
 } from "@superset/shared/dev-app-profile";
 import { app } from "electron";
@@ -21,7 +22,7 @@ const IS_DEV = process.env.NODE_ENV === "development";
  * `@superset/shared/dev-app-profile` alongside the name derivation.
  *
  * Age alone does not prove a profile is idle, so a live Chromium lock vetoes
- * removal — see isProfileInUse.
+ * removal — see isProfileLockHeld.
  */
 export function sweepDevAppProfiles(): void {
 	// Packaged builds never rename themselves, so they have no per-workspace
@@ -50,7 +51,7 @@ export function sweepDevAppProfiles(): void {
 					mtimeMs: stats.mtimeMs,
 					now,
 				});
-				if (!stranded || isProfileInUse(target)) return;
+				if (!stranded || isProfileLockHeld(target)) return;
 				fs.rm(target, { recursive: true, force: true }, (rmError) => {
 					if (rmError) {
 						console.warn(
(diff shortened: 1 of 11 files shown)


saddlepaddle:
Both reviewers independently flagged the same two P1s and both were right. All four findings fixed in 9aa6b0fc7.

**Profile removal was ordered before the delete could roll back.** Moved from `runTeardown` (step 2) to the end of `runDestroyPhases` (step 5). Every throw between those points un-archives the workspace and hands it back to the user, so a *failed* delete was returning a live workspace whose logins and browser storage had already been wiped. Step 5 is past every rollback point. This also closes two coverage gaps the old placement had and that I had written up as known limitations: the `skipTeardown` force-retry path and project-less session workspaces now reap as well.

**The sweep no longer trusts age alone.** I had measured directory mtime against the newest file inside every profile on the machine and found no lag, but the reviewers are right that this does not prove an app running *continuously* past the threshold keeps its root fresh, and the cost of being wrong is someone's live profile. Chromium's `SingletonLock` is a symlink to `<hostname>-<pid>`, so the owning process is now probed directly and vetoes removal. Verified the format against the real profiles on this machine — `Satyas-MacBook-Pro.local-78859`; the hostname contains hyphens, so the pid is parsed from the last one. It fails closed: a lock that can't be read or parsed counts as in use, which at worst strands a directory the teardown path still reaps on a real delete.

**Empty `APPDATA` / `XDG_CONFIG_HOME` now fall back.** `??` preserved `""`, and `path.join("", dirName)` would have resolved the target relative to the process working directory. Now `||`, with a test for both platforms.

**`app.setName` validates the derived name.** A name carrying a separator would nest `userData` in a directory neither the sweep nor teardown could ever reap. It now keeps the default profile and logs, rather than minting an unreclaimable one.

Re-verified: shared 951 pass, host-service 1569 pass / 0 fail (the `workspace router integration` flake passed this run too), desktop main 686 pass, typecheck clean in all three, biome and sherif clean. The end-to-end sweep fixture now covers the liveness cases directly — a 400-day-old profile with a live lock survives, one with an unparseable lock survives, one with a dead-pid lock is reaped.

The manual two-real-dev-apps walkthrough is still not done, and the `appDataDir` default remains a documented footgun for future tests.

saddlepaddle:
Three more findings from the last round, two fixed in 479d70d41 and one deliberately not.

**Non-decimal pid suffix (cubic P2) — fixed.** `Number()` reads `0x10` as 16 and `1e3` as 1000, so a malformed lock target resolved to an invented pid that could read as dead and defeat the fail-closed contract. Decimal digits only now; every other shape counts as held, with a test over `0x10`, `1e3`, `+7`, `" 7"` and `""`.

**Whitespace-only name difference (cubic P2) — fixed.** Good catch, and it was a genuine inconsistency in my own code: the profile directory is keyed by the *trimmed* name, but the sharing lookup matched on the raw one, so `"foo"` and `" foo"` shared a directory the guard would have missed. Now compared trimmed in JS over the host's non-archived rows — the row count is per host and this path already does far heavier work.

**Create-versus-delete race on a shared profile name (CodeRabbit, Major) — not fixed, deliberately.** The finding is correct: `registerLocalWorkspace` can insert a same-named live row between the sharing check and the `rm`. Serializing profile-name lifecycle across create and destroy is the architectural change CodeRabbit's own "heavy lift" label implies, and it is out of proportion to the exposure here. Hitting it needs a same-named workspace created inside the millisecond window, and the loss is then browser cache for a workspace nobody has opened yet — no tokens, no database, no window state, all of which live in `~/.superset/`. Worth a ticket, not a blocker on a disk-reclamation PR; I would rather leave it visible in review than paper over it with a lock that only narrows the window.

Re-verified after the fixes: shared 958 pass, host-service 1571 pass / 0 fail, desktop main 686 pass, all three typechecks clean, biome and sherif clean repo-wide, and the end-to-end sweep fixture still green on all 10 assertions.

PR #7192, opened 2026-09-05
fix(desktop): keep the sidebar project order manual under every sort

saddlepaddle:
## What

The Projects sort menu sorted the project list as well as the rows inside it. Picking **Last active** therefore reshuffled the projects themselves every time an agent touched a workspace — the landmarks people navigate the sidebar by moved under them.

Sorting now stops at a project's boundary:

- **Projects** are always the manual drag order, in every mode.
- **Workspaces and folders inside a project** sort by the selected mode (Manual / Last active / Date created), unchanged.

I took "always manual" to cover **Date created** too, not just Last active — a project list that reorders itself is the thing that's wrong, and the mode you picked shouldn't decide whether your drag order survives.

## Drag follows

Because the project list is no longer a transformed view of the manual order, **projects stay draggable while a sort is active** — otherwise you'd have to switch to Manual, drag, and switch back to reorder the list the app claims is manual. The single sensor-level drag gate splits in two:

| | blocked by |
|---|---|
| project drags | filter only (it hides projects, so a drop would rewrite `tabOrder` against the view) |
| row drags (workspaces, folders, pinned, sessions) | any non-manual sort, or the filter |

Row drags stay blocked in the unsorted Pinned and Sessions lanes too, since a row dragged out of them can land in a sorted project list. The sensors can no longer be the single gate, so each sortable reads the row gate off the dnd context.

Menu copy follows the behaviour: "Sort projects" → "Sort workspaces", and the hint reads "Drag workspaces to reorder in Manual order". 16 locales translated.

## Testing

- `bun test` in `apps/desktop`: 3489 pass, 1 pre-existing env-dependent failure (`LeaderboardRank`, passes under `env -i`; unrelated to this change).
- Sort unit tests rewritten: project order preserved in both sort modes, with the tie-break and garbage-timestamp coverage moved down to the row level where the comparator still runs. Dead `getProjectActivityTimestamp` removed.
- `typecheck`, `lint`, `check:i18n` clean.
- **Not driven in the app** — the drag split deserves a minute of hands-on: with Last active selected, drag a project (should move and stick) and drag a workspace (should be inert).

https://claude.ai/code/session_01GvK1ZzjBWK3iCMjzY48oRR

---
## Summary by cubic
Fixes the Projects sort menu reordering the project list itself, so picking **Last active** no longer reshuffles projects under you. Sorting now stops at a project's boundary: projects always keep the manual drag order, while workspaces and folders inside sort by the selected mode.

- Projects stay draggable while a sort is active; only the filter blocks project drags.
- Row drags (workspaces, folders, pinned, sessions) remain inert under any non-manual sort or the filter, enforced per sortable instead of at the sensor level.
- Folder headers no longer show a grab cursor when row drags are inert, since their sortable is disabled too.
- Sort menu copy now reads "Sort workspaces" across 16 locales, and the manual-order hint says "Drag workspaces to reorder in Manual order".
- Removed the dead `getProjectActivityTimestamp` helper and moved its test coverage down to the row comparator.
- Hardened the `LeaderboardRank` test to build expected hrefs from `COMPANY.MARKETING_URL` instead of a literal host, so it passes on machines with `NEXT_PUBLIC_MARKETING_URL` set locally.

<sup>Written for commit bbe123b35b4d3cecada0bea4e89cd77eb0ccea88. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Projects now remain in manual drag order while workspace sorting options reorder workspaces within each project.
  - Project and workspace dragging is managed independently, including filtered views and non-manual sorting modes.

- **User Interface**
  - Updated sort labels, tooltips, and guidance to clearly refer to workspaces and show the active sort mode.

- **Localization**
  - Updated workspace sorting and reordering translations across supported languages.

- **Tests**
  - Expanded coverage for project ordering, workspace sorting, and timestamp handling.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/DashboardSidebar.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/DashboardSidebar.tsx
@@ -62,6 +62,7 @@ interface DashboardSidebarProps {
 interface SortableProjectWrapperProps {
 	project: DashboardSidebarProject;
 	isCollapsed: boolean;
+	isDragDisabled: boolean;
 	workspaceShortcutLabels: Map<string, string>;
 	onWorkspaceHover: (workspaceId: string) => void | Promise<void>;
 	onToggleCollapse: (projectId: string) => void;
@@ -70,10 +71,12 @@ interface SortableProjectWrapperProps {
 const SortableProjectWrapper = memo(function SortableProjectWrapper({
 	project,
 	isCollapsed,
+	isDragDisabled,
 	workspaceShortcutLabels,
 	onWorkspaceHover,
 	onToggleCollapse,
 }: SortableProjectWrapperProps) {
+	const { activeType } = useDashboardSidebarDnd();
 	const {
 		attributes,
 		listeners,
@@ -81,8 +84,7 @@ const SortableProjectWrapper = memo(function SortableProjectWrapper({
 		transform,
 		transition,
 		isDragging,
-	} = useSortable({ id: project.id });
-	const { activeType } = useDashboardSidebarDnd();
+	} = useSortable({ id: project.id, disabled: isDragDisabled });
 	const isDraggingProject = activeType === "project";
 
 	// useSortable re-renders this wrapper on every pointer move of any drag in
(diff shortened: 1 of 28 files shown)


saddlepaddle:
Bot review triage:

**Fixed — folder grab cursor (`0da292b32`).** `SortableSectionHeader` passed `isDraggable` unconditionally, so a folder header advertised `cursor-grab` while its sortable was inert. Now gated on the same `isChildDragDisabled` flag. Worth noting this was pre-existing rather than introduced here — row drags were already dead under a non-manual sort, the cursor just didn't say so. Folder headers are the only row with that affordance (`cursor-grab` appears once, in `DashboardSidebarGroupHeader`); workspace rows and project headers never had one.

**Not doing — DnD interaction tests** (raised by both bots). dnd-kit drives everything off measured rects and pointer events; under happy-dom every rect is zero, so a "drag" test would assert against the mocked harness rather than the behaviour, and would keep passing if the real gating broke. The parts that are unit-testable are already covered (sort order), and the gating itself is now three explicit `disabled` props rather than implicit sensor state. This wants the minute of hands-on noted in the PR description instead.

PR #7204, opened 2026-09-05
fix(billing): make failed-payment emails recoverable

saddlepaddle:
Involuntary churn — subscriptions Stripe cancels after it gives up collecting — is now **33% of all cancellations**, up from 10% in June. That's 21 subscriptions and $420 MRR last month, $2,360 MRR (~$28k ARR) cumulative.

Stripe's own recovery emails are **disabled** on the account, so our emails are the only dunning mail these customers get. All three defects below were in that one path.

### The pay link was dead on arrival

Both emails embedded a `billingPortal.sessions.create` URL created at send time. Stripe documents those as [short-lived](https://docs.stripe.com/api/customer_portal/sessions/object), expiring if the customer doesn't visit them — so by the time anyone opened the mail, the button was spent.

Now uses the invoice's hosted page: valid 30 days, and it takes the payment directly instead of routing through a portal. The in-app `usePaymentFailedCard` already did exactly this.

### The retry line was always wrong

`nextRetryDate` was never passed from the handler, so the template's default (`now + 3d`) made it truthy and **"We'll automatically retry the payment in a few days" rendered on every attempt — including the ninth and final one**, when the subscription was about to be canceled.

Now passes Stripe's real `next_payment_attempt`. When it's `null`, the mail becomes a final notice with the actual consequence instead of a false reassurance.

### The cancellation mail told churned users they still had access

`accessEndsAt` comes from `subscription.periodEnd`, which better-auth sets from `current_period_end` — the period Stripe was *trying* to bill. So every payment-failure cancel promised access *"until"* a date **~17 days in the future** (verified across the last 10) while the subscription was already canceled and the org back on free. It also read as a voluntary-cancel confirmation ("Changed your mind?"), wasting the best win-back moment.

Involuntary cancels now get their own copy — no access-until row, accurate explanation, and a resubscribe link that doesn't expire.

### Also

Stopped mailing on all nine retries. Stripe fires `invoice.payment_failed` per attempt (225 events across 58 invoices in the last 30 days, mean 3.9, max 9); now only the first and final go out. Slack notifications are unchanged.

## Verification

- Rendered all four variants (first attempt / final attempt / involuntary cancel / voluntary cancel); links resolve to the hosted invoice and app URL, no `billing.stripe.com` session anywhere.
- Voluntary-cancel copy is byte-identical to before.
- `typecheck` clean on auth, email, trpc, api, web. `lint.sh` clean. `@superset/auth` tests 8/8.
- No other consumers of the renamed props.

## Not in this PR

- **Enable Stripe's recovery emails / hosted recovery surface** — both off today. Worth turning on, but they'd overlap with these emails, so that's a deliberate choice about which channel owns dunning rather than a code change.
- **126 open invoices, $2,927 outstanding**, 110 of which Stripe has abandoned (`auto_advance: false`). A one-off win-back list, not a code fix.
- **The in-app card disappears at cancellation** — it keys off `past_due`, so the moment access is actually lost there's no in-app signal at all.

---
## Summary by cubic
Fixes the failed-payment email path so customers on the verge of involuntary churn get working links and honest copy instead of expired pay links, a false retry promise, and reassurance that access continues after it's already gone.

**Bug Fixes**
- Emails now link to Stripe's hosted invoice page (valid 30 days, pays directly) instead of a short-lived billing portal session created at send time.
- Passes Stripe's real next payment attempt; the final notice states access is lost today instead of claiming an automatic retry is coming.
- Involuntary cancels get dedicated copy with no future "access until" date and resubscribe instructions pointing to Settings → Billing in-app; voluntary-cancel copy is unchanged.
- Payment-failed emails now go out only on the first and final attempts instead of on all nine retries.

<sup>Written for commit e8a267b5811125c780d2c6c0a3c5fef58ae9b49e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7204?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Improvements**
  - Payment failure emails now indicate whether another payment attempt is scheduled or the subscription has ended.
  - Retryable notices include the next retry date.
  - Recipients can use a **Pay now** link to access the hosted invoice.
  - Subscription cancellation emails distinguish payment-failure cancellations from voluntary cancellations and direct recipients to Settings → Billing for billing actions.
  - Cancellation emails no longer include resubscription links or access-expiration details for payment-failure cancellations.

### DIFF
--- a/packages/auth/src/server.ts
+++ b/packages/auth/src/server.ts
@@ -1211,24 +1211,30 @@ export const auth = betterAuth({
 					);
 					const accessEndsAt = subscription.periodEnd ?? new Date();
 
-					const portalSession =
-						await stripeClient.billingPortal.sessions.create({
-							customer: org.stripeCustomerId,
-							return_url: env.NEXT_PUBLIC_WEB_URL,
-						});
+					// periodEnd is the period Stripe was trying to bill for, so on a
+					// collection failure it sits weeks in the future while access has
+					// already stopped. Only a voluntary cancel keeps access until then.
+					const dueToPaymentFailure =
+						(cancellationDetails ?? stripeSubscription.cancellation_details)
+							?.reason === "payment_failed";
 
 					await resend.batch.send(
 						recipients.map((recipient) => ({
 							from: "Superset <noreply@superset.sh>",
 							to: recipient.email,
-							subject: `Your ${subscription.plan} subscription has been cancelled`,
+							subject: dueToPaymentFailure
+								? `Your ${subscription.plan} subscription ended`
+								: `Your ${subscription.plan} subscription has been cancelled`,
 							react: SubscriptionCancelledEmail({
 								recipientName: recipient.name,
 								organizationName: org.name,
 								planName: subscription.plan,
 								accessEndsAt,
-								billingPortalUrl:
-									recipient.role === "owner" ? portalSession.url : undefined,
+								dueToPaymentFailure,
+								resubscribeUrl:
+									recipient.role === "owner"
+										? env.NEXT_PUBLIC_WEB_URL
+										: undefined,
 							}),
 						})),
 					);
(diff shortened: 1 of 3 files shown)


saddlepaddle:
Addressed the P1 in e8a267b58.

Verified the finding: `NEXT_PUBLIC_WEB_URL`'s root renders `apps/web/src/app/(agents)/page.tsx`, so the Resubscribe button did land on the agents page. Resubscribing only exists on the desktop billing page — there's no web route for it — and Stripe's durable portal `login_page` is disabled on the account (`enabled: false`), so there's no working https link I can build from code today.

Rather than ship a button that doesn't do what it says, the cancellation email now names the destination in text (**Settings → Billing**) and the CTA is gone. Payment-failure emails are unaffected — their "Pay now" still points at `hosted_invoice_url`, which works.

Enabling the portal login page would give us a durable link and let the button come back; filing that alongside the recovery-emails toggle rather than doing it in this PR, since both are account-config decisions.

On the P2 (no tests for retry classification): skipping deliberately — this repo doesn't add tests unprompted, and the branch logic is two null checks on fields the webhook payload owns. Happy to add coverage if you'd rather have it.

PR #7207, opened 2026-09-05
fix(desktop): stop the sidebar holding a live git watch on every workspace row

Kitenite:
## Problem

GitWatcher watches a workspace only while a client holds interest (#6848), but `DashboardSidebarWorkspaceStatusProvider` declared interest for every row it rendered, main workspaces included. Its own comment flagged this as a follow-up. For a heavy user that keeps most of their non-archived workspaces under a recursive `.git` watch plus git subprocess fan-out on every change, for as long as the dashboard is open. The reporter behind this had 39 projects and 9 worktrees, all 48 watched, while working in one.

## Change

- The provider keeps its agent and terminal lifecycle listeners (renderer-side, no host cost) and drops the `git:changed` listener and the `watchGit`/`unwatchGit` loop for sidebar rows.
- Only the active row renders diff stats, and `useDiffStats` already holds its own git watch for it, so live updates for the active row are unchanged.
- An inactive row's cached diff count (staleTime Infinity) is invalidated when the row becomes active, so changes that happened while it was unwatched still show. A row with nothing cached is fetching for the first time anyway and is left alone.

Net effect for the reporter: 48 watched repos down to the one that's open.

## Testing

- `bun run typecheck` in apps/desktop
- `bun test src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar` (173 pass)
- biome clean

https://claude.ai/code/session_01UWyngsh3WSdEM6kmf1P8s5

## Summary by CodeRabbit

* **Bug Fixes**
  * Sidebar workspace diff counts now refresh when a workspace becomes active, including changes made while it was inactive.
  * Improved lifecycle handling for workspace status updates.

---
## Summary by cubic
Stops the dashboard sidebar from holding a live git watch on every workspace row. The provider previously declared watch interest for each rendered row, which kept most of a heavy user's non-archived workspaces watched while the dashboard was open (48 repos in the reported case). Now only the active row's diff stats are live, via `useDiffStats`'s existing subscription; inactive rows' cached diff counts are invalidated when they become active, so changes made while unwatched still show. Rows with nothing cached are left alone, and the renderer-side lifecycle listeners are unchanged.

<sup>Written for commit b519912eb090850444df1a978b803f129f34b910. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/providers/DashboardSidebarWorkspaceStatusProvider/DashboardSidebarWorkspaceStatusProvider.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/components/DashboardSidebar/providers/DashboardSidebarWorkspaceStatusProvider/DashboardSidebarWorkspaceStatusProvider.tsx
@@ -171,10 +171,6 @@ export function DashboardSidebarWorkspaceStatusProvider({
 	const [store] = useState(() => new SidebarWorkspaceStatusStore());
 	const queryClient = useQueryClient();
 	const { cache: hostWorkspacesCache } = useHostWorkspaces();
-	// workspaceId -> hostUrl for whatever's currently git-watched, so the
-	// subscription effect below can diff instead of unwatch-then-rewatch
-	// everything on every targets change.
-	const gitWatchedRef = useRef<Map<string, string>>(new Map());
 
 	const computedTargets = useMemo<WorkspaceStatusTarget[]>(
 		() =>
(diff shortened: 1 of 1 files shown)


Kitenite:
CDP verification against the running dev app from this worktree (renderer on Vite 4245, CDP 9555, real data: 44 non-archived workspaces in host.db). Watched set measured by hooking the renderer's WebSocket `send`, forcing the host event bus to reconnect, and counting the `git:watch` frames it re-sends.

Before (main @ 78a8bbdd6c):
- no workspace open: 44 watched
- one workspace open: 44 watched

After (this branch):
- no workspace open: 0 watched
- one workspace open: 1 watched, the active one; switching sends exactly `git:unwatch <old>` then `git:watch <new>`

Journeys, all with real CDP mouse clicks on sidebar rows:
- Active workspace live update: reverted a file on disk while the workspace stayed active, no interaction. Changes 3 to 2, top bar +359 −1 to +357 −1.
- Row changed while unwatched, first activation this session: 2 files shown on activation.
- Row with cached stats, changed while unwatched: activation refreshed it. Changes 2 to 3 with the new file listed.
- Adversarial: 12 A/B switches with a 120 ms busy-loop every 400 ms injected on the main thread. Zero console.error, no "Maximum update depth", final counts correct. One extra watch/unwatch pair came from the row hover card mounting its own git status as the pointer crossed a row, which is existing behavior and was released.

Test edits were reverted and the dev stack stopped.

PR #7218, opened 2026-09-05
feat(billing): tell annual subscribers before we charge them

saddlepaddle:
**Stacked on #7217** — base is `billing-email-gaps`, since both touch the same `onEvent` region. Retarget to `main` once that merges.

A yearly subscriber hears from us at signup and then **nothing for twelve months**, until a charge an order of magnitude larger than a monthly one lands. Stripe fires `invoice.upcoming` **7.0 days** ahead (measured across the last 30 days, not assumed) and we were ignoring it — 413 a month, none handled.

## Annual only, on purpose

**Not monthly.** 363 of those 413 events are flat $20 monthlies. A weekly heads-up before a charge people already expect mostly reminds them to cancel.

**Not seat changes**, even though that was the original motivation. `MemberAddedBillingEmail` and `MemberRemovedBillingEmail` already fire the moment seats change, to the same billing recipients, quoting seat count, new total, proration, and the exact next invoice via `previewNextInvoice` — whose own comment names the failure this would have duplicated:

> *"A seat-change email stating a total the customer will not be charged is how a card limit gets set too low, which is the whole reason this exists."*

That leaves the genuine gap: **141 live annual subscriptions** with no signal between signup and renewal.

## Details worth reviewing

- **No CTA.** Nothing is payable yet — it's a renewal notice, not an invoice. Also sidesteps the missing web billing route.
- **Subscription matched on the Stripe id**, not the organization: an organization that resubscribed has several `subscriptions` rows and the wrong one can win. Same reasoning as the guard in #7217.
- **Seats take the max quantity across lines, not the sum** — a proration line carries its own quantity, and summing reports more seats than exist.
- **Interval comes from our `billingInterval` column**, because on the current API version invoice lines carry only a price *id* (`pricing.price_details.price`), with no expanded `recurring.interval` to read.
- Goes to `BILLING_ROLES` (owners + admins), not the whole team.

## Verification

- Rendered the email and read the output; preview sent to Satya for copy review before this ships.
- `lint.sh` clean across 7,214 files. `typecheck` clean on auth, email, trpc, api, web. `@superset/auth` tests 8/8.
- Payload shape confirmed against a real `invoice.upcoming` event rather than assumed.

---
## Summary by cubic
Sends annual subscribers a renewal notice before we charge them by handling Stripe's `invoice.upcoming` events, which we previously ignored. Only yearly plans get the email; monthly plans and seat changes are deliberately excluded because they already receive timely notifications or the charge is expected.

**Details**
- Subscriptions are matched on the Stripe subscription id, not the organization, to avoid hitting the wrong row for resubscribed organizations.
- Seat count uses the max quantity across invoice lines instead of the sum, because proration lines double-count.
- The email goes to billing roles (owners and admins) and contains no call to action since nothing is due yet.
- Owners are told to update the card in Settings → Billing; admins are told to ask an owner, since payment methods are owner-only.

<sup>Written for commit c7ec186d2daa49370d0f5901a8adca13f6a428ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added renewal reminder emails for annual subscriptions before the upcoming renewal date.
  - Reminders include the organization, plan, renewal amount, renewal date, and seat count.
  - Payment-method instructions are tailored for account owners and administrators.

### DIFF
--- a/packages/auth/src/server.ts
+++ b/packages/auth/src/server.ts
@@ -11,6 +11,7 @@ import { WelcomeEmail } from "@superset/email/emails/activation/00-welcome";
 import { MemberAddedBillingEmail } from "@superset/email/emails/billing/member-added";
 import { MemberRemovedBillingEmail } from "@superset/email/emails/billing/member-removed";
 import { PaymentFailedEmail } from "@superset/email/emails/billing/payment-failed";
+import { RenewalUpcomingEmail } from "@superset/email/emails/billing/renewal-upcoming";
 import { SubscriptionCancelledEmail } from "@superset/email/emails/billing/subscription-cancelled";
 import { SubscriptionStartedEmail } from "@superset/email/emails/billing/subscription-started";
 import { OrganizationInvitationEmail } from "@superset/email/emails/team/invitation";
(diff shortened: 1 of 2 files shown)


saddlepaddle:
Fixed — valid catch. The notice goes to owners and admins (BILLING_ROLES), but `requireOwnerWithCustomer` refuses billing changes for admins, so the Settings → Billing instruction was unfollowable for half the recipients. Admins now get "ask an owner to update it", matching `usePaymentFailedCard`'s existing owner/member split.

Worth distinguishing from #7217, which removed an owners-only gate: that one was on a *hosted invoice* link, payable by anyone holding it. Our billing settings genuinely are owner-only, so this gate stays.

PR #7220, opened 2026-09-06
fix(billing): keep warning people after the subscription is gone

saddlepaddle:
**Draft — I have not seen this render.** Typechecked, linted, tests at baseline, but the card itself is unverified in a running app. See "What still needs proving" below before merging.

## The gap

`usePaymentFailedCard` keyed on `past_due`, so it vanished at exactly the moment it mattered: when Stripe stops retrying, cancels, and the organization drops to free.

**All 87 organizations this happened to recorded zero `payment_failed_banner_shown`** (PostHog, 90 days). They weren't ignoring a warning — they were never shown one. 40 of them are still active in the last 30 days, and many are hitting the Pro paywall repeatedly, unaware their triggers stopped because a card failed months ago.

## Why the lapsed state says something different

Paying the old invoice **cannot** undo a cancellation — Stripe: *"You can't reactivate a canceled subscription."* So a "Pay now" button here would lead nowhere useful. The lapsed card offers **Restart Pro** into Billing instead, and its copy names the actual consequence (triggers stopped), which is what these users are visibly running into.

The `past_due` path is unchanged.

## The load decision, because I got it wrong first

My first version gated the client on `plan === "free"`. That's every free user, and `outstandingInvoice` makes a live Stripe call for any org with a customer id — a Stripe round-trip on every app load for every past customer, a set that only grows.

Now `activePlan` decides `lapsed` server-side from rows it already reads, so the extra queries stay off for everyone still paying *and* everyone who never paid. It counts `unpaid` as well as `canceled`, because which one Stripe lands on after retries is a dashboard setting and checking only `canceled` would let this silently stop working. `incomplete` is excluded — that subscription never began.

## Verification so far

- `typecheck` clean on desktop and trpc; `lint.sh` clean; `check:i18n` compiles.
- Desktop suite **3516 pass / 2 fail**, and those 2 (`LeaderboardRank`) fail identically with this change stashed — pre-existing and env-dependent, confirmed by running both ways.

## What still needs proving

I have not seen either state render. Doing it properly needs the desktop app built from this branch plus an organization in the lapsed state — a cancelled subscription with an unpaid invoice — which I can't seed without database access. The only Superset running locally is shipped Canary, which per `cdp-verification`'s first rule isn't equivalent.

Worth exercising both states before this merges: `past_due` (should be unchanged) and lapsed (new).

---
## Summary by cubic
Fixes the payment-failed warning card so it stays visible after Stripe cancels the subscription and the organization drops to free, instead of vanishing at that exact moment.

- `activePlan` returns a `lapsed` flag computed server-side from rows it already reads, avoiding extra Stripe calls for users still paying or who never paid.
- Lapsed is true for both `canceled` and `unpaid` statuses; `incomplete` is excluded.
- The lapsed card offers "Restart Pro" to Billing instead of a "Pay now" that leads nowhere, since Stripe won't reactivate a cancelled subscription.
- Its copy states the plan ended and the amount went uncollected as two facts, not one cause, because voluntary cancellations can also land on `canceled` with an unpaid closing invoice.
- The `past_due` card behavior and its direct-to-invoice link are unchanged; both states track `payment_failed_banner_shown` and `payment_failed_banner_clicked` with a `state` field.

<sup>Written for commit 0f858b34f36d9ccf87b3f9de123146bed72ca718. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added billing notices for organizations with lapsed subscriptions.
  * Lapsed subscription notices guide owners to restart Pro instead of paying an outdated invoice.
  * Billing status now distinguishes between lapsed subscriptions and past-due payments.
  * Notices clarify that the organization is on the free plan and that automated triggers have stopped.
  * Unpaid invoice amounts are shown only when applicable.

### DIFF
--- a/apps/desktop/src/renderer/components/SidebarCardSlot/cards/usePaymentFailedCard/usePaymentFailedCard.ts
+++ b/apps/desktop/src/renderer/components/SidebarCardSlot/cards/usePaymentFailedCard/usePaymentFailedCard.ts
@@ -8,9 +8,20 @@ import { electronTrpc } from "renderer/lib/electron-trpc";
 import type { SidebarCardEntry } from "../../types";
 
 /**
- * Stripe keeps retrying for ~14 days before canceling, and access continues
- * for that whole window, so a failed charge is otherwise invisible in-app
- * until the plan abruptly disappears. Returns no `onDismiss` for that reason.
+ * Two states, because a failed charge is otherwise invisible in-app.
+ *
+ * `past_due` — Stripe is still retrying and access continues, so the card
+ * warns while paying the invoice still saves the subscription.
+ *
+ * `lapsed` — Stripe gave up, cancelled, and the organization dropped to free.
+ * Keying only on `past_due` meant the card vanished at exactly that moment:
+ * every one of the 87 organizations this happened to recorded zero
+ * `payment_failed_banner_shown`. They were never told, they just quietly lost
+ * their triggers. Paying the old invoice cannot undo it either — Stripe will
+ * not reactivate a cancelled subscription — so this state sends people to
+ * Billing to resubscribe instead of to the invoice.
+ *
+ * Neither state gets an `onDismiss`.
  */
 export function usePaymentFailedCard({
 	surface,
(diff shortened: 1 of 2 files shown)


saddlepaddle:
Closed the main correctness risk, so moving this out of draft.

I'd flagged that `lapsed` depends on our `subscriptions.status` being exactly `canceled`, and that I hadn't proven our rows say that. Now traced: better-auth's `onSubscriptionDeleted` (`@better-auth/stripe/dist/index.mjs:401`) looks the row up by `stripeSubscriptionId` and writes `status: "canceled"` as a hardcoded literal on `customer.subscription.deleted`. So every organization in this cohort has exactly that value, and `lapsed` will be true for them.

What's left is **visual only** — nobody has watched either state render. The logic, the load characteristics, and the data path are all verified. Worth clicking through `past_due` (should be unchanged) and lapsed (new) before merge, but I no longer think there's a hidden correctness problem in here.

saddlepaddle:
Fixed. Valid catch, and I have the data for it: **8 of the 87** organizations in this cohort ended with `cancellation_requested` and still had an open invoice, so this was not hypothetical.

Both clauses were individually true. The `so` joining them was the problem: it asserted the failed charge caused the downgrade, which is wrong for someone who chose to leave. Now stated as two separate facts, which reads correctly either way.

Worth noting this is the same defect class that #7204 fixed in the cancellation *email* (it told involuntarily-churned users they had access until a date ~17 days out). I reintroduced it in the card. Getting the cancellation reason properly would need a column we do not have, so the honest fix is to not claim a reason.

saddlepaddle:
Pushing back on the "render the lapsed card without requiring an invoice" suggestion, because the invoice requirement is load-bearing rather than incidental.

`lapsed` alone only says a subscription ended. It cannot say **why**, and that is the whole difficulty in this area: `canceled` covers both "the card failed" and "they deliberately cancelled". The unpaid invoice is what separates the two.

Drop the requirement and every organization that cancelled cleanly and settled its final bill gets a permanent "Action needed: Pro ended, restart it" card with no way to dismiss it. That is nagging people who made a deliberate decision, and it is a worse bug than the one being fixed.

The specific null cases raised:
- **No Stripe customer** then there are no subscription rows either, so `lapsed` is already false. No card, correctly.
- **No open or uncollectible invoice** means they paid up or it was voided. Correctly no card.
- **Query still loading** resolves to the card appearing once data arrives, which is ordinary async rendering for a sidebar slot.

The greptile P1 above is stale, addressed in 0f858b34f: the copy no longer claims the failed charge caused the downgrade.

PR #7234, opened 2026-09-06
feat(mobile): pick the agent's model and effort when starting a session

saddlepaddle:
## Summary
- The iOS home composer gets a model dropdown after the agent (`✱ Claude ⌄ Opus ⌄`). It opens a sheet titled with the agent: a short model list, the pinned releases behind a "Pinned releases ›" row, and the reasoning effort below.
- Picks persist per launch preset and apply to every launch of that agent from the phone: new workspaces (host and cloud), the workspace "+" sheet (which shows the pick under the agent), the PR "Ask agent" buttons and finish-review.
- Unset sends nothing, so the agent's own default is unchanged.

## Why / Context
Slack request from an iOS user whose Fable credit ran out mid-day with no way to switch to Opus from the phone. The desktop's new-workspace form already has model and effort pickers; the phone had none.

## How It Works
- No new API. The host's `agents.run` and `workspaces.create*` agent entries and `cloudWorkspace.create` already accept `model`/`effort` and validate them against `packages/shared/src/agent-models.ts`. The app reads that same catalog, so the pickers are per-preset and dynamic (Codex's per-model effort ladder included), and a Pi config running OMP resolves to OMP's options the way the host does.
- `newSessionPreferencesStore` gains `modelByAgent`/`effortByAgent`, keyed by launch preset id like the desktop's per-preset maps. `useAgentLaunchPreferences` resolves a pick against the catalog; an id that fell out, or an effort the picked model rejects, reads as Default rather than a failing launch.
- Native Composer gains a `launchOptions` prop plus `onLaunchOptionPress`, rendered after the agent button in the expanded toolbar.
- `AgentOptionsScreen` shows the catalog up to and including its first group inline (Claude: Fable, Opus, Sonnet, Haiku); every later group is a row that pushes `ModelGroupScreen`. Picks stick immediately; the sheet stays open.

## Manual QA Checklist
Verified on an iPhone 17 Pro simulator against prod (App Review org, acme-devbox):
- [x] Composer shows `Claude ⌄ Default model ⌄`; picking Opus and High updates the dropdown to `Opus`
- [x] Sheet: Model section with Default/Fable/Opus/Sonnet/Haiku, Pinned releases › (shows the picked pinned name + check), Effort section Default…Max
- [x] Workspace "+" sheet shows "Claude / Opus" and the launched session's Claude Code banner reads "Opus 5"
- [x] Agents without an effort flag show no Effort section; agents without a model flag show no dropdown
- [x] Preference survives app relaunch
- [ ] Cloud workspace launch with a pick (not exercised; same `cloudWorkspace.create` fields the desktop sends)

## Testing
- `bun run typecheck` (apps/mobile; needs `bun run build:types` in packages/host-service first in a fresh worktree)
- `bun run lint`
- `bun run test` (apps/mobile, 66 pass)
- `bun run check:i18n` ("Effort" is the only new string, translated in all 16 locales)
- `xcodebuild` Debug simulator build of the app

## Design Decisions
- **Picks apply to every launch of that agent, not only new workspaces**: the requester's credit ran out, so a pick covering only the home composer would still burn on "Ask agent" and finish-review. The workspace "+" sheet reflects the pick rather than offering its own picker.
- **No change to running sessions**: Claude Code and Codex already expose `/model` in the terminal composer's slash-command list, which keeps the conversation. A restart-in-place path was prototyped and dropped.
- **Short list + grouped sub-screen** instead of the desktop's flat select: twelve pinned versions buried the four aliases almost everyone wants.

## Known Limitations
- The Claude "Latest" aliases (`opus`, `fable`) resolve on the host's CLI, so a pick means whatever that CLI calls Opus today. Same as the desktop.

https://claude.ai/code/session_01X7E4Rre3GzgfkQCPjJPzQ1

---
## Summary by cubic
Adds model and effort pickers to the iOS composer so you can start an agent on a specific model instead of its default. Picks persist per launch preset and apply to every launch of that agent from the phone, including new host/cloud workspaces, the workspace "+" sheet, PR "Ask agent", and finish-review. Leaving a pick unset sends nothing, so the agent's own default is unchanged.

**Notes**
- Agents without model or effort support show no picker.
- A remembered pick that is no longer valid falls back to Default instead of failing the launch.
- Picks still apply before the host's agent configs load; the preset id stands in until then.
- No new API; launches reuse the existing `model` and `effort` fields.

<sup>Written for commit 7755db46644f6a62dac69f87257dd6957c2cafde. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added model and reasoning-effort selection when starting agent sessions.
  - Users can browse available models, model groups, or restore default settings.
  - Selected preferences are remembered per launch preset and shown in session controls.
  - Model and effort choices apply to cloud, terminal, review, and follow-up agent sessions.
  - Added localized “Effort” labels across supported languages.

- **Improvements**
  - Added clearer indicators for selected options and launch preferences.
  - Added navigation from the new-session flow to model and preference screens.

### DIFF
--- a/apps/mobile/app/(authenticated)/(home)/new-session/_layout.tsx
+++ b/apps/mobile/app/(authenticated)/(home)/new-session/_layout.tsx
@@ -24,6 +24,13 @@ export default function NewSessionLayout() {
 				}}
 			/>
 			<Stack.Screen
+				name="model"
+				options={{
+					title: t({ message: "Model" }),
+				}}
+			/>
+			<Stack.Screen name="model-group" />
+			<Stack.Screen
 				name="project"
 				options={{
 					title: t({ message: "Project" }),
--- a/apps/mobile/app/(authenticated)/(home)/new-session/model-group.tsx
+++ b/apps/mobile/app/(authenticated)/(home)/new-session/model-group.tsx
@@ -1,0 +1,3 @@
+import { ModelGroupScreen } from "@/screens/(authenticated)/components/ModelGroupScreen";
+
+export default ModelGroupScreen;
--- a/apps/mobile/app/(authenticated)/(home)/new-session/model.tsx
+++ b/apps/mobile/app/(authenticated)/(home)/new-session/model.tsx
@@ -1,0 +1,12 @@
+import { AgentOptionsScreen } from "@/screens/(authenticated)/components/AgentOptionsScreen";
+
+export default function NewSessionModelRoute() {
+	return (
+		<AgentOptionsScreen
+			groupHref={(params) => ({
+				pathname: "/(authenticated)/(home)/new-session/model-group",
+				params,
+			})}
+		/>
+	);
+}
--- a/apps/mobile/modules/composer/ios/ComposerModel.swift
+++ b/apps/mobile/modules/composer/ios/ComposerModel.swift
@@ -38,6 +38,10 @@ final class ComposerModel {
   /// `ComposerModelPicker`.
   var selectedModel: ComposerMenuOption?
 
+  /// Per-agent launch settings drawn after the agent — model, effort — each a
+  /// chevron button reporting its id. Empty for agents that have none.
+  var launchOptions: [ComposerMenuOption] = []
+
   /// A submit is in flight. The caller owns this — only it knows when delivery
   /// finished — and while it is true the send button shows a spinner and the
   /// mic gets out of the way.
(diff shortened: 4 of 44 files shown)


saddlepaddle:
Addressed in 7755db466:
- **Preferences dropped while agent configs load** (greptile, CodeRabbit): the store's preset id now stands in for the launch preset until the host's configs answer, in the home composer, "Ask agent" and finish-review. A send made before they arrive carries the pick; the two ids only differ for a config whose executable is not its preset's, where the resolver reads Default rather than a bad flag.
- **OptionRow accessibility**: rows expose `accessibilityRole="button"` and `accessibilityState.selected`.
- **id/ja "Effort"**: now "Upaya" and "推論の強度", matching the catalogs' reasoning terms.

Skipped: restoring composer focus after the options sheet closes. The agent, project and branch pickers pushed from the same composer behave the same way today; changing that belongs to all of them together, not this sheet alone.

PR #7242, opened 2026-09-06
feat(skills): add an integrations skill for calling plugin MCP tools

harshithmullapudi:
## Summary
- New `plugins/superset/skills/integrations/SKILL.md`: how to find a connected integration, list the tools it exposes, and call one through `superset mcp`. Its description is written to fire on integration questions ("do X in Linear", "what tools do I have", "why is this tool failing").
- Fixes the stale command surface in the `plugins` skill and points its "Call its tools" section at the new skill.

## Why / Context
The `plugins` skill covers install → connect → marketplace. Nothing covered the other half: an agent that has a connected integration and needs to actually use it. Worse, the four `superset mcp` lines that skill did carry no longer match the CLI, so an agent following them writes a failing command.

Verified against the live CLI, not just the source:

| Skill said | CLI today |
| --- | --- |
| `superset mcp tools --plugin-id <id>` | `superset mcp tools <plugin>`; `--plugin-id` is a deprecated alias for `--connection` |
| `superset mcp call-tool <tool> --plugin-id <id>` | `call-tool <plugin> <tool> [json]` — plugin first, and `--connection` does not stand in for it (`Missing required argument: <tool>`) |
| `superset marketplace list/install/remove` | `superset plugins marketplace list/add/remove` — the old form errors with `Unknown command: marketplace` |
| `superset plugins remove` | `uninstall` (`remove` survives as an alias) |

## How It Works
The bundled `superset` plugin ships every directory under `plugins/superset/skills/` automatically — `packages/agent-setup/src/managed-skills.ts:300` enumerates them, so a new skill needs no code change, no version bump, and no marketplace publish (it is a built-in marketplace, absent from `.agent-marketplace.json`).

The new skill deliberately stops where the `plugins` skill starts: it never installs or connects anything, it hands that back. Content worth flagging for review:

- `superset mcp tools` returns name and description only — **no input schema**. The skill says so and tells the agent to read the full description rather than the `jq`-filtered name list.
- A plugin with several connected accounts has no default; the CLI refuses to guess and prints one `--connection <id>` line per account. The skill makes choosing an account the user's call.
- Results come back as an MCP `content` array whose text parts are themselves JSON strings, so the skill shows the `jq -r '.content[0].text' | jq` unwrap.

## Manual QA
Every command in both skills was run against the real CLI in this workspace:

- [x] `superset mcp tools linear` lists tools; `jq -r '.[].tool'` filters them
- [x] `superset mcp call-tool linear list_agent_skills` returns a result; the `'{}'` third positional parses
- [x] `superset mcp call-tool --connection <id> list_agent_skills` fails with `Missing required argument: <tool>` — confirms the plugin positional is required alongside `--connection`
- [x] `superset mcp tools github` (not connected) → `"github" is not connected. Connect an account first.`
- [x] `superset mcp tools` (no target) → `Name a plugin, or pass --connection <id>.`
- [x] `superset plugins marketplace list` works; `superset marketplace list` → `Unknown command: marketplace`
- [x] Multi-account and 401/403 rows in the failure table read from `packages/cli/src/lib/plugins/connection-ref.ts` rather than reproduced live
- [x] "disable keeps the install but stops provisioning skills" checked against `packages/agent-setup/src/installed-plugins.ts:48`

## Testing
- `bun test` in `packages/agent-setup` → 304 pass. This suite validates every bundled skill (`bundled-skills.test.ts`), and I wrongly skipped it on the first push as "markdown, no code" — it caught three real failures: em dashes (a repo prose rule) in both files, and a missing `agents/openai.yaml` for the new skill. Fixed in 021f0ce.
- `bun run check:plugins` → "All plugins are valid and published."

## Review feedback (021f0ce)
All three CodeRabbit findings were valid and are applied:
- `plugins list` prints a row for an installed plugin with **no** connection yet (`list/command.ts` returns `pluginId: ""` for it), so "one row per connected account" hid exactly the integration that needs connecting.
- A tool's 401/403 is not always an expired credential. A 403 naming a scope or resource is a permissions problem on the account you already have, and reconnecting it changes nothing. The table and anti-patterns now say so.
- `uninstall` removes the plugin locally even when the account call fails (`uninstall/command.ts:19-46`). Since the credential is revoked only as part of that account call (`plugins.ts:532-557` clears the tokens), an unconfirmed removal leaves the connection live. The skill now says to re-check and re-run.

## Known Limitations
- The failure table's multi-account and expired-credential rows are read from the source, not reproduced — I have one connected account and did not want to revoke a credential to see the error.

---
## Summary by cubic
Adds an `integrations` skill so agents can discover and call the MCP tools of a connected plugin, updates the `plugins` skill's CLI examples to match the current `superset` CLI, and points report-producing skills at `superset pages` so their output persists as a link instead of scrollback.

- The `integrations` skill covers listing tools, picking an account, unwrapping MCP `content` results, and landing agents when a plugin's skill expects native `mcp__*__` tools that aren't available; it stops at install and connect.
- Failure guidance now distinguishes permission-based 403s from expired credentials and covers partially-completed uninstalls.
- The `plugins` skill now uses plugin-first `call-tool` syntax, `--connection` for account selection, and current `uninstall`, `disable`, and `plugins marketplace` subcommands.

**Publishing behavior**
- Page identity is workspace plus path; history accumulates only when the workspace stays constant or `--page <id>` is passed.
- `standup` and `automate` render the report to `.html` before publishing.
- `browser` and `computer` require reading captures, cropping, and `--visibility just_me` unless org-wide is intended; `computer`'s grant is narrowed to `superset pages publish`.

No code changes; the bundled `superset` plugin auto-loads every skill in its directory, and an `agents/openai.yaml` ships with the new skill.

<sup>Written for commit 59d1cfedf6d5d385b39484bc87bdbbac9f187ed3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for discovering connected integrations, available tools, account selection, secure secret handling, and fallback behavior.
  - Clarified integration authentication, argument, JSON, and permission errors, including when reconnection is appropriate.
  - Updated plugin workflows, management commands, and local uninstall behavior.
  - Added OpenAI integration skill guidance for discovering and using tools.
  - Documented publishing reports, screenshots, walkthroughs, run summaries, and recurring digests as shareable pages with comments and versioning.
  - Added guidance for workspace-based page identity, evidence redaction, visibility controls, and publishing confirmation.

### DIFF
--- a/plugins/superset/skills/10x/SKILL.md
+++ b/plugins/superset/skills/10x/SKILL.md
@@ -35,7 +35,7 @@ For each recommendation in order: a two-sentence pitch, then ask (use the ask_us
 | Terminal remote-control | Read and drive any agent's terminal from anywhere | `superset terminals list / read / send` |
 | Custom slash commands | Your repo's own workflows as commands every agent can run | create `.agents/commands/<name>.md` in their repo |
 | MCP servers | Give every workspace agent the same extra tools | add servers to `.mcp.json` at their repo root |
-| Pages | Turn a report, dashboard, or digest into a link the org can read and pin comments to, versioned on every republish | `superset pages publish report.html --title "..."`, then `superset pages comments list` |
+| Pages | Turn a report, dashboard, or digest into a link the org can read and pin comments to, versioned on every republish | `superset pages publish report.html --workspace <id> --title "..."`, then `superset pages comments list` |
 | Feedback loop | Report bugs or ideas without leaving the agent | the feedback skill |
 
 ## Rules
--- a/plugins/superset/skills/10x/SKILL.md
+++ b/plugins/superset/skills/10x/SKILL.md
@@ -35,6 +35,7 @@ For each recommendation in order: a two-sentence pitch, then ask (use the ask_us
 | Terminal remote-control | Read and drive any agent's terminal from anywhere | `superset terminals list / read / send` |
 | Custom slash commands | Your repo's own workflows as commands every agent can run | create `.agents/commands/<name>.md` in their repo |
 | MCP servers | Give every workspace agent the same extra tools | add servers to `.mcp.json` at their repo root |
+| Pages | Turn a report, dashboard, or digest into a link the org can read and pin comments to, versioned on every republish | `superset pages publish report.html --title "..."`, then `superset pages comments list` |
 | Feedback loop | Report bugs or ideas without leaving the agent | the feedback skill |
 
 ## Rules
--- a/plugins/superset/skills/automate/SKILL.md
+++ b/plugins/superset/skills/automate/SKILL.md
@@ -13,6 +13,13 @@ Turn "I keep doing X every morning" into an automation that does X on a schedule
 
 Pin down: the outcome, the cadence, the inputs it reads, and what "done" looks like. Then draft the automation prompt as instructions for an agent with zero context. If the task has rules that will evolve (triage criteria, formats), put them in a document the automation reads at runtime so they can be edited without touching the prompt.
 
+Nobody watches a run, so a chore whose product is a digest, a report, or a scorecard needs somewhere for that product to land. End the prompt by publishing a page to the same path every run: the user opens one link instead of digging through run logs, and each run adds a version, so the page becomes the history of the thing it tracks.
+
+```
+...then publish the digest: superset pages publish digest.html \
+  --workspace $SUPERSET_WORKSPACE_ID --title "Nightly triage" --label "what changed today"
+```
+
 ## 2. Pick the target
 
 - `superset projects list`: a project target creates a fresh workspace per run (most tasks)
@@ -36,4 +43,4 @@ superset automations create \
 
 (`--workspace <id>` instead of `--project` for reuse mode; `--host <id>` if it should run on another machine; prefer `--prompt-file` for multiline prompts.)
 
-Then trigger a first run now with `superset automations run <id>`, review `superset automations logs <id>` with the user, and refine the prompt via `superset automations prompt set <id>` until the run output is right. An automation isn't done until one real run looked good.
+Then trigger a first run now with `superset automations run <id>`, review `superset automations logs <id>` with the user, and refine the prompt via `superset automations prompt set <id>` until the run output is right. An automation isn't done until one real run looked good. If the prompt publishes a page, open the published page as part of that review, not just the run log.
--- a/plugins/superset/skills/automate/SKILL.md
+++ b/plugins/superset/skills/automate/SKILL.md
@@ -13,12 +13,14 @@ Turn "I keep doing X every morning" into an automation that does X on a schedule
 
 Pin down: the outcome, the cadence, the inputs it reads, and what "done" looks like. Then draft the automation prompt as instructions for an agent with zero context. If the task has rules that will evolve (triage criteria, formats), put them in a document the automation reads at runtime so they can be edited without touching the prompt.
 
-Nobody watches a run, so a chore whose product is a digest, a report, or a scorecard needs somewhere for that product to land. End the prompt by publishing a page to the same path every run: the user opens one link instead of digging through run logs, and each run adds a version, so the page becomes the history of the thing it tracks.
+Nobody watches a run, so a chore whose product is a digest, a report, or a scorecard needs somewhere for that product to land. End the prompt by writing the report to an `.html` file and publishing it, so the user opens one link instead of digging through run logs.
 
 ```
-...then publish the digest: superset pages publish digest.html \
-  --workspace $SUPERSET_WORKSPACE_ID --title "Nightly triage" --label "what changed today"
+...write the digest to digest.html, then publish it:
+superset pages publish digest.html --title "Nightly triage" --label "what changed today"
 ```
+
+A page is identified by its workspace plus its path, so **which target you picked in step 2 decides whether history accumulates**. A project target creates a fresh workspace per run, which means a new page every run rather than a new version of one. For a report meant to build up history, use a workspace target, or capture the page id from the first run and have the prompt pass `--page <id>` from then on.
 
 ## 2. Pick the target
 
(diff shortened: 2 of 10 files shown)


saddlepaddle:
AH i see why this is useful - otherwise you can't use the plugins in the superset cli. Mb for pushing to drop it!

harshithmullapudi:
Addressed in 021f0ce.

CI was right and I was wrong to skip `bun test` as "markdown, no code" — `packages/agent-setup/src/bundled-skills.test.ts` validates every bundled skill and caught three failures: em dashes (repo prose rule) in both files, and a missing `agents/openai.yaml` for the new skill. All fixed; the package's 304 tests pass locally.

All three CodeRabbit findings were valid:

1. **`plugins list` rows** — `list/command.ts` returns a row with `pluginId: ""` for an installed plugin with no connection, so "one row per connected account" hid exactly the integration that needs connecting. Reworded, and the `needs connection` case now says its tools fail until someone connects.
2. **401/403** — agreed, a 403 naming a scope or resource is a permissions problem on the account you already have. The table now says to read the error and reconnect only when it names an invalid, revoked, or expired credential, and "reconnecting on any 403" is listed as an anti-pattern.
3. **Partial uninstall** — `uninstall/command.ts:19-46` removes locally even when the account call fails, and `plugins.ts:532-557` shows the credential is revoked only as part of that account call. So an unconfirmed removal leaves the connection live while the skills are gone. The skill now says to check with `plugins connections --plugin <name>` and re-run.

harshithmullapudi:
4148d59 adds the case that motivated this skill in the first place.

A plugin's own skills are written against native MCP tools: linear's three all declare `allowed-tools: mcp__linear__*`. Those tools exist only where the plugin's MCP server was written into the agent's config, and that happens in exactly two places today — `syncManagedMcpServers` targets Claude Code and Codex, and its only caller is the desktop install path (`plugin-installs.ts:93`). A CLI or account install provisions skills and never wires MCP.

Reproduced on my machine: `~/.agents/skills/` has all three linear skills, `~/.claude.json` has no `linear` server, and `~/.superset/plugins/mcp-ledger.json` is `{"files": {}}`. Ask the agent to file a Linear issue and it loads `linear-file-issue`, reaches for `mcp__linear__*`, finds nothing, and stops — the integrations skill never gets consulted, because "file this in Linear" matches the vendor skill's description far more closely.

So the skill now names that recovery explicitly and says to keep following the skill you were reading, since only the transport changes. Its description covers the case too.

The complete fix wants a fallback line inside the vendor skills themselves, since those are what the agent reads first. That means a `plugins publish` and a tag for `linear` and `github`, so I've left it out of this PR.

harshithmullapudi:
560489e adds a second, separable piece of work to this branch: wiring pages into the skills that produce something worth reading. Happy to split it into its own PR if you'd rather review them apart.

Before this, nothing outside `page/SKILL.md` mentioned pages at all. The two `plugins` <-> `integrations` links added earlier in this branch were the only skill-to-skill references in the whole bundled set. So every skill that produces a reader-facing artifact stopped at the terminal, and skill selection is over by the time one of them is running: the agent that just built a standup digest is not going to go shopping for the page skill. Each producer now names the handoff itself, with the command.

- `standup`: offer the digest as a page; republishing the same path versions one page instead of a new one daily. Carries the workspace caveat, since a digest often runs from outside a workspace and `page/SKILL.md:150` says a publish without one is refused.
- `orchestrate`: a run across several workspaces is worth a link reviewers can pin comments to.
- `automate`: nobody watches a run, so digest-shaped chores should end by publishing. Placed in section 1, where the prompt is actually drafted, not section 4 where I first put it.
- `10x`: Pages was missing from the feature catalog entirely.
- `browser` / `computer`: publish verification screenshots as a directory publish, so the images ride along at their relative paths.
- `page`: names the producers, so the link runs both ways.

`computer` had no superset entry in `allowed-tools` (`Bash(cua-driver:*) Bash(peekaboo:*)`), so the command it now suggests would not have run; added `Bash(superset pages:*)`.

`doctor` is deliberately untouched. Its escalation path already warns that logs carry paths and project names, and pages default to org-visible, so publishing diagnostics quietly widens who sees them.

Every flag checked against `superset pages publish --help` rather than written from memory. `bundled-skills.test.ts` 92/92, full agent-setup suite 304/304, `check:plugins` clean.

harshithmullapudi:
All seven addressed in 59d1cfe. The `automate` one was a real error on my part, not a nit.

**Page identity (automate, page, standup).** Confirmed against `packages/trpc/src/router/page/publish.ts:484-497`: an existing page is resolved by `(workspaceId, entryPath, organizationId, createdByUserId)`. `automate` section 2 offers a *project* target that creates a fresh workspace per run, so my "each run adds a version" claim was backwards for that case: it would mint a new page every night. `automate` now says the target choice in step 2 is what decides whether history accumulates, and points at a workspace target or a stable `--page <id>`. `page` and `standup` no longer imply the path alone is the key.

**Publishing a file nothing writes (standup, automate).** Correct, and it would have failed with ENOENT: neither skill ever produces `standup.html` / `digest.html`. Both now say to render the report to an `.html` file first.

**Evidence redaction (browser).** Fair, and I should have caught it. The browser is signed into the user's accounts and pages default to org visibility. `browser` now says to read every screenshot before it goes up, crop or redact, use `--visibility just_me` when the org has no need for it, and confirm first. Same for `computer`, where a desktop capture can catch mail, messages, or a password manager.

**Confirmation (orchestrate).** Added, matching what `standup` already said.

**Permission scope (computer).** Narrowed `Bash(superset pages:*)` to `Bash(superset pages publish:*)`. The skill only publishes.

Unrelated to this diff: SkillSpector's two findings on `10x` lines 16 and 24 are pre-existing content, untouched here.

`bundled-skills.test.ts` 92/92, `check:plugins` clean, `--visibility` verified in `superset pages publish --help`.

PR #7252, opened 2026-09-07
fix(agent-setup): preserve Cursor identity through tool execution

owieschon:
## What & why

Cursor imports Claude hook settings. Superset’s imported command sets `SUPERSET_AGENT_ID=claude`, so a duplicate event can relabel a Cursor terminal as Claude and change its runtime identity. This reproduces with a GPT model selected in Cursor; the terminal label identifies the agent application, not the model.

Suppress imported events only when the command identifies Claude, `CURSOR_VERSION` is present, and the event uses Cursor’s lowercase naming. Genuine Claude events remain supported, including Claude launched from Cursor’s environment. Bump the generated notification hook marker to v10.

Register native `postToolUse` and `postToolUseFailure` as `Start` so shell/MCP activity does not remain at `PermissionRequest` after the imported duplicate is suppressed. Match only `Shell` and `MCP:<tool>`, whose existing before-hooks set that state. Existing user hooks are preserved.

Cursor documents [third-party hook imports](https://cursor.com/docs/reference/third-party-hooks) and [native hook matchers](https://cursor.com/docs/hooks).

## How I tested it

Real interactive Cursor CLI `2026.09.02-c22c1a3`, launched inside Superset with `gpt-5.6-terra-low`. The same terminal shows Claude after a completed turn with the base notification hook, then Cursor with the candidate hooks. Captures are cropped to the terminal header and response; private metadata is omitted.

**Before: Cursor running, Claude icon**

Cursor incorrectly labeled Claude

**After: Cursor running, Cursor icon**

Cursor identity preserved after completion

- Native CLI tool runs verified shell success/nonzero exit, MCP success/error results, file-read exclusion, and session-end detachment against host bindings. Temporary hook configuration was restored byte-for-byte. Tool qualification used print mode; completed-turn captures used the interactive CLI.
- Regression fails before the post-tool registrations: `PermissionRequest` remains where `Start` is required. Removing the matcher in an isolated test copy fails both child Read assertions.
- Bun **1.3.14**, current commit `a0ee299`: **303 package tests passed**, 1,262 assertions across 14 files. Repository lint and all 38 typecheck tasks passed locally. Executable hook tests also cover interruption, duplicate imports, genuine Claude events, and other providers. Interruption is replay coverage, not a live user-cancel claim.
- [CI on the current commit `a0ee299`](https://github.com/owieschon/superset/actions/runs/34077476906): **all 11 jobs passed**, including lint, tests, typecheck, desktop build, and four CLI builds.
[...]

PR #7263, opened 2026-09-07
fix(agent-setup): preserve agent identity and enable OpenCode session resume

Kitenite:
## What & why

Cursor Agent could acquire a Claude binding and then fail to resume: Cursor replays `~/.claude/settings.json`, whose managed hook command overwrote `SUPERSET_AGENT_ID` with `claude`. The host accepted that identity together with Cursor's session ID, so a later terminal restore ran `claude --resume <cursor-session-id>` and reported no conversation found.

This change preserves the terminal's agent identity across foreign hook activity and captures OpenCode's native session ID so OpenCode conversations can resume after terminal loss.

- Managed hook commands identify their source with `SUPERSET_HOOK_HARNESS` instead of overwriting the wrapper's `SUPERSET_AGENT_ID`. Wrappers preserve an existing identity and skip nested launch reports; hook adapters drop foreign-harness events. Wrapper-less Cursor replay is also filtered using Cursor's environment markers.
- Cursor's native `postToolUse` and `postToolUseFailure` hooks restore Start after a Shell/MCP permission event completes. This incorporates @owieschon's Cursor work from #7252 and supersedes that PR.
- OpenCode's plugin sends its native session ID and explicit harness marker, rejects missing or unverified session events, and prevents child or competing-root events from replacing the active root. It retains legacy permission-hook support and captures resumed sessions without requiring `session.created`. The root session stays bound while idle, so unrelated session creation/deletion cannot replace or end its resume binding; a new busy root can still switch conversations.
- Provisioning markers are bumped so existing installations receive the updated wrappers/hooks, including OpenCode plugin v10. Kimi's managed-block matcher still replaces the old command format.

## How I tested it

**Real desktop app, driven through CDP:** both Cursor runs used this worktree's renderer on port 6545, CDP 9226, the expected signed-in development organization, and real keyboard/paste/mouse input. Host SQLite bindings and hook logs were read as supporting evidence; no synthetic lifecycle events were injected into the live app.

| Journey | Before | After |
| --- | --- | --- |
| Cursor 2026.09.02: launch, prompt a shell command, approve, kill the terminal shell, switch tabs and return | On parent commit `4fff6157b2`, `postToolUse` relabeled the binding to Claude while preserving Cursor session `01753eba-…`. Remount ran `claude --resume` with that ID and failed with “No conversation found.” | Binding retained `cursor-agent` and session `43d60bf7-…`. Remount ran `cursor-agent --resume` with that ID and restored the original conversation. PermissionRequest → Start also retained identity. |
| OpenCode 1.18.29: launch from preset, submit a no-tools prompt, kill the terminal shell, switch tabs and return | The old plugin omitted the session ID; previously observed bindings had no resume ID. | Start → Stop recorded `ses_f8158c475ffegEXjg1Qm6p4cNM`. Remount created a new OpenCode terminal with the same session ID and displayed the original prompt and answer. |

Screenshots and binding records were captured for the failing Cursor resume and both successful restores. The Cursor continuation does not claim a clean observed Stop transition: additional tool requests were declined before terminal-loss testing. The original report of a top-level OpenCode terminal becoming Claude remains unconfirmed.

**Automated checks:**

- 74 tests pass in `agent-wrappers.test.ts`; 34 pass in `notify-hook.test.ts` (108 total), run separately.
- New OpenCode regressions failed before implementation and pass afterward: lifecycle session IDs, child/competing-root filtering, missing-session rejection, resumed/subsequent sessions, wrapper-less attribution, and nested OpenCode event rejection.
- Final review added an idle-root regression: unrelated creation, permission, and deletion events after Stop caused incorrect lifecycle notifications before the fix and are now ignored. All 74 wrapper tests pass with that additional case.
- Agent-setup typecheck and Biome checks for changed TypeScript files pass.
- `git diff --check` and `bun run check:plugins` pass.

## Review findings addressed

Final verification at `765c24e405`: all five review threads are resolved, CodeRabbit completed with no concrete current-head merge-blocking risk, and required Test/Lint/Typecheck/Build checks pass. GitHub still requires an approving review before merge.

- [Competing root creation before busy](https://github.com/superset-sh/superset/pull/7263#discussion_r3953785457): select the root before SessionStart dispatch and test two competing session.created events before busy. The root also stays bound after Stop so unrelated idle-time creation/deletion cannot corrupt resume.
- [Legacy permission without a session ID](https://github.com/superset-sh/superset/pull/7263#discussion_r3953785462): reject the notification when neither input nor tracked root supplies an ID. Verified-root fallback still works for older permission hooks.
- [Concurrent deletion and replacement](https://github.com/superset-sh/superset/pull/7263#discussion_r3953835923): serialize plugin events and legacy permission callbacks so a pending SessionEnd cannot discard the next root. A regression holds SessionEnd in flight while creating the replacement and requesting permission; it fails before the fix and passes afterward.
- Ownership design now requires atomic validation and commit before ordered lifecycle effects, and explicitly preserves the verified-root fallback for legacy OpenCode permissions.
- All 74 wrapper tests, package typecheck, changed-file Biome checks, plugin validation, and diff checks pass after the final fixes.

## Follow-ups outside this PR

Proposed design: [Terminal agent session ownership](https://github.com/superset-sh/superset/blob/fix-agent-identification/plans/agent-session-ownership.md). It specifies the host acceptance model, compatibility rollout, and verification matrix; implementation is deferred.

**Host ownership validation:** an inherited terminal ID identifies a destination, but does not prove that the reporting process owns its top-level agent session. The host currently accepts hook identity/session replacements. A future check should tie the binding to a host-owned launch generation and corroborating process evidence, and validate events before both lifecycle broadcast and persistence. It must distinguish nested child processes from legitimate sequential top-level launches in the same shell.

That distinction also covers **same-harness nesting**: a standalone child using the same harness may emit its own session ID without a native subagent marker. An isolated synthetic check confirms the generic script still accepts that shape; it is not a live nested-CLI reproduction. Native subagents carrying explicit subagent identity already use a separate path.

**Missing agent configuration:** a detected built-in agent with no saved host agent config still reports resume unsupported. The fallback is unchanged.

Source comparison informed these boundaries: Orca preserves active parent identity against foreign hooks; cmux validates hook delivery against live process evidence; ccmux rechecks and repairs session ownership. These comparisons were code reviews, not end-to-end tests of those applications.

## Checklist

- [x] PR title follows conventional commits
- [x] Package tests/typecheck, changed-file Biome checks, plugin validation, and diff checks pass
- [x] Repository-wide test, lint, typecheck, and build pass on `765c24e405`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent identity handling during nested launches, reducing duplicate or misattributed activity.
  * Prevented unrelated, replayed, or mismatched hook events from being processed.
  * Refined Cursor tool-event handling to focus on Shell and MCP activity.
  * Improved OpenCode session tracking and serialized lifecycle notifications and permission handling.
  * Enhanced compatibility when upgrading or replacing managed hook configurations.

* **Maintenance**
  * Updated integration hook definitions and markers across supported agents.

* **Documentation**
  * Clarified agent session ownership and hook-event handling requirements.

### DIFF
--- a/packages/agent-setup/src/agent-wrappers-claude-codex-opencode.ts
+++ b/packages/agent-setup/src/agent-wrappers-claude-codex-opencode.ts
@@ -24,7 +24,7 @@ import { getOpenCodeConfigDir, getOpenCodePluginDir } from "./paths";
 export const OPENCODE_PLUGIN_FILE = "superset-notify.js";
 
 const OPENCODE_PLUGIN_SIGNATURE = "// Superset opencode plugin";
-const OPENCODE_PLUGIN_VERSION = "v9";
+const OPENCODE_PLUGIN_VERSION = "v10";
 export const OPENCODE_PLUGIN_MARKER = `${OPENCODE_PLUGIN_SIGNATURE} ${OPENCODE_PLUGIN_VERSION}`;
 
 /**
--- a/packages/agent-setup/src/agent-wrappers-common.ts
+++ b/packages/agent-setup/src/agent-wrappers-common.ts
@@ -4,7 +4,7 @@ import { SUPERSET_MANAGED_BINARIES } from "./agent-setup-targets";
 import { NOTIFY_SCRIPT_NAME } from "./notify-hook";
 import { getBinDir } from "./paths";
 
-export const WRAPPER_MARKER = "# Superset agent-wrapper v4";
+export const WRAPPER_MARKER = "# Superset agent-wrapper v5";
 export { SUPERSET_MANAGED_BINARIES };
 
 /** Path (under SUPERSET_HOME_DIR) of the runtime notify hook script. */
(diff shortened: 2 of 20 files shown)


owieschon (approved):
Validated against my Cursor identity captures behind #7252 — cross-harness fix matches. Same-harness child overwrite still reproduces; out of scope here, fine as follow-up. LGTM.

PR #7274, opened 2026-09-08
feat(mcp): agents_presets add/edit/remove

owieschon:
## What & why

I could list and launch host agent presets over MCP, but I couldn't author one — still Settings → Agents by hand. That gap hits anyone driving agents from MCP (and anyone stacking on #7269 CLI presets): you can run a preset you can't create remotely.

This closes it with `agents_presets_add|edit|remove` on the same `settings.agentConfigs.*` / `hostAgentConfigs` path as the CLI. `agents_create` stays launch; CRUD is `agents_presets_*`. Verb `edit` matches CLI (host procedure still `update`). Launch shape: structured `command`+`args` primary, or `launchCommand` (tokenized); dual-reject. `tokenizeAgentCommand` from `@superset/shared`. Out of scope: reorder/restoreDefault/iconId/resumeArgs/forkArgs/SDK.

Also carries the #7276 agent-setup Test isolation cherry-pick so owned Test means what it says on this tip.

## How I tested it

Tip: `19387b99c644d81d85864acca2a560b28d8ad1f9` (= `dfe48dbdf` + #7276 isolation cherry-pick).

- `bun test` in `packages/mcp`: 20/20 (mock `hostServiceCall` — structured/`launchCommand` add, dual-reject, empty binary/patch, pair replace, remove id echo). Typecheck (mcp + host-service), shared suite, biome, sherif.
- Tip-native live receipt PASS 12/12 @ tip above: real MCP Client → handlers → relay → host `447e38a1…`; add/edit/remove round-trip; dual-shape + empty-patch reject; cleaned after.
- Owned fork CI @ exact tip, Test SUCCESS: https://github.com/owieschon/superset/actions/runs/34178960314. Lint/Typecheck/Sherif/Build/Build CLI/Version Sync/Vale green. Neon/translations = secret allowlist.
- Upstream Actions rollup is thin (CodeRabbit) — environment fact, not claimed as green.

## Checklist

- [x] MCP-only (+ trivial `tokenizeAgentCommand` export in `@superset/shared`); no CLI / host-service / dock / SDK CRUD
- [x] Same host procedures; no second store
- [x] Unit tests + typecheck/lint/sherif
- [x] Tip-native live-host receipt PASS @ tip SHA above
- [x] Fork CI exact-commit observed

Approach issue: https://github.com/superset-sh/superset/issues/7370

PR #7276, opened 2026-09-08
fix(agent-setup): isolate disabled-hooks SUPERSET_HOME_DIR

owieschon:
## What & why

I kept seeing agent-setup Test go red for no real reason. Same tip: sometimes green, sometimes `disabled-agent-hooks` expects `[]` and gets leftover `["claude", "codex"]` because bun shared a hooks state file across tests.

That blocked honest green on stacked hooks/MCP work (#7274), and anyone reading CI couldn't tell flake from a break. If you run `packages/agent-setup` tests, you can hit it.

This is test-only isolation (per-test home + clear at `getAgentHooksStateFilePath()`, same shape as `disabled-skills`) so that Test means what it says. No product change.

## How I tested it

Tip: `2f484bc27d2af5782ebcbc4a3a69f8188f8f4a26`

Race = unrestored `mock.module("./paths", …)` in `agent-wrappers.test.ts` + shared process file order. Fix = per-test `mkdtemp` home + `beforeEach` unlink/mkdir at `getAgentHooksStateFilePath()` (same shape as `disabled-skills.test.ts`).

Forced-order local probe (preload installs the same `./paths` mock agent-wrappers leaves behind):

```
bun test --preload <paths-mock> src/disabled-agent-hooks.test.ts
# pre-fix HEAD~1: 3 pass, 1 fail (received leftover ["claude","codex"])
# post-fix tip: 4 pass, 0 fail (same preload)
```

- Full package: `bun test` in `packages/agent-setup` → 279 pass / 0 fail
- `bun run typecheck` clean; `bunx biome check` clean
- Owned fork CI @ exact tip, Test SUCCESS: https://github.com/owieschon/superset/actions/runs/34176864239

## Checklist

- [x] Test-only change; no product behaviour touched
- [x] Owned CI green at tip `2f484bc27`
- [x] Failure reproduced before the fix, confirmed gone after
- [x] Typecheck and lint clean

### DIFF
--- a/packages/agent-setup/src/disabled-agent-hooks.test.ts
+++ b/packages/agent-setup/src/disabled-agent-hooks.test.ts
@@ -9,13 +9,28 @@ import {
 	writeSharedDisabledAgentIds,
 } from "./disabled-agent-hooks";
 
-const TEST_HOME = fs.mkdtempSync(path.join(os.tmpdir(), "superset-hooks-"));
 const ORIGINAL_HOME_DIR = process.env.SUPERSET_HOME_DIR;
 const ORIGINAL_DISABLED = process.env.SUPERSET_DISABLED_AGENT_HOOKS;
 
+let testHome: string;
+
 beforeEach(() => {
-	process.env.SUPERSET_HOME_DIR = TEST_HOME;
+	// A fresh directory per test (rather than wiping/reusing one shared dir)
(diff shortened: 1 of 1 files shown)


owieschon:
Closing: this test-isolation change is already included, byte for byte, in #7274 (`packages/agent-setup/src/disabled-agent-hooks.test.ts`). One PR is enough.

PR #7296, opened 2026-09-08
DRAFT docs(recipe): Needs review is not shipped without a tip-native proof

owieschon:
Draft. Do not merge; do not undraft.

## What

Adds one docs page: `docs/proof-gate-needs-review-receipt.md`. It states an operator rule — a workspace in the **Needs review** board column is not shipped until there is a tip-native proof artifact.

Tip-native proof = the pipeline produced it and it is pinned to the exact commit SHA at the branch tip: check runs that CI actually ran at that SHA (`gh checks <sha>`, all green), or a receipt that names the SHA.

The rule adds no board column. It cites only the columns the app already has (Needs review, Needs attention, Merged, Deleted) and leaves `agentStatus` untouched — that stays the single source of truth in its own work (#7007).

## Why — first-person stakes

I am the operator who merges. If I read "Needs review" as "ready", I ship whatever the agent left: possibly failing CI, possibly a commit I never verified. A bot saying "looks good" is an opinion. A check run pinned to a concrete SHA is proof, because I can read it back and pin it to the same commit I will merge.

## Not in scope

No product code, no new columns, no `agentStatus` changes, no hook changes.

---
## Summary by cubic
Fixes premature completion announcements by deferring a parent's Stop until all its subagents finish. Adds a docs page stating that a workspace in "Needs review" requires a tip-native proof before shipping.

**Behavior changes**
- Parent Stop is held while subagents run and released exactly once when the last child clears.
- Child events that straggle after a stop no longer revive the child, except for Codex, which genuinely resumes via `send_input`.
- Permission requests still broadcast while children run; a new parent event supersedes the held Stop.
- Held Stops are dropped if the terminal dies or the status escape hatch fires.

<sup>Written for commit 8c61d203116feb7ef18ee9d60422e58e3f1ded7e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #7007, opened 2026-08-29
feat(sdk): expose terminal agent lifecycle status

toddagent01:
## FOR TODD — 1 decision, 0 actions

**Decision:** expose the existing normalized lifecycle record on `terminals.list`, rather than add a second status endpoint.

I recommend this because the terminal is already the public CLI/SDK object consumers use to read and send messages, while the host service remains the single owner of agent-state normalization.

## What & why

Superset already knows when Claude, Codex, and other supported agents are working or waiting for attention, and the workspace board renders that state. CLI and SDK consumers could only see that the PTY existed, though, so they could not distinguish a working agent from one waiting for input.

This adds an optional `agentStatus` field to each live terminal returned by `terminal.list` / `terminals.list()`. It exposes agent/session identity, timestamps, and the latest normalized lifecycle event. In particular, `PermissionRequest` gives integrations the same needs-attention signal used by the workspace board.

The response is joined from the existing `TerminalAgentStore`; this does not introduce another detector or status model. The CLI and SDK reference docs describe the additive field.

Downstream context: [todd-studio/workstation#252](https://github.com/todd-studio/workstation/issues/252)

## How I tested it

- `bun test test/integration/terminal.integration.test.ts --test-name-pattern "createSession uses the configured shell and list includes live agent status"` — passed
- Full terminal integration file — 8 passed; the process-group disposal test timed out once and passed when rerun alone
- `bun run typecheck` in `packages/host-service` — passed
- `bun run typecheck` in `packages/sdk` — passed
- `bun run build` in `packages/sdk` — passed
- `bun run lint` — passed
- `bun run typecheck` — passed
- `bun run test` — not fully green: two dynamic-import tests in the untouched `@superset/agent-setup` package could not find their shared temporary module after another suite removed it; the same failures remained under the repository-pinned Bun 1.3.14

## Checklist

- [x] PR title follows conventional commits (`type(scope): subject`)
- [x] `bun run lint` and `bun run typecheck` pass (CI fails on lint warnings too)
- [x] "Allow edits from maintainers" is checked on fork PRs

---
## Summary by cubic
Exposes the terminal agent's lifecycle status on `terminals.list` responses, so SDK and CLI consumers can tell when a supported agent is working or waiting for input instead of only knowing the PTY exists.

- Adds an optional `agentStatus` field with agent identity, timestamps, and the latest normalized lifecycle event, populated only when listing a specific workspace; workspace-wide listings still return just PTY presence.
[...]

PR #7297, opened 2026-09-08
docs(recipe): Needs review is not shipped without tip-native proof

owieschon:
## What & why

Adds one docs page: `docs/proof-gate-needs-review-receipt.md`.

**Needs review** on the board is not shipped. Tip-native proof (check runs or a receipt pinned to the exact branch-tip SHA) is required first. Without that bar, Operators treat chrome as done and the product keeps lying about finished work.

## How I tested

- Docs-only diff vs `main` (single file)
- Stay **DRAFT** — no undraft without CoS→Owen

## Checklist

- [x] Thin docs brick only (no host-service / product code)
- [x] Attested board nouns; no invented ReadyVerified column
- [x] Fence Todd #7007 agentStatus SoT; no hooks #7252/#7263

Supersedes contaminated draft #7296 (same docs intent; #7296 head accidentally included terminal-agents files).

---
## Summary by cubic
Stops a parent terminal from announcing completion while its subagents are still running, and adds a docs rule that Needs review on the board is not shipped without tip-native proof.

**Bug Fixes**
- A parent Stop now waits until the last child exits; previously it flipped the terminal green while children still worked.
- A stopped child no longer revives when its own events straggle in after the stop, unless the harness opts in to revival (Codex `send_input`).
- Deferred completions fire exactly once when the roster drains, and are dropped when the parent starts a new turn or the terminal dies.
- A child that goes quiet without a stop hook now delays the completion by the stale window instead of losing it; the sidebar poll releases it.
- Permission requests still raise while children run.

**Docs**
- Adds `docs/proof-gate-needs-review-receipt.md`: Needs review means not shipped without a check run or receipt pinned to the branch-tip SHA.

<sup>Written for commit 0c964b650dce03b47b7c063dcad8582e46b1f5a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7297?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


owieschon:
Superseded: head was based on fork main (included unpublished host-service commits vs upstream). Thin replacement cut from upstream/main.

PR #7322, opened 2026-09-08
fix(desktop): say 'Stopped' not 'Complete' in agent lifecycle notifications

owieschon:
## What & why

Native lifecycle notifications used “Complete” for every non-Failed `Stop`, including interrupts. That title claims the agent finished when it only stopped.

Maps those events to “Stopped.” Failed stays Failed (#7294). PermissionRequest stays Needs Attention. No board/status detection change.

## How I tested it

- Tip: `a7ceb279a9d297a7dbcd830e95951f29f9742887`
- `bun test` on `notificationContent.test.ts` (string + branch update)
- Scope: `notificationContent.ts` + test only

## Checklist

- [x] Non-failed Stop title = Stopped
- [x] Failed path unchanged (#7294)
- [x] Draft

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/components/V2NotificationController/lib/notificationContent.test.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/components/V2NotificationController/lib/notificationContent.test.ts
@@ -23,7 +23,7 @@ describe("getV2NativeNotificationContent", () => {
 				}),
 			}),
 		).toEqual({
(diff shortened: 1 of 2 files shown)


owieschon:
Closing in favor of #7294. Both edit the same title expression; #7294 adds the `Failed` case, which is the correctness fix, and leaves the existing `Complete` wording alone. The rename here was an opinion, not a bug.

PR #7326, opened 2026-09-08
fix(desktop): keep account names out of host-service crash tails

saddlepaddle:
## Problem

The host-service crash capture attaches the child's last 16 KB of output as an
`outputTail` extra. That output is mostly absolute paths under the user's home
directory — worktree paths, log paths, socket paths — so every crash report
carries the machine's account name, which on macOS is usually a person's real
name. In a 75-event sample read back from the desktop project, one account name
appeared 274 times in a single issue's tails.

The desktop SDK runs with `sendDefaultPii: false`. That governs SDK-collected
PII, not free-form extras, so this capture quietly sends what the SDK config
says it does not. Sentry's default `@userpath` scrubber does not catch it
either: the names are present verbatim in the stored events.

## Root cause

`handleChildExit` redacts the secrets it handed the child and nothing else. The
`redactions` array is exactly the right mechanism — it just never had the home
directory in it.

## Fix

Redact the home directory at the same capture site, via a `redactCrashTail`
helper extracted into the (electron-free) `host-service-utils` sibling so it is
testable without stubbing electron or the lazy `@sentry/electron/main` import.

Two deliberate constraints:

- **`~`, not removal.** Which worktree the child died under is part of the
  diagnosis; `~/.superset/worktrees/<id>/<name>` keeps that and drops only the
  name.
- **Exact strings, never patterns.** The tail is the only evidence a hard kill
  (SIGSEGV, OOM) leaves, so a regex that reached past what it named would redact
  the diagnosis along with the account name. The helper substitutes only values
  we already hold — the known secrets and `os.homedir()`.

Negative tests were written first and confirmed failing before the change. They
pin the two inputs a redactor must *not* touch:

- a root home directory (`/`, `C:\`, `""`) — it prefixes nearly every absolute
  path in the tail, so substituting one would erase the report. `path.dirname(h)
  !== h` rejects roots.
- an empty secret — `"".split("")` splits between every character, which would
  leave nothing but `[redacted]` separators. This hazard pre-dates the change;
  the old inline `reduce` was guarded only by a `.filter(Boolean)` at the two
  construction sites.

A third test asserts the fatal-error line, native frames, `/Applications/...`
and `/usr/lib/dyld` all survive redaction.

Behaviour is otherwise unchanged: same message, same tags, same extras, same
`[redacted]` placeholder for secrets. The capture is not silenced or downgraded.

## Scope

This is only the PII half of the DESKTOP-H1 triage. **The `[Filtered]` premise
that opened that ticket does not hold and no code here addresses it** — see the
note below. The crash itself (host-service V8 heap OOM) is being fixed
separately and deliberately not touched here.

### On `outputTail` arriving as `[Filtered]`

Worth recording, because it was the original request and the answer is "don't":

- It is not the norm. Across 75 events read back from the API: DESKTOP-H1
  **50/50 full tails, 0 filtered**; DESKTOP-HX 15/20 full; SIGKILL 2/3; SIGSEGV
  and SIGTRAP 1/1 each.
- Sentry's own `_meta` names the rule — `@password:filter`, whole-value
  substitution. Project config is stock (`dataScrubberDefaults: true`,
  `safeFields: []`, `sensitiveFields: []`, `relayPiiConfig: null`).
- It is content-triggered, not field-name-triggered: the same release appears on
  both sides, and 0 of 69 unfiltered tails contain any of Sentry's password-like
  key words. So renaming the field or splitting it into sub-fields would not
  reliably dodge it.
- Adding `outputTail` to the project's `safeFields` would work, and should
  **not** be done — that switch turns scrubbing off for the field, and this PR
  exists because the field carries personal data. Fixing it at the source is
  strictly better anyway: server-side scrubbing happens on ingest, after the
  names have already left the user's machine.

Note this cannot repair already-stored events, which retain the names.

## Verification

`bunx biome check` on the three changed files (after one `--write` for
formatting):

```
Checked 3 files in 16ms. No fixes applied.
```

`cd apps/desktop && bun run typecheck` — clean:

```
$ bun run generate:icons && bun run generate:routes
$ tsr generate
$ tsc --noEmit
```

`cd apps/desktop && bun test` (full package suite):

```
 3676 pass
 1 fail
 1 error
 2 snapshots, 59385 expect() calls
Ran 3677 tests across 398 files. [24.51s]
```

The one failure is pre-existing and unrelated —
`useSidebarDnd.test.ts`, `TypeError: React.createContext is not a function` from
`@dnd-kit/core`. It passes in isolation (8 pass), so it is full-suite ordering,
not a file-level break. Re-running the **full suite** with this diff fully
reverted (clean tree) reproduces it exactly:

```
 3669 pass
 1 fail
 1 error
 2 snapshots, 59370 expect() calls
Ran 3670 tests across 397 files. [23.77s]
```

Same file, same `React.createContext` frame. The delta is exactly this PR's
additions — +7 tests, +1 file, +15 expect() calls — with the failure count
unchanged.

New tests alone:

```
 7 pass
 0 fail
 15 expect() calls
Ran 7 tests across 1 file. [98.00ms]
```

Refs DESKTOP-H1
Refs DESKTOP-HX

https://claude.ai/code/session_01XtMkp7dcyp7JzCy6PgyMih

---
## Summary by cubic
Host-service crash tails previously included absolute paths under the user's home directory, exposing macOS account names to Sentry despite `sendDefaultPii: false`. They now replace the user's home directory with `~` at capture time while preserving secret redaction and diagnostic paths; stored events are not changed.

- Matches only the exact home directory at a path-segment boundary, including Windows separators, so sibling paths remain intact.
- Ignores filesystem-root home directories and empty secrets to avoid over-redacting the tail.
- Adds tests for boundary matching, literal special characters, combined redaction, and diagnostic preservation.
- Covers the PII portion of DESKTOP-H1; it does not change Sentry's `[Filtered]` behavior.

<sup>Written for commit 058cd6ee833750e7c9f3d05920212b2b1984eee4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Crash reports now better protect sensitive information by redacting known secrets and replacing the user’s home directory with `~`.
  - Path handling preserves crash diagnostics, avoids altering filesystem root paths, and prevents unintended redaction of similarly named sibling paths.
  - Empty secrets are safely ignored during sanitization, including across supported path separator styles.

- **Tests**
  - Added coverage for secret redaction, home-directory replacement, root-directory handling, path boundaries, unchanged output, and combined sanitization scenarios.

### DIFF
--- a/apps/desktop/src/main/lib/host-service-coordinator.ts
+++ b/apps/desktop/src/main/lib/host-service-coordinator.ts
@@ -34,6 +34,7 @@ import {
 	MAX_HOST_LOG_BYTES,
 	openRotatingLogFd,
 	pollHealthCheck,
+	redactCrashTail,
 } from "./host-service-utils";
 import { localDb } from "./local-db";
 import { HOOK_PROTOCOL_VERSION } from "./terminal/env";
@@ -1092,10 +1093,10 @@ export class HostServiceCoordinator extends EventEmitter {
 							pid: childPid,
 							version: app.getVersion(),
 							uptimeMs: Date.now() - current.spawnedAt,
-							outputTail: current.redactions.reduce(
-								(tail, secret) => tail.split(secret).join("[redacted]"),
-								current.outputTail,
-							),
+							outputTail: redactCrashTail(current.outputTail, {
+								secrets: current.redactions,
+								homeDir: os.homedir(),
+							}),
 						},
 					}),
 				)
--- a/apps/desktop/src/main/lib/host-service-utils.test.ts
+++ b/apps/desktop/src/main/lib/host-service-utils.test.ts
@@ -56,6 +56,61 @@ describe("redactCrashTail", () => {
 				expect(redactCrashTail(CRASH_TAIL, { homeDir: root })).toBe(CRASH_TAIL);
 			}
 		});
+
+		test("substitutes a home directory that is the whole path", () => {
+			expect(redactCrashTail("/Users/ada", { homeDir: "/Users/ada" })).toBe(
+				"~",
+			);
+		});
+
+		// `/Users/ada` is a literal substring of `/Users/adam`, but that is a
+		// different account's directory — outside this home, and none of our
+		// business. `/Users/ada-old` from a migration is the same shape.
+		test("leaves a sibling path that merely starts with the home dir intact", () => {
+			for (const sibling of [
+				"/Users/adam/project",
+				"/Users/ada-old/project",
+				"/Users/adam",
+			]) {
+				expect(redactCrashTail(sibling, { homeDir: "/Users/ada" })).toBe(
+					sibling,
+				);
+			}
+		});
+
+		test("substitutes the home dir but not its sibling in the same tail", () => {
+			expect(
+				redactCrashTail("/Users/ada/mine and /Users/adam/theirs", {
+					homeDir: "/Users/ada",
+				}),
+			).toBe("~/mine and /Users/adam/theirs");
+		});
+
+		// Windows: os.homedir() is `C:\Users\...`, whose separators are regex
+		// metacharacters, and paths beneath it get logged with either separator.
+		test("handles a Windows home directory and both separators", () => {
+			const homeDir = "C:\\Users\\ada";
+
+			expect(redactCrashTail("C:\\Users\\ada\\project", { homeDir })).toBe(
+				"~\\project",
+			);
+			expect(redactCrashTail("C:\\Users\\ada/project", { homeDir })).toBe(
+				"~/project",
+			);
+			expect(redactCrashTail("C:\\Users\\adam\\project", { homeDir })).toBe(
+				"C:\\Users\\adam\\project",
+			);
+		});
+
+		// A HOME pointing somewhere odd must not turn into a wildcard.
+		test("treats regex metacharacters in the home dir as literal", () => {
+			expect(
+				redactCrashTail("/Users/axb/project", { homeDir: "/Users/a.b" }),
+			).toBe("/Users/axb/project");
+			expect(
+				redactCrashTail("/Users/a.b/project", { homeDir: "/Users/a.b" }),
+			).toBe("~/project");
+		});
 	});
 
 	describe("secrets", () => {
--- a/apps/desktop/src/main/lib/host-service-utils.test.ts
+++ b/apps/desktop/src/main/lib/host-service-utils.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, test } from "bun:test";
+import { redactCrashTail } from "./host-service-utils";
+
+/**
+ * A tail shaped like the ones host-service crashes actually attach: worktree
+ * paths under the user's home, then the V8 fatal error and its native frames.
+ */
+const CRASH_TAIL = [
+	"[host-service:pull-request-runtime] Worktree missing on disk; pausing branch sync until it reappears {",
+	"  worktreePath: '/Users/ada/.superset/worktrees/de7ee5cf/rift-menu'",
+	"}",
+	"[workspace-fs/watch] nested-repo scan hit cap { absolutePath: '/Users/ada/Desktop/flow', found: 3 }",
+	"",
+	"<--- Last few GCs --->",
+	"",
+	"FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory",
+	"----- Native stack trace -----",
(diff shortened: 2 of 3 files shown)


saddlepaddle:
Fixed in 058cd6ee8. CodeRabbit's finding was correct: `split(homeDir).join("~")` matched the home directory as a bare substring, so with a home of `/Users/ada` the path `/Users/adam/project` became `~m/project` — corrupting a path belonging to a *different* account, outside this home, which is the exact property the helper exists to protect.

**Fix.** Substitute only where the home directory ends the path or is followed by a separator, with the literal escaped before it reaches the regex:

```ts
const PATH_SEGMENT_END = "(?=$|[/\\\\])";

function escapeRegExp(literal: string): string {
	return literal.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
}

redacted = redacted.replace(
	new RegExp(`${escapeRegExp(homeDir)}${PATH_SEGMENT_END}`, "g"),
	"~",
);
```

Still the same exact literal — the lookahead can only *refuse* matches mid-segment, never admit new ones, so there is no drift toward matching `/Users/<anything>`.

**Separators: both `/` and `\`.** `os.homedir()` is `C:\Users\...` on Windows, and paths beneath it get logged with either separator there. Accepting both only narrows the match further, so it costs nothing. The backslashes in a Windows home are themselves regex metacharacters, which is the second reason the escape is not optional.

**Secrets deliberately unchanged** — still plain `split`/`join`. A secret is an opaque token with no boundary semantics; asserting one would make it miss.

**Tests.** Five added, and I confirmed all three sibling cases fail against the old implementation first, reproducing the reported corruption verbatim:

```
error: expect(received).toBe(expected)
Expected: "/Users/adam/project"
Received: "~m/project"
(fail) redactCrashTail > home directory > leaves a sibling path that merely starts with the home dir intact

error: expect(received).toBe(expected)
Expected: "~/mine and /Users/adam/theirs"
Received: "~/mine and ~m/theirs"
(fail) redactCrashTail > home directory > substitutes the home dir but not its sibling in the same tail

error: expect(received).toBe(expected)
Expected: "C:\Users\adam\project"
Received: "~m\project"
(fail) redactCrashTail > home directory > handles a Windows home directory and both separators
 9 pass
 3 fail
```

They assert the sibling survives **intact** (`/Users/adam/project` still reads `/Users/adam/project`), not merely that the output differs. Also covered: the exact-match case (bare `/Users/ada` → `~`), `/Users/ada-old`, a mixed-separator Windows path, and a home containing a regex metacharacter (`/Users/a.b` must not match `/Users/axb`). The existing "keeps the diagnosis" test passes unchanged.

---

**One measurement worth your call before merge.** I checked what actually follows the home directory in 401 real occurrences across the 69 unfiltered tails I pulled from the desktop project:

| next character | count |
|---|---|
| `/` (separator) | 342 |
| `'` (closing quote) | 59 |
| name character (would corrupt) | 0 |

The 59 are logged as `absolutePath: '/Users/ada'` — the home directory with nothing after it *in the path*, terminated by a quote rather than by end-of-input. This boundary does not redact those, so ~15% of occurrences still carry the account name.

I did **not** widen it to treat a quote as a boundary, because a quote is a legal character in a macOS path segment, so that would be a heuristic about filename charsets rather than a structural assertion — the drift you ruled out. The current boundary is safe by construction; the quote-aware one would only be safe in practice.

Two options if the 15% matters: leave it (a miss leaks a name, an over-match corrupts evidence — the asymmetry favours strictness), or add `'` and `"` to the lookahead and accept a documented heuristic. My recommendation is to leave it as-is and land this; the 85% is the part that repeats hundreds of times per report, and the remaining shape is a single bare path per occurrence.

**Verification after the change** — `bunx biome check` on the three files clean after one `--write`; `bun run typecheck` clean; full desktop suite:

```
 3681 pass
 1 fail
 1 error
 2 snapshots, 59395 expect() calls
Ran 3682 tests across 398 files. [30.36s]
```

Same single pre-existing failure as before (`useSidebarDnd.test.ts`, `React.createContext` from `@dnd-kit/core`), already proven on a clean tree at 3669 pass / 1 fail. Delta is exactly the 5 new tests.

PR #7335, opened 2026-09-08
fix(desktop): stop trpc-electron reading a disposed frame on navigation

saddlepaddle:
## What

`trpc-electron@0.1.2` attaches a `did-start-navigation` listener to every window's
webContents and reads `frame.routingId` unconditionally to scope its subscription
cleanup. Patch it to skip the cleanup when that frame is missing or already destroyed.

## Root cause

Electron types `details.frame` on `did-start-navigation` as `WebFrameMain | null` —
"May be `null` if accessed after the frame has either navigated or been destroyed"
(`electron.d.ts`, `WebContentsDidStartNavigationEventParams`) — and reading *any*
property of a `WebFrameMain` whose render frame has been disposed throws
`Render frame was disposed before WebFrameMain could be accessed`.

trpc-electron does neither check:

```js
win.webContents.on('did-start-navigation', ({ isSameDocument, frame }) => {
  if (!isSameDocument) {
    this.#cleanUpSubscriptions({ webContentsId, frameRoutingId: frame.routingId });
  }
});
```

So a frame torn down as its navigation starts throws out of `webContents.emit` in the
main process. Two consequences: an unhandled `fatal` in Sentry, and **no** subscription
is aborted for that navigation — the throw happens while building the argument object,
before `#cleanUpSubscriptions` is called.

The library was written against Electron 25, where `frame` was non-nullable
(`electron: ^25.1.1` in its devDependencies), and is unmaintained — 0.1.2 is the latest
release, published 2025-01-06. There is no version to upgrade to.

## Evidence

- **DESKTOP-16T** (2026-09-08, 1.27.0-canary, Electron 41.10.3) and **DESKTOP-K8**
  (2026-08-06, 1.18.1, Electron 40.8.5) are the same error, both terminating at
  `trpc-electron/dist/main.mjs:667:25` on the source line `frameRoutingId: n.routingId`.
  K8's culprit is `menu.ts:72` — a main-frame `reloadIgnoringCache()` from Force Reload.
- Verified against the real Electron 41.10.3 binary (the runtime from the reports) that a
  disposed `WebFrameMain` answers `isDestroyed() === true` **without throwing**, while
  `routingId` on the same object throws verbatim
  `Error: Render frame was disposed before WebFrameMain could be accessed`. That is what
  makes `isDestroyed()` the correct predicate here rather than a try/catch.
- The production stack traces name `dist/main.mjs`, so the ESM bundle is what ships;
  `dist/main.cjs` is patched identically so the fix cannot silently disappear if the
  main-process build ever resolves the `require` condition.

## Why this shape

Skipping the cleanup is not a behaviour change: today the throw means no cleanup runs for
that navigation either. When the frame is gone there is nothing left to scope the abort
to — a disposed *sub*frame must not abort the main frame's subscriptions — and those
subscriptions are already reaped by the handler's own `destroyed` listener when the
webContents goes away.

## Blast radius / signal

Small and honest: 2 events in 90 days, 0 users impacted, and the app does not crash —
`apps/desktop/src/main/index.ts:334` installs a `uncaughtException` handler that logs and
continues. What the patch buys is a recurring unhandled main-process exception removed at
its cause, permanently: DESKTOP-K8 was archived on 2026-08-06 and the identical error came
back today as a *new* issue, because grouping keys on the calling frame. Archiving does
not hold for this one.

I could not force the race locally in three attempts (iframe removed mid-navigation,
reload storms with slow unload handlers, superseded navigations) — consistent with two
events in 90 days. The production stack traces are the reproduction.

## Not in scope

`getInternalId` in the same file reads `event.senderFrame.routingId`, which is nullable
for the same reason. No Sentry evidence for that path, and handling it means deciding
whether to drop an IPC message — a behaviour change. Left alone deliberately.

## Per patches/README.md

New patch entries need a CI guard test, because `patchedDependencies` is keyed to an exact
version and a bump would silently drop the fix while everything still builds.
`apps/desktop/src/trpc-electron-frame-patch.test.ts` asserts the liveness check sits
between the listener and the `routingId` read in both bundles; verified it goes red
against the unpatched text.

Sentry: DESKTOP-16T — https://superset-sh.sentry.io/issues/7720030372/

https://claude.ai/code/session_01WN7egRgkoq2PAfqqJNZeSj

---
## Summary by cubic
Patches `trpc-electron` to skip subscription cleanup when the navigation event's frame is missing or already destroyed. Previously, reading `frame.routingId` on a disposed frame threw "Render frame was disposed before WebFrameMain could be accessed", causing an unhandled exception in the main process and preventing cleanup. Now the listener checks `frame && !frame.isDestroyed()` before reading. This does not change behavior because the throw already skipped cleanup. Adds a guard test in `apps/desktop/src/trpc-electron-frame-patch.test.ts` that pins both the null check and the `isDestroyed()` check against the frame's own minified binding name, keeping the patch applied across version bumps.

<sup>Written for commit 6809e1beb39acbedd9caf584957176c5f61147f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved desktop navigation handling to avoid errors when a renderer frame is unavailable or has been destroyed.
  - Preserved subscription cleanup for valid frames while preventing failures during disposed-frame navigation.
  - Improved stability during rapid navigation and window or frame lifecycle changes.

- **Documentation**
  - Added documentation covering the desktop compatibility patch, its safeguards, testing, and maintenance instructions.

- **Tests**
  - Added coverage to verify the safeguard remains present in both desktop bundles.


saddlepaddle:
Addressed the CodeRabbit finding on the guard test in 6809e1beb.

The assertion only pinned `isDestroyed()`, so a regenerated patch that kept the liveness check but dropped the null check would have gone green while `frame === null` throws a TypeError — the patch guards two conditions and the test pinned one. It now reads the frame's minified binding name out of the listener's destructuring and asserts both `!<frame>` and `<frame>.isDestroyed()` sit between the listener and the `routingId` read, which also keeps the assertions working across a re-minify.

Checked it discriminates by mutating the installed bundle four ways — dropping the null check, dropping the liveness check, dropping both, and the patched text — and each dropped guard fails its own assertion.

https://claude.ai/code/session_01WN7egRgkoq2PAfqqJNZeSj

PR #7341, opened 2026-09-08
fix(automations): continue a pinned automation's agent instead of stacking a new session per run

saddlepaddle:
## Problem

A user reported that opening a workspace on mobile "duplicates the same agent a bunch of times" — 6 agent marks on the home row, then 13 forty minutes later. It is not the mobile app, and it is not reconnecting: the count grows on a wall clock, not on taps.

They had a scheduled automation (`FREQ=MINUTELY;INTERVAL=5`) **pinned to a workspace**. Every duplicate was a real PTY running a real agent, spending real tokens.

## Root cause

A workspace pin means "reuse this workspace every run" — the `--workspace` flag is documented as *"existing v2 workspace id — reuses it every run"*. Only the directory was ever reused.

- `packages/trpc/src/router/automation/dispatch.ts:249` — `runAgent(workspaceId)` fires unconditionally on every tick
- `packages/host-service/src/trpc/router/agents/agents.ts:580` — `runTerminalAgent` mints a fresh `crypto.randomUUID()` terminal on every call

A TUI agent never exits; it sits at its prompt after finishing a turn. So each tick left another live agent behind, forever:

```
schedule tick → dispatchAutomation → agents.run(workspaceId: <pinned>)
              → runTerminalAgent → new terminalId → new pty → another live agent
```

### Evidence

From `automation_runs` for the affected automation (read-only query against a branch copy):

| status | runs | with a session | **distinct sessions** |
|---|---|---|---|
| dispatched | 181 | 181 | **181** |
| skipped_offline | 23 | 0 | 0 |
| dispatch_failed | 13 | 0 | 0 |

One session per successful run — **181 of them, all in a single `v2_workspace_id`**, over 18 hours until the owner paused the automation. The counts match the two user reports exactly: **5** runs dispatched by the time of the first screenshot (plus the workspace's own original session = the 6 shown), and **13** forty minutes later.

Failed runs produced **zero** sessions, so this is not the retry-storm class of #7284 / #7280 — it is one session per *successful* run, by construction.

## Fix

`agents.run` takes an optional `continueTerminalId`. When that terminal still holds a **live binding for the same agent**, the prompt is delivered into it via `writeFramedInputToSession` — the same path `terminal.send` already uses — instead of launching beside it. Automation dispatch names the terminal its own previous run reported, and only for a pinned workspace.

The host makes the decision, in one round trip. Every "no" falls through to a normal launch rather than failing the run:

- the terminal is gone, or its agent exited (a bare shell must never be typed into — the send adopts but never respawns, so a dead pty errors instead of returning as an empty shell)
- a different agent now owns it
- launch-only options are set (`model` / `effort` / `mode` / attachments / resume / fork), which cannot be applied to a running process
- the send raced the session's death

The candidate is always the automation's **own** previous session, never "any agent of this kind in the workspace", so it can never hijack a session the user started. Two automations pinned to one workspace keep their own sessions.

**Unchanged:** create-on-attach (the desktop still inserts panes optimistically), resume, `restartAccountSessions`, and unpinned automations — those branch a fresh workspace per run, so they have no previous session to continue. An older host simply strips the unknown key and launches, exactly as before.

**No host-side cap.** Parallel agents in one workspace are legitimate and deliberate; the host is not where the unwanted repetition originates. The dispatcher is.

**Trade-off, stated plainly:** a pinned recurring automation now accumulates turns in one conversation instead of accumulating sessions. That is the right side of the trade against N concurrent processes on one worktree, and it also gives a recurring monitor the continuity it would otherwise hand-roll through a state file.

## Verification

End to end against a real host-service and pty daemon (`createProjectScenario` + `agents.run` + the real `notifications.hook` agent lifecycle hook), 4 cases:

1. **the bug** — 4 runs without the field → 4 live ptys
2. **the fix** — 6 ticks with `continueTerminalId` → exactly **1** live pty
3. an unknown terminal → falls through to a launch
4. an agent that ended (pty still alive as a bare shell) → falls through; the prompt is never typed into the shell

```
 4 pass
 0 fail
Ran 4 tests across 1 file.
```

Across three repeat runs on a loaded machine, cases 2–4 passed every time; the reproduction case flaked twice on git-fixture setup (`fatal: not a git repository` in the temp worktree), unrelated to this change.

The test is not in this commit, per the usual "no tests unprompted" rule — it is written and green, say the word and I will add it.

```
$ bunx biome check packages/host-service/src/trpc/router/agents/agents.ts packages/trpc/src/router/automation/dispatch.ts
Checked 2 files in 40ms. No fixes applied.

$ bunx tsc --noEmit -p packages/host-service/tsconfig.json && bunx tsc --noEmit -p packages/trpc/tsconfig.json
TYPECHECK OK

$ cd packages/host-service && bun test src/trpc/router/agents src/trpc/router/terminal-agents src/terminal-agents src/trpc/router/terminal
 148 pass
 0 fail
Ran 148 tests across 9 files.
```

`packages/trpc` has 4 pre-existing failures (growth sitemap, search-console, github, plugins oauth) that fail identically with `origin/main`'s `dispatch.ts` — missing env vars in the worktree, unrelated. The wider `host-service` suite timed out on `workspaceCleanup` and `host-worker-pool` under load; both pass when run alone.

## Not fixed here

The sessions already accumulated on an affected machine. Reaping them is a user action, not a migration.

https://claude.ai/code/session_01HqtqS723QGrk6esbZ3gpA5

---
## Summary by cubic
Fixes pinned automations leaving one live agent session behind per run by continuing the previous run's terminal instead of launching a new one.

**Notes**
- Applies only to pinned workspaces; unpinned automations branch fresh workspaces and are unchanged.
- Any mismatch falls back to a fresh launch — no run fails.
- Already-accumulated sessions on affected machines aren't cleaned up here.
- Older hosts ignore `continueTerminalId` and launch as before.

<sup>Written for commit fd90817ad0f07d2754aeaa140616fa5277a5399b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7341?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Automation runs can now continue an agent session already active in the relevant terminal.
  * Prompts are delivered to an existing session when compatible; otherwise, a new session starts automatically.
  * Automation dispatch can preserve continuity for runs associated with a pinned workspace.

### DIFF
--- a/packages/host-service/src/trpc/router/agents/agents.ts
+++ b/packages/host-service/src/trpc/router/agents/agents.ts
@@ -24,7 +24,10 @@ import { z } from "zod";
 import type { HostDb } from "../../../db";
 import { hostAgentConfigs, workspaces } from "../../../db/schema";
 import { hasHarnessSession } from "../../../terminal/harness-transcript";
-import { createTerminalSessionInternal } from "../../../terminal/terminal";
+import {
+	createTerminalSessionInternal,
+	writeFramedInputToSession,
+} from "../../../terminal/terminal";
 import type { HostServiceContext } from "../../../types";
 import { protectedProcedure, router } from "../../index";
 import { resolveAttachmentPath } from "../attachments/storage";
@@ -217,6 +220,13 @@ export interface AgentRunInput {
 	resumeSessionId?: string;
 	/** Session id to clone into a new provider-owned session. */
 	forkSessionId?: string;
+	/**
+	 * A terminal an earlier run of this same caller left behind. When it still
+	 * runs this agent, the prompt is delivered into that session instead of
+	 * starting a second one beside it. Anything else — terminal gone, agent
+	 * exited, a different agent now owns it — launches as usual.
+	 */
+	continueTerminalId?: string;
 }
 
 export type AgentRunResult = {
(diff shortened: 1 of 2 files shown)


saddlepaddle:
Closing unmerged — deliberate product decision, not a rejection of the diagnosis.

The root cause above stands and is worth keeping: a workspace-pinned automation reuses the directory but not the agent session, so a recurring schedule accumulates one live agent per tick, unbounded (measured: 181 dispatched runs → 181 distinct terminal sessions → 1 workspace, over 18 hours).

We're not adding a per-automation concurrency policy for this. The general answer is rate limits + spend limits as part of the move to cloud, which subsumes this case and the others like it.

Known gap that budget limits will *not* cover, for whoever designs them: a spend cap bounds the cost but not the correctness — concurrent agents on one worktree still overwrite each other's edits.

Branch kept for the analysis and the diff if we revisit.

PR #7346, opened 2026-09-08
feat(mobile): keep switched-to terminals warm with a cached buffer

saddlepaddle:
Mason (#ext-trainloop): *"Can you maintain this connection, at least for a little bit. It's frustrating to switch between these and have to wait for it to load every time."* He cycles between trainloop-core, apex-nemo-rl-h200, taskhopper and depot, and every entry pays a cold terminal load.

## What the wait actually is

Measured on device (iPhone 16e sim, prod relay, host on the same Mac — so a real phone on cellular is worse), instrumenting the page's own bridge and socket. Median of 5 cold opens:

| stage | ms | fixable by |
| --- | ---: | --- |
| WebView creation + handing over the 435 KB HTML | ~970 | keeping a WebView alive |
| HTML parse + xterm init | ~35 | — |
| `_whoowns` probe round trip | ~220 | not doing it |
| WebSocket upgrade | ~790 | keeping the socket alive |
| scrollback replay (~105 KB) | ~440 | attaching by sequence |
| **total to painted** | **~2.5–3.0s** | |

Two of those were surprises worth writing down. The 435 KB HTML string is *not* the WebView cost — swapping in a one-line page moved mount→document-start only from ~970ms to ~780ms, so ~780ms is WKWebView instantiation and nothing in our control shortens it. And the probe buys nothing on the happy path: WebKit opens a WebSocket on its own connection, so the probe's TLS work is not reused, and its only job is spotting a 403.

## What this does

The page hands RN a serialized copy of its buffer whenever the session goes quiet, paired with its position in the host's PTY stream. Reopening writes that buffer **before the socket exists** and attaches with `?seq=<epoch>:<n>`, so the host sends only the bytes that arrived while we were away. The probe moves off the happy path to the first socket that closes without ever attaching.

None of that machinery is new — `?seq=` catch-up has been in host-service since #6298 and desktop's renderer has used it (`terminal-seq-anchor.ts`) all along. The mobile page was the one client still speaking the legacy `replay=0|1` dialect. This teaches it the same contract rather than inventing a second one.

| | before | after |
| --- | ---: | ---: |
| time to content on a reopen | ~2.5–3.0s | **~0.9s** |
| bytes on the wire | ~105 KB | **~300 B** |
| time to live socket | ~2.5–3.0s | ~1.4–2.0s |

A first open improves too: `seq=new` returns the host's ring tail, where the legacy FIFO it replaces is *drained* by whichever client attaches first (`replayBuffer` empties `session.buffer`) — so the second client to attach got nothing, and the phone was often that client.

The remaining ~0.9s is almost entirely WKWebView instantiation, and roughly a third of it hides behind the push transition.

## I did not build the LRU connection pool — deliberately

The brief scoped ~5 sessions warm for ~2 minutes, and invited me to say so if the pool was the wrong shape. It is, for three reasons:

1. **It cannot be built cleanly.** React Native cannot reparent a native view, so terminals that outlive a popped screen have to live in a host outside the navigator — an overlay above the `Stack` (which a natively-presented `formSheet` then covers, and which floats stationary while screens slide) or an underlay behind it (which shears during every push and back-swipe). Both trade a visible regression on the most common gesture for the win.
2. **It keeps receiving.** A hidden socket is still a subscribed socket: the host broadcasts to hidden clients, `hidden` only excludes them from sizing. Five warm agent sessions means five live output streams on someone's cellular connection.
3. **It puts the visibility contract permanently at risk.** The failure mode is silent, remote, and lands on *other people's* screens.

The cache gets most of the win at none of that cost, because the expensive part of a reopen is the replay and the round trips, not the socket handle. What a pool would additionally buy is the ~790ms to a live socket — and mobile input already goes over HTTP (`terminal.send`), so you can type and submit during that window; only quick keys wait.

**Numbers, adjusted with reasons.** 5 entries stands. 2 minutes I raised to 10: a snapshot is 33 KB (five is ~165 KB), serializing costs 2–5ms (7–9ms at 4× CPU throttle), nothing is held open, and coming back after a five-minute detour is exactly the case worth serving. Beyond the host's 2 MB catch-up ring an anchor degrades to a reanchor, which repaints rather than corrupting.

## The visibility contract holds

The hazard the brief flags — pinning every desktop pane to phone width — is *structurally* absent here: leaving the screen still closes the socket, and `releaseSocketDims` drops the constraint outright rather than relying on a `hidden` flag being correct.

Proven anyway, on the exact build in this PR, with a second client attached to the same terminal declaring 120x40 while the shell reported `tput cols` on a loop:

| action | PTY |
| --- | --- |
| baseline, phone showing the terminal | 53x37 |
| phone → home (screen unmounts) | **120x40** within ~1.2s |
| phone → reopen | 53x37 |
| app backgrounded | **120x40** within ~0.7s |
| app foregrounded | 53x37 |

## Also verified

- Warm reopen paints the restored buffer at ~1.2s while the "Connecting…" banner is still up, then converges with a ~300 B catch-up — no flash, no reset, no doubled scrollback.
- Tab switching inside a workspace: leaving a tab snapshots it under the right terminal id, returning restores it and attaches `mode=exact` for 79 bytes.
- A cold tab still takes the ring tail (`mode=tail`).
- While the sim was starved mid-run and sockets failed for minutes, every retry re-sent the *same* anchor and recovered exactly once it reconnected.
- `bun test` (66 pass), typecheck, biome, `check:i18n` (no catalog churn — no new strings), sherif.

## Not done

- Tests. The LRU's eviction and expiry are the obvious unit-test target; say the word and I'll add them.
- The ~780ms WKWebView instantiation is measured and left alone; the only lever on it is keeping a WebView alive, which is the pool question above.

https://claude.ai/code/session_011UhaTg7uhefyGtG9neVtTB

---
## Summary by cubic
Switching terminals on mobile now paints a cached buffer before the socket exists, cutting time to content from ~2.5–3s to ~0.9s and catch-up from ~105KB to ~300B. Reopens attach with the same `?seq=` catch-up contract desktop uses, replacing the legacy `replay=0|1` dialect.

- Added a memory-only `warmTerminalCache` (LRU, 5 entries, 10-minute expiry) storing serialized xterm buffers paired with their stream position (`epoch:seq`); cleared on sign-out and account deletion.
- The terminal page exports a snapshot when output goes quiet (300ms, min 2s apart) and on hide, so a popped WebView still leaves a fresh buffer.
- Reopen writes the restored buffer before dialing and requests `?seq=<epoch>:<n>`; first open uses `seq=new` for the ring tail, and unknown positions reanchor with `seq=none` instead of risking a gap.
- The host probe now runs only when a socket closes without ever attaching, saving ~220ms on every normal dial.
- A session-start guard stops a tab switch racing the first attach from opening a second socket, and switches drain xterm's write queue before resetting so bytes from the session being left can't parse into the session being entered.
- Leaving the screen still closes the socket, so no connection pool is introduced and the visibility contract is preserved.
- Added `@xterm/addon-serialize` dependency.

<sup>Written for commit 12a9a1af3687ce8dc05592e6da8cc9b878eab1e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7346?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added warm terminal restoration, allowing recently used terminal sessions to reopen with scrollback and position preserved.
  - Improved terminal tab switching by restoring cached content before reconnecting and loading only missing output.
  - Terminal snapshots refresh during session changes and when leaving the screen for a smoother return experience.

- **Bug Fixes**
  - Cleared cached terminal data when signing out or deleting an account to prevent previous session content from reappearing.


saddlepaddle:
Addressed all four cubic findings in 12a9a1af3. Three were real bugs; I verified each against the code rather than taking the review's word for it, and reproduced two of them.

**P1 — switch racing the initial attach (`TerminalWebView.tsx:310`).** Real. Only `beginSession` claimed the start guard, so a `switch` arriving before RN's answer to `ready` connected, and the later `attach` connected again. The window is not theoretical: the WebView mounts as soon as `activeTerminalId` is non-null, and that id can still change while the page boots (~900ms), e.g. as `params.tab` or the tab-order store resolves.

Reproduced headlessly against the *generated* page with a stubbed bridge and socket:

| scenario | before | after |
| --- | ---: | ---: |
| switch arrives before attach | **2 sockets** | 1 |
| attach then switch (normal) | 2 | 2 |
| switch drains queued output | 3000 stale lines | 0 |

The failing case is worse than a duplicate connection: the second dial went out as `seq=E1:100` — an anchor harvested from the *other* socket's `synced` — which is exactly the corruption the review described. Fix is a shared `claimStart()` that the switch path also takes.

**P1 — `term.reset()` and the write queue (`generate-terminal-html.ts:498`).** Real, and I confirmed the underlying behaviour directly against the bundled xterm rather than assuming: after `write(big)` then `reset()`, the queued data still parses and the buffer ends up holding the old content. So output from the session being left was landing in the one being entered, above its restored content.

Fixed by riding an empty write's callback as a barrier — also verified against the bundle: `write("", cb)` fires strictly after everything queued before it (`["big-parsed", "barrier"]`). A second switch while one is pending is guarded by generation.

Worth noting this **predates this PR** — the old switch called the same bare `reset()` — but the restore made it both more likely and more visible, so it belongs here.

**P2 — account deletion (`warmTerminalCache.ts:50`).** Real. `useSignOut` cleared the cache; `useDeleteAccount` is a separate path that cleared the query cache but not terminal buffers. Now clears both.

**P2 — late attach after the 1s backstop.** Real but narrower than described. The session is already running correctly at that point, so re-painting or re-dialling would be wrong. What was genuinely lost is the terminal id: a backstop start has none, and without one that page can never file a snapshot again. It now adopts the id without touching the running session.

Not changed: the LRU is still 5 entries / 10 minutes, and there are still no unit tests for its eviction and expiry — happy to add them if wanted.

PR #7353, opened 2026-09-09
refactor(desktop): remove the v2 new-workspace modal

saddlepaddle:
The new-workspace screen won its experiment and its flag is now at 100%, so v2 has exactly one create surface: the `/new-workspace` route. This removes the modal branch and everything that existed only to choose between the two.

## Why now

Experiment [387127](https://us.posthog.com/project/264803/experiments/387127) ended 2026-08-11: **55.7%** of the test arm created a real workspace vs **44.3%** of control (n=499, 98.8% chance to win). The flag has since served the screen to every account created after 2026-08-05 plus half the older base — ~81% of everyone who opened the surface last week.

The other ~19% was a live randomised holdout on *existing* users, the population the original experiment never tested. On a clean window (exposure Sep 1–5, 3-day workspace creation) the arms are indistinguishable — screen 39.4% (n=259) vs modal 40.4% (n=230), difference −1.0pp, 95% CI ≈ [−9.7, +7.7]. No benefit for existing users, and no sign of harm. Flag set to 100% before this PR.

## What goes

- `DashboardNewWorkspaceModal` — its only remaining job was to render nothing, close the store, and redirect to the route.
- `DashboardNewWorkspaceModalContent` and the modal's `PromptGroup` (856 lines), plus the two children only it used, `AttachmentButtons` and `UploadingAttachmentPill`.
- `useNewWorkspaceScreenVariant`, and the `new-workspace-screen` / `new-workspace-screen-override` flag constants.
- The stash/restore machinery on the new-workspace store — a `StashedDraft` type, three actions and three exported hooks, with **zero consumers** anywhere in the repo.

Net **−1,427 / +108**.

## What changes

`useOpenNewWorkspace` (and `useOpenNewSession`) is now the single entry point for "start creating a workspace". It navigates to the route on v2 and keeps v1 on its dialog, so the 13 call sites stop reaching into a store to set a flag that a component elsewhere watched in order to redirect.

The command palette gets `openNewWorkspace` on `CommandContext` rather than calling `useNewWorkspaceModalStore.getState()` — `Command.run` has no hook access, and this is the same shape as the existing `navigate` helper.

`useSelectedHostProjectIds` moved to `renderer/hooks/` since it has four consumers and its parent directory is gone.

## Deliberately left alone

- **v1.** `components/NewWorkspaceModal/` still ships for non-v2 installs and still owns the store's `isOpen` / `preSelected*` state, so the store stays.
- **`new-workspace-prompt-cards`.** That experiment is still in draft and lives inside the screen; its `useNewWorkspacePromptCardsVariant` hook stays. Worth a separate decision on whether it's still happening — its release-condition cutoff is still the `2026-09-01` placeholder.
- **Directory names.** `components/DashboardNewWorkspaceModal/` and `.../DashboardNewWorkspaceForm/PromptGroup/` are now named after deleted components. Renaming them touches ~20 files of imports including v1's, which would bury the deletion — better as a follow-up.
- **`openSessionModal` / `preSelectedSession`** on the store are only read by v2 today, but I couldn't cheaply prove the sessions section never renders under v1, so the v1 fallback stays.

## Verification

- `bun run typecheck` — 38/38 tasks pass.
- `bun test` (desktop) — 3712 pass, 1 fail, 1 error. **Identical on the base commit** (`git checkout HEAD~1 -- apps/desktop packages/shared`, same counts), and both suspects pass in isolation: a `@dnd-kit`/React interop error leaking between files. Pre-existing, not from this change.
- `bun run lint`, `bun run check:i18n` (0 missing in every locale, no catalog churn), `sherif` (one pre-existing `apps/review-host` warning).
- Persisted-key registry test passes; the deleted code only *read* `AGENT_STORAGE_KEY`, which the screen still writes, so no key is orphaned and `DEAD_KEYS` needs nothing.

Not verified by hand in a running app — the route path this makes universal is already what ~81% of users get, so the behaviour is the shipped one, not a new one.

https://claude.ai/code/session_01476rbYah5x4eZ31ksKX3fP

---
## Summary by cubic
The new-workspace screen won its experiment (flag now 100%), so v2 has exactly one create surface: the `/new-workspace` route. This removes the modal branch, the redirect shim that existed only to choose between the two, and everything the modal alone used.

- Deletes `DashboardNewWorkspaceModal`, its `PromptGroup` composer, `DashboardNewWorkspaceModalContent`, `AttachmentButtons`, and `UploadingAttachmentPill`.
- Removes the `useNewWorkspaceScreenVariant` hook and the `new-workspace-screen` / `new-workspace-screen-override` flag constants.
- Removes the stash/restore draft machinery from the new-workspace store — declared and exported, with zero consumers.
- `useOpenNewWorkspace` (and `useOpenNewSession`) is now the single entry point for starting a workspace: it navigates to the route on v2 and keeps the dialog for v1, replacing the call sites that set a store flag a component elsewhere watched. `useOpenNewWorkspaceForHost` covers the Cloud sidebar "+", navigating with a `?host=` search param.
- The command palette gets `openNewWorkspace` on `CommandContext`, and `useSelectedHostProjectIds` moves to `renderer/hooks/`.

**Deliberately left alone**
- v1's `NewWorkspaceModal` and its store state stay for non-v2 installs; the store keeps `isOpen`, `preSelected*`, and the open actions. The `usePreSelectedSession`, `usePreSelectedHostId`, `useOpenNewSessionModal`, and `useOpenNewWorkspaceModalForHost` selectors are gone — the deleted modal was their only reader.
- The `new-workspace-prompt-cards` experiment is still in draft and lives inside the screen; its hook stays.
- Old directory names (`components/DashboardNewWorkspaceModal/`) persist to avoid churning ~20 imports; renaming them is a good follow-up.

The route behavior this makes universal is already what ~81% of users get, so it is the shipped behavior, not a new one. Typecheck, lint, and i18n pass; the one failing test also fails on the base commit.

<sup>Written for commit fae75f7b29a30bf1c732a4b9afb459cd830f5a67. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7353?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Unified new-workspace actions across the command palette, sidebar, onboarding, tasks, issues, pull requests, and keyboard shortcuts.
  * New workspace and session actions now open the appropriate creation experience based on cloud availability, with project and host context preserved.
  * Added shared project-selection support for workspace actions.

* **Changes**
  * The v2 workspace creation experience is accessed through a dedicated route.
  * Removed the legacy dashboard workspace modal, draft-stashing support, and related experiment flags.


saddlepaddle:
Both cubic findings reviewed.

**P3 (stale doc comment)** — valid, fixed in 8f0f1a4: the quick-create fallback comment now says it opens the `/new-workspace` route on v2 and the dialog on v1.

**P2 (unhandled rejection on `navigate`)** — not taking this one. `void navigate(...)` is the codebase convention (38 call sites against 3 using `.catch`), and it is exactly what the deleted `DashboardNewWorkspaceModal` did on this same navigation. Adding a catch here would introduce a second way of doing it in the one place that has no more claim to it than the other 38. Worth doing repo-wide as its own change if the team wants it, not as a one-off here.

PR #7355, opened 2026-09-09
fix(automations): name the organization instead of a dead-end 404

saddlepaddle:
Follow-up to #7347. Automation reads are scoped to the **active** organization and threw `NOT_FOUND` for everything, so a member who opened a shared link while a different organization was active saw "Automation not found." — indistinguishable from a deleted row, with no hint that switching would show it. Deep links make that the likelier of the two failures, since you no longer have to already be in the organization that lists it.

## The rule

`automationNotFound(id, userId)` mirrors `pageNotFound` in `router/page/page.ts`: on a miss, re-query by id `innerJoin`ed to `members` on the caller's own userId.

- caller is a member of the owning organization → `FORBIDDEN`, naming it
- otherwise → `NOT_FOUND`, unchanged

The join is the leak guard — it can only ever name an organization the caller already belongs to. Used by `get` and `getPrompt`, the two reads behind the detail route.

## Notes

- **One-click switch.** The organization id rides along in `i18nParams` so the client can call `switchOrganization` directly instead of only naming the organization and leaving you to find the picker. Commented at the throw site so it doesn't get tidied away as unused.
- **First parameterised server error.** `userError`'s `params` and the `i18nParams` plumbing through `formatError` → `errorMessage()` already existed but had no callers; this is the first, so the catalog entry takes `params`.
- **Every error state keeps the breadcrumb.** They used to replace the whole pane, so a deep-linked visitor had no way back to the list — true of the existing not-found branch too. `AutomationBreadcrumbBar` is extracted from `AutomationDetailHeader` (which now composes it) and rendered on its own by the error branch.
- 2 new messages, 32 translations written; all 17 locales at 0 missing.

## Verification

Driven over CDP against a dev build, using a real 4-organization membership:

| Case | Server | UI |
| --- | --- | --- |
| Own automation, right organization active | 200 | detail page |
| Own automation, **wrong organization active** | `FORBIDDEN` naming it | notice + switch button |
| Automation in an organization the caller is **not** in | `NOT_FOUND` | "Automation not found." |
| Nonexistent id | `NOT_FOUND` | "Automation not found." |

Clicking **Switch organization** flipped the active organization and rendered the automation in one click.

Worth recording: the first UI pass on the two no-leak rows *looked* like it showed the wrong-organization notice. That was a stale render — `window.location.hash` does not remount the route, and a reload is overridden by the persisted history. Reading the raw tRPC codes through the app's authed client, then re-testing via the deep link, produced the table above.

https://claude.ai/code/session_012i4sML8GdjiCrKypTanYLS

---
## Summary by cubic
Opening a shared automation link under the wrong active organization now shows that organization's name and a one-click switch, instead of a dead-end "Automation not found." The server returns `FORBIDDEN` naming the organization only when the caller is a member of it; otherwise it stays `NOT_FOUND`. Error states keep the breadcrumb so deep-linked visitors can get back to the list. Translations cover all 17 locales, with German using the catalog's informal address.

<sup>Written for commit d9ecd509c2e735b889d46a193b1066463cb1abbf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Automation detail pages now include breadcrumb navigation with contextual actions.
  - Shared automation links identify when an automation belongs to another organization and offer an option to switch organizations.

- **Bug Fixes**
  - Automation access errors now distinguish unavailable automations from those belonging to another organization.

- **Localization**
  - Added translated cross-organization automation messages across supported languages.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/automations/$automationId/components/AutomationBreadcrumbBar/AutomationBreadcrumbBar.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/automations/$automationId/components/AutomationBreadcrumbBar/AutomationBreadcrumbBar.tsx
@@ -1,0 +1,56 @@
+import { Trans } from "@lingui/react/macro";
+import {
+	Breadcrumb,
+	BreadcrumbItem,
+	BreadcrumbLink,
+	BreadcrumbList,
+	BreadcrumbPage,
+	BreadcrumbSeparator,
+} from "@superset/ui/breadcrumb";
+import { Link } from "@tanstack/react-router";
+import type { ReactNode } from "react";
+
+interface AutomationBreadcrumbBarProps {
+	/** Omitted when the automation could not be loaded — the crumb stops at the list. */
+	name?: string;
+	/** The header's right-hand actions, if there are any. */
+	children?: ReactNode;
+}
+
+/**
+ * The detail route's top bar. Error states render it without a name so a
+ * deep link that lands somewhere unusable still offers the way back.
+ */
+export function AutomationBreadcrumbBar({
+	name,
+	children,
+}: AutomationBreadcrumbBarProps) {
+	return (
+		<header className="flex h-11 shrink-0 items-center justify-between border-b border-border px-4">
+			<Breadcrumb>
+				<BreadcrumbList className="text-sm">
+					<BreadcrumbItem>
+						<BreadcrumbLink asChild>
+							<Link to="/automations">
+								<Trans>Automations</Trans>
+							</Link>
+						</BreadcrumbLink>
+					</BreadcrumbItem>
+					{name !== undefined && (
+						<>
+							<BreadcrumbSeparator />
+							<BreadcrumbItem>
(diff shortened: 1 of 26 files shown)


saddlepaddle:
Both reviewers flagged the same thing and they were right — the German string was the only `Sie` in a catalog that is `du` throughout (117 informal vs 1 formal, and the 1 was mine). Fixed in d9ecd509c.

I checked the other T/V locales for the same mistake rather than assuming it was German-only. It was: the French, Russian and Czech catalogs are formal by their own convention (88, 114 and 99 formal imperatives against ~0 informal), so `Basculez`, `Смените` and `Přepněte` match their catalogs and stay as written. German is the outlier.

PR #7357, opened 2026-09-09
fix(desktop): stop the terminal IME composition overlay clipping CJK

zime78:
## What & why

Composing CJK text (Korean / Japanese / Chinese) in the v2 terminal renders the **in-progress syllable clipped** — the Korean `글` shows only its left half while you are still composing it. It snaps back to normal once committed, so the *live* composition is the unreadable part.

This is not a rare edge case: it hits **every CJK IME user on macOS who has a Nerd Font "Mono" variant installed**, which is extremely common among terminal users (Starship/eza/LazyGit setups almost always install one). For those users the IME preedit is broken on every keystroke of every CJK word. That's why I think this is worth fixing rather than leaving for later.

## Root cause

xterm.js draws the in-progress composition in a DOM overlay (`.composition-view`) whose box is sized by normal **DOM text layout** and then clipped with `overflow: hidden`.

`installed-nerd-fonts.ts` auto-detects installed Nerd Fonts and appends them to the terminal font stack with the **Mono variants sorted first** (correct for the WebGL cell renderer — single-cell icons line up). But Nerd Font *Mono* variants force one advance width on **every** glyph, so a full-width CJK glyph gets a **half-width advance in DOM text**:

| font (14px) | advance of `글` |
|---|---|
| `D2CodingLigature Nerd Font Mono` | **7px** |
| `D2CodingLigature Nerd Font` (non-Mono) | 14px |
| Menlo cell width | 8px |

When the primary terminal font has no Hangul (e.g. Menlo), the glyph falls through to the Mono variant, the overlay box becomes **narrower than one cell**, and `overflow: hidden` clips it. The WebGL cell renderer is unaffected (it draws per cell), which is exactly why the committed line looks fine and only the overlay breaks.

## The fix

Give the composition overlay its **own font stack** (`imeFontFamily`): same validation as the terminal stack, but Nerd Font Mono variants are dropped from the icon fallbacks. A user-chosen primary is never dropped. It is delivered as a CSS variable on the terminal-runtime wrapper and applied to `.composition-view` with `!important`, because xterm writes that element's `font-family` **inline** on every update. The WebGL body stack keeps Mono-first, so icons still fit one cell.

- No behavior change for non-CJK input, or for users without a Nerd Font Mono in the stack.
- Scoped to the v2 terminal runtime via a wrapper class; the v1 path is untouched.

## Before / after

- **Before:** `.composition-view` box = 7px (< 1 cell) → `글` clipped to its left half.
- **After:** box = one full cell (14px) → `글` renders whole.

Reproduce in ~30s: set `terminalFontFamily` to a Nerd Font Mono (or just install one so it's auto-detected), keep a Latin primary like Menlo, switch to a Korean/Japanese IME, and compose in the terminal. Screenshots/crops from an Electron harness that drives the real xterm bundle via `Input.imeSetComposition` are available on request — happy to add them inline.

## Upstream relationship

The underlying cause is in xterm.js and is tracked by **xtermjs/xterm.js#6161** with fix PR **xtermjs/xterm.js#6162** ("Align the IME preedit to the renderer cell grid"). That is not released yet and this repo pins an xterm beta, so this small app-level fix restores CJK composition now; once #6162 ships and the pinned xterm is bumped, this override can be removed — the code comments say so.

## Testing

- `bun test` — terminal appearance/runtime suites, **44 pass / 0 fail** (5 new cases: the Mono-variant predicate, the IME stack dropping Mono while the body stack keeps it, a user-chosen Mono primary preserved, the runtime setting the CSS var on both runtimes, and a single-validation warn-once guard).
- `bun run typecheck` — clean.
- Biome — clean.

Allow edits from maintainers is enabled.

---
## Summary by cubic
Fixes the terminal IME composition overlay clipping CJK characters while typing (e.g., Korean '글' shows only left half). The cause is Nerd Font Mono variants in the autodetected fallback stack forcing half-width advance for full-width glyphs in the DOM overlay. The overlay now gets its own font stack that drops Mono variants from auto-detected fallbacks, while the main terminal stack keeps them for icon alignment. No behavior change for non-CJK input or users without a Nerd Font Mono in the stack.

<sup>Written for commit eaa9e097bb14ca1ca105e26d5e5e6464e833f89e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7357?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal input method editor (IME) text rendering for full-width CJK characters.
  * Fixed clipping and incorrect character widths when composing text with Nerd Font Mono terminal fonts.
  * Improved handling of abbreviated and full-name Nerd Font Mono variants.

* **Improvements**
  * Terminal appearance settings now apply consistently to active and inactive terminal sessions.
  * IME text uses a dedicated font configuration while preserving the selected terminal font.

### DIFF
--- a/apps/desktop/src/renderer/globals.css
+++ b/apps/desktop/src/renderer/globals.css
@@ -209,6 +209,25 @@
 		font-family: var(--superset-terminal-font-family, ui-monospace, monospace);
 	}
 
+	/*
+	 * IME 조합 중인 글자("글" 등)는 WebGL 셀이 아니라 `.composition-view` DOM
+	 * 오버레이에 그려지고, 그 박스는 일반 DOM 텍스트 폭으로 재서 잘린다.
+	 * Nerd Font Mono 변형은 전각 CJK 글리프에도 반각 advance 를 주므로
+	 * (14px 기준 "글" 7px, 비-Mono 는 14px) 오버레이가 한 셀보다 좁아져 글자가
+	 * 잘린다. WebGL 셀 렌더러는 셀 단위로 그려서 영향이 없다.
+	 * xterm.js 가 이 오버레이의 font-family 를 인라인으로 써넣기 때문에
+	 * `!important` 가 필요하다. 업스트림 xterm.js PR#6162(open)이 반영되면
+	 * 이 재정의는 불필요해진다.
+	 * var() 폴백을 일부러 두지 않았다: 변수가 없으면 이 선언이
+	 * invalid-at-computed-value-time → `unset` 이 되어 위 `.xterm` 의
+	 * `--superset-terminal-font-family` 스택을 그대로 상속한다. 즉 사용자 폰트를
+	 * 잃지 않고 변경 이전 동작으로 degrade 된다.
+	 */
+	.superset-terminal-runtime .xterm .composition-view {
+		/* biome-ignore lint/complexity/noImportantStyles: xterm 이 이 오버레이의 font-family 를 인라인으로 써넣으므로 !important 없이는 덮어쓸 수 없다 */
+		font-family: var(--superset-terminal-ime-font-family) !important;
+	}
+
 	.xterm .xterm-rows a {
 		color: inherit;
 		cursor: pointer;
--- a/apps/desktop/src/renderer/lib/terminal/appearance/appearance.test.ts
+++ b/apps/desktop/src/renderer/lib/terminal/appearance/appearance.test.ts
@@ -1,10 +1,19 @@
-import { afterEach, describe, expect, mock, test } from "bun:test";
+import { afterEach, describe, expect, mock, spyOn, test } from "bun:test";
 import {
 	DEFAULT_TERMINAL_FONT_FAMILY,
 	NERD_FONT_FALLBACK_FAMILIES,
 	resolveTerminalAppearance,
 	sanitizeTerminalFontFamily,
 } from "./index";
+
+/**
+ * IME 오버레이 기본 스택 — 터미널 기본 스택에서 Nerd Font Mono 폴백만 뺀 값.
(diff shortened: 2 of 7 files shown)


Kitenite:
@zime78 Follow-up: the fix has merged in #7488. Thank you for providing the D2Coding Mono reproduction and linking the upstream cause.

We used the cell-layout backport from xtermjs/xterm.js#6162. It fixes the composition overlay using the terminal renderer's Unicode cell widths while preserving the configured font stack, including an explicitly configured secondary Mono fallback. In a real Superset v2 terminal with `Menlo, "D2CodingLigature Nerd Font Mono"` at 14px, `글` changed from a clipped 7px overlay to a complete 16px glyph box; `글한글` changed from 21px to 48px. Committed text, Latin composition and combining accents remained intact.

[Fresh before/after screenshots, exact commands, and bundle hashes](https://app.superset.sh/page/pr-7488-verified-actual-xterm-bundles-before-and-a-nqp4nv) are linked in #7488. These captures use CDP IME input, not a physical keyboard/input-method session. The final runtime bundles have behavioral tests for both CJS and ESM.

This supersedes the font-filtering implementation in this PR. Upstream-backport removal is tracked in #7490: after the upstream fix is released and Superset upgrades to it, remove only the IME CompositionHelper hunks, retain our independent WriteBuffer/RenderDebouncer fixes, and keep the regression tests.

PR #7361, opened 2026-09-09
fix(host-service,desktop): prefer origin when importing v1 projects (#7241)

### LINKED ISSUE #7241: [bug] v1→v2 import hijacks another repo's v2 project via a non-origin remote, stranding the rightful repo's workspaces
## Summary

On a repo that has **more than one GitHub remote**, the v1→v2 import claimed the *wrong* cloud project: it matched a non-`origin` remote, hijacked the v2 project belonging to a **different** local repo, re-pointed that project's `repoPath` at the wrong directory, and left the rightful repo with no v2 project at all. The rightful repo's workspaces silently vanished from the sidebar.

App version 1.26.0, macOS.

## What I observed

`~/Kogan/kogan` (a monorepo, `origin` = `kogan/kogan`) has a second remote used for upstream syncs:

```
$ git -C ~/Kogan/kogan remote -v
oms-service  git@github.com:kogan/oms-service.git (fetch)
oms-service  git@github.com:kogan/oms-service.git (push)
origin       git@github.com:kogan/kogan.git (fetch)
origin       git@github.com:kogan/kogan.git (push)
```

`~/Kogan/oms-service` is a *separate* local repo (`origin` = `kogan/oms-service`) that was **already a v2 project** (`7efcc5af-…`, created 2026-05-11).

After the import ran:

- `v1_migration_state` recorded `01be230b (v1 "kogan", ~/Kogan/kogan) -> 7efcc5af` — i.e. the monorepo was imported **into the oms-service v2 project**.
- Cloud project `7efcc5af` was renamed `oms-service` → `kogan`, its repo flipped to `https://github.com/kogan/kogan`, and its path re-pointed to `~/Kogan/kogan`.
- The 8 monorepo workspaces were migrated onto `7efcc5af`.
- `~/Kogan/oms-service` had **no** `v1_migration_state` row at all, and none of its 7 v1 workspaces migrated. Its previously-existing v2 workspaces on `7efcc5af` were gone too.
- In the sidebar the monorepo appeared *labelled* `oms-service` with its own workspaces underneath, and the real oms-service project was absent. Re-adding `~/Kogan/kogan` as a project reported "already added".

Every worktree and branch survived on disk — this is a metadata/ownership bug, not disk loss — but all sidebar rows, names and workspace records for the displaced project were lost.

## Root cause

Three things compose into the failure. (Line refs from `dist/renderer/assets/layout-*.js` and `dist/main/host-service.js` in the 1.26.0 asar.)

**1. The importer opts into walking every remote.**

```js
function findProjectByPath(hostClient, project) {
  return hostClient.project.findByPath.query({
    repoPath: project.mainRepoPath,
    walkAllRemotes: true,
    expectedRemoteUrl: expectedRemoteUrlFor(project)
  });
}
```

`findByPath` then queries the cloud for **every** GitHub remote on the repo and merges the results:

```js
const allRemotes = await getGitHubRemotes(usage.createUserSimpleGit(gitRoot));
for (const parsed of allRemotes.values()) urlsToQuery.set(parsed.url.toLowerCase(), parsed);
...
const candidates = Array.from(byId.values()).sort((a, b) => {
  if (a.matchesExpected !== b.matchesExpected) return a.matchesExpected ? -1 : 1;
  return a.name.localeCompare(b.name);
});
```

`origin` gets no preference anywhere — not in the query set, not in the sort.

**2. The one disambiguation hint is dead in practice.**

```js
function expectedRemoteUrlFor(project) {
  if (!project.githubOwner) return void 0;
  ...
}
```

`projects.github_owner` is empty for **every** row in my v1 `local.db` (13/13). So `expectedRemoteUrl` is always `undefined`, `matchesExpected` is always `false`, and the sort silently degrades to `name.localeCompare` — an alphabetical tiebreak between unrelated repos.

**3. A lone wrong candidate is accepted without question.**

```js
function decideProjectImport(result) {
  const local = result.candidates.find((c) => c.source === "local-path");
  if (local) return { kind: "already-imported", v2ProjectId: local.id };
  if (result.candidates.length > 1) return { kind: "skip", reason: "multiple-candidates" };
  ...
  return { kind: "import" };
}
```

The `multiple-candidates` guard only fires at ≥2. In my case `kogan/kogan` had no v2 project yet, so exactly **one** candidate came back — the one found via the *non-origin* remote. `importV1Project` then takes `candidates[0]` blindly and calls `project.setup`:

```js
const targetCandidate = linkToProjectId ? candidates.find(c => c.id === linkToProjectId) : candidates[0];
...
await hostClient.project.setup.mutate({
  projectId: targetCandidate.id,
  origin: { repoCloneUrl: targetCandidate.repoCloneUrl, name: targetCandidate.name },
  mode: { kind: "import", repoPath: project.mainRepoPath, allowRelocate }
});
```

In `setup`'s `import` branch, `resolveMatchingSlug(repoPath, "kogan/oms-service")` happily finds the `oms-service` remote on the monorepo and returns the **monorepo** path, and `persistLocalProject` re-points the v2 project there. `rejectIfRepoint` doesn't help: it only guards when a local row for that project id already exists on this host.

Net: a repo that merely *references* another repo's remote can take that repo's v2 project.

## Repro

1. v2 project **A** exists for repo `owner/a`, checked out at `~/a`.
2. Separate v1 project **B** at `~/b`, `origin` = `owner/b`, plus a second remote pointing at `owner/a`.
3. No v2 project exists for `owner/b`.
4. Run the v1→v2 import.
5. **B** is imported into project **A**. **A** is renamed/re-pointed to `~/b`. `~/a` ends up with no v2 project and no migrated workspaces.

Multi-remote setups like this are ordinary: monorepo-to-service syncs, forks with an `upstream` remote, mirrors.

## Suggested fixes

- Prefer `origin` when building the candidate query set, and rank `origin`-derived candidates above other remotes in the sort.
- Don't accept a lone candidate that came from a non-`origin` remote without confirming with the user — treat "matched only via a secondary remote" the same way `multiple-candidates` is treated today.
- Never let `project.setup` re-point an existing cloud project's repo path/URL as an implicit side effect of import. If the resolved repo path differs from the project's recorded one, that's a conflict to surface, not a relocate.
- Populate/derive `githubOwner` (or just read `origin` directly) so `expectedRemoteUrl` isn't silently `undefined`; alternatively drop the hint and always resolve `origin` on the spot.
- Refuse to strand a v1 project: if another v1 project's import consumed the v2 project this one would have matched, record a `skip` with a reason rather than leaving no ledger row at all.

## Workaround (for anyone hit by this)

Worktrees and branches are untouched. Workspaces can be re-attached to the correct project — `workspaces create` adopts an existing worktree for an existing branch instead of creating a new one:

```
superset projects list          # find the correct project id
superset workspaces create --local --project <PROJECT_ID> \
  --name "<name>" --branch "<existing-branch>" --skip-branch-prefix
superset workspaces get <NEW_ID>   # confirm worktreePath is the pre-existing dir
```

Verified: `worktreePath` came back as the original worktree, `git worktree list` count unchanged, and uncommitted work in those worktrees survived.

andyst-dev:
## What & why

fixes #7241 — v1→v2 importer silently hijacks another repo's v2 project on multi-remote repos.

The importer's discovery (`findByPath` with `walkAllRemotes`) queried **every** GitHub remote with no preference for `origin`, and then accepted a **lone** candidate even when it was found only via a secondary remote. On a repo that merely references another repo as a remote (monorepo + upstream-sync remotes, forks with `upstream`, mirrors), this incorrectly claimed the other repo's existing v2 project, re-pointed its `repoPath`/repo metadata at the wrong directory, and left the rightful repo with no v2 project.

Two coordinated changes:

- **Host (`project.findByPath`)**: each cloud candidate is now tagged with `viaOrigin` (whether it was discovered through the repo's `origin` remote), the result reports `hasOriginRemote`, and origin-derived candidates sort **first**.
- **Renderer (`decideProjectImport`)**: a lone `remote` candidate that did **not** come from `origin`, on a repo that **has** an `origin` remote, is refused with a new `non-origin-only` skip reason — the wizard surfaces it for explicit human confirmation instead of silently relocating the wrong project. Repos with no `origin` remote at all (fork/mirror-only, `origin` absent) keep the old behavior.

## How I tested it

- `packages/host-service`: `bun test src/trpc/router/project/project-import.test.ts` → 5/5 pass, including two new cases: (1) the #7241 shape — a secondary `oms-service` remote resolving to a cloud project while `origin` has none — yields a single candidate marked `viaOrigin: false` with `hasOriginRemote: true`; (2) two remotes each resolving to a project rank the origin one first.
- `apps/desktop`: `bun test src/renderer/lib/v1-migration/v1-migration.test.ts` (+runV1Migration) → 59/59 pass, including the new `decideProjectImport` matrix: lone non-origin + origin→skip `non-origin-only`; lone origin→import; lone non-origin without origin→import.
- `bun run lint` (Biome) clean on all touched files; typecheck clean on the touched packages (the full desktop typecheck run has a pre-existing `@codemirror` 6.6.0/6.7.1 duplicate-version error in `CodeEditor.tsx` unrelated to this diff).

## Checklist

- [x] `bun run lint` and `bun run typecheck` pass on the touched code (CI-clean; note: full desktop typecheck has a pre-existing @codemirror dup unrelated to this diff)

---
## Summary by cubic
Fixes the v1→v2 importer silently hijacking another repo's v2 project when the repo has multiple remotes, e.g. an `upstream` or sync remote pointing at an already-imported project (#7241). The importer now prefers candidates found via `origin` and refuses a lone candidate found only through a secondary remote when `origin` exists, so the wizard asks instead of re-pointing the wrong project.

- `findByPath` tags cloud candidates with `viaOrigin`, returns `hasOriginRemote`, and sorts origin-derived candidates first; `hasOriginRemote` is read from the raw remote list so the guard also fires when `origin` points at a non-GitHub host (GitLab, self-hosted, SSH alias).
- `decideProjectImport` adds a `non-origin-only` skip reason for a lone non-origin candidate on a repo that has an `origin`.
- Repos without an `origin` remote (fork/mirror-only clones) keep the old import behavior.

<sup>Written for commit 1ca0046d0b47feee354575771e44929602831604. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7361?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project migration matching for repositories with multiple remotes.
  - Origin-based project matches are now prioritized during discovery.
  - Prevented automatic linking to a project found only through a non-origin remote when an origin remote is present.
  - Projects found through secondary remotes can still be imported when no origin remote exists.
  - Correctly recognizes an origin remote even when it uses a non-GitHub host.


andyst-dev:
Addressed both cubic findings in `69505b1` (host + renderer, 59/59 desktop + 6/6 host-service tests pass).

1. **P1 — non-GitHub origin:** `hasOriginRemote` was derived from `getGitHubRemotes()`, which discards a remote that isn't GitHub. A repo whose `origin` points at GitLab/self-hosted forge/SSH alias therefore reported `hasOriginRemote: false`, letting a lone candidate found only through a secondary GitHub remote fall through the hijack guard. Now `findByPath` reads origin *existence* from the raw remote list (`getAllRemoteUrls`) — `hasOriginRemote = rawRemotes.has("origin")` — while `viaOrigin` stays scoped to parsed GitHub remotes, exactly as suggested. So a repo whose canonical identity lives on another host now triggers the guard: the lone GitHub candidate is `viaOrigin: false` and is rejected unless origin itself matches.

2. **P3 — test fixtures:** the `decideProjectImport` fixtures used the non-existent `source: "github-remote"`, so none of them exercised the `viaOrigin` guard (which keys on `source === "remote"`). Switched the fixtures to the real `"remote"` value (and the multi-candidate/import path fixtures too, for type-correctness).

Added a host-service regression test: `origin → gitlab.com/owner/kogan.git` + secondary `oms-service → github.com/owner/oms-service.git` (the cloud project) → asserts `hasOriginRemote === true` and the lone candidate is `viaOrigin: false` so the renderer skips it.

Validation:
- `bun test src/trpc/router/project/project-import.test.ts` → **6 pass** (5 old + 1 new), 25 expects.
- `bun test src/renderer/lib/v1-migration/{v1-migration,runV1Migration}.test.ts` → **59 pass**, 0 fail.
- Head `69505b1` pushed; PR reads `mergeable=true` (CI `blocked` while the checks run on the new commit).

PR #7312, opened 2026-09-08
feat(pages): split comment mode from the comments panel, theme published pages

harshithmullapudi:
## Summary

- **Comments**: the comment button now only enters comment mode; a floating "All comments" button opens the thread list, which stays open until you close it.
- **Theme**: every published page is served a default stylesheet built from Superset's own tokens, injected at view time, so pages look like they belong to the product.
- **Marketing** (two unrelated fixes, called out below): the contribution graph opens on the latest week, and the leaderboard profile OG image no longer 500s.

## Why / Context

The comment button did two jobs at once — arming the page for pin placement *and* opening the list. On web the list was a permanent 300px rail; on desktop it appeared and vanished with comment mode, so reading a thread meant arming the page for a stray click. Published pages, meanwhile, were whatever HTML an agent wrote against browser defaults, so two pages from the same workspace looked like different products.

## How It Works

**Comments panel.** `enabled` (pin placement) and `panelOpen` (the list) are now independent state on `CommentProvider`; leaving comment mode no longer closes the panel, and closing the panel clears the thread selection so the on-page popover cannot inherit it. `CommentsSidebar` became `CommentsPanel`: a floating card over the top-right of the page from `md` up, a sheet over the pane below it (absolute, not fixed — a desktop window can be narrower than `md`). Threads group by day and each is a card with its own reply box; past three comments the middle folds behind a count. While the panel is open it is the only reading surface — a pin click selects the thread there rather than also drawing a popover.

**Page theme.** `apps/usercontent` already rewrites each document on the way out to inject the comment runtime; it now also inlines a `<style>` at the top of `<head>`, which reaches pages published before it existed. Every rule is wrapped in `:where()` (zero specificity), so any selector a page writes wins outright — that is what makes it safe to apply retroactively. Palette is the app's own oklch values from `packages/ui/src/globals.css`. The theme is light unless a page opts in: `dark` or `light` on `<body>` pins it, `auto` follows `prefers-color-scheme`. It is not automatic, because a page that declared a background but no text colour would otherwise take the two halves from different palettes and end up unreadable. The page skill documents the structure, tokens and theming.

**OG image.** The profile OG route loaded its font with `fetch(new URL(…, import.meta.url))` — the edge-runtime pattern — but declares no runtime, so it runs on Node where `fetch` cannot read `file://`. Every share of a leaderboard profile had no image. Now read off disk with `fs` from `public/`, matching the blog OG route.

### Comments (web + desktop)
[...]

PR #7367, opened 2026-09-09
feat(mobile): read and comment on pages from the phone

harshithmullapudi:
## Summary

- Adds Pages to the mobile app: a native list, a WebView viewer, and the element-picker comment flow (remove / comment / quick feedback / approve).
- The shared comment runtime is **unchanged** — host and frame already talk over `postMessage`, never the DOM, so a WebView is a drop-in for desktop's iframe.

## Why / Context

Pages were desktop-only. The viewer is the whole point of the feature on a phone: an agent publishes a page, and you want to read it and leave feedback without opening a laptop.

## How It Works

**Viewer.** `page.pull` already returns a signed `viewUrl` on the usercontent worker — exactly what desktop's `<iframe src>` loads (`PageFrame.tsx:24` on the desktop side). Mobile loads the same URL in a `WebView`, so there is no auth or asset-resolution problem: the tRPC call is already authenticated and the page keeps its own origin.

**Bridge.** The runtime posts with `parent.postMessage` (`page-comments-runtime.ts:69`). In a WebView the page *is* the top document, so `parent === window` and that lands as a message event on its own window. An injected listener forwards those to React Native; host messages go back as a plain `window.postMessage`, which the runtime's own listener (line 242) already handles. Nothing in `packages/shared` changed.

**Comment mode.** The runtime's `pick` is driven by a `click` with `clientX/Y` (line 191) — hover only drives the highlight preview — so a tap already produces a pick with anchor + rect. No new gesture, no runtime change.

## Manual QA Checklist

Validated on iPhone 16e simulator against **production** data.

- [x] Pages list renders with thumbnails, owner, visibility icon, date
- [x] Pull-to-refresh works
- [x] Tapping a page opens the viewer; page renders with assets
- [x] Links inside a page open in the system browser, not the frame
- [x] Focus (comment-mode) toolbar button toggles the picker
- [x] Tapping a block highlights it and shows the toolbar
- [x] Toolbar clamps inside the frame near edges
- [x] Quick feedback sheet posts a preset
- [x] Write-a-comment sheet posts and the pin appears
- [x] All comments sheet lists threads; tapping opens the thread
- [x] Thread opens scrolled to the composer
- [ ] **Not verified:** resolve/reopen round-trip, agent-authored comment rendering, a page whose `viewUrl` ticket expires while open

## Testing

- `bun run typecheck` — 38/38 tasks pass
- `bun run lint` — clean
- `bun run check:i18n` — all 17 locales compile `--strict`
- `bunx sherif` — clean
- `bun test apps/mobile/screens/.../pages` — 11 pass (new: `pinLayout`, `toolbarPlacement`)

`bun test` repo-wide has 40 failures, all in `packages/agent-setup`, the node-pty patch guard, and host-service process/socket utils — tests that spawn processes or open sockets. None touch this diff, but I did not verify them against a clean tree.

## Design Decisions

- **WebView on `viewUrl`, not inlined HTML.** A page version is a document plus assets served by the worker, so `source={{ html }}` would break relative asset refs and lose the page's own origin.
- **Sheets as `formSheet` routes** per `apps/mobile/AGENTS.md`, with a transient zustand store carrying the pick. Route params would have to carry a 2000-char anchor path.
- **Absolute dates instead of `formatRelativeTime`.** Hermes ships no `Intl.RelativeTimeFormat`, and `apps/mobile/lib/intl-polyfills` installs only `Locale` and `PluralRules`. `formatDate` uses `Intl.DateTimeFormat`, which Hermes has. Adding the polyfill means a new dependency plus 16 locale-data imports — deferred deliberately.

## Known Limitations

- **No `intent` on remove/approve.** The enum lands in #7359, which is not on main and whose migration is unapplied. Those two buttons post their preset text as ordinary comments, so nothing downstream can distinguish them yet. Wiring `intent` is one line per action once that merges.
- **Catalog collision with #7359.** Twelve msgids overlap (the ten `QUICK_PRESETS`, "Looks good", "Delete this"). Lingui keys on English text, so every `messages.po` will conflict on merge. **#7359's translations should win** — they went through review alongside the intent work.
- **`pinLayout` is duplicated**, not shared. It belongs in `@superset/shared` next to the runtime types it imports, but #7359 rewrites `pinLayout.ts` heavily and moving it now guarantees a conflict. Follow-up.
- **Pins lag a fling.** Rects cross the RN bridge per frame; desktop gets this free from a same-process iframe. Acceptable at rest, visible while scrolling fast.

## Follow-ups

- Wire `intent` and consolidate `pinLayout` into `@superset/shared` after #7359.
- Search and scope tabs on the list (desktop has All/Pinned/Team/Just me).
- Edit and delete a comment; agent-activation affordance.

---
## Summary by cubic
Adds native mobile browsing and commenting for published Pages. Links to same-origin `/page/<slug>` URLs that previously opened in Safari now open an in-app preview sheet; all other links still open externally.

- Uses the existing signed `viewUrl` and comment runtime through a WebView bridge.
- Supports element comments, quick feedback, replies, threads, pins, and optimistic resolve with rollback; remove and approve still post preset text as regular comments.
- Pages now live in the organization sheet, with org switching on its own screen.
- Comment composition is a native pill composer; sharing is an in-app sheet with copy link, access, and version, read-only for non-creators.
- Keeps page links and subframes in the reader, retries failed or expired loads with a fresh ticket, preserves scroll position when expanding a preview, and lets only http/https links open externally while worker redirects for expired tickets pass through.
- Clips off-screen pins, caches pulled pages per slug and organization, distinguishes offline from empty in the list, and surfaces resolve failures.
- Adds loading, empty, error, retry, pull-to-refresh, localized copy, pin layout, toolbar placement, and link parsing tests.
- Keeps the Pages shortcut available when the selected host is offline.

<sup>Written for commit 695925fb7c39bd5472d690c43ad4bff22d277b20. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7367?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added mobile page browsing with persistent filters for all, team, or personal pages.
  - Added page sharing controls, visibility settings, version selection, and link copying.
  - Added comment threads, replies, inline comment pins, quick feedback, and review actions.
  - Added clearer offline, loading, and error states.
  - Added organization switching and improved page navigation.
- **Localization**
  - Added and updated translations for page, comment, review, sharing, and connection-related messages across supported languages.
- **Bug Fixes**
  - CLI tables now display full URLs without truncation while preserving layout.
  - Mobile link handling now validates URLs before opening them.


apps/mobile/screens/(authenticated)/pages/[slug]/hooks/usePageComments/usePageComments.ts
```diff
@@ -0,0 +1,112 @@
+import type { CommentAnchor } from "@superset/shared/page-comments-runtime";
```
Comment (saddlepaddle) [inline apps/mobile/screens/(authenticated)/pages/[slug]/hooks/usePageComments/usePageComments.ts:1]: This is the third implementation of page comments, and the thing worth raising is not that it is a third one, it is that all three now behave differently.

Desktop and web both do optimistic writes into the query cache and coordinate invalidation with an `isMutating` count scoped by `meta`. This invalidates on success and does no optimistic write at all. So the same action reads as instant on desktop and as a round trip on a phone, on the same page, in the same product.

To be clear about which way I would resolve it: this one is probably right. An optimistically created comment carries a client-invented id and timestamp that the server owns, so it has to be reconciled by refetch, and coordinating that refetch across five mutations is where the counter came from. Invalidating on success has none of that and is what the rest of the ecosystem does for creates, reserving optimistic writes for actions whose result the client can compute exactly, like a like or a resolve toggle.

The problem is not the choice, it is that three surfaces made it independently and landed in two different places. Whichever way it settles, the decision wants one home rather than three, which is the same conclusion the `usePageCommentStore` threads on #7312 reached from the desktop and web side.

Not a change request for this PR. Worth linking to that discussion so the mobile case is counted when it gets picked up, since it is now the third data point rather than a hypothetical one.

packages/cli-framework/src/output.ts
```diff
@@ -103,17 +103,14 @@ export function table(
 			const val = getNestedValue(row, col);
 			const str = val === null || val === undefined ? "—" : String(val);
 			const cap = caps[i]!;
-			const text = str.length > cap ? `${str.slice(0, cap - 1)}…` : str;
-			return URL_CELL.test(str) ? { text, link: str } : { text };
+			if (URL_CELL.test(str)) return { text: str, link: str };
+			return { text: str.length > cap ? `${str.slice(0, cap - 1)}…` : str };
 		}),
```
saddlepaddle, line 108: This change is right, and the reasoning in the comment is the good part: a terminal that ignores OSC 8 detects links from visible text, so truncating a URL cell produces a link that does not work rather than a shorter one.

It also has nothing to do with mobile pages and comments, and it is two small files with a test that stands on its own. Bundled here it waits on review of an 89-file feature branch; split out it lands today. Same for anyone bisecting a CLI output change later and finding it inside a mobile PR.

Worth pulling into its own PR unless there is a dependency I am not seeing.

PR #7374, opened 2026-09-09
feat(admin): company metrics — star chart and paywall funnel

saddlepaddle:
Puts the two company numbers we watch week to week at the top of the admin dashboard's Company tab (the page already carries the **Company Metrics** heading, so these lead it rather than starting a second area with the same name): **GitHub star growth**, and the **paywall → paid funnel** that #7338 made followable end to end.

## Star chart — reused, not reimplemented

`/starchart` and the admin tile now render the same series from the same code:

| moved to | what |
|---|---|
| `@superset/ui/star-chart` | `StarChart` (granularity toggle, range picker) + `StarHistoryChart`, `StarPaceChart`, and the pace maths |
| `@superset/trpc/star-history` | the stargazer walk (`fetchStarHistory({ token })`), env-free so either app can own the token |
| `@superset/shared/github-stars` | `StarHistoryPoint`/`StarHistory`, `formatStarCount`, `githubRepoSlug` |

Marketing keeps only its env boundary (`getStarHistory()` = the shared fetcher + `env.GITHUB_TOKEN`). The slug derivation existed in three copies (marketing's `lib/github.ts`, trpc's `growth/github.ts`, and the fetcher); now there is one, and `apps/marketing/src/lib/github.ts` is gone. `formatRangeLabel` moves to `@superset/ui/lib/format-range-label` since the picker and the leaderboard both use it.

The admin tile reads through a `growth.starHistory` procedure with the usual 6-hour metric cache. Verified against the live repo: 324 daily points, 0 → 13,999 stars, monotonic, ~25 s for a cold walk (the tRPC route allows 60 s; cached for six hours after).

**Needs one thing from you:** `GITHUB_TOKEN` on the **API** Vercel project — the same read-only PAT the marketing project already has (GitHub's stargazers endpoint is closed even for public repos, and the name is reserved in Actions, which is why marketing carries it as a project env rather than through the deploy workflow). Until it is set the tile says `Unavailable: GITHUB_TOKEN is not set` rather than drawing a fabricated curve.

## Paywall funnel

HogQL in code (like the growth tiles), not a saved insight, so the definition is reviewable and versioned next to the tiles. Two tiles, one query (`usePaywallFunnel`): the stage funnel, and weekly cohorts over time.

**Verified against Production (264803), last 12 weeks:**

| stage | event | people |
|---|---|---|
| Saw the paywall | `paywall_opened` | 10,611 |
| Clicked upgrade | `paywall_upgrade_clicked` | 1,802 (17.0%, median 35 s later) |
| Started checkout | `checkout_started` | 0 |
| Paid | `subscription_started` | 4 |

## Assumptions about the paywall data model

- **One person, both sides.** The desktop identifies with the Better Auth user id, and Better Auth stamps that same id into Stripe metadata (`captureBillingEvent` in `packages/auth/src/lib/billing-analytics.ts`), so client events and webhook events land on one PostHog person and a person-level funnel needs no id mapping.
- **"Paid" is `subscription_started`, not `payment_succeeded`.** Only four Stripe webhook event types are configured, so the `invoice.paid` and `checkout.session.expired` handlers have never run in production — `payment_succeeded` and `checkout_abandoned` have zero rows and are not funnel steps here.
- **`checkout_started` is instrumented but not yet shipped** (added in #7338; it rides the desktop release), so it reads 0 today. Stages therefore count anyone who reached them *after* the paywall view rather than requiring the stage before it — otherwise an unshipped client step would zero out "paid" while people are visibly paying. The tile names any such empty stage in its description instead of letting it read as a conversion cliff, and `FunnelChart` no longer renders a negative drop-off when a later stage exceeds an earlier one.
- **Cohorts are open-ended:** membership is fixed by the first paywall view in the window, conversions keep accruing after, so trailing weeks are still maturing. Said in the tile copy, not just here.
- Attribution for organizations that subscribed outside the app falls back to the lowest-id owner (`attribution: "owner"` on the event) — inherited from #7338, not introduced here.

## Lingui (no repeat of the starchart Polish bug)

- Every new string is a `<Trans>` / `useLingui()` macro in a **client** component; stage labels are `msg()` descriptors translated at render, never a module-scope `t`.
- `formatUTCDate` stays pinned to `en-US` **on purpose**, and now says why: `/starchart` formats those dates from a server component that awaits GitHub, and the shared `i18n` instance is process-wide — a locale-aware formatter there is exactly the read-after-await that showed Polish. Revisit when server locale is bound per request.
- Moving components between packages preserved every message id (catalog diff is +42/-0 in `en`). All 14 new messages are translated into the 16 non-English locales; `bun run check:i18n` is green. Added a `paywall` row to the glossary since this change settles that term.

## Also: org adoption replaces the active-orgs insight

The `Active Organizations with 2+/5+ Users` tile (PostHog `IlEQoT55`) was measuring seats inside paid orgs, not team adoption — a free organization cannot have a second member, so "2+ active members" is a paid-org proxy with extra steps. It is now **Teams vs individual accounts**, computed in Postgres by `business.getOrgAdoption`, and it stays on the Product tab where org adoption belongs.

Every account gets a personal organization on signup, so an organization gaining a second member is the moment an account became a team — the same definition the growth page's `teams` series already uses. Per week: teams, individual accounts, and teams per 1,000 organizations.

The ratio is the point rather than the team count: team count alone rises with any growth, while the ratio only rises when teams outpace signups. Against production it does — **teams 82 → 125 (+52%)** over twelve weeks while **individual accounts went 48.2k → 60.0k (+24%)**, so the ratio moved **1.70 → 2.08 per thousand**. The chart gives the ratio the labelled axis and lets team counts ride along as bars, because 125 against 60,019 on one scale is a flat line at zero.

Membership is reconstructed from `members.created_at`; removing a member deletes the row, so a team that shrank back to one person is missing from the weeks it was a team — history biased down, never up, and the tile says so. `activeOrgs` leaves the insight registry with its last caller; the saved PostHog insight is untouched.

## Finding: the cloud database stopped recording workspaces on 2026-08-12

While sourcing the tile: `v2_workspaces` (last row 2026-08-12 15:27) and v1 `workspaces` (2026-08-12 08:01) both stopped taking rows the same day, and no replacement table exists. Everything else keeps writing — `auth.users`, `auth.sessions`, `tasks`, `automation_runs` and `subscriptions` all have rows from tonight — so this is specific to workspaces.

Consequence: the cloud database can no longer answer "was this organization active", and `workspace_created` in PostHog is now the only record that a workspace was created. Either the local-first move landing or a silent regression; worth a look either way, and not something this PR tries to fix.

## One bug this caught: the React Compiler runs before the Lingui macro

Worth knowing beyond this PR, and the same family as the starchart write-up. Screenshotting the tiles showed them rendering raw message ids (`_-Z5u4`, `01fO6B`) instead of copy, with a green typecheck, a green `check:i18n`, and a green build.

The admin app has `reactCompiler: true`. It rewrites a component before the Lingui SWC plugin sees it, and two rewrites break the macro:

- **A `t({ message: \`…${x}…\` })` whose result is stored in a variable** gets its template hoisted into a memo slot. The macro then has no literal to read: it emits an *empty* message — one shared id (`_-Z5u4`) for two different strings — and drops the values. Inline in JSX it survives, which is why the existing tiles were fine and mine were not.
- **A `const weeks = PAYWALL_FUNNEL_WEEKS` alias** gets const-folded, so the macro named the placeholder `{PAYWALL_FUNNEL_WEEKS}` while `lingui extract` (plain Babel, no React Compiler) named it `{weeks}`. Different message, different id, no catalog hit.

Fix in the second commit: interpolated copy goes through `<Trans>` JSX (read from the JSX tree, not a string literal), and the constant goes into the message directly. `description` widens from `string` to `ReactNode` on both tile frames. Verified in the built bundle, not by eye — every call site now emits an id the compiled catalog carries (`cfWk2W` + {stars, perDay, peakPerDay}, `jaon8D` + {gaps}, `01fO6B` + {PAYWALL_FUNNEL_WEEKS}) and `_-Z5u4` is gone.

Worth a lint rule or a note in `packages/i18n/README.md` for the React Compiler apps — happy to follow up if you want it.

## Verification

- `bun run typecheck` — 38/38 tasks pass.
- `bun run lint`, `sherif` — clean (sherif's one warning is the pre-existing `apps/review-host` note).
- `bun run check:i18n` — green, catalogs committed.
- `next build` for **admin** and **marketing** — both succeed.
- Full suite: desktop 3720, host-service 1703, marketing 139, cli 234, api 69, web 38, mobile 66, agent-setup 323 … all pass. `@superset/trpc` reports 4 failures **locally only** — three growth test files import the env module and `bun test` loads `.env` from the package dir, so it fails identically at `main` (checked) and passes in CI.
- Both new data sources were run against the real upstreams (numbers above).
- Ran the built admin app locally against the dev API and read the rendered DOM: descriptions resolve to real copy ("14k stars · 67/day this week · 419/day at peak"), all eight charts paint, and the funnel matches the numbers above.

https://claude.ai/code/session_01F2YrD1iP54m4XFyRnmB2Rz

## Summary by CodeRabbit

* **New Features**
  * Added admin dashboard insights for GitHub star history, paywall conversion trends, and organization adoption.
  * Added funnel reporting for paywall views, upgrades, checkout starts, and paid conversions across weekly cohorts.
  * Shared GitHub star charts now power both the admin dashboard and marketing star chart.
* **Bug Fixes**
  * Funnel drop-off metrics no longer show negative values or irrelevant tooltip rows.
* **Localization**
  * Added and updated translations for star, paywall, cohort, team, and analytics metrics across supported languages.


saddlepaddle:
Went through CodeRabbit's 21 findings. Taken, in `70cb778d6`:

- **Stargazer walk had no timeout** (the one Major). It used bare `fetch` while every sibling GitHub read goes through `fetchWithTimeout`. Fixed, but as a caller opt-in rather than always on: the API passes 15s per request, marketing keeps today's behaviour — it reaches the same fetcher through Next's fetch, where an AbortSignal risks the data cache, and a public page re-walking 140 stargazer pages per visitor is the worse failure.
- **Four locales collapsed two funnel stages.** `Started checkout` and `Paid` read as the same idea in es/it/id/tr ("Inició el pago" vs "Pagó", etc.), which defeats the tile that exists to separate them. Also swapped the masculine past-tense stage labels in cs/pl/ru ("Klikl", "Zaplatil") for ungendered noun phrases.
- **Ambiguous ratio copy.** "how many of every 1,000 organizations are one" could be read as the single-member count; now "teams per 1,000 organizations", with the 16 translations updated.

Not taken, with reasons:

- **Stricter funnel ordering** (a stage's first event *after* the paywall view, rather than its first in the window). Ran both against production: 1805/1805 upgrade clicks, 4/4 paid — no difference, because an upgrade click only exists inside an already-open paywall. Kept the simpler query.
- **Append-only membership history** to back the adoption series. That's a schema change and a backfill, well past this PR. The hard-delete limitation (a team that shrank back to one is missing from past weeks) is disclosed in the tile copy itself.
- **`.env` wording about the stargazers endpoint** and the `kohor`/`kohort` spelling: pre-existing text I only touched adjacently.

PR #7383, opened 2026-09-10
fix(mobile): hold Home's first paint until it has content, and keep it across launches

saddlepaddle:
On a cold start Home drew its chrome — search, the sort chip, the composer — over several seconds of empty list, and not even the empty-state text, which was gated behind the same unfinished queries. Nothing was broken; everything was still loading. It read as a broken app rather than a loading one.

Write-up with the frames: https://app.superset.sh/page/mobile-home-first-paint-t0ttpa

## Why it was empty

Four waits in series before the first row can draw: the session (seeded synchronously from SecureStore, so ~a tick), the AsyncStorage rehydrate of the saved host/sort pick (~165ms), one batched API round trip, and then `workspace.list` / `project.list` over the relay to the user's own laptop.

Underneath all four: nothing survived a launch. The app built a bare `QueryClient` with no persister, so every cold start re-derived the whole screen from the network. `useHostWorkspaces` documents a "cache-first rule", but that cache only lived in memory.

## Changes

**Home waits for an answer before it paints**, showing its mark until then. The gate lives in `HomeScreen` rather than the route above it — a gate in the route deadlocks, because the queries it would wait on are only started by HomeScreen's own hooks. It's latched, so an organization switch never blanks a list someone is already reading.

**The hold gives up after 1200ms**, because the last leg is unbounded: a host the API still calls online, reached over the relay, answers whenever it answers, and the tRPC client sets no timeout. Past the cap the list draws with a spinner rather than nothing — rendering nothing there is what made a still-loading Home read as a broken one, and "No projects" would be the same screen telling a lie.

**The queries first paint waits on are persisted** to AsyncStorage, matching desktop's `ElectronTRPCProvider`. The whitelist matches full key prefixes rather than just the first segment: sandbox access lives under `["cloud", "sandbox-access"]` and is a credential, so a first-segment match (as desktop does) would have written brokered sandbox URLs to disk. Verified against the blob actually written: 8 queries, 164KB, with `sandbox-access`, `diff-stats` and terminal lists absent. `useSignOut` already calls `queryClient.clear()`, which wipes the disk copy too.

`gcTime` is raised to 24h per persisted prefix rather than globally — a query evicted from memory is absent from the next write, but holding a day of per-row diff stats on a phone is what the 5-minute default protects against.

## Verification

On an iPhone 17 Pro Max simulator against production. Same build, same device, only the persisted cache differing:

- **With cache:** splash → rows directly, no intermediate state.
- **Cache cleared:** splash → chrome and a spinner, still spinning seconds later.

Full package suite green (66 tests), lint clean, `check:i18n` clean, sherif clean, and `turbo typecheck --filter=@superset/mobile` reports nothing in `apps/mobile` (the 24 remaining errors are pre-existing, in `packages/port-scanner`, `packages/pty-daemon` and `packages/workspace-fs`).

## What this does not claim

The splash did not get shorter — ~1.8s in both runs. Restored data was expected to satisfy `contentReady` inside the 1200ms cap and didn't, and that's unexplained. The win is that the first content frame is complete rather than a spinner, not that the wait shrank. All timings are from a dev bundle over Metro; a release build is untested.

## Follow-up, not included

`useNewChatTargets` requests the same `project.list` from the same host under a different key than `useHostProjects` (`[…machineId]` vs `[…machineId, hostUrl]`). It predates this change, but it's a duplicate network call and now 66KB of the 164KB stored twice. The fix is to have it call `useHostProjects`, which it nearly duplicates — left out because it sits in the composer's target path and deserves its own change.

https://claude.ai/code/session_01WiJ7HCo2cVHYvyvQqRBpr2

---
## Summary by cubic
Fixes Home's cold start so it shows one splash from launch until it has content — previously boot was blank, then the mark, then several seconds of chrome over an empty list — persists the queries first paint waits on so returning launches open on the last rows, and stops the composer animating in from the origin on mount.

**Bug Fixes**
- Adopts `expo-splash-screen` with the mark on a dark ground from process start; the hold and its 5s backstop live in `app/_layout`, and RootLayout hides the splash immediately for any landing other than Home.
- Home holds its first paint until content is ready, giving up after 1200ms to draw a spinner rather than a misleading empty state; the gate is latched, so organization or host switches never blank an already-drawn list.
- With no active organization the hosts query is disabled and stays pending forever, which kept the gate closed and left the list on a permanent spinner; that case is now treated as ready.
- The queries first paint waits on persist to AsyncStorage, so a returning launch skips the round trip; the whitelist matches full key prefixes so `sandbox-access` credentials stay off disk.
- `gcTime` is 24h for persisted prefixes only, so per-row decoration like diff stats and terminals still evict at the 5-minute default.
- Home no longer refetches on initial focus, only on re-focus, so a cold start doesn't fetch every query twice.
- The composer lays out on mount before any prop animation, so its card no longer travels in from the origin; navigating back re-arms the behavior.

<sup>Written for commit c1432927270880cf110129d3d4999e03850ec860. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added offline persistence for selected mobile app data for up to 24 hours.
  * Added a configured splash screen with a dark background and branded splash mark.

* **Bug Fixes**
  * Improved startup loading so the app waits for content before displaying the home screen.
  * Added loading feedback when content is delayed.
  * Prevented the splash screen from remaining visible indefinitely during startup.
  * Reduced unnecessary data fetching when returning to the home screen.


saddlepaddle:
Thanks — one of the two landed.

**Disabled hosts query (`contentReady`)**: real, and fixed in 6da2759. With no `activeOrganizationId` the query is disabled and `isPending` stays true for the life of the screen, so `contentReady` never flipped: the 1200ms cap painted the list and `ListEmptyComponent` showed its spinner permanently, where before this branch the empty text was shown. `useCloudWorkspaces` already treats a null organization as ready for exactly this reason, so the fix follows that precedent rather than inventing one.

**`lib/splash.ts` hide-retry**: no longer applicable — that file was deleted in a656f35. `hideAsync` is already idempotent, so the wrapper only added a flag and a timer; the hold and its backstop now sit in `app/_layout.tsx` and callers invoke `SplashScreen.hideAsync()` directly. The retry concern doesn't survive the deletion: there's no `hidden` flag to latch, and the backstop is an independent `setTimeout` that fires regardless of whether an earlier call rejected.

PR #7411, opened 2026-09-10
feat(automations): let an automation continue its own agent session

saddlepaddle:
A workspace pin means "reuse this workspace every run" — the CLI documents `--workspace` as exactly that. But only the *directory* was ever reused. `agents.run` mints a fresh terminal on every call, and a TUI agent never exits; it sits at its prompt after a turn. So a five-minute schedule left a live agent per tick in one worktree, each with its own context, all editing the same files.

`continueAgentSession` opts an automation into delivering each run's prompt into the session its own previous run left behind. **Off by default**, and only valid with a pinned workspace, which is where that session lives.

## Whose session it continues

The automation's **own** previous terminal, read from `automation_runs.terminal_session_id` — never "any live agent of this kind in the workspace." Someone may be working in there too, and a scheduled prompt must never land in a session they started.

## What the host refuses

The cloud only *nominates* a terminal; `continuationTarget` decides. It declines a terminal with no live binding, one bound to a different agent, and any call carrying launch-only options (model/effort/mode/resume/fork/attachments) that a running process cannot adopt. Each falls through to a normal launch rather than failing the run, so a stale nomination costs exactly what today costs. Older hosts strip the unknown key and launch.

Two properties this rests on, both verified rather than assumed:

- **`listByWorkspace` is live-only** — it routes through `listLiveByWorkspace`, and the in-memory fallback reads `byTerminal`, which `endBinding` deletes from.
- **The send adopts but never respawns** — `getOrAdoptSession` passes `adoptOnly: true`. This is the one that matters: a respawn would type the automation's prompt into a bare shell as a command. A dead pty errors and we launch instead.

Known limitation, pre-existing to the send mechanism: an agent that dies *without* firing its end hook leaves a live-looking binding, and a prompt could reach the shell behind it. The opt-in keeps that risk with automations that asked for this.

## Explicit failure reasons

The desktop picked which guidance to show by grepping the failure text — `"No host agent config matching"`, `"target host offline"` / `"no host available"` — and host-service pinned its wording in a comment to keep that working. Three packages coupled through prose, and it breaks the moment those strings are translated.

Runs now carry `error_code`, and the live "run now" path carries the same code on the thrown error beside the message it cannot translate. Both matchers are gone, and the guidance moved to one table keyed on the code — which also covers two failures nothing explained before: a pinned workspace that no longer exists, and an automation with no instructions.

`host_offline` is redundant with `status = 'skipped_offline'` and kept anyway, so a client switches on one field rather than two.

## Prior art

PR #7341 diagnosed this as a bug and proposed making it automatic for every pinned automation; it was closed unmerged, deliberately — the answer to *runaway* sessions is rate and spend limits with the cloud move. This is the opt-in feature instead, and it takes that branch's key insight (own previous terminal, not any agent in the workspace).

## Testing

`continuationTarget` is split from the send so the policy is testable without a pty — with no pty every path returns null, which is how the first version of these tests passed **with the safety guard deleted**. Mutation-checked: removing the agent-identity comparison now fails exactly one test. That needed a second fix too — the test originally seeded only a claude config, so a codex binding resolved to `null` and the `!bound` half was doing the rejecting rather than the comparison.

Host suite 1724 pass, trpc 274, CLI 237. Typecheck, lint, `check:i18n` and sherif green. Migration 0112 is additive (enum + two columns), safe for a rolling deploy.

https://claude.ai/code/session_01MXaz3nDCA1BDNJ3pChLkNV

---
## Summary by cubic
Lets an automation deliver each scheduled run's prompt into the agent session its own previous run left, instead of starting a fresh agent beside it. Also replaces the desktop's English-message matching for run failures with a structured `error_code`.

**Session continuation**
- `continueAgentSession` is off by default and only valid with a pinned workspace.
- Only the automation's own previous terminal is continued; a scheduled prompt never lands in a session a person started.
- The host falls back to a normal launch when the terminal is gone, bound to a different agent, or the call carries launch-only options.
- Continued prompts are sanitized like launch prompts, so embedded paste-end sequences can't inject keys.
- Unpinning drops the flag in the editor and on stale-pin recovery, and the CLI refuses `--session` with `--continue-session` before any mutation.
- Exposed via a new desktop picker, `--continue-session` CLI flags, and MCP/SDK fields.

**Error codes**
- Stored runs and the live "run now" path now carry `error_code`; clients branch on it instead of grepping message prose.
- Covers two failures that had no guidance before: a deleted pinned workspace and an automation without instructions.
- Migration 0112 is additive (enum + two columns) and safe for a rolling deploy.

<sup>Written for commit a548038bf6ddf41c8a457e24dce2643c93911f17. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7411?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Automations can continue the agent session left by the previous run or start a fresh session each time.
  * Added session-mode controls across the dashboard, CLI, SDK, and automation tools.
  * Continuing a session requires a pinned workspace.
  * Eligible runs can reuse an existing terminal session.
  * Automation creation, manual runs, retries, and resuming paused automations now require the Pro plan.

* **Bug Fixes**
  * Improved localized guidance for offline hosts, missing agents, invalid workspaces, and missing instructions.
  * Session continuation settings reset when the selected workspace, project, or host changes.


saddlepaddle:
Review triage — 14 of 15 applied, one declined with reasoning.

**Applied**
- `TriggersCard`: every path that clears `v2WorkspaceId` now clears `continueAgentSession` too. The draft is sent whole on save, so a stale `true` was refused server-side and the automation couldn't be saved after unpinning. Real bug.
- CLI `update`: `--session --continue-session` is rejected before any mutation, so `--enabled` can no longer land while the update fails.
- **CWE-77, prompt sanitisation** — real, and a regression from this PR, not pre-existing. Automation prompts embed third-party content through `promptWithTriggerContext` (an email body, a PR title). The launch path runs that through `sanitizePromptForPty`; the continuation path wrote it raw into a bracketed paste, so a `\x1b[201~` inside an email would have closed the frame and injected keystrokes into the agent. Now sanitised on both paths.
- All 11 locale findings: consistency with each catalog's established terms (ピン留め, 智能体, área de trabalho, `prompt` in fr, infinitive in nl, etc.). I'd translated against the glossary rather than each locale's existing entries.

**Declined: CWE-862, terminal ownership**
This PR adds no capability the host doesn't already expose. `terminal.send` is `protectedProcedure` with no owner check and writes into any terminal by id — used today by the CLI, the MCP server and page-watch. `continuationTarget` is strictly narrower than that route: it additionally requires a live binding *and* an agent-identity match. The host-service model is "authenticated to this host = may act on its terminals" (`v2UsersHosts` grants that per member), and changing it belongs to that model, not to this feature. Worth its own discussion if the member role should be tightened.

saddlepaddle:
Re-review finding applied: the stale-pin recovery in `dispatch.ts` now clears `continueAgentSession` alongside `v2WorkspaceId`. Dispatch itself was already safe — its guard requires both the flag *and* a pin — but the row was left in a state the API's own `update` validation refuses, and the UI would have read "continuing the last session" against nothing pinned. Same rule as the client fix: the flag goes with the pin, everywhere the pin goes.

Also merged `main` — #7434 (Pro gating) extended `I18nErrorCause` / `formatError` with `requiredPlan` using the same shape this PR uses for `automationErrorCode`; both kept.

PR #7412, opened 2026-09-10
fix(desktop): clamp negative leaderboard delta after stale standings

Kitenite:
## Summary
- Clamp the leaderboard's rank-neighbor deltas ("N to pass #rank" / "N behind you") at zero so they can never render as negative.

## Why / Context
Reported in Discord: https://discord.com/channels/1446776342577283114/1547717959516950689/1547717982807920707 — the desktop Settings → Usage leaderboard card could show a negative number like "-42.3K to pass #7".

`RankNeighbors` computes those deltas as `above.tokens - me.tokens` and `me.tokens - below.tokens`. `me.tokens` comes from a live, uncached tRPC query (`leaderboard.me`), but the neighbor rows come from `leaderboard.public.standings`, which is CDN-cached (`stale-while-revalidate`, up to 5 min per `apps/api/src/app/api/trpc/[trpc]/route.ts`). Right after a usage publish, the live total can momentarily exceed a stale cached neighbor's total (or a stale neighbor can appear to have overtaken the live total), flipping the subtraction negative. `formatTokens`/`formatScaled` have no negative-number handling, so the sign passed straight through to the UI.

## How It Works
Wrap both subtractions in `Math.max(0, ...)`. Once the standings cache catches up (≤5 min later), the numbers self-correct back to accurate deltas; in the interim the UI just shows 0 instead of a negative value, rather than inventing a new UI state for the stale-cache window.

## Manual QA Checklist
- [x] Reasoned through both flip cases (`above.tokens <= me.tokens`, `below.tokens >= me.tokens`) — both now clamp to 0 instead of negative.
- [ ] Not manually reproduced live in the app (would require timing a usage publish against the CDN cache window); relying on code-level verification below.

## Testing
- `bunx tsc --noEmit -p apps/desktop/tsconfig.json` (no errors in changed file)
- `bunx biome check` on the changed file (clean)
- `bun run check:i18n` (passes, 0 missing across all 16 locales — no new translatable strings, only the auto-generated placeholder-expression comments in `messages.po` updated to reflect the new `Math.max(0, ...)` source)

## Known Limitations
- This clamps the display, not the underlying rank ordering — if the cached standings snapshot is stale enough that "above"/"below" labels themselves are wrong (not just the delta), that's unaffected by this fix and self-corrects when the cache refreshes.

---
## Summary by cubic
Clamps the leaderboard's "N to pass #rank" and "N behind you" deltas at zero so they can no longer render negative values like "-42.3K to pass #7". Previously, the live per-user token total could momentarily exceed a CDN-cached standings snapshot (stale up to 5 minutes), flipping the subtraction negative; now the UI shows 0 until the cache catches up.

**Bug Fixes**
- Both delta calculations now use `Math.max(0, ...)`.
- The clamp only affects the display; underlying rank ordering is unchanged and self-corrects when the cache refreshes.
- No new translatable strings — `messages.po` files only update placeholder comments.

<sup>Written for commit bc9d5eaa603876b3d850f513d62dea2a6cb105f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Leaderboard neighbor comparisons no longer display negative token counts for “behind you” or “to pass” values when standings are outdated.

* **Localization**
  * Updated leaderboard placeholder metadata across supported languages to reflect the corrected non-negative token counts. Translation text remains unchanged.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/settings/usage/components/LeaderboardCard/components/LeaderboardRank/components/RankNeighbors/RankNeighbors.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/settings/usage/components/LeaderboardCard/components/LeaderboardRank/components/RankNeighbors/RankNeighbors.tsx
@@ -74,7 +74,7 @@ export function RankNeighbors({ me, rows }: RankNeighborsProps) {
 			label: <Trans>You</Trans>,
 			detail: above ? (
 				<Trans>
-					{formatTokens(above.tokens - me.tokens)} to pass #
+					{formatTokens(Math.max(0, above.tokens - me.tokens))} to pass #
 					{formatNumber(above.rank)}
 				</Trans>
 			) : null,
@@ -85,7 +85,11 @@ export function RankNeighbors({ me, rows }: RankNeighborsProps) {
 			tokens: below.tokens,
 			label: aliasFor(below.rank),
 			detail: (
-				<Trans>{formatTokens(me.tokens - below.tokens)} behind you</Trans>
+				// Clamped: standings are CDN-cached, so right after a publish
+				// the live "me" total can already have overtaken a stale neighbor.
+				<Trans>
+					{formatTokens(Math.max(0, me.tokens - below.tokens))} behind you
+				</Trans>
 			),
 			isMe: false,
 		},
--- a/packages/i18n/locales/cs/messages.po
+++ b/packages/i18n/locales/cs/messages.po
@@ -310,7 +310,7 @@ msgstr "Akce {0}"
 msgid "{0} and {1} settle it as terminal dinosaurs. Stats decide the winner."
 msgstr "{0} a {1} si to vyříkají jako termináloví dinosauři. O vítězi rozhodnou statistiky."
 
-#. placeholder {0}: formatTokens(me.tokens - below.tokens)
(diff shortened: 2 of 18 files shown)


Kitenite:
Note for follow-up: this clamps the display-level symptom (negative delta), not the root cause — \`me.tokens\` is a live/uncached query while \`above\`/\`below\` come from a CDN-cached standings snapshot (up to ~5 min stale), so the two can momentarily disagree. The smallest true fix would be to prefer the caller's own row from the already-fetched \`rows\` window (same cached source as the neighbors) for the delta calc, falling back to live \`membership.tokens\` only when absent from \`rows\` — keeping the "You" row's displayed total live, but making the delta math internally consistent with its neighbors. Deferring that for now since it's out of scope for the reported bug (negative number), which this PR fully resolves.

PR #7427, opened 2026-09-11
refactor(pages): one page-comment store for web, desktop and mobile

harshithmullapudi:
## Summary

- Collapses three near-identical page-comment stores (web, desktop, mobile) into one `@superset/cloud-client` package, with the shared types moved to `@superset/shared/page-comments` and server shaping extracted to `trpc/router/page-comment/shape.ts`.
- Opens the comment composer the moment a block is selected, with the quick actions on the card's top row; deletes the standalone `SelectionToolbar` and `toolbarPlacement` on both surfaces.
- Fixes the hover highlight surviving a scroll, and two defects found while reviewing this branch.

## Why

Each surface had grown its own optimistic React Query store against the same `pageComment` router — 247 lines on web, 208 on desktop, 112 on mobile — so every change to comment behaviour had to be made three times and kept in step by hand. They had already drifted. One store, one set of optimistic row helpers, one place to fix a bug.

## How it works

`@superset/cloud-client` owns the store. `usePageComments` runs the optimistic mutations (create / reply / edit / resolve / delete) over pure row helpers in `threadRows` and `optimisticRows`, each mutation snapshotting the cache in `onMutate` and restoring it in `onError`. `CloudClientProvider` supplies the caller, so each app only has to hand it a tRPC client: web and desktop through their existing providers, mobile through `createCloudCaller(apiClient)`.

Server-side, `shapeThread`/`shapeComment` are now the single definition of the wire shape, and `create`/`reply` return the shaped row rather than bare ids so the client can swap its placeholder for the real row without a refetch.

## Review findings fixed in this branch

Two defects found reviewing the first two commits, each fixed in its own commit:

1. **`reply` reported the wrong failure** (`page-comment.ts:270`) — the new empty-insert guard reused `failedToCreateThread`, so a failed reply told the reader their *thread* could not be created, about a thread already on screen. Added `serverError.pageComment.failedToPostReply` with translations for all 16 non-English locales.
2. **The store was rebuilt every render** — `usePageComments` memoised on the `useMutation` results, which TanStack Query recreates each render. The memo never held, so `CommentProvider`'s callbacks and context value churned and re-rendered every consumer. Now destructures the stable `mutateAsync` references.

## Behaviour change worth flagging

`pageComment.reply` now returns a full shaped comment instead of `{ id }`. Two callers outside the apps pass the result straight through, so their output gains `body`, `authorName`, `authorImage` and `createdAt`:

- `packages/cli/src/commands/pages/comments/reply/command.ts` → `superset pages comments reply --json`
- `packages/mcp/src/tools/pages/comments/reply.ts` → the MCP tool result

Additive — `id` is still present, so anything reading it keeps working — but it is an agent- and CLI-facing contract change.

## Testing

- `packages/cloud-client` 14/14, `packages/shared` 1021/1021, `packages/ui` 71/71
- `typecheck` clean: `cloud-client`, `shared`, `ui`, `i18n`, `trpc`, `apps/web`, `apps/desktop`
- `bun run check:i18n` — 5358 messages, 0 missing across all 17 locales, `compile --strict` passes
- `bunx sherif` — 0 errors
- `biome check` clean on every file touched

## Known limitations

- `usePageCommentThreads` sets `isLoading` from `list.isPending`, which stays true while the query is disabled (`version > 0`). A page at version 0 would spin forever. No caller appears to pass 0, so this is likely unreachable — left as-is rather than guessed at.
- Concurrent mutations each snapshot the whole cache, so if two are in flight and the first fails, its rollback discards the second's optimistic row. Pre-existing in all three stores this replaces; not made worse here.
- Manual QA covered the desktop pane only (hover, scroll, comment, reply). Web and mobile were not exercised by hand.

## Pre-existing failures, not from this branch

- `packages/trpc` `lib/growth/sitemap.test.ts` fails at module load on env validation; passes under `dotenv -e ../../.env`.
- `apps/mobile` typecheck reports 5 errors for `importFromCloud`, `materializeIntoWorkspace`, `customTitle` and `rename`. Those procedures do not exist in `@superset/trpc` at all, none of the files are touched here, and this branch's only mobile change is an additive `export const cloudCaller`. Worth confirming `main` is already red before landing.

https://claude.ai/code/session_01YQXXfpsXYTcQwj6PiSptGm

---
## Summary by cubic
Consolidates web, desktop, and mobile page comments into one optimistic data store in `@superset/cloud-client`, replacing three near-identical stores that had drifted apart. Selecting a block now opens the comment composer immediately, with quick actions on the card's top row.

Apps hand `CloudClientProvider` their plain tRPC client (React, vanilla, or mobile proxy), and the hooks call `client.pageComment.*` directly, so the package has both runtime and type-only independence from the router. The session-to-user mapping is folded into `pageCommentUser()` in `@superset/shared/page-comments`.

**Behavior changes**
- `pageComment.create` and `reply` now return the shaped row instead of bare ids, so the optimistic placeholder is replaced without a refetch.
- `reply` now returns `body`, `authorName`, `authorImage`, and `createdAt` in addition to `id`; this changes the `superset pages comments reply --json` output and the MCP tool result. The author now comes from the session instead of a post-insert lookup, removing a duplicate-on-retry ordering hazard.
- Failed replies now show a "Failed to post reply" message, translated in all 17 locales, instead of the thread-creation error.
- Fixed the store being rebuilt on every render, which re-rendered every consumer via the provider.
- Fixed hover highlighting: the outline dismisses on scroll, and its geometry no longer flashes at the previous position for a frame.
- Mobile now sends the comment intent (`SUPER-2237`) and clears a stale pending selection after the composer is dismissed.
- Desktop reads through `cloudTrpcClient` instead of `apiTrpcClient`, keeping queries scoped to the window's organization; the org-switch purge and 30s stale time now recognize the store's flat `cloud` query keys.
- Mutations now drain per page and invalidate the list once the last in-flight settles, and rollback removes the query when nothing was cached yet.

**Notes for review**
- Tests now cover the optimistic rows, cache transitions, and the in-flight settlement race.
- Manual QA covered only the desktop pane; web and mobile were not exercised by hand.
- Pre-existing, unrelated failures: `packages/trpc` sitemap test fails without env vars, and mobile typecheck reports 5 errors for procedures missing from `@superset/trpc`.

<sup>Written for commit 59fad8816cee41c7b203ac8d6d5ac1ca15a4cda4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7427?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added consistent page comments across web, desktop, and mobile.
  * Added quick feedback actions for approvals, deletion requests, and preset responses.
  * Mobile selections now open the composer with the selected page anchor.
  * Added localized error messaging when replies fail to post.

* **Bug Fixes**
  * Improved optimistic comment updates, refresh behavior, and rollback handling.
  * Improved comment hover outlines and selection highlighting.
  * Hover indicators now clear correctly while scrolling.
  * Improved comment refresh when returning to pages or switching collections.


saddlepaddle:
The consolidation is the right move and the coordination inside it is correct: `isMutating` scoped by `meta` for freshness, `create`/`reply` returning the shaped row so the placeholder swap needs no refetch, pure tests on the row helpers. Keep all of that.

The layer to remove is `createCloudCaller`. Every call site already hands it a vanilla tRPC client:

```
desktop  createCloudCaller(cloudTrpcClient)
mobile   createCloudCaller(apiClient)
web      createCloudCaller(trpcClient)
```

The vanilla client is already binding-agnostic; that is what it is for. So `PageCommentProcedures`, `PageCommentCaller`, `CloudCaller` and the wrapper exist only to rename `.query()` / `.mutate()` into a flat call, at the cost of three hand-listed procedure tables that all change to add a seventh procedure, where tRPC's proxy needs zero.

Built and verified on a branch off this head: `cloud-client`, `shared`, `desktop` and `web` typecheck clean, the same 23 tests pass, net **-43 lines**, two files and two interfaces deleted. Everything below is complete enough to apply as-is.

### 1. The provider carries the client itself

`packages/cloud-client/src/providers/CloudClientProvider/CloudClientProvider.tsx`, whole file. `AppRouter` and `TRPCClient` are type-only, so the package gains no runtime dependency on the router.

```ts
import type { AppRouter } from "@superset/trpc";
import type { TRPCClient } from "@trpc/client";
import { createContext, type ReactNode, useContext } from "react";

const CloudClientContext = createContext<TRPCClient<AppRouter> | null>(null);

export function CloudClientProvider({
	client,
	children,
}: {
	client: TRPCClient<AppRouter>;
	children: ReactNode;
}) {
	return (
		<CloudClientContext.Provider value={client}>
			{children}
		</CloudClientContext.Provider>
	);
}

export function useCloudClient(): TRPCClient<AppRouter> {
	const client = useContext(CloudClientContext);
	if (!client) {
		throw new Error("useCloudClient must be used within CloudClientProvider");
	}
	return client;
}
```

Delete `src/lib/createCloudCaller/`, remove its export from `src/index.ts`, and remove `PageCommentCaller` and `CloudCaller` from `src/types.ts`. The `ServerThread` / `*Args` types there stay; they are inferred from the router and are fine.

### 2. The hooks call tRPC directly

`const client = useCloudClient();` then, inputs and outputs inferred:

```ts
queryFn:    () => client.pageComment.list.query({ pageId }),        // usePageCommentThreads
mutationFn: (input) => client.pageComment.create.mutate(input),     // and reply, edit, resolve, delete
```

### 3. Each app hands over the client it already has

```diff
- const [caller] = useState(() => createCloudCaller(trpcClient));
- <CloudClientProvider caller={caller}>
+ <CloudClientProvider client={trpcClient}>
```

Same shape on desktop (`ElectronTRPCProvider`, `client={cloudTrpcClient}`) and mobile (`RootLayout`, `client={apiClient}`), and the `cloudCaller` exports in `cloud-trpc.ts` and `lib/trpc/client.ts` go. Desktop's is the one worth knowing about: `cloudTrpc.createClient()` is the react-query binding's factory, but a probe assigning it to `TRPCClient<AppRouter>` compiles, so it is the proxy in v11.

### 4. One user derivation instead of three

The `{ id, name ?? "You", image }` block is byte-identical in `PageViewer.tsx`, `page.tsx` and mobile's `usePageCommentUser`. As a flat file beside the type it produces (this directory uses flat files, not folders):

```ts
// packages/shared/src/page-comments/pageCommentUser.ts
import type { PageCommentUser } from "./types";

export function pageCommentUser(
	session: { user: { id: string; name?: string | null; image?: string | null } } | null | undefined,
	fallbackName: string,
): PageCommentUser {
	return {
		id: session?.user.id ?? "",
		name: session?.user.name ?? fallbackName,
		image: session?.user.image ?? null,
	};
}
```

The structural `session` type is deliberate: importing one app's `Session` would break the other two. Export it from `page-comments/index.ts`, then:

```diff
- user={{ id: session?.user.id ?? "", name: session?.user.name ?? i18n._(msg({ message: "You" })), image: session?.user.image ?? null }}
+ user={pageCommentUser(session, i18n._(msg({ message: "You" })))}
```

`user` stays a hook argument rather than moving onto the provider, on purpose: every app mounts `CloudClientProvider` above auth, and on web the user is derived server-side in `page.tsx`, so there is no session at the root mount. The client is app-lifetime; the user is page-lifetime. Mobile's `usePageCommentUser` becomes a four-line binding of `useSession()` to this function, which is a fine reason for it to exist.

### 5. The test fake becomes a fake tRPC client

Only `makeCaller` and `setup`'s mount line change; the three `makeCaller({ create: … })` call sites do not. Note the second parameter, which is easy to drop:

```ts
function makeCaller(
	overrides: Partial<PageCommentFns> = {},
	serverRows: ServerThread[] = [],
): TRPCClient<AppRouter> {
	const fns: PageCommentFns = { /* the six existing defaults */ ...overrides };
	return {
		pageComment: {
			list: { query: fns.list },
			create: { mutate: fns.create },
			reply: { mutate: fns.reply },
			edit: { mutate: fns.edit },
			resolve: { mutate: fns.resolve },
			delete: { mutate: fns.delete },
		},
	} as unknown as TRPCClient<AppRouter>;
}
```

### Things that bite

- Add `"@trpc/client": "11.16.0"` to `packages/cloud-client/package.json` and re-run `bun install`. Without it the type import fails with "Cannot find module", which looks like a stale install rather than a missing dep.
- `apps/web/src/app/page/[slug]/page.tsx` has no `i18n` import to anchor on; it gets `i18n` from `initServerI18n()` in the body. Add `import { pageCommentUser } from "@superset/shared/page-comments";` after the `@lingui/core/macro` line.
- Done when: `cloud-client`, `shared`, `desktop`, `web` typecheck clean and `cloud-client` reports 23 pass. Mobile's typecheck currently fails on an unrelated `@superset/host-service/router` resolution in files this does not touch.

packages/cloud-client/src/hooks/usePageComments/usePageComments.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+		{ wrapper },
+	);
+	return { result, queryClient };
+}
+
+function rows(queryClient: QueryClient): ServerThread[] | undefined {
+	return queryClient.getQueryData<ServerThread[]>(queryKey);
+}
+
+describe("usePageComments", () => {
```
saddlepaddle, line 126: Worth keeping this file: three of the four tests pin React Query lifecycle ordering that cannot be expressed as a pure-function test, and "does not strand the optimistic row when the cache was never populated" is the `previous === undefined` rollback no-op that shipped in #7312. That one is the reason it cannot come back.

The gap is the thing the `isMutating` counter exists for. Nothing here fires two sends before the first settles and checks that the invalidate waits for the last one. That is the race the `=== 1` boundary guards, and it fails silently in both directions: flip it to `=== 0` and the list just stops refreshing, with no error anywhere and all four of these still green.

A fifth test that resolves `create` twice with a deferred promise, settles the first, and asserts no invalidation happened until the second settles would cover it. Everything it needs is already in `makeCaller` and `setup`.

PR #7435, opened 2026-09-11
fix(auth): validate the desktop sign-in callback target

PascalOrdano18:
The desktop sign-in handoff builds its redirect out of query params — `protocol` on `GET /api/auth/desktop/connect`, and `desktop_protocol` / `desktop_local_callback` on the web success page — and used them as given, so the redirect target was not necessarily the desktop app.

Both ends now validate before the redirect is built:

- **apps/api** — `protocol` has to be a scheme the desktop registers (`superset`, `superset-dev`, `superset-<workspace>`). Anything else answers 400.
- **apps/web** — same scheme allowlist, and the local callback has to be `http://127.0.0.1|localhost/auth/callback` with no credentials, query or fragment. Rejected values render the error page the route already has.

Unit tests cover the accepted and rejected shapes on both sides.

Part of an audit pass over the v2 stack on macOS.

https://claude.ai/code/session_01HmVz1yzkCEfnfDYDQxjN8u

---
## Summary by cubic
Fixes the desktop sign-in handoff so the session token can only be redirected to the desktop app, not an attacker-controlled URL.

- API now requires `protocol` to be `superset`, `superset-dev`, or `superset-<workspace>`; anything else answers 400.
- Web applies the same scheme allowlist to `desktop_protocol`, and `desktop_local_callback` must be `http://127.0.0.1|localhost/auth/callback` with no credentials, query, or fragment; rejected values render the existing error page.
- Unit tests cover accepted and rejected shapes on both sides.

<sup>Written for commit f63bc0c3acbad77106e89401a5dfde68ff3c5fc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved desktop sign-in security by accepting only supported desktop protocol schemes.
  - Added validation for local authentication callback URLs, blocking malformed, external, or unsafe destinations.
  - Authentication now reports a clear failure instead of proceeding when protocol or callback details are invalid.
  - Callback tokens are handled through validated URL parameters, reducing the risk of unintended redirects.

- **Tests**
  - Added coverage for supported protocols and invalid callback targets, including external hosts, unsafe schemes, credentials, and malformed URLs.

### DIFF
--- a/apps/api/src/app/api/auth/desktop/connect/desktopProtocol.test.ts
+++ b/apps/api/src/app/api/auth/desktop/connect/desktopProtocol.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, test } from "bun:test";
+
+import { isDesktopProtocol } from "./desktopProtocol";
+
+describe("isDesktopProtocol", () => {
+	test("accepts the schemes the desktop registers", () => {
+		expect(isDesktopProtocol("superset")).toBe(true);
+		expect(isDesktopProtocol("superset-dev")).toBe(true);
+		expect(isDesktopProtocol("superset-my-worktree2")).toBe(true);
+		expect(isDesktopProtocol("Superset")).toBe(true);
+	});
+
+	test("rejects anything that could redirect the token elsewhere", () => {
+		expect(isDesktopProtocol("https")).toBe(false);
+		expect(isDesktopProtocol("https://attacker.example/?x=")).toBe(false);
+		expect(isDesktopProtocol("superset://evil")).toBe(false);
+		expect(isDesktopProtocol("superset-")).toBe(false);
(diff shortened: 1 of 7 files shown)


Kitenite (approved):
Reviewed and verified locally: tests + typecheck pass, and the behavior was reproduced end-to-end. Small, correct, solves a real problem.

PR #7442, opened 2026-09-11
feat(relay,desktop,trpc): presence from the relay, roster never polled

saddlepaddle:
## Why

The desktop polled `v2Host.list` every 30s per window. That procedure minted a JWT, called the relay's `/presence`, and the relay called back into `host.checkAccess` once per host on a cache miss: the API asking itself through Cloudflare. It was 27% of api requests plus a share of the 14% access checks on the Vercel spend page.

## What

**Roster and presence are separate now.**

- `v2Host.list` returns membership only, takes the organization id as input (org-scoped query key), is fetched on mount, focus and reconnect, and is persisted through the query client's IndexedDB persister. The hand-rolled `useKnownHosts` snapshot from the Electric era is deleted.
- Presence is the state of each host's relay-routed event socket. `useKnownHosts` holds one open per roster host; the relay preflight distinguishes not-connected (503) from denied (403), and an offline host is re-probed every 30s instead of on partysocket's seconds-scale backoff.
- Every desktop consumer reads `useKnownHosts()`; the presence polling hook is deleted. The host settings page keeps "last seen" through one on-demand relay read.

**Relay**

- The access decision is cached in the host's Durable Object storage keyed by user, so every isolate and colo shares one API call per (user, host) per TTL. A host that has never connected has no object and answers offline with no access check at all.
- `/presence` trusts API-minted scoped tokens (`automation-presence`, `host-presence`), so server-side reads skip access checks.

**Also**

- Host-service keeps the relay URL across reconnects and re-asks only after a connect attempt fails. Desktop holds it for the life of the window.
- Mobile merges relay presence in `useOrgHosts`; `isOnline` no longer comes from the API.
- The automation failure badge polls every 120s instead of 30s; the automations pages every 60s.

## Verified

Local dev stack (wrangler dev relay + api + desktop) driven over CDP:

- One `v2Host.list` fetch per load, none over the following 100s. Zero `/presence` calls from the renderer.
- Offline roster hosts re-probed at 33s, 64s, 96s; one event socket for the online host; never-placed hosts answered 503 in a few ms with no API call.
- Roster present in the persisted IndexedDB cache; Hosts settings shows online/offline groups from socket state; offline host page renders with one on-demand presence read.
- Durable Object cache: after an isolate reload with the front cache cold, the probe returned 200 with no `host.checkAccess` on the API and no "access check via api" log line, while the first cold probe did go to the API once.

Suites: desktop 3751 pass, host-service 1720 pass, mobile 91 pass, workspace-client 29 pass, trpc 253 pass (4 pre-existing `sitemap.test.ts` env failures).

Plan: `plans/20260911-remote-hosts-realtime.md` (local, not committed). PR 2 adds the realtime nudge channel.

https://claude.ai/code/session_01LrqbaXvTkvex2tarYQPd59

---
## Summary by cubic
Separates host roster membership from live presence so the desktop and mobile stop polling the API every 30s, which was driving 27% of API requests and repeated access checks.

**Roster and presence**
- `v2Host.list` now returns membership only, is org-scoped by input, fetched on mount/focus/reconnect, and persisted via the query client's IndexedDB persister.
- Desktop and mobile read presence from each host's relay event socket; offline hosts are re-probed every 30s instead of using partysocket's seconds-scale backoff.
- The hand-rolled `useKnownHosts` snapshot and the Electric-era sidebar resilience doc are deleted; `isOnline` no longer comes from the API.
- Automation lists and failure badges poll every 60−120s instead of 15−30s.

**Relay**
- Each host's Durable Object authorizes its own calls, caching the decision keyed by user, so all isolates and colos share one API call per (user, host) per TTL.
- A host that has never connected returns 503 with no access check.
- `/presence` skips the access check for API-minted `automation-presence` tokens, so server-side reads skip access checks.
- Host-service re-asks for the relay URL only after a connect attempt fails.

The `organizationId` input to `v2Host.list` is optional so older clients keep scoping by the active-org header; the fallback is dated for removal once the fleet has moved past 1.28.

<sup>Written for commit b181af1f3711122826e6702f63ba9b10c90fcd38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7442?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Host settings now show when an offline host was last seen.
  * Relay requests now apply access permissions consistently.

* **Improvements**
  * Host availability is presented more consistently across desktop and mobile host pickers, workspace views, and settings.
  * Automation and host data refresh less frequently to improve efficiency.
  * Connection retries better handle temporarily offline hosts.
  * Host lists can be restored after reloads through cached query data.


saddlepaddle:
Added `49db34c4f` refactor(relay): the host's object authorizes its own calls.

The Worker no longer keeps a per-isolate LRU in front of the object's cache, and `checkHostAccess`/`checkHostAccessForRegister` no longer differ in shape. Every client request already lands on the host's Durable Object, so `isConnected`, `presenceInfo`, `prepareStream` and `proxyHttp` now take the caller and authorize inside, with the same 15-minute allow and 30-second deny TTLs in object storage. The Worker decides only what the token settles: a valid JWT and a host in one of the caller's organizations. A host's own control connect still asks the API, since its object may not exist yet. The unused `host-presence` scope is gone, and JWT verification moved to `@superset/shared/verify-jwt` so the realtime Worker in #7443 shares it.

Re-verified on the local stack: offline hosts still answer 503 with no access check, the online host's event socket opens, presence calls from the renderer stay at zero, and a placed host probed with a cold object went to the API once (185ms, "access check via api" logged) and served from object storage after (4ms).

saddlepaddle:
Added `84743bb28`: `v2Host.list` takes its organization id input optionally and falls back to the active-org header. Installed desktops and phones call it with no input, and requiring the id would have emptied their host list until they updated. The sidebar resilience doc is deleted; it narrated the Electric-era fixes and nothing referenced it.

Rollout notes for older clients: the relay's routes keep their status codes (401/403/503), `/presence` keeps its shape, and a `host-presence`-scoped token from an old API still works, it just goes through the access check. The relay deploy drops every host tunnel once, as any relay deploy does; hosts reconnect on their own backoff.

saddlepaddle:
On the CodeRabbit findings, none is a regression against what this replaces:

- **Persisted roster in the account-independent cache.** The IndexedDB snapshot this PR deletes was keyed by organization the same way, and the persisted PR and issue lists already share that property. The roster refetches on mount, so a second account on the same machine sees the prior roster only until that fetch lands. Clearing the persisted cache on sign-out is worth doing, as its own change, for every persisted key.
- **`lastSeenAt` cached across identity.** Same key shape the deleted presence hook used; a 60-second window on the settings page for an offline host's timestamp.
- **Non-HTTPS relay URL.** The dev stack runs the relay on `http://localhost`; the deleted presence hook sent the same bearer to the same URL.
- **Relay URL held for the window's life.** Deliberate: it is a deployment constant and the desktop relaunches on update. A relay move mid-session is not a case we serve today either.
- **Mobile `?? false` when presence is unavailable.** Matches the old steady state, where the API's own relay read returned null and fell through to `false`. The only change is the boot window before the presence query answers.

PR #7463, opened 2026-09-12
fix(desktop): reject diff hydration when the loaded contents don't match the patch

Kitenite:
## Summary
- The Changes pane no longer crashes with `DiffHunksRenderer.processDiffResult: deletionLine and additionLine are null` when it expands context on a file whose loaded contents don't match its cached patch.
- `isDiffContentStale` now verifies every line the patch carries against the hydrated contents, on both sides, instead of only comparing trailing line counts.

## Why / Context
Expanding context (or editing) in the Changes pane asks the host for both sides of the file and splices them under the patch's hunks. The host answers any failed `git show` with an empty string rather than an error, and a file can be rewritten after its patch was cached. The existing guard only compared the line counts after the last hunk, so empty, truncated, or shifted contents passed as fresh. @pierre/diffs then walked rows that exist on neither side and threw from a layout effect, which no error boundary inside the pane can catch, so the whole pane went down (the stack trace in the report ends in `walkContextLines`).

## How It Works
- After the trial `hydratePartialDiff("clone")`, `hunkLinesMatchContents` walks each hunk and compares the patch's compact line arrays (offset by `hunk.deletionLineIndex` / `additionLineIndex`) with the hydrated file lines at `hunk.deletionStart` / `additionStart`. Any missing or different line marks the contents stale, and the diff stays partial with its hunks still rendered.
- Lines the patch doesn't carry (a trailing blank context line trimmed off) are skipped rather than failed; the patch's final line is compared without its newline since the parser drops it.
- The trailing-count check is kept for the append-after-patch case, where hunk lines still match but the sides end unevenly.
- The Changes pane's rejection message now covers unreadable contents as well as files that moved on.

## Manual QA Checklist
Verified over CDP in this worktree's dev desktop, same journey on both builds: unstaged edit on a file's last line (hunk reaches EOF on both sides, the shape the old guard could not see), Changes pane in unified view, working-tree file truncated from a script, expand chevron clicked within ~20ms.
- [x] Before the fix: `deletionLine and additionLine are null` thrown twice from the pane's layout effect
- [x] After the fix: one `story.txt no longer matches its diff` console line, no library error, hunks still rendered
- [x] Correct contents still hydrate: guard sweep over every file in five real commits (680 checks across raw/trimmed patches and empty/short/grown content variants) reports no missed crash and no correct file rejected; the only rejections are new files, which the library itself refuses to hydrate, unchanged from before

## Testing
- `bun test .../diffLoadingGuards.test.ts` (11 pass; three new cases: empty contents, contents shorter than the hunks, lines moved under the patch)
- `bun run typecheck` in `apps/desktop`
- `bunx biome check` on the changed files

## Known Limitations
- The PR Code tab is unaffected: it never hydrates files, so it cannot reach this path.

## Follow-ups
- The diff card header keeps stale +/- counts after the patch refetches (a plain edit the sidebar reports as +2 −2 leaves the header at −1 +1). Present with or without this change.
- A renamed-and-edited file renders as a full new-file addition in the Changes pane because the per-file patch is pathspec-limited, so git cannot see the rename.

https://claude.ai/code/session_01MHDxzd3qBph1BQwhkmjpkP

---
## Summary by cubic
Fixes the Changes pane crash when expanding context on a file whose loaded contents don't match its cached patch. The old staleness guard only compared trailing line counts, so empty, truncated, or shifted contents passed as fresh, and `@pierre/diffs` threw `deletionLine and additionLine are null` from a layout effect, taking down the whole pane.

The guard now reads every line the patch carries back from the hydrated contents at the position its hunk header claims, on both sides. Any missing or different line keeps the diff partial, so hunks stay rendered. The trailing-count check remains for the append-after-patch case.

<sup>Written for commit ad3d6bbaaf0137ec1110dcff5bab28f0474ac0e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diff validation to detect when displayed file contents no longer match the associated patch.
  - Prevented stale or incomplete diff content from being shown, including cases involving missing lines, shifted lines, or unreadable file content.
  - Updated error messaging to clarify when file content no longer matches its diff.

- **Tests**
  - Added coverage for malformed and stale diff content scenarios.

### DIFF
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/DiffPane/DiffPane.tsx
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/DiffPane/DiffPane.tsx
@@ -475,12 +475,13 @@ export function DiffPane({
 					? { oldFile: null, newFile }
 					: { oldFile, newFile };
 			if (isDiffContentStale(fileDiff, loaded)) {
-				// The file moved on after its patch was cached, so these lines
-				// don't line up with the hunks they'd be hydrated into.
-				// Staying partial keeps the patch's hunks on screen instead of
-				// tearing the pane down; the `git:changed` that follows the
-				// write refetches the patch, and expanding works again.
-				throw new Error(`${file.path} changed since its diff was loaded`);
+				// The file moved on after its patch was cached, or a side
+				// couldn't be read at the patch's ref, so these lines don't
+				// line up with the hunks they'd be hydrated into. Staying
+				// partial keeps the patch's hunks on screen instead of tearing
+				// the pane down; the `git:changed` that follows the write
+				// refetches the patch, and expanding works again.
+				throw new Error(`${file.path} no longer matches its diff`);
 			}
 			return loaded;
 		},
--- a/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/DiffPane/utils/diffLoadingGuards/diffLoadingGuards.test.ts
+++ b/apps/desktop/src/renderer/routes/_authenticated/_dashboard/v2-workspace/$workspaceId/hooks/usePaneRegistry/components/DiffPane/utils/diffLoadingGuards/diffLoadingGuards.test.ts
@@ -169,6 +169,50 @@ describe("diff loading guards", () => {
 		).toBe(true);
 	});
 
+	test("rejects contents the host could not read at the patch's ref", () => {
+		// loadFileDiffContent answers a failed `git show` with "" instead of an
+		// error; hydrated from nothing, every hunk line is one @pierre/diffs
+		// walks straight off the end of.
+		const unreadable: FileContents = { name: FILE_PATH, contents: "" };
+		expect(
+			isDiffContentStale(REWRITES_MIDDLE, {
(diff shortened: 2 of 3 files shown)


Kitenite:
Live verification, second pass, with the pane's `git.getDiffPatch` refetch held back so nothing re-renders behind the throw (the state a user sees when the mismatch comes from git rather than a disk edit):

- **main's guard:** the `story.txt` diff card is replaced by `DiffHunksRenderer.processDiffResult: deletionLine and additionLine are null, something is wrong` with the stack from the report (`VirtualizedFileDiff.render` → `renderItem` → `CodeView.computeRenderRangeAndEmit`). Thrown twice.
- **this branch:** same click, same truncated contents; one `story.txt no longer matches its diff` console line, the card stays rendered with its hunk and collapsed context.

Journey for both: unstaged edit on the file's last line, Changes pane in unified view, working-tree file truncated to five lines, then the "36 unmodified lines" chevron clicked within ~20ms, all through real input over CDP in this worktree's dev desktop. Holding the refetch was done with a `window.fetch` wrapper on that one procedure; everything else was the real UI.

PR #7530, opened 2026-09-14
fix(pages): only let the first frame reveal itself by handshake

harshithmullapudi:
## Summary
- Scope the `ready` handshake to the frame it exists for, so a stale answer cannot reveal an iframe that is still navigating.

## Why / Context
Follow-up to #7422 (SUPER-2233). This commit was written in response to a review comment on that PR but was pushed after it had already merged, so it never landed.

#7422 added a `ready` handshake because a server-rendered iframe can finish loading before React hydrates, and React never replays the `load` event it missed. The reveal therefore cannot depend on `load` alone.

The remaining hole: the host posts its ask on mount, and a same-origin navigation keeps the same `WindowProxy`. If `src` flips while that ask is still in flight to the previous document, the old document's answer passes both the origin and `event.source` checks, and the handler — whose closure now reads the new `src` — marks the new one ready.

## How It Works
`initialSrc` captures the src present at mount; a `ready` only sets `readySrc` when it still matches. Every later `src` change navigates a frame whose `onLoad` handler is already attached, so `load` reveals it and the handshake is not needed.

## Why this isn't already covered
`PageFrame` on `main` guards with:

```ts
const loaded = loadedSrc === src || (ready === true && loadedSrc === null);
```

That closes the case where a load has already been observed — after any `load`, `ready` alone cannot reveal. But in the hydration race `loadedSrc` is still `null` (that is the whole premise: React missed the event), so a stale `ready` arriving in that window still satisfies `ready === true && loadedSrc === null` and reveals a frame mid-navigation. The two guards are complementary; this one covers the case the other cannot see.

## Manual QA Checklist
- [ ] Open a published page cold — appears, no lingering spinner
- [ ] Switch versions from the header — spinner during the swap, then the new version
- [ ] Switch versions immediately after first paint — no flash of the previous version
- [ ] Desktop PageViewer unchanged — frame loads, comments and pins work

## Testing
Not run — no `node_modules` in this worktree, so `bun run typecheck` / `lint` / `test` were not executed. CI covers them. No strings changed.

## Notes
The cherry-pick onto current `main` needed a conflict resolution: `main` has since added the pinch-zoom branch to the same `ready` handler and a `pinchZoomEnabled` dep. Both are preserved alongside the guard.

## Risks / Rollout
- Risk: low. One conditional narrowing when `readySrc` is set; the reveal path is otherwise untouched.
- Rollback: revert the commit.

https://claude.ai/code/session_01NbPEmduQUgqMcg476BHEnV

---
## Summary by cubic
Scopes the `ready` handshake to the initial frame src so a stale handshake answer can no longer reveal an iframe that is still navigating. Previously, if `src` changed while the mount-time ask was in flight, the old document's answer would pass the origin and source checks and reveal the new frame prematurely. Now only the server-rendered src (captured at mount) can be revealed by handshake; every later navigation gets a real `load` event and reveals normally. This closes a gap in the `SUPER-2233` handshake fix where `loadedSrc` is still null during the hydration race.

<sup>Written for commit 4de7abeccc40552271d9a70a6aba1c52104fb76c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved embedded content loading reliability by preventing outdated frame readiness messages from marking a newer source as ready.


harshithmullapudi:
Closing — I opened this on a wrong premise.

I concluded `3b0799271` "never landed" because it isn't an ancestor of `main`. It isn't an ancestor because #7422 was squash-merged, not because the commit was missed. It *was* in the PR, and @kiet-ho removed it before merge in d465ba5e7, deliberately.

That commit also identifies a real bug in the guard I was re-landing: `src === initialSrc` is true again whenever the reader switches **back** to the version the page opened on. The frame then reveals at full opacity with no spinner while the new document is still loading, with the previous version painted underneath. That is ordinary use — switch to v2, switch back to v1 — whereas the race my guard closed needs a version switch inside the sub-second window before any `load` has been observed.

Trading a common visual bug for an almost unreachable one is the wrong trade. The `loadedSrc === null` guard in `PageFrame` is the better fix and it is already on `main`.

PR #7532, opened 2026-09-14
fix(mobile): reply without swapping sheets, and post comments at all

harshithmullapudi:
## Summary

- Replying to a page comment on mobile closed the list and opened a different sheet. The all-comments sheet is now the only comment surface, with a pinned composer and inline replies.
- Selecting a block opened a sheet too. It now opens a popover anchored to the block, matching web, reusing web's placement math from `packages/shared`.
- **Commenting from mobile was broken outright** — `optimisticId()` called `crypto.randomUUID()`, which Hermes does not have. Fixed at the source.

## Why / Context

The reply flow was the reported problem: tapping **Reply** did `router.replace` from `/comments` to `/thread`, so the list visibly closed and a second sheet opened just to type one line.

While verifying on a simulator against production, two more things surfaced that the reply work would have papered over:

**Nothing posts from mobile.** `optimisticId()` lives in `packages/shared` and calls `crypto.randomUUID()`. Hermes ships neither that nor `getRandomValues`, and `apps/mobile` polyfills only the latter via `react-native-get-random-values`. It threw `undefined is not a function` inside `onMutate`, before any network call, surfacing as a bare "Comment not posted" alert. Every comment and every reply from iOS has failed since #7427 moved this code into shared. The id is a local placeholder the server row replaces, so it never needed to be a UUID.

**Page-level threads were unreachable.** The API has always offered `anchorKind: "page"` with a null anchor and validates that pairing (`page-comment/schema.ts:38`), but no client could produce one — the store hardcoded `"element"` and dereferenced `anchor.path`.

## How It Works

**One sheet.** `/pages/[slug]/thread` and `/pages/[slug]/compose` are deleted along with their routes. `AllCommentsSheet` holds a `replyingTo` target; **Reply** sets it, shows a `Replying to X ✕` chip, focuses the composer and scrolls the thread into view. ✕ clears it and the composer falls back to a page-level comment. Pin taps set a focus thread and open the same sheet.

**Anchored popover.** Web's `popoverPlacement` was pure math with no DOM, so it moved to `packages/shared/src/page-comments/popoverLayout.ts`; `packages/ui` re-exports it and **its existing tests still run against it unchanged**, which is the evidence the move is behaviour-preserving. Two new optional params, both defaulted so web and desktop are untouched: `pinSize` (24 web, 28 mobile) and `maxWidth` (capping a 390pt phone at web's 350 clamps `left` to 28 against 12 on the right — visibly off-centre).

The popover anchors to the block's **rect**, not `pinPointOf`, which returns a point *inside* the element and put the card on top of tall headings. Passing the block's height as `pinSize` makes the existing flip-above branch land correctly without touching the formula.

**Resolved comments** sit behind a `View N resolved comments` disclosure at the foot of the list, expanding inline.

## Manual QA Checklist

Verified on an iOS 26.3 simulator against production.

### Reply flow
- [x] Tapping Reply keeps the sheet open — no swap
- [x] Chip names the author; ✕ returns to page-level
- [ ] Reply lands inline and the list scrolls to it
- [ ] Composer sits flush on the keyboard at full detent

### Popover
- [x] Opens anchored below the selected block
- [x] Full width with symmetric margins
- [x] Avatar aligns with the input pill
- [ ] Flips above a block near the bottom
- [ ] Lifts only as far as the keyboard overlap
- [ ] Quick actions (trash / lightning / thumbs-up) post

### Comments
- [ ] A typed comment posts (this is the `randomUUID` fix — was failing 100%)
- [ ] A comment with no reply target creates a page-level thread and draws no pin
- [ ] Long-press → Resolve/Reopen
- [ ] `View N resolved comments` expands inline; a resolved pin auto-expands it

### Web / desktop regression
- [ ] Popover placement unchanged on web
- [ ] Anchored threads still create with `anchorKind: "element"`

## Testing

- `bun run typecheck` — 41/41
- `bun run lint` — clean
- `bun run check:i18n` — all 17 locales compile under `--strict`; the 7 new strings are translated by hand in every catalog, matching the existing "Hide resolved" wording already there from web, with proper plural categories for ru/pl/cs
- `bun run sherif` — no issues
- `bun test` for shared, cloud-client, ui, web, mobile, i18n — 0 fail
- New tests: `createThread` sends the page-kind payload with a null anchor, the element path still sends its anchor, and `optimisticThread` reports the kind it has

## Known Limitations

- The unticked QA boxes above are **not verified**. Building the app locally needed three throwaway workarounds for an Xcode 26.3 / Expo SDK 57 incompatibility (see below), and I would rather flag that than claim runtime behaviour I did not watch.
- Page-level comments are new to mobile. Web and desktop render them (`toThreads` drops anchorless rows from the pin layer), but neither has ever created one, so they are an unfamiliar shape there.
- Scrolling a focused thread to the top can tuck its first line under the transparent glass header. `contentInsetAdjustmentBehavior="automatic"` means the true offset needs the header height, and `@react-navigation/elements` is not installed, so there is no honest way to read it. Left rather than hardcoded.

## Follow-ups

**Xcode 26.3 cannot build apps/mobile.** Not addressed here, and it will hit anyone who updates:

1. `expo-modules-jsi` annotates a constructor `SWIFT_RETURNS_RETAINED` while the class is declared `SWIFT_SHARED_REFERENCE` after its closing brace; 26.3 evaluates the former first and rejects it. Present identically in 57.0.5, 57.0.8 and 57.1.0, so no bump fixes it.
2. `expo-observe` fails Swift 6 strict-concurrency checking.
3. Expo CLI cannot read `app.config.ts` — Node 22.17.1 is one minor below 22.18, where TS type-stripping became default. `NODE_OPTIONS=--experimental-strip-types` works around it.

I patched all three locally (node_modules + gitignored `ios/Podfile`) purely to get a simulator. Worth a committed `bun patch` plus a build-settings change, which is a decision for the team rather than something to slip into this PR.

Also worth lifting: `pinLayout` is still duplicated between `packages/ui` and `apps/mobile`, same as `popoverLayout` was.

https://claude.ai/code/session_01HEKrvgfxuBaNLb9Fp9VdZE

---
## Summary by cubic
Fixes mobile page comments so replying stays in one sheet, comments actually post from iOS (the optimistic ID relied on `crypto.randomUUID()`, which Hermes doesn't provide), and the comment popover no longer covers the block being commented on. Also adds page-level thread support and cuts comment-runtime scroll traffic.

**Bug Fixes**
- Replies happen inline in the all-comments sheet with a "Replying to X" chip that names the comment; the separate `/thread` and `/compose` routes are gone.
- `optimisticId()` now uses `getRandomValues` and returns a hex string, so it works on all platforms.
- The all-comments sheet no longer opens blank, now distinguishes loading, failed, and empty states, and shows its composer only when a reply target is set.
- Agent comments are attributed to the agent (from the MCP session id) instead of the person whose session wrote them.
- The popover places against the space above the keyboard, so it flips above the block instead of drifting onto it while typing.

**New Features**
- Selecting a block opens a popover anchored to it, using placement math moved from `packages/ui` to `packages/shared` (with new `pinSize` and `maxWidth` options).
- Page-level threads now render on web and desktop, draw no pin, and resolved threads sit behind a "View N resolved comments" disclosure.
- The comment runtime skips rect syncs when no pins are tracked and throttles scroll posts to 150ms, cutting bridge traffic from ~120 to 6 messages per 60 frames.

<sup>Written for commit eb7ee2a41e2d89047ec52db78e5146124bb31cf0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added inline mobile commenting with quick replies, reply context, thread focusing, and keyboard-aware composition.
  * Added comment resolving/reopening, resolved-comment filtering, expandable replies, and loading-error retry handling.
  * Added workspace page browsing with page cards and direct navigation.
  * Added a workspace details toolbar shortcut.
  * Improved agent-authored comments by displaying their specific agent labels.

* **Bug Fixes**
  * Page-level comments now display correctly without element anchors.
  * Improved comment popover positioning when the keyboard appears.

* **Localization**
  * Added updated translations for comment, reply, error, resolved-comment, and workspace-detail messages across supported languages.


harshithmullapudi:
Correcting a claim in the description above, and fixing what was behind it.

The description said page-level threads "are rendered by web and desktop". **That was wrong.** `toThreads` dropped every anchorless row, and `usePageComments` hands that list out as `store.threads` — which web and desktop read for the whole surface, not just pin placement: the `CommentsPanel` list, the `AllCommentsButton` count, the `CommentModeToggle` count. Mobile reads the raw `rows`, so a page-level comment was visible only to its author and to nobody on web or desktop.

I verified the pin path, saw `toThreads` drop anchorless rows, and generalised without checking the list path.

Fixed in f86a3b0c0: `toThreads` keeps those rows with a null anchor, and `CommentThread.anchor` is now nullable so the compiler locates every site that assumed otherwise. Three of them place or track pins and now skip a null anchor — a page thread has no rect to measure and draws no pin, which was always the intent. The fourth reads `anchor.text` for quoted context and optional-chains it.

The `drops a thread with no anchor` test encoded the bug, so it now pins the new contract instead.

`bun run typecheck` 42/42, `bun run lint`, `bun run check:i18n`, and tests across shared, cloud-client, ui, web, mobile, i18n all clean.

saddlepaddle:
@harshithmullapudi I ran this branch on a simulator against prod. Two things still block, both reproduced:

- Review (updated for f86a3b0c0): https://app.superset.sh/page/pr-7532-review-mobile-comment-replies-t3yca9
- Recordings: https://app.superset.sh/page/pr-7532-on-a-simulator-ajoy79

**1. The All comments sheet opens blank on every open.** It shows the title and the ReplyBar but no threads and no empty state. react-native-screens logs `FormSheet with ScrollView expects at most 2 subviews. Got 3` each time. Removing only the ReplyBar is still blank; making the ScrollView the sheet's only child renders the threads. That check is upstream `RNSScreenContentWrapper.mm` code that runs on iOS 26+ and isn't touched by our patch, so any build of this branch hits it. A fast refresh remounts the sheet and makes it look fine, so please try a cold launch: kill the app, relaunch, open All comments. The `apps/mobile/AGENTS.md` formSheet rule, with the nested `_layout` Stack recipe, covers this.

**2. VoiceOver sees each thread as one "Comment options" element.** The sheet's accessibility tree has exactly two of them for two threads, and no author, body, or Reply. The `accessibilityLabel` on the wrapper `Pressable` swallows its children, and resolve is long-press only.

Smaller:
- With the keyboard up, the popover lifts over the block you picked and hides it while you type (clip 2).
- `pendingScroll` is set after the `await`, so it misses the optimistic insert (from reading the code, not reproduced).
- The MCP `pages_comments_list` description still says every thread is anchored to one element.

f86a3b0c0 looks right for the web/desktop gap, and posting works now: every create, reply, 👍 and resolve in the recordings went through.

harshithmullapudi:
Pushed six commits addressing @saddlepaddle's review and two bugs the simulator found. Verified on device: **pins render and the All comments sheet shows its threads.**

## The blank All comments sheet

Reproduced, and it was not a data problem. A probe at the fetch showed `pageComment.list` returning `rows: 1` with query `status: "success"` — the comments were arriving the whole time.

`react-native-screens` sizes a formSheet from a ScrollView that is its *direct* child. Ours sat inside a `flex-1` View, so it resolved to zero height, and everything drawn inside it — threads, empty state, spinner, error — had nowhere to go. That is what made it read as a fetch failure.

The sheet now returns the ScrollView and the ReplyBar as siblings. Two subviews, which is what the sheet allows. The log went from ten `FormSheet with ScrollView expects at most 2 subviews. Got 3` warnings to **none across two opens**.

Worth recording that the nested `_layout` Stack alone did *not* fix this — I tried that first and the warning kept firing after a full reload. The constraint isn't the number of children, it's that the ScrollView must not be wrapped.

A related defect this exposed: `usePageCommentThreads` did `rows = list.data ?? []` and exposed no error, so a failed fetch was indistinguishable from a page with no comments. It returns the error now, and the sheet tells three states apart — loading, could-not-load with a retry, and genuinely empty.

## Pins are drawn by the app again

The runtime-pin approach is reverted. The page runtime is served by the usercontent Worker at `/_superset/runtime.js`, not bundled into the app, so a published page keeps loading the deployed copy no matter what the app sends it. I fetched the live script: zero occurrences of `render-pins`. Pins could not have appeared on any device until that Worker shipped, and only merging to `main` ships it — there is no preview Worker for usercontent, so the change was unverifiable before merge.

Pins are back on the app-side overlay, which is how web and desktop draw them and what the already-deployed runtime supports (`track`, `rects`, `scroll`, `pick` are all live).

Sending rects in **page coordinates** is still the right way to stop the per-frame measuring, and it does not require pins to live in the page. The host can translate against a scroll offset locally — and since `PageFrame` is a `react-native-webview`, its native `onScroll` can drive a Reanimated value on the UI thread, so a drag costs no bridge traffic at all. That wants device verification, so it is left for its own change. The reverted work is at `cdf30e7dd` when we pick it up.

## Review findings

| # | Finding | Status |
|---|---|---|
| 1 | Page-level threads dropped by `toThreads` | Fixed in `f86a3b0c0` (before this push) |
| 2 | Sheet breaks the formSheet layout rule | Fixed — ScrollView is a direct child, cold launch verified |
| 3 | `pendingScroll` set after the `await` | Fixed — set before the mutation, cleared on throw; a reply scrolls to the thread's bottom |
| 4 | VoiceOver reads a thread as one element | Fixed — plain `View` wrapper, resolve is a visible control, drops the app's only `ActionSheetIOS` |
| 5 | `popoverLayout` re-export + test in the wrong package | Fixed — folder deleted, test moved next to the code, covers `pinSize`/`maxWidth`; `pinSize` is now required so web and mobile cannot drift |
| 6 | Tap-outside dismisses mid-post | Fixed — waits for the submit, as web already did |
| 7 | Leftovers from the one-sheet-per-step flow | Fixed — three anchored `createThread` calls collapse into one, `openCompose` renamed for what it does |
| 8 | Ship the `randomUUID` fix separately | **Not done** — still your call |

Also: the MCP `pages_comments_list` description documents page-level threads instead of claiming every thread is anchored to an element, and the Indonesian `selesai` → `terselesaikan` wording is fixed. CodeRabbit's Turkish `not sabitlemek` nit became moot — that string was retired by the composer change below.

## Two product changes

**The All comments composer is reply-only.** A comment on mobile belongs to a block, so a bare composer at the bottom had nothing to attach to — it created a page-level thread instead, which draws no pin. Creating a thread now starts where it should: tap a block on the page.

This answers review question 3 for mobile, and it answers it differently from web and desktop, which can still create page-level threads. Deliberate, but flagging it: **the three clients now disagree about what a comment is.** Worth settling as a product decision rather than leaving as a per-client accident.

**Agent comments no longer show as the person.** An agent writes under the credential of whoever's session it is, so `author_user_id` — and every name joined from it — is that person. Mobile printed it straight, so a reply Claude wrote appeared under a human's name with a bot avatar beside it.

Web and desktop already had this right via `commentAuthor()`; mobile had grown a second, wronger copy of the same decision and now calls the helper. The helper could also only ever say "Agent", because the agent's identity never left the database — it is in `agentSessionId` as `mcp:<label>`, which the server sets from the transport rather than from the request body. `shapeComment` derives a label from it now, so an MCP write from Claude reads as **Claude**. A CLI agent's session id is opaque and self-reported, so that still reads as "Agent" — per `agent-access.ts` the CLI marking is a claim, not proof.

## Still open

- **Keyboard avoidance.** The recordings showing a buried input are the old `/thread` sheet, which this PR deletes, so they do not describe the current flow. Not changed — it needs a fresh look against the new reply-only bar, and `InputAccessoryView` is the native fix if it is still rough.
- **Popover over the picked block with the keyboard up** (@saddlepaddle's clip 2) — unaddressed.
- **Splitting out `128f613fd`** — finding 8.
- `CommentPin` and `apps/mobile/.../utils/pinLayout` are live again after the revert, so nothing is dead there.

## Checks

`bun run typecheck` 42/42 · `bun run lint` clean · `bun run check:i18n` 0 missing across 17 locales · shared 1083 · cloud-client 27 · mobile 94 · web 49 · 7 new tests on the agent-label derivation.

`@superset/desktop` has 2 failures in `terminal-ws-transport` — `jest.advanceTimersByTime is not a function`, a bun API incompatibility that is pre-existing on `main` and unrelated to these paths.

harshithmullapudi:
Correcting my last comment: clip 2's popover **is** addressed now, in `eb7ee2a41`. I had listed it as unaddressed.

@saddlepaddle's observation was that the popover "lifts over the block you picked and hides it", leaving only the blue selection edge visible. The cause is that the card was placed against the *whole* container — so it chose to hang below the block — and then a separate `lift` translated it up by whatever the keyboard overlapped. That amount is unbounded, so on a block low on the page the card travelled far enough to land on the thing it was anchored to.

The keyboard now takes the bottom of the container away rather than being corrected for afterwards:

```
container: { width, height: container.height - keyboardHeight }
```

That lets `popoverPlacement` make the decision it already knows how to make — flip above the block — and `pinSize` is already the block's height here, so the flip clears the whole block rather than the 24px a pin would need. The manual `lift` transform is gone.

Placement has to be a render-time number for this, so the keyboard height comes from `keyboardWillShow`/`keyboardWillHide` rather than `useAnimatedKeyboard`, and `top` settles with a timing curve so the card moves with the keyboard instead of snapping.

Pinned with a shared test modelling the recorded case — a block low enough that the card cannot fit beneath it once the keyboard is up ends with the card's bottom above the block's top:

```ts
test("the keyboard taking the bottom flips the card clear of its block", ...)
expect(top + cardHeight).toBeLessThanOrEqual(blockTop);
```

Worth a look on device, since the geometry is verified by unit test but the keyboard timing is not.

Still open from that list: **keyboard avoidance in the reply bar** (deliberately left — the recordings of a buried input are the old `/thread` sheet this PR deletes, so they don't describe the current flow, and the composer now only mounts while replying), and **finding 8**, splitting out `128f613fd`.

saddlepaddle:
@harshithmullapudi Merged this — the fixes hold up. I re-verified `eb7ee2a41` on an iPhone 17 Pro Max (iOS 26.0) against production: the sheet renders its threads with zero `FormSheet with ScrollView` warnings, VoiceOver reaches every author, body, Reply and Resolve (it saw only two "Comment options" elements before), and a reply posts and lands inline under its thread. Block comments post correctly too.

One thing to pick up separately, found in the same run: **the block popover now opens behind the keyboard on the lower half of a page.**

- Evidence, with measurements and screenshots: https://app.superset.sh/page/mobile-block-comment-popover-hides-behind-the-keyb-gd7pmv
- Ticket, assigned to you: **SUPER-2384**

Short version: shrinking the container by the keyboard height makes `popoverPlacement` take the flip-above branch, which is right, but the flipped `top` is measured from the block — `point.y - pinSize - GAP - height` — so when the block sits below the keyboard's top edge, the card does too. Composer at y 777 with the keyboard starting at y 611 for a block ~87% down; an upper block is fine. It's a swap of symptom rather than a regression: before it stayed on screen but covered the block.

Clamping the flipped top into the keyboard-reduced container fixes the common case; scrolling the frame so the block clears the keyboard is better for a block near the very bottom. The test added in `eb7ee2a41` passes through this because it only asserts the card is above the block, not that it stays inside the visible container.

Also worth splitting out next time: the workspace ⋯ button and the Pages section in the workspace actions sheet came along in `52fa7126b` / `c7b58ddc8` and have nothing to do with comment replies.

harshithmullapudi:
yeah I also saw them but didn't make this PR clumsy I will work on this in the next one

PR #7556, opened 2026-09-14
fix(host-service): classify an unaccepted Xcode license as environmental

saddlepaddle:
## Problem

`git.getStatus` returns a 500 on Macs where Xcode was updated and its new license has not been accepted yet. Every status refresh on such a machine reports to Sentry.

HOST-SERVICE-X: 14 events from at least four machines on 1.29.0 within about 90 minutes today, all on `git.getStatus`, and the issue is escalating. Nothing in our code can make git run there: git stays unusable until the user accepts the license. HOST-SERVICE-X is also the catch-all group for every unclassified simple-git failure, since they all share one stack. Archiving it would also mute the real git failures in that group, so the fix has to be at the classification.

## Root cause

`/usr/bin/git` on macOS is Apple's shim. After an Xcode update it will not run git until the license is accepted. Instead it prints a single sentence asking for `sudo xcodebuild -license` and exits. simple-git turns that stderr into the error message. `rethrowEnvironmentalGitError` already classifies the shim's two other refusals, no developer tools (#6737) and a broken Xcode (#7278), as `PRECONDITION_FAILED / GIT_ENVIRONMENT`. It had no case for this third wording, so the error fell through as an `INTERNAL_SERVER_ERROR`.

## Fix

One more pattern on the existing `GIT_ENVIRONMENT` branch in `classify-git-error.ts`. The message is kept verbatim.

Unlike the two siblings, this sentence has no `xcode-select:` prefix and names no tool. A hook that reaches any shimmed tool (clang, python3, make) prints the same line. So the pattern anchors the **whole message** (the sentence plus an optional trailing newline) rather than a line. It matches only when the refusal is everything git produced, which means git never ran.

**Still a 500 after this change:** the same sentence mixed into any other git output, and this condition on procedures that do not run the classifier.

## Verification

Tests written before the pattern. Before the change, the positive case failed and the negative passed:

```
(fail) rethrowEnvironmentalGitError > Xcode license not accepted → PRECONDITION_FAILED / GIT_ENVIRONMENT [0.15ms]
 32 pass
 1 fail
Ran 33 tests across 1 file.
```

The negative case defends the anchor. Against a line-anchored variant (`/^…\.$/im`) it fails:

```
(fail) rethrowEnvironmentalGitError > does not claim the license refusal alongside a failure of git's own [1.02ms]
 32 pass
 1 fail
```

With the whole-message anchor:

```
$ bun test src/trpc/router/git/utils/classify-git-error.test.ts
 33 pass
 0 fail
Ran 33 tests across 1 file. [37.00ms]

$ bunx biome check packages/host-service/src/trpc/router/git/utils/classify-git-error.ts packages/host-service/src/trpc/router/git/utils/classify-git-error.test.ts
Checked 2 files in 14ms. No fixes applied.

$ cd packages/host-service && bun run typecheck
(no errors)

$ cd packages/host-service && bun test src/trpc/router/git/utils
(fail) getGitStatusPartial: renames and index interplay > long converging sequence of mixed batches [5395.20ms]
 234 pass
 1 fail
Ran 235 tests across 17 files. [563.52s]
```

The one failure does not involve this change. It is a 5 s timeout in a real-git adversarial test that never imports the classifier. It fails identically with main's `classify-git-error.ts` restored (`this test timed out after 5000ms`) on this machine under load.

Refs HOST-SERVICE-X

https://claude.ai/code/session_011UM7NxZCKwiBiDneLt7Qd4

---
## Summary by cubic
Fixes `git.getStatus` returning a 500 on Macs where the Xcode license hasn't been accepted after an update. The refusal is now classified as `PRECONDITION_FAILED / GIT_ENVIRONMENT` instead of an internal error, so it stops reporting to HOST-SERVICE-X.

- Matches only when the license sentence is all git produced; the same sentence mixed into real git output still returns 500.

<sup>Written for commit 4e0c077bd00908de469fb13a34586df583ba10db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Git environment error handling for systems where the Xcode license has not been accepted.
  - Users now receive a clear setup-related error instead of an incorrectly reported application issue.
  - Other Git failures, including network and hook errors, continue to be classified and reported appropriately.
  - Added coverage to ensure Xcode license errors are recognized consistently without affecting unrelated Git error handling.


saddlepaddle:
Verified both CodeRabbit findings against the code:

- **Add a regression test for the Xcode license refusal:** false positive. The test exists: `Xcode license not accepted → PRECONDITION_FAILED / GIT_ENVIRONMENT` at `classify-git-error.test.ts:425` asserts the code, the cause kind, and that the message is preserved. A negative case keeps the refusal mixed into other git output reporting.
- **Push path never reaches the classifier:** accurate about the code, not addressed here. `gitPushTask` swallows any `revparse --abbrev-ref HEAD` failure (`.catch(() => "")`) and returns `detached-head`. That predates this PR and hides the two existing Xcode shim refusals the same way. The result is a 400 that never reports to Sentry, and it is not the `git.getStatus` failure HOST-SERVICE-X records. Changing it would change push behaviour for every revparse failure, which is a separate change from this classification fix.

https://claude.ai/code/session_011UM7NxZCKwiBiDneLt7Qd4

PR #7570, opened 2026-09-15
feat(pages): public pages, and thumbnails in link unfurls

harshithmullapudi:
**Links**
- Fixes SUPER-2189 — Have pages show their thumbnails on opengraph
- Fixes SUPER-1998 — an anonymous fetch of a just_me page 307s to sign-in rather than 404ing

## Summary

- `everyone` becomes a selectable visibility ("Anyone with the link"), on web, desktop, mobile, the CLI and the MCP tool.
- A public page shares **only the version its link resolves to**. The rest of the history stays behind a ticket.
- Anonymous readers of a public page now get the page, read-only. Everyone else gets a 200 not-found screen instead of a redirect.
- Shared page links unfurl with the page's own thumbnail.
- **Also bundled (not in the tickets):** the page title menu is rebuilt around read-only version previews, plus rename and refresh.

## Why / Context

The two tickets are the same bug from two sides.

`apps/web/src/proxy.ts:45` redirects any unauthenticated request for a non-public route to `/sign-in`, and `/page` was not in `otherPublicRoutes`. A crawler asking for a shared page got a **307** and never reached `generateMetadata` — so no amount of `og:` tags would ever have unfurled (2189), and a signed-out human got bounced to a login wall instead of being told the page isn't available (1998).

The thumbnail machinery already existed: `packages/trpc/src/router/page/thumbnail.ts` captures a 1280×880 JPEG per published version, and `apps/usercontent` served it ticket-free when visibility is `everyone`. Nothing pointed at it from the web viewer, and no page could be `everyone` in the first place.

## How It Works

**`everyone` end to end.** `OFFERED_VISIBILITIES` gains the third value — one constant feeding `page.create`, `page.publish`, `setVisibility`, the CLI and the MCP tool schemas. `PageVisibility` now has a single definition in `@superset/shared/usercontent`; the UI, mobile and CLI copies that had drifted are gone. Three read paths silently excluded `everyone` and had to be fixed with it:

| Site | Was | Effect |
|---|---|---|
| `page.ts` `visibilityFilter()` | `org` + own `just_me` | public page vanished from `page.list`, `pageNotFound()` reported a bogus cross-org FORBIDDEN |
| desktop `matchesScope()` | `team` was `=== "org"` | public page fell out of **both** "team" and "mine" chips, reachable only under "all" |
| mobile `matchesScope()` + `PageRow` | same, plus `visibility === "org" ? Globe : Lock` | public page dropped out of "team" **and** rendered with a **Lock**, telling the owner a live public page was private |

`PageCard` had already spent `Globe` on `org`, so the three states now read Lock / Building2 / Globe everywhere, matching the share popover. The mobile share sheet gains the third row, so "public" can be seen and undone from the phone.

**A public page shares one version, not its history.** The Worker's `everyone` branch returned `"public"` for *any* `/versions/N/` request. The page id is not secret on a shared page — it is the host of the `og:image` and the iframe `viewUrl` — so anyone holding the link could walk `/versions/1/`, `/2/`, …, including a version the owner had deliberately moved away from and every version an agent published later, while the share popover promised *"Everyone stays on v3 until you change this."* `publiclyReadable(manifest, version)` now gates this on `servedVersionOf(manifest)`; every other version needs the ticket the owner's own clients mint.

Because public readability is now **revocable** — it moves when the served version moves — public responses can no longer cache immutably, or a shared CDN would keep answering for a version the owner has already left. They carry `public, max-age=300` instead.

**`page.publicView`** is the only page read with no session behind it. It answers for `everyone`-pages and returns `null` for everything else, so a private page stays indistinguishable from one that never existed. An R2 head check keeps `og:image` from pointing at a capture that never ran.

**`generateMetadata`** reads `publicView` *first*, so what it emits is by construction what a crawler may see — real title/description/thumbnail for a public page, generic Superset card otherwise. It falls back to the authenticated pull for the browser tab title only, so signed-in users keep page titles on `org` pages while unfurlers learn nothing. Every branch now returns `robots: { index: false, follow: false }`; unfurlers ignore it, so previews still work.

**Anonymous rendering.** Public pages render read-only (no share popover, version picker, or comments panel — comments stay `protectedProcedure`). A signed-in user from another org now sees a public page rather than *"This page belongs to Acme."*, with an "Open in Superset" affordance rather than a "Sign in" link that would bounce them home. Everything else gets `PageUnavailable`.

**Version history (the bundled part).** Picking a version from the title menu used to republish it as the shared version; it now opens a read-only preview — `?v=N` on web, per-pane state on desktop — with the comment composer off, because a thread created against an unserved version would send a watching agent feedback about content nobody can see. Both surfaces show a "Back to current" banner. The menu also gains rename (via a new `page.update`) and refresh.

## Design Decisions

- **`PageUnavailable` reuses the existing not-found copy.** My first pass said *"This page is shared with a Superset organization. Sign in with an account that has access."* — which confirms to a stranger holding a guessed slug that the page **exists** and belongs to an org. That is precisely the disclosure `assertPageReadable` avoids by throwing `NOT_FOUND` rather than `FORBIDDEN` for `just_me` pages. Same screen for missing, private, and unauthorised; only the button differs (Sign in / Go to Superset). Both now render one shared `PageMissingScreen` so the two copies cannot diverge.
- **200, not 404, for the unavailable screen.** A 404 suppresses unfurls entirely. 200 keeps the generic branded card.
- **Comments stay org-only.** Letting anonymous readers comment means inventing anonymous identity and spam controls; letting them *read* comments makes commenter names and avatars public. Neither belongs in this PR.
- **The public-read rate limit lives in the web route, not in `page.publicView`.** On a public page the slug is the credential, and it is only six characters of `[a-z0-9]` — about 2.2×10⁹. Throttling moves enumeration of a guessable base (`untitled-`, `report-`) from roughly 12 days at 1000 req/s to roughly 34 years. It cannot sit inside the API the way `leaderboard.ts` does it: the web app calls `publicView` server-side, so every reader would arrive from the web function's egress address and share one bucket — the same mistake the leaderboard hit from the marketing server. `/page/[slug]` is the last hop that can see the real client IP, and the check sits in front of the call so a guessed slug costs no DB lookup and no R2 HEAD. It fails open.
- **`page.versions` emits thumbnail URLs unchecked.** It is fetched on every web render, every desktop header and the mobile share sheet; a paginated R2 ListObjects per call bought nothing the client's own `onError` placeholder does not already handle. Only `publicView`'s `og:image` still does a server-side HEAD.

### Visibility
- [ ] Share popover offers "Anyone with the link"; picking it copies the link
- [ ] Page stays listed in `page.list` and in the desktop grid after going public
- [ ] Public page appears under the "team" chip, not only "all"
- [ ] Card icon reads Lock / Building2 / Globe for just_me / org / everyone
- [ ] Mobile: same icon, same "team" filter, and the share sheet can set **and undo** "Anyone with the link"
- [ ] `superset pages publish --visibility everyone` is accepted

### Version scoping
- [ ] On a public page, `/<pageId>.usercontent…/versions/<not-the-served-one>/` does **not** serve without a ticket
- [ ] The owner can still open every version from the title menu
- [ ] Moving the shared version stops the old one being publicly readable

### Anonymous access (signed out, fresh profile)
- [ ] Public page renders read-only — no share popover, version picker or comments
- [ ] Private page shows "This page isn't here" + **Sign in**, HTTP **200**, no redirect
- [ ] Nonexistent slug shows the identical screen (no existence signal)
- [ ] Signing in from that screen lands back on the page

### Unfurls
- [ ] Paste a public page link into Slack/Linear — title, description and thumbnail appear
- [ ] Paste an `org` page link — generic Superset card, no title or thumbnail leak
- [ ] Public page whose thumbnail has not been captured yet degrades to `summary`

### Cross-org
- [ ] Signed-in user in another org sees a public page, with "Open in Superset"
- [ ] Signed-in user in another org still gets WrongOrganization for an `org` page

## Testing

- `bun run typecheck` — 42/42
- `bun run lint:fix` — clean
- `bun run check:i18n` — all 17 locales compile, 0 missing; 11 new strings translated into all 16 non-English catalogs
- `bunx sherif` — clean
- `bun test` — `@superset/web` 52, `@superset/trpc` page + page-comment routers 67, `@superset/shared` usercontent 56, `@superset/ui` 71, desktop `filterPages` 15, all passing

**Known red on `bun test`, none of it from this change's logic:**

| Suite | Count | On `main`? |
|---|---|---|
| desktop `terminal-ws-transport` | 2 | yes — pre-existing |
| desktop `Redirect` | 3 | **no** — see below |

The three `Redirect` failures pass in isolation and fail in the full suite: `Redirect.test.tsx` shares process-wide happy-dom globals with 32 other desktop test files, and the one extra `filterPages` case this PR adds is enough to shift the ordering that exposes it. It is test-suite pollution, not a behaviour change — nothing in this PR touches `Redirect` or the router. **A CI re-run will not clear it.**

## Known Limitations

- **No automated test for `page.publicView`.** It needs DB + R2; the router's only integration file is `publish.integration.ts`, which `bun test` does not pick up. Covered by the QA checklist instead. The Worker's public branch has no live coverage either — `publiclyReadable` is unit-tested in `@superset/shared`, but the Worker itself has no test harness.
- **Going public has no confirm step.** It is one click in the select, and `--visibility everyone` from an agent is one flag. Worth a confirmation before this reaches users — see the open questions below.
- **Flipping `everyone` → `org` is not instant.** Public responses cache for five minutes, so already-fetched bytes survive that long. Down from "indefinitely", which is what `IMMUTABLE` meant before this PR.
- **`X-Robots-Tag: noindex, nofollow` still applies to every page document** on the usercontent origin, including public ones — "anyone with the link" is not obviously "index me in Google". The `app.superset.sh/page/<slug>` wrapper now says the same thing via `generateMetadata`. Thumbnails never carried that header, which is why unfurls work regardless.

## Open Questions

1. **Should agents be able to make pages public?** The MCP tool and `superset pages publish --visibility everyone` publish to the open internet with no human in the loop. That is a real blast radius for a prompt-injected agent.
2. **Org policy.** Should there be an org-level switch that disables public pages, and what happens to already-public pages when a plan lapses or the owner leaves the org? Today `publicView` and the Worker keep serving them.
3. **Slug entropy.** Six characters is thin for what is now a bearer credential. The rate limit compensates; a longer suffix for public pages would fix it properly, at the cost of changing a page's URL when it goes public.

## Risks / Rollout

- **Risk:** the `/page` entry in `otherPublicRoutes` removes the proxy auth gate for that route subtree. The route itself is the gate now — `getPagesAccess()` then `publicView`, which only ever answers for `everyone`-pages. Worth the closest read in review.
- **Risk:** `mintPageTicket` now issues a ticket whenever a version is named, including on public pages. Without it the version-scoping change above would have locked owners out of their own history.
- **Rollout:** no migration, no feature flag change. `everyone` is opt-in per page and no existing page has it.
- **Rollback:** revert the eight commits on this branch. No data to unwind — a page whose visibility was set to `everyone` before a revert keeps that value and becomes unselectable again in the UI. Note that reverting also restores the pre-existing Worker behaviour where *every* version of such a page is publicly readable.

---
## Summary by cubic
Adds opt-in “Anyone with the link” sharing and page thumbnails to link unfurls, addressing SUPER-2189 and SUPER-1998. Anonymous requests that previously redirected to sign-in now show public pages read-only or a 200 not-found screen, and SUPER-1990’s version-preview changes are included.

**Public sharing**
- Adds `everyone` visibility to web, desktop, mobile, CLI, and MCP sharing controls and page filters.
- Public links expose only the currently served version; other versions still require tickets.
- Unfurls include a public page’s title, description, and thumbnail, while private pages keep generic metadata.
- Public reads are IP-rate-limited, cache for five minutes, and disable comments; failed manifest writes restore the database visibility.
- No migration or feature flag is required; existing pages remain unchanged until made public.

**Version history**
- Version selection now opens a read-only preview with a “Back to current” action and version-specific comment counts.
- The title menu now supports renaming without creating a version, refreshing page data, and showing version thumbnails with a file-icon fallback.

<sup>Written for commit 905bd627ddef7c6779de1b6468a4f5290f8d2be3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7570?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Pages can be shared publicly with anyone who has the link, including public viewing, sign-in redirects, unavailable-page messaging, and sharing controls.
  - Owners can copy public links, open sharing settings, rename pages, refresh content, and preview earlier read-only versions with version thumbnails.
  - Added live-page and version-preview banners with clearer visibility indicators.
  - Team filters and publishing tools now support public pages.

- **Bug Fixes**
  - Page visibility is handled consistently across the app.

- **Localization**
  - Added translations for sharing, navigation, renaming, and version-preview messages.


saddlepaddle (commented):
The public-sharing approach looks right to me. The route is its own gate, `publicView` only answers for `everyone`, and a private page looks the same as a missing one. A few things to fix before merge.

### Must fix: mobile never learned `everyone`

You fixed this on desktop, but the mobile copies still assume two states:
- `apps/mobile/screens/(authenticated)/pages/components/PageRow/PageRow.tsx:54`: `visibility === "org" ? Globe : Lock` shows a **Lock** on a public page. That tells the owner a live public page is private.
- `apps/mobile/screens/(authenticated)/pages/stores/pagesFilterStore/pagesFilterStore.ts:37`: public pages drop out of "team".
- `apps/mobile/screens/(authenticated)/pages/[slug]/hooks/usePageSharing/usePageSharing.ts:6` and `share/PageShareSheet.tsx:145-160`: `PageVisibility` is still `"just_me" | "org"`, so neither row is selected and you can't see or undo "public" from the phone. It needs an "Anyone with the link" row.

### Bugs in the version-history / rename rework

- **Desktop pane preview has no indicator and no way back.** `PagePaneTitle.tsx:63` sets `previewVersion`, and `PagePane.tsx:47` quietly loads the old version. Only `PageHeader` renders `PageVersionBanner`, and the pane doesn't use `PageHeader`. The title bar still pulls the served version, so its thread count disagrees with the panel below. Render the banner in `PagePane` when previewing a non-served version.
- **"Read-only" previews still accept comments.** `PageVersionBanner.tsx:17` says read-only, but web (`page.tsx:152-155,187`) and desktop (`PageViewer` → `usePageComments`) keep comment mode on. New threads land on the historical version, and a watching agent gets feedback about content that isn't served anymore. Hide the composer while previewing, or drop "read-only" from the copy.
- **Rename leaves stale titles.** `usePageHeaderData.ts:94-100` patches only the unversioned `page.pull` entry. In a pane the header prefers `data.title` (line 62), so the new name doesn't show. `page.list` isn't invalidated either, so the grid keeps the old name. Invalidate `page.pull` and `page.list`, and push the title through `onDataChange`.
- **Preview version survives a page change.** `PageDetailView.tsx:19` keeps `previewVersion` in state, and `pages/$slug/page.tsx:20` doesn't key the view by slug, so navigating to another page sends the old `version` to `page.pull`. `<PageDetailView key={slug} slug={slug} />`.

Going forward: the title-menu rework is a separate change from SUPER-2189/1998, and most of the bugs above came in with it. Please keep that kind of thing in its own PR next time so the public-sharing part can ship or roll back on its own.

### Rate-limit `page.publicView`

For public pages the slug is the only thing between a stranger and the page. That's fine for "anyone with the link", but slugs are a title-derived base plus 6 chars of `[a-z0-9]` (~2×10⁹ combinations), and `publicView` is an unauthenticated DB lookup plus an R2 HEAD with no throttle. Two notes on where to put the limit:
- Keep it on the slug lookup, not the usercontent asset path, which legitimately fetches many files per page.
- Don't key it by `x-forwarded-for` inside the API the way `leaderboard.ts` does. The web app calls `publicView` server-side, so on Vercel every reader arrives from the web function's egress IP and shares one bucket. The leaderboard limit already hit this from the marketing server. Limit in the `/page/[slug]` route, where the real client IP is visible, or pass a trusted client-IP header.

### Cleanups

- **`page.tsx:58-102`:** no `robots` on any branch. Public pages and the 200 `PageUnavailable` soft-404 are indexable, even though usercontent deliberately sends `noindex`. Add `robots: { index: false, follow: false }`; unfurlers ignore it.
- **`page.tsx:29-32, 92, 135`:** the `pullPage` cache now misses. Metadata calls `pullPage(slug)` and the component calls `pullPage(slug, undefined)`. React's `cache` keys on argument count too, so every org-page render pulls twice, which contradicts the comment above it. Make `version: number | null` required and pass it from both call sites (`generateMetadata` gets `searchParams` too).
- **`page.tsx:133-144`:** the pull `catch` falls back to `publicView()` for any error. A same-org member hitting `?v=999` or a presign failure silently gets the anonymous view. Only fall back on NOT_FOUND/FORBIDDEN.
- **`PublicPageView.tsx:28`:** "Sign in" is shown to signed-in cross-org users too, and the proxy bounces them to `/`. Show `OpenInSupersetButton` when there's a session. Also build the href as `{ pathname: "/sign-in", query: { redirect: `/page/${slug}` } }` (same in `PageUnavailable.tsx:27`).
- **`packages/trpc/src/router/page/page.ts:697-713`:** `versions` now does a paginated R2 ListObjects on every call (web render, desktop header, mobile share sheet). `list` already emits thumbnail URLs unchecked and lets `PageThumbnail`'s `onError` fall back. Do the same in the version menu and delete `listObjectKeys`. Only `publicView`'s `og:image` needs the server-side HEAD.
- **`packages/ui/src/components/PageComments/components/PageHeader/types.ts:1`:** `PageVisibility` is now hand-copied in ui, `shared/usercontent/manifest.ts` and mobile. The mobile copy is exactly what drifted. Re-export it from `@superset/shared/usercontent`. In `packages/cli/src/commands/pages/publish/command.ts:21`, drop the `VISIBILITIES` alias and use `OFFERED_VISIBILITIES` directly.
- **`packages/ui/src/components/PageComments/index.ts:20`:** `PageFrame` now has two consumers but is exported from `PageCommentsView/components/`. Move it up to `PageComments/components/PageFrame/`. The missing `"use client"` in c74deaeb9 was a symptom of this.
- **`PageUnavailable.tsx`:** it duplicates `page/[slug]/not-found.tsx` except for the button. Share one screen that takes an `action`.
- **PR body:** the rollback plan ("revert the two commits") and "3 new strings" are stale now that there are six commits.

CI: the one Test failure is the host-service `gitAuthorNameTask` timeout. It passes locally and this PR doesn't touch host-service, so a re-run should clear it.

harshithmullapudi:
Thanks — pushed `a9caea275` and `225b19dff`, and rewrote the PR body (the rollback plan said "two commits" and the string count was stale; both are fixed, along with the cache-revocation and robots notes that this change made inaccurate).

Everything in the review is addressed. Three things worth calling out rather than leaving in the diff:

**The rate limit was wrong in the way you predicted.** My first pass put it inside `page.publicView` keyed on `x-forwarded-for`, exactly copying `leaderboard.ts`. That would have collapsed every reader into the web function's egress bucket — throttling all public pages in aggregate to 60/min while doing nothing to an actual enumerator. It now lives in `/page/[slug]` in front of `pullPublicPage`, so a guessed slug is refused before it costs a DB lookup and an R2 HEAD, and `cache()` means metadata and the component share one check. `apps/web` already declared `KV_REST_API_*` and the deploy workflows already pass them, so this needed no env plumbing.

**Rename: I fixed the precedence rather than threading `onDataChange`.** `PagePaneTitle` doesn't receive `onDataChange` — only `PagePane` does — so pushing the title through it meant new plumbing. Instead `usePageHeaderData` now prefers `resolved.title` over the pane's cached label, and `PagePane`'s existing `onResolved` already syncs pane data once the pull settles. `page.pull` and `page.list` are both invalidated. Same outcome, no new prop.

**One extra finding, on a version-history bug you didn't flag:** the Worker's `everyone` branch returned `"public"` for *any* `/versions/N/`. Since the page id is the host of the `og:image` and the iframe `viewUrl`, it isn't secret on a shared page — so anyone with the link could walk the whole history, including versions the owner had moved away from and every version an agent published later, while the popover promised "Everyone stays on v3". `publiclyReadable()` now scopes public access to the served version. That also made public readability revocable, so those responses can't cache `IMMUTABLE` any more (`max-age=300` instead), and `mintPageTicket` has to issue a ticket whenever a version is named or owners lose their own history.

---

**On CI — the Test job needs more than a re-run.** There are two independent failures, not one:

| Suite | Count | On `main`? |
|---|---|---|
| host-service `gitAuthorNameTask` | 1 | flake, as you said — a re-run clears it |
| desktop `Redirect` | 3 | **no** |

`main` is stable at 2 local failures across repeated runs (`terminal-ws-transport`); this branch is stable at 5, and was already at 5 at `df0a4bbd4` before either of my commits. The three `Redirect` cases pass in isolation and fail in the full suite — `Redirect.test.tsx` shares process-wide happy-dom globals with 32 other desktop test files, and the single extra `filterPages` case this PR adds shifts the ordering enough to expose it. Nothing here touches `Redirect` or the router, so it's suite pollution rather than a behaviour change, but it won't go green on a retry. Happy to chase the polluting file in a separate PR if you'd rather not hold this one for it.

I've left the title-menu rework in place per your process note being about next time rather than this PR — say the word if you'd rather I split it before merge.

PR #7575, opened 2026-09-15
Feature (desktop): qualify pinned sidebar rows with their project name

### LINKED ISSUE #7574: [feat] Pinned workspaces named "local" are indistinguishable — the PINNED section drops the project name
### What problem are you trying to solve?

The PINNED section pulls a workspace out of its project group. That's the whole point — but it also means the project name is no longer anywhere near the row.

Every project's main workspace is called `local`. So pin two of them and the sidebar shows this:

```
PINNED
  💻 🌍 local
  💻 🌑 local
```

Two identical rows. The only thing telling them apart is a 14px project avatar, which for most projects is a generated circle — I can't tell Flamaster from Bifrost at that size (or whatever projects I have). Inside a project group `local` is unambiguous because the group header says which project. Pinned, that context is gone.

I have to hover each row and wait for the tooltip to find out which is which, every time. Pinning is supposed to be the fast path.

### Proposed solution

Show the project name in the row label, dimmed, as a path-style prefix:

```
PINNED
  💻 🌍 Flamaster/local
  💻 🌑 Bifrost/basic_optimization
```

Readable at a glance, no hover. Keep the avatar — it's still a useful color cue once you know which is which.

The data is already there: `pinnedContext` on the row carries `projectName` and `projectIconUrl`, and today only the icon is rendered (`DashboardSidebarExpandedWorkspaceRow.tsx#L311`). The name is passed in and thrown away.

Only pinned rows need this. Rows inside a project group already have their header. Project-less sessions can read `Session/…`.

### Workarounds you've tried

- **Hover the row** — the tooltip does show the project name, but one row at a time, and it's the thing I'm trying to avoid.
- **Rename the workspace** — not possible for a main workspace: `onRename` is undefined when `isMainWorkspace`, and the stored `name` is ignored in favour of `"local"`.
- **Move it into a section I could name myself** — the *Create section* / *Move to section* submenu is gated on `!isLocalMainWorkspace`, and grouping is hidden for pinned rows anyway.
- **Unpin and rely on the project groups** — works, but gives up the feature.

### Related

- #6520 — same "every row says `local`" ambiguity, scoped to rows inside a project group, proposing the branch as a subtitle. Complementary: that one disambiguates checkouts of the *same* project, this one disambiguates *across* projects.

Superset 1.25.0, macOS (Darwin arm64).

pdolega:
> **WIP — not ready for review.** The code is complete and green, but it has not been verified in the running app yet (see *Verification* below). Opening early for direction on the presentation.

Closes #7574.

## The problem

The PINNED section lifts a workspace out of its project group — that's the point — but it also strips the only thing that said which project the row belongs to. Every project's main workspace is named `local`, so two pinned mains render as two identical rows:

```
PINNED
  💻 🌍 local        →   💻 🌍 Flamaster/local
  💻 🌑 local        →   💻 🌑 Bifrost/basic_optimization
```

The 14px project avatar is meant to disambiguate them, but for projects without a custom icon it's a generated circle — not something you can tell apart at that size. The fix is to say it in words.

## The change

This is how it looks like in the app:
image

`pinnedContext` already carried `projectName` next to `projectIconUrl`; only the icon was ever rendered. Now the name renders too, dimmed, as a path-style prefix.

**`WorkspaceNameMarquee`** gains an optional `prefix`. It renders *inside* the marquee's scrolling span, not beside it — so overflow measurement, the hover-scroll reveal and the `title` attribute all operate on the full `Flamaster/local` string rather than just the tail. The prefix span carries `text-muted-foreground`; as a declaration on a child it beats the row's inherited `text-foreground/80` regardless of class order.

**`DashboardSidebarExpandedWorkspaceRow`** passes `prefix` only when `pinnedContext` is set, reusing the existing `"Session"` message for project-less workspaces. Rows inside a project group are untouched — their header already says the project. The other two `WorkspaceNameMarquee` call sites (v1 sidebar, v2 workspaces list) pass no prefix and render exactly as before.

The avatar stays: once you know which row is which, the colour is still a useful cue.

Collapsed rail is icon-only, so it needs no change.

## Verification

- 5 new `WorkspaceNameMarquee` tests, including that an absent or empty prefix renders byte-identically to today.
- `bun run typecheck`, `bun run lint` clean.
- `bun run check:i18n` reports 0 missing across all 17 locales and `compile --strict` passes — no new string, `"Session"` already existed in this file.

**Not yet verified over CDP.** This worktree has never been provisioned (no `.env`, no dev stack), so there was no renderer to attach to, and `apps/desktop/AGENTS.md` wants UI changes confirmed in the real app. That's the remaining work before this leaves draft: pin two same-named workspaces from different projects and capture before/after screenshots.

---
## Summary by cubic
Fixes #7574 by showing the project name next to pinned workspace rows in the sidebar, so two same-named workspaces from different projects are no longer identical. The prefix renders inside the marquee, so the title and hover reveal cover the full label. Other `WorkspaceNameMarquee` call sites remain unchanged.

<sup>Written for commit 281a70e7812fda42a8b07d878c64a0e286c2a8ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7575?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Workspace names in the dashboard sidebar now display their associated project or session prefix, formatted as `prefix/name`.
  - Prefixes are visually dimmed for clearer hierarchy.
  - Full prefixed workspace labels are available as hover text when needed.

- **Tests**
  - Added coverage for workspace names with and without prefixes, including empty-prefix handling and display styling.


theblondealex:
I love this and actually hadnt pinned workspaces because of this

theblondealex:
gave this a lil review, I am a fan

PR #7584, opened 2026-09-16
fix(desktop): classify an unspawnable git as an environment failure

saddlepaddle:
## Problem

On machines where no `PATH` entry holds a git binary, Node fails the spawn and
every git command in the desktop dies identically until the machine is fixed.
The desktop's git classifier knew only git's own stderr for a missing worktree
and a missing repository, so this arrived at the router unclassified and became
an `INTERNAL_SERVER_ERROR`.

Two consequences, plus the reporting noise:

- The user got a generic failure instead of the "Git can't read this workspace"
  copy that already exists for exactly this state (`gitChangesUnavailableCopy`).
- `PRECONDITION_FAILED` is in `DETERMINISTIC_FAILURE_CODES`, which is what makes
  a query back off to the 30s slow poll. Unclassified, this condition — which
  cannot heal mid-session — kept polling on the transient-error interval.

## Root cause

`classifyEnvironmentalGitError` / `rethrowEnvironmentalGitError` matched only
`unable to read current working directory` and `not a git repository`. A failed
spawn of git itself is the same class of environmental failure (the sibling
classifier in `packages/host-service` already treats the macOS CLT/Xcode shim
refusals this way) but had no pattern.

Matching is on the message rather than `code === "ENOENT"` because simple-git
rebuilds the spawn failure as a `GitError` carrying only the stringified
original — the code is gone by the time the classifier sees it. Verified
against real errors on Node rather than assumed:

```
=== node v26.8.1 | via simple-git (the getBranches path) ===
name: "Error"
code: undefined
message: "Error: spawn git ENOENT\n    at ChildProcess._handle.onexit (...)"

=== via execFile (execGitWithShellPath path) ===
name: "Error"
code: "ENOENT"
message: "spawn git ENOENT"
```

Both real shapes are covered; the `code` is only present on one of them, which
is why a code check would have missed the reported path entirely.

## Fix

One pattern, used in the two functions that already share
`CWD_UNREADABLE_PATTERN`. No new abstraction, no new error kind — it reuses the
existing `GIT_ENVIRONMENT` cause the renderer already renders and backs off on.

The line is anchored (`/^(?:Error: )?spawn git ENOENT$/m`) so it cannot swallow
neighbouring failures. Checked against real errors:

```
pattern under test: /^(?:Error: )?spawn git ENOENT$/m

REAL simple-git getBranches failure -> matches: true
REAL execFile git failure        -> matches: true
REAL file-read ENOENT            -> matches: false (must be false)
REAL working git still succeeds  -> "sentry/desktop-73-git-spawn-enoent"
```

Negative tests name what it must not match: an ENOENT from a file a task read,
another binary git spawns (`spawn git-lfs ENOENT`), and the phrase quoted inside
a larger failure. All three must keep reporting as bugs.

## Verification

Tests were written first and failed before the change:

```
(fail) classifyEnvironmentalGitError > maps an unspawnable git to GitEnvironmentError
(fail) rethrowEnvironmentalGitError > throws PRECONDITION_FAILED when git cannot be spawned
 8 pass
 2 fail
```

After:

```
$ bun test src/lib/trpc/routers/workspaces/utils/git-errors.test.ts
 10 pass
 0 fail
 25 expect() calls
```

```
$ bunx biome check <the two changed files>
Checked 2 files in 15ms. No fixes applied.

$ cd apps/desktop && bun run typecheck
$ tsc --noEmit          (clean)

$ bun test src/lib/trpc/routers/workspaces src/lib/trpc/routers/changes
 327 pass
 0 fail
 586 expect() calls
Ran 327 tests across 30 files. [15.54s]
```

`check:i18n` not run: no user-facing string added or changed — the copy this
now reaches already existed.

Refs DESKTOP-73

---
## Summary by cubic
Classifies an unspawnable git — no PATH entry holds a git binary — as an environment failure instead of an internal server error. Users now get the existing "Git can't read this workspace" copy, and queries back off to the deterministic-failure slow poll instead of the transient-error interval. Fixes DESKTOP-73.

**Notes**
- Matches `/^(?:Error: )?spawn git ENOENT(?:\r?\n|$)/` anchored to the message start rather than `code === "ENOENT"` because simple-git stringifies the spawn error and loses the code.
- Other `ENOENT`s — file reads, missing `git-lfs`, and a git that ran but relayed the words from a failing hook — still classify as bugs.

<sup>Written for commit cc3254e5ac553c508e58b92de98551191802acef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of Git environment failures when Git cannot be started.
  - Git startup errors now return a clear precondition failure with the appropriate environment error classification.
  - Unrelated missing-file errors and embedded error text remain unaffected.

- **Tests**
  - Added coverage for multiple Git process-start failure formats.
  - Added validation that unspawnable Git failures are reported correctly while unrelated errors remain unclassified.


saddlepaddle:
Closing this — the review found the matcher unsound, and I verified all of it
against the code rather than taking it on faith.

**1. It reclassifies a git command that ran fine.** `rethrowCommitFailure` calls
`rethrowEnvironmentalGitError` *before* the `exitCode === 1` branch, so a commit
declined by a pre-commit hook whose own git spawn failed gets relabelled from
`COMMIT_DECLINED` to `GIT_ENVIRONMENT` — the user loses the hook's reason and is
told "Git can't read this workspace". `commit-failure.ts` states the rule this
breaks in its own comment: hook text "is arbitrary user program output and
matching it would classify on the hook author's prose."

My negative test gave false assurance. Its fixture was
`"fatal: pre-commit hook failed\nspawn git ENOENT\n"`, but git prints no such
prefix for a pre-commit failure — the message is purely the hook's output, so
the real shape starts with the matched line. The test passed for the wrong
reason.

**2. The message cannot identify the condition.** A missing cwd produces a
byte-identical error — `code: ENOENT`, `syscall: "spawn git"`,
`message: "spawn git ENOENT"` — which is `WORKTREE_MISSING`, a different kind
consumers branch on. The must-match and must-not-match inputs are literally the
same string, so no message-only matcher can separate them. That is not a regex
to tighten; the information isn't in the message.

**3. It covered 28% of the condition.** Over 30 days `spawn git ENOENT` is 135
events: 97 in host-service (which routes through the other classifier, untouched
here) and 38 in desktop. I scoped my volume check to one project and undercounted
by 3.5x.

The likelier root cause is ours, not the user's: `getProcessEnvWithShellPath`
*overwrites* PATH with the login shell's, and `COMMON_MACOS_PATHS` re-adds only
the homebrew and `/usr/local` dirs — not `/usr/bin`, where the macOS git shim
lives. A shell config that assigns rather than appends to PATH would lose git
because of how we build the env. `packages/host-service/src/runtime/login-shell-env.ts`
already does this correctly by unioning.

That's the fix worth making, but it changes the env for every git invocation and
I can't confirm it from the events (no PATH is captured), so it should not land
on a hunch attached to this PR. Leaving it for a human with the evidence written
up rather than shipping a label for a cause I hadn't found.

saddlepaddle:
Superseded by the analysis above: the classifier is unsound and the real cause is likely our own PATH construction. Not merging a relabel in its place.

PR #7598, opened 2026-09-16
fix(desktop): keep viewport emulation on a crashed browser pane from crashing the app

saddlepaddle:
## Problem

The desktop main process can segfault (null dereference on the main thread), which closes the whole app. It happens when an agent's CDP client sends a viewport-emulation command to a browser pane whose page renderer has already crashed. In the one field event, a guest page crashed in Chromium's device-emulation teardown and the main process died 14 seconds later on the next `Emulation.setDeviceMetricsOverride`. That renderer-side crash (DESKTOP-17A) has hit 4 different machines in the last 8 days, so the first half of the sequence is not rare.

## Root cause

When a guest renderer crashes, its `WebContents` stays alive (`isDestroyed()` is false) but its `RenderWidgetHostView` is torn down. Chromium 146's `WebContentsImpl::SetDeviceEmulationSize` reads `GetPrimaryMainFrame()->GetView()` and dereferences it without a null check. Its counterpart `ClearDeviceEmulationSize` does check. `EmulationHandler::SetDeviceMetricsOverride` and `SetVisibleSize` both call it. Electron's `WebContents.enableDeviceEmulation` / `disableDeviceEmulation` dereference the same missing view.

In our code there are two ways to reach that:
- **Agent CDP bridge:** `attachCdp`'s `send` forwards any method to `wc.debugger.sendCommand` whatever state the guest is in. The existing `isDestroyed()` check (DESKTOP-ZS) doesn't cover a crashed guest.
- **Device toolbar:** `setDeviceEmulation` calls the Electron APIs whatever state the guest is in.

Reproduced with plain Electron 41.10.3 (the shipping version): a `<webview>` guest, `debugger.attach`, `forcefullyCrashRenderer()`, then one command. Results:

| Command on a crashed guest | Result |
| --- | --- |
| `Emulation.setDeviceMetricsOverride` (debugger attached before *or* after the crash) | **SIGSEGV** (exit 139) |
| `Emulation.setVisibleSize` | **SIGSEGV** |
| `wc.enableDeviceEmulation(...)` / `wc.disableDeviceEmulation()` | **SIGSEGV** |
| `Emulation.clearDeviceMetricsOverride` | `{}` |
| `Page.navigate` | succeeds and brings the guest back (`isCrashed()` flips to `false`) |
| `Page.reload` | succeeds, but reloads the **host window** (Chromium's `PageHandler::Reload` reloads the outermost frame / embedder), not the guest. Pre-existing and not changed here. |
| `Input.dispatchMouseEvent`, `Page.captureScreenshot` | CDP error |
| `Runtime.evaluate`, `Emulation.setTouchEmulationEnabled` | no reply (no crash) |

The macOS crash report from the repro has the same signature as the field event: `EXC_BAD_ACCESS / KERN_INVALID_ADDRESS at 0x0` on thread 0.

## Fix

- **CDP bridge:** while `wc.isCrashed()`, refuse only the two methods that reach `SetDeviceEmulationSize` (`Emulation.setDeviceMetricsOverride`, `Emulation.setVisibleSize`) with a CDP error. Everything else is still forwarded. I didn't close the session on crash because `Page.navigate` over CDP is how a client recovers a crashed guest today (see the table), and that keeps working. The refusal message says to navigate, not reload: an agent following "reload" would send `Page.reload` and reload the whole app window.
- **Device toolbar:** `setDeviceEmulation` returns early while the guest is crashed. The renderer already ignores this mutation's result.

In real Electron 41.10.3: `isCrashed()` is `true` right after `render-process-gone`. It flips back to `false` once `Page.navigate` recovers the guest, and `Emulation.setDeviceMetricsOverride` then succeeds. So the guard only applies while the view is actually missing:

```
after crash: isCrashed = true
guard: refuse Emulation.setDeviceMetricsOverride (not forwarded)
Page.navigate on crashed guest -> frameId ok
after recovery: isCrashed = false
emulation after recovery -> {}
main process alive
```

The real fix belongs upstream (a null check in `SetDeviceEmulationSize` like the one in `ClearDeviceEmulationSize`). This change only stops our bridge and toolbar from driving the app into it.

## Verification

Tests: one positive, one negative. The negative test covers what must **not** be refused: `Page.navigate` on a crashed guest, and `Emulation.setDeviceMetricsOverride` on a live guest. Before the source change, the positive test failed and the negative test passed:

```
(fail) CDP session on a crashed guest renderer > viewport resizes are refused instead of forwarded [0.88ms]
error: expect(received).not.toHaveBeenCalled()
Expected number of calls: 0
Received number of calls: 2
 1 pass
 1 fail
```

After:

```
$ bunx biome check apps/desktop/src/main/lib/browser/browser-manager.ts apps/desktop/src/main/lib/browser/browser-manager.test.ts
Checked 2 files in 13ms. No fixes applied.

$ cd apps/desktop && bun run typecheck
$ tsr generate
$ tsc --noEmit
(exit 0)

$ bun test src/main/lib/browser/
 112 pass
 0 fail
Ran 112 tests across 9 files. [258.00ms]
```

No user-facing strings were touched (the refusal text is a CDP error message for agent clients), so `check:i18n` was not run.

Refs DESKTOP-195

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented viewport resizing and device emulation commands from being sent when a pane’s renderer has crashed.
  * Added clear errors for unsupported viewport operations during renderer crashes.
  * Preserved navigation and other crash-recovery commands so affected panes can recover normally.


saddlepaddle:
Re the two CodeRabbit "outside the diff" findings (the `isCrashed()` check could be stale by the time `sendCommand` / `enableDeviceEmulation` / `disableDeviceEmulation` runs): dismissing both as false positives, for these reasons.

- **Nothing can run between the check and the call.** Each check sits in the same synchronous JS stack on the browser UI thread as the native call, with no `await` between them. `Debugger::SendCommand` dispatches synchronously all the way to `WebContentsImpl::SetDeviceEmulationSize` (the field stack shows that whole chain in one call). The Electron emulation calls are synchronous too.
- **The view disappears later, in its own task.** The view goes away when Chromium processes the renderer's death (`RenderProcessHostImpl::ProcessDied` → `RendererExited`), and that runs as a separate UI-thread task. It can't run between two statements of one JS task.
- **A renderer that died but isn't processed yet is still safe.** In that case `isCrashed()` is still `false`, but the view hasn't been torn down either. The dereference is safe and the IPC just goes nowhere. The crash needs the view to be *gone*, and `isCrashed()` becomes `true` in the same task that removes it.

So there's no check-then-call race to close. An "atomic" Electron-level check doesn't exist and isn't needed.

PR #7614, opened 2026-09-16
feat(cli): cap 'pages list' output with --limit

saddlepaddle:
## What

Adds a `--limit <n>` option to `superset pages list` that caps how many pages are printed.

```
superset pages list --limit 10
```

## Notes

- `page.list` takes no limit server-side, so the cap is applied to the result in `run` before `display`. That way it bounds both the table and `--json` output, which `display` alone could not do (it receives only `data`, not `options`).
- The option has no default, so omitting it returns every page and leaves the current output byte-for-byte unchanged.
- `.int().min(1)` matches the validation style used by `automations logs --limit`.

## Verification

- `bun run typecheck` in `packages/cli` — passes.
- `biome check --write` on the touched file — clean.

---
## Summary by cubic
Adds a `--limit <n>` option to `superset pages list` so large orgs can cap how many pages are printed. Without the flag, every page is still printed and the output is unchanged.

**New Features**
- The limit is applied before display, so it bounds both table and `--json` output.
- Validation (`--limit` must be an integer ≥ 1) matches `automations logs --limit`.

<sup>Written for commit 1f38068b58cf401240a415fb50321843c73cddc9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7614?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a `limit` option to the `pages list` command, allowing users to display only the first specified number of pages.
  - The option requires a positive integer.
  - Without a limit, the command continues to display all pages as before.


saddlepaddle:
Closing: this PR was an end-to-end validation run for cloud workspaces, not a change we want to land.

PR #7616, opened 2026-09-17
feat(cli): add --limit to pages list

saddlepaddle:
## What

`superset pages list` takes a `--limit N` flag. Omitting it is unchanged: every page in the org (or workspace) comes back, newest-updated first.

```bash
superset pages list --limit 10
superset pages list --workspace my-ws --limit 5
```

## How

The limit goes through the tRPC input and becomes a SQL `LIMIT` after the `updatedAt desc` ordering, instead of the CLI slicing the array it got back. `page.list` mints a signed page ticket (and a thumbnail ticket) per row, so trimming client-side would do that signing work for rows nobody sees.

`listPagesSchema` gains an optional positive int `limit` — no max cap, since the default with the flag absent already returns more than any cap would.

- `packages/cli/src/commands/pages/list/command.ts` — the flag, `number().int().min(1)`
- `packages/trpc/src/router/page/schema.ts` — `listPagesSchema.limit`
- `packages/trpc/src/router/page/page.ts` — `.limit()` on the ordered query

## Testing

- `turbo typecheck --filter=@superset/cli --filter=@superset/trpc` passes
- `bun test src/commands/pages` in `packages/cli` passes (26)
- biome clean on the three files

Not run end-to-end against a live API: `bun run dev` in `packages/cli` fails in this workspace on a missing `babel-plugin-macros`, unrelated to this change.

---
## Summary by cubic
`superset pages list` now accepts a `--limit N` flag to cap the number of pages returned. Omitting the flag keeps the old behavior of returning every page, newest-updated first.

**New Features**
- The limit travels through the tRPC input and becomes a SQL `LIMIT` after the `updatedAt desc` ordering, so ticket signing only runs for rows actually returned.
- `listPagesSchema` accepts an optional positive integer `limit`; the CLI flag rejects zero and negative values, and no cap is applied when the flag is absent.

<sup>Written for commit ff8f88a7d77fdfda5f5cfa14d7aca6985f4bdcb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7616?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an optional limit to page listings in the CLI, allowing users to request a specific maximum number of results.
  - Page listing requests now validate that the limit is a positive whole number.
  - The page listing API applies the requested result limit while preserving the existing sorting order.


saddlepaddle:
Closing: this PR was an end-to-end validation run for cloud workspaces, not a change we want to land.

PR #7617, opened 2026-09-17
feat(desktop): add optional custom branch name input

Kitenite:
## What & why

Adds a **Show custom branch name input** toggle directly below Branch prefix in **Settings → Git & Worktrees → Branch naming**. It defaults to off and persists on the device.

When enabled, a borderless branch icon and editable text above the workspace creation prompt let users enter a custom branch name (mockup option 2). An empty field keeps automatic naming. V1 and V2 share the same input helpers: whitespace becomes hyphens while typing, blur sanitizes with case preservation and V1 length limits, and clearing restores automatic naming. Submission applies the same sanitization without requiring blur; provider-supplied names retain their existing handling. The field uses the existing draft and submission logic and is hidden for sessions, local checkouts, PR checkouts, and cloud workspaces. Includes Settings search support and translations for all shipping locales.

Custom branch input above the composer

## How I tested it

- `bun run lint`, `bun run typecheck`, `bun run check:i18n`, and `bun run check:plugins` passed; desktop typecheck rerun after sharing the V1 input logic.
- Focused branch-input, name-resolution, settings-search, and persisted-key tests: 33 passed.
- Extensive real desktop CDP checks on this worktree's renderer (`localhost:4785`, CDP `9357`): 16 input edge cases, keyboard/blur, 10k-character insertion, repeated toggles, settings search/reload persistence, checkout modes, PR link/unlink, and provider branch/manual override/clear behavior.
- Submitted actual workspaces with Cmd+Enter before blur: verified the resulting branch with Git, invalid-only fallback, draft reset, and duplicate branch reopening. Removed the test-created workspaces afterward.
- Found and fixed two Git-invalid truncation cases (`.` and `.lock` suffixes); reproduced before and passed after, with regression tests in the shared V1/V2 helper.
- Captured screenshots and checked a simulated 900×620 viewport for overflow. Cloud creation, remote hosts, and a separate V1 UI session were not exercised.
- `bun run test` did not pass: failures in CLI desktop-settings acknowledgement and host-service workspace integration tests outside the changed files.

## Checklist

- [x] "Allow edits from maintainers" is checked on fork PRs (not a fork PR)

## Summary by CodeRabbit

- **New Features**
  - Added an optional custom branch-name field above the workspace creation prompt.
  - Added a Git setting to show or hide the custom branch-name field.
  - Added automatic branch-name cleanup for whitespace, invalid characters, length limits, and Git restrictions.
- **Bug Fixes**
  - Ensured manually entered branch names are sanitized consistently, including when submitted without leaving the field.
- **Localization**
  - Added translations for branch-naming controls across supported languages.
- **Tests**
  - Added coverage for branch-name editing, validation, sanitization, and workspace submission behavior.


harshithmullapudi:
Rebased onto `main` (was 29 commits behind; rebase was conflict-free) and addressed the review.

**Settings search — fixed.** The finding was right but narrower than the real bug: `V2GitSettings` ignored `visibleItems` entirely, so *no* v2 git row filtered on search, only highlighted. Rather than special-casing the new row, `page.tsx` now passes `visibleItems` to `V2GitSettings` and all three rows gate on `isItemVisible`, matching what `GitSettings` (v1) already does. The "Branch naming" heading hides when both rows under it are filtered out. `BranchNameInputSection` also moved onto the shared `SettingsRow` instead of hand-rolling the same flex row.

**V1 `PromptGroup` — half accepted.**
- *Gate on `linkedPR`:* fixed. Confirmed the bug — with a PR linked, submission goes through `createFromPr({ projectId, prUrl })`, which never receives `branchName`, so a typed name was silently dropped. The field is now hidden when `linkedPR` is set, which also matches what the v2 screen already does.
- *Gate on `showBranchNameInput`:* declined. V1 has always shown this field; putting it behind a setting that is off by default would *remove* a field from v1 users, which is the complaint this PR exists to answer. The setting is scoped to the v2 creation screen, where the field is new.

**Translations — fixed.** All three verified against the catalogs' own terminology before changing: `pl`/`ru` use `prompt`/`промпт` for the AI prompt and reserve `Polecenie`/`Команда` for "Command", so `poleceniem`/`запросом` were inconsistent; `nl` already uses `Naamgeving` for "naming" (`Naamgevingsinstructies`), and `Branchnamen` read as "branch names" rather than the activity.

### Verification

`lint`, `typecheck` (42/42 packages), and `check:i18n` pass — the latter reports 0 missing across all 17 locales and `compile --strict` succeeds. Catalogs regenerated clean: the only `.po` delta is the three intended lines. Focused tests 21/21; persisted-key registry 12/12.

Drove the real dev app over CDP (renderer on this worktree's vite port, authenticated session with a real `activeOrganizationId`), 13/13 checks:

| | |
|---|---|
| baseline, setting off | `#new-workspace-branch-name` absent |
| `Branch naming` section | renders under Git & worktrees |
| search `creation prompt` | filters to the one row, other two hidden |
| clearing search | all 3 rows return |
| toggle | `data-state=checked`, persisted to the settings store |
| creation screen | field now renders |
| typing `ENG 1234 fix login` | → `ENG-1234-fix-login` (spaces hyphenated live) |
| blur | value held, case preserved |
| clearing the field | empties, restoring automatic naming |

Navigation used the app's own links throughout — an earlier attempt via `location.hash` changed the URL without remounting the route, so every route assertion now also checks that the route's content actually mounted.

**Not covered:** I did not submit a workspace in this pass, so the typed-name→actual-git-branch path rests on the original verification plus `resolveNames` unit tests. V1's modal was also not exercised live (the `linkedPR` change is a render guard).

PR #7636, opened 2026-09-18
fix(host-service): classify an index its storage cannot deliver as environmental

saddlepaddle:
## Problem

`git.getStatus` returns a 500 when git cannot map the repository's index because the storage under it could not deliver the file. Every status refresh on such a machine reports to Sentry.

HOST-SERVICE-6N is 3 events from one machine on 1.29.0. The condition is wider than that group: over the last 30 days the same failure shows up as 29 `git.getStatus` events from 7 machines on 1.22.0, 1.27.0 and 1.29.0. Most of those land in an older group that also collects unrelated git failures such as `spawn git ENOENT`, so archiving that group would also mute real failures. Three of the seven machines had the repository under the user's Desktop folder. The other four report a relative `.git/index`, so their location is unknown.

## Root cause

git's `do_read_index` (read-cache.c) opens the index, then `mmap`s it, and on failure calls `die_errno("%s: unable to map index file%s")`. On macOS `mmap_os_err()` is empty, so the message ends in `: <strerror(errno)>`.

On macOS, `mmap` asks the filesystem (`VNOP_MMAP_CHECK` in XNU's `kern_mman.c`) whether the file can be mapped. A cloud-synced folder can evict a file and leave a dataless placeholder. Mapping that placeholder asks the sync client to fetch the contents back. In XNU's resolver (`vfs_syscalls.c`), the kernel waits on the sync client and returns the error the client replies with. The kernel supplies `ETIMEDOUT` itself only when the sync service cannot be reached or exits. A timed-out or cancelled fetch reaches git as whatever errno the client chose. APFS is closed source, so the step where its mmap check starts that fetch cannot be read. The classification does not depend on it, only on the error coming from the storage layer. A network volume that stops answering also fails with `ETIMEDOUT`. The observed errnos are exactly these two: "Operation timed out" and "Operation canceled".

Nothing Superset does can make that index readable. The index stays unreadable until the storage recovers or the repository moves off it. `rethrowEnvironmentalGitError` already maps the neighbouring case, where macOS refuses to read the working directory, to `PRECONDITION_FAILED / GIT_ENVIRONMENT`. It had no case for this one, so the error fell through as an `INTERNAL_SERVER_ERROR`.

## Fix

The change adds one pattern to the existing `CWD_UNREADABLE` → `GIT_ENVIRONMENT` branch in `classify-git-error.ts`, and the message is kept verbatim. The pattern requires git's own `unable to map index file: ` prefix immediately before one of the two errno strings, at the end of a line.

The client treats `PRECONDITION_FAILED` and `INTERNAL_SERVER_ERROR` the same way: `WorkspaceClientProvider` retries both once, and `useGitStatus` has no code-specific handling. So this changes reporting only, not what the user sees.

**Still a 500 after this change:** the same line with any other errno, such as `Cannot allocate memory`. A timeout reading any other file, such as an object or a worktree file, also stays a 500.

## Verification

Tests were written before the pattern. Before the change, the positive case failed and the negative case passed:

```
(fail) rethrowEnvironmentalGitError > index its storage cannot deliver → PRECONDITION_FAILED / GIT_ENVIRONMENT [0.12ms]
 34 pass
 1 fail
Ran 35 tests across 1 file. [34.00ms]
```

After:

```
 35 pass
 0 fail
Ran 35 tests across 1 file. [33.00ms]
```

The negative test is load-bearing. Against a deliberately broad variant (`/unable to map index file|Operation timed out/m`) it fails:

```
(fail) rethrowEnvironmentalGitError > does not claim other index mapping failures or network timeouts [0.20ms]
 34 pass
 1 fail
```

I also ran each distinct real message from the 7 machines through the classifier locally; the messages are not committed. All six classify as `PRECONDITION_FAILED/GIT_ENVIRONMENT`.

`bunx biome check` on both files:

```
Checked 2 files in 16ms. No fixes applied.
```

`cd packages/host-service && bun run typecheck` exits 0 (`tsc --noEmit --emitDeclarationOnly false`, no output).

`bun test src/trpc/router/git` (packages/host-service):

```
(fail) getGitStatusPartial: renames and index interplay > long converging sequence of mixed batches [5371.05ms]
 290 pass
 1 fail
Ran 291 tests across 25 files. [592.33s]
```

That one failure is a 5s test timeout on this machine, and it also times out when run alone. It is in `git-status-partial.adversarial.test.ts`, which neither imports nor reaches the classifier. CI is the check of record for it.

Refs HOST-SERVICE-6N

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Git errors caused by unavailable or unresponsive index storage, including cloud-synced or network-mounted repositories.
  * These environmental failures now return a clear precondition error instead of a generic server error.
  * Memory-related index failures and unrelated network timeouts continue to be handled separately, preserving their existing error behavior.


saddlepaddle:
@coderabbitai re the outside-diff finding *Restrict this matcher to Git index paths*: not applying it. Here's why.

git prints `unable to map index file` from exactly one place, `do_read_index()` in `read-cache.c`. A code search of git/git finds the wording only in that file. The path in the message is always the file git is reading as an index:

- `.git/index`
- a linked worktree's `.git/worktrees/<name>/index`
- the split-index base `.git/sharedindex.<oid>` (see `read_index_from`, read-cache.c:2396/2406)
- whatever `$GIT_INDEX_FILE` names

So `fatal: src/file: unable to map index file: …` can only happen when `src/file` is being read as an index. That is the same condition: storage that can't deliver the file git is mapping. A `.git/index` or worktree-path requirement wouldn't exclude any other failure. It would only stop matching shared indexes and custom index files that hit the same errno.

PR #7671, opened 2026-09-19
fix(db,ci): make migration 0119 online-safe and bound production migrations

Kitenite:
## Summary

Unblocks production deploys after the 17 September API outage. Deploy Production is still disabled; this PR is what has to merge before it is turned back on.

- **Migration 0119 no longer rewrites `ingest.webhook_events`.** The enum to text change on `provider` was a full rewrite of the largest table in production, run while the same transaction held ACCESS EXCLUSIVE locks on `auth.users`, `auth.organizations` and `integration_connections`. 0119 is regenerated with `drizzle-kit generate` and is identical to the original minus that one statement. It rolled back during the incident and never applied in production.
- **`provider` stays the `integration_provider` enum.** Every writer passes `linear`, `github`, `notion` or `slack`, all already in the enum. `recordWebhookDelivery` goes back to the enum type, so a non-enum provider is a compile error.
- **Production migrate job is bounded.** A `concurrency` group (three migrators ran at once during the incident) and `PGOPTIONS: -c lock_timeout=5s -c statement_timeout=60s`. drizzle applies every pending migration in one transaction, so splitting files does not isolate locks; timeouts are what make a blocked migration fail and roll back instead of stalling traffic.
- **Preview migrate step gets the same timeouts**, so the preview database check runs what production will run.

## Testing

Rehearsed with the exact CI command on a Neon branch of production (122 migrations applied, 540 GB `webhook_events`, 2 CU):

| Scenario | Result |
|---|---|
| No timeouts, one held read lock on `auth.users` | Migrator still waiting on `DROP TABLE ... CASCADE` after 12s, killed |
| With timeouts, same blocker | Exit 1 after 7.0s, `canceling statement due to lock timeout`, clean rollback, still 122 applied |
| Clean run | Exit 0 in 2.3s, new tables present, old table and FK gone, `provider` still the enum |
| Re-run | No-op, exit 0 |

`PGOPTIONS` confirmed to reach the session through Neon's direct endpoint with the `@neondatabase/serverless` driver. db and api typecheck, ingest tests, lint and `drizzle-kit check` pass. The Deploy Database (Neon) check on this PR applies 0119 to a fresh branch of production with the same timeouts.

## Notes

- The journal timestamp for 0119 is kept, so a dev database that already applied the old 0119 skips this one. Its `provider` column stays text until the branch is reset; inserts still work.
- #7589 carries its own `0119_slack_agent_launches.sql` and needs a rebase and regenerate.
- The concurrency group gets its first real exercise on the first merge after the workflow is re-enabled.

---
## Summary by cubic
Fixes the 17 September production outage by regenerating migration 0119 without the `webhook_events` table rewrite and ordering its locks to match the application. Unblocks production deploys; Deploy Production stays disabled until this merges. Production migration runs are now bounded and retried.

**Bug Fixes**
- 0119 no longer changes `provider` to text; the column stays the `integration_provider` enum, and `recordWebhookDelivery` types it as the enum so invalid providers fail compilation.
- The `automation_events` foreign key is dropped before `plugin_connections`, child table first like inserts, avoiding the deadlock seen under traffic.
- The journal timestamp for 0119 is kept, so dev databases that already applied the old 0119 skip this one.

**Reliability**
- Production migrations run under a `concurrency` group with `lock_timeout=5s` and `statement_timeout=60s` via `PGOPTIONS`; preview gets the same timeouts.
- Because drizzle applies pending migrations in one transaction, those timeouts make a blocked migration roll back instead of holding traffic.
- The production migrate step makes up to eight attempts, retrying only on deadlock or lock timeout.

<sup>Written for commit d9218aaf1cf2f654f4ce302e5b6219fae7bcd364. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/superset-sh/superset/pull/7671?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Webhook delivery records now accept only supported integration providers, preventing invalid values.
  - Webhook event data now consistently enforces the supported provider set at the database level.

- **Reliability**
  - Database deployments now apply lock and execution time limits to reduce stalled migrations.
  - Production migrations run one at a time and retry select transient failures automatically.
  - Migration ordering has been adjusted to reduce deployment conflicts.


Kitenite:
Follow-up after rehearsing under load (16 clients against the tables 0119 locks, on a Neon branch of production, migration applied mid-run):

- **Found:** the generated statement order reproduced the 08:53 deadlock from 17 September, 3 attempts out of 3. `DROP TABLE plugin_connections CASCADE` takes ACCESS EXCLUSIVE on `auth.organizations`, then the migration asks for `automation_events`, while an in-flight insert into `automation_events` holds that table and waits on `auth.organizations` for its FK check. Safe failure (0 failed transactions, about 1.1s stall on auth reads per attempt), but the migration would not land under real traffic.
- **Fixed:** the `automation_events` FK drop now runs first, so locks are taken child table first like every insert. Same end schema, same snapshot. Same load: applied first try on 4 runs out of 4, 1.7 to 2.0s, 0 failed transactions.
- **Added:** the production migrate step retries up to 5 times, only on `deadlock detected` or `lock timeout`. Tested with the exact `run:` block under `bash -eo pipefail`: a held lock retried once then applied, bad credentials failed in 0.7s with no retry, nothing pending exits 0.

The Neon preview branch for this PR was deleted before this push so the Deploy Database check applies the reordered 0119 to a fresh branch of production rather than no-op on the old one.

